PR #5, opened 2021-02-04
Integration test setup for parachain

h4n0:
Several changes here:

1. Added litentry/polkadot as submodule of litentry parachain, to lock the version dependency between polkadot and parachain.
2. Updated integration test of parachain, to run through our test scenarios with two relay nodes as validators.
3. Added scripts and CI workflow for running through integration tests on GitHub Actions.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "polkadot"]
+	path = polkadot
+	url = https://github.com/litentry/polkadot
```
Satoshi-Kusumoto, line 3: why not using upstream?

ts-tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Install
+
+`npm i`
+
+## Run
+
+For now, you need to start the node manually first by the command
+
+`./target/debug/litentry-node --dev --tmp`
```
Satoshi-Kusumoto: These are obsolete I think

ts-tests/scripts/export_env.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+export etherscan="RF71W4Z2RDA7XQD6EN19NGB66C2QD9UPHB"
```
Satoshi-Kusumoto, line 2: hard code api tokens?

Satoshi-Kusumoto (commented):
I am afraid some changes to the parachain, if not fix dependency, will result the test to fail. But we may just leave it fail when we want to move fast until the next check point

ts-tests/scripts/export_env.sh
h4n0: Yes, it's a good point. We had this since the litentry node test. I would write down an issue to encrypt the config file like this one.

.gitmodules
h4n0: We will try to set up test first with a local proved polkadot version (chain spec, Cargo.lock etc) first and once our test runs stably, It's good idea to migrate this dependency to upstream.

Satoshi-Kusumoto (approved):
Maybe also write an issue about the obsolete README

ts-tests/README.md
h4n0: solved with new commit

PR #6, opened 2021-02-09
Structure refactor

Satoshi-Kusumoto:
Now the code structure is almost the same as node. While there are still some minor differences.
Wait for #5 to be merged first.
@h4n0 after #5 is merged, you may add some more things here?

@suinuj Currently MultiAddress is introduced, therefore some changes are made. And in runtime/lib.rs
```rust
let address = Self::Lookup::unlookup(account);
```
this is added. Please check carefully if these are right


runtime/src/lib.rs
```diff
@@ -154,13 +180,21 @@ parameter_types! {
 	pub const SS58Prefix: u8 = 42;
```
suinuj: ss58prefix for litentry is 31, could we use it for parachain

runtime/src/lib.rs
```diff
@@ -154,13 +180,21 @@ parameter_types! {
...
+	type BlockWeights = RuntimeBlockWeights;
+	/// The maximum length of a block (in bytes).
+	type BlockLength = RuntimeBlockLength;
 	/// The identifier used to distinguish between accounts.
 	type AccountId = AccountId;
 	/// The aggregated dispatch type that is available for extrinsics.
 	type Call = Call;
 	/// The lookup mechanism to get account ID from whatever is passed in dispatchers.
-	type Lookup = IdentityLookup<AccountId>;
+	type Lookup = AccountIdLookup<AccountId, ()>;
```
suinuj, line 197: why not use AccountIndex as second parameter here.

runtime/src/lib.rs
```diff
@@ -383,10 +425,10 @@ construct_runtime! {
 		AccountLinkerModule: pallet_account_linker::{Module, Call, Storage, Event<T>},
 		OffchainWorkerModule: pallet_offchain_worker::{Module, Call, Storage, Event<T>, ValidateUnsigned},
 	}
-}
+);
 
 /// The address format for describing accounts.
-pub type Address = AccountId;
+pub type Address = sp_runtime::MultiAddress<AccountId, ()>;
```
suinuj, line 431: MultiAddress definition here is fine.

runtime/src/lib.rs
Satoshi-Kusumoto: I am just asking since address is now MultiAddress, this only affects ocw. So I am not sure if the correction is right. And here we may also use Indices. I think you know about the system, may you give a brief idea about potential candidates?

suinuj: AccountIndex is fine, AccountIdLookup is simpler than introducing indices pallet.

h4n0:
I will create a new PR with adding ocw updates and token server from @suinuj , to enable full integration test. 

This needs however this PR to be merged first, as a precondition, to compile token server as well.

runtime/src/lib.rs
Satoshi-Kusumoto: Currently seems we may omit the AccountIndex as second parameter here. When it's necessary, we may add it. Or you may provide why it is needed now?

suinuj (approved):

PR #8, opened 2021-02-10
Setup self hosted server as CI runner

h4n0:


ts-tests/scripts/smoke-test.ts
```diff
@@ -0,0 +1,287 @@
...
+  console.log(
+    `Initiating the API (ignore message "Unable to resolve type B..." and "Unknown types found...")`
+  );
+
+  // Initiate the polkadot API.
+  const api = await ApiPromise.create({
+    provider: wsProvider,
+    types: {
+      // mapping the actual specified address format
+      Address: "AccountId",
```
chenzongxiong, line 48: what's the difference between "AccountId" and "MultiAddress"

I use "MultiAddress" in registrar. I tried "AccoundId", but failed.

ts-tests/scripts/smoke-test.ts
```diff
@@ -0,0 +1,287 @@
...
+    encodedPrefix.length + alice.addressRaw.length + encodedExpiredBlock.length
+  );
+  encodedMsg.set(encodedPrefix);
+  encodedMsg.set(alice.addressRaw, encodedPrefix.length);
+  encodedMsg.set(
+    encodedExpiredBlock,
+    encodedPrefix.length + alice.addressRaw.length
+  );
+
+  // To use manual hash and sign method, a prefix of \x19Ethereum ... is also needed to be prefixed
```
chenzongxiong, line 108: Those comments should be removed before merged

ts-tests/tests/utils.ts
```diff
@@ -0,0 +1,353 @@
+import 'mocha';
+
```
chenzongxiong, line 2: Please use prettier to prettify your source code.
https://prettier.io/

Some configuration, refer to litentry-registrar or push-notification repo

chenzongxiong (changes requested):

ts-tests/scripts/smoke-test.ts
h4n0: I never used MultiAddress, and this config is from the polkadot js api. Did you use the same polkadot version here?

Satoshi-Kusumoto: Multi address is a new type of address that supports also say ethereum like address and many others.
Check https://github.com/substrate-developer-hub/substrate-node-template/pull/130

Satoshi-Kusumoto: We also tried to introduce it in our lates parachain code #6

h4n0:
This was done together with substrate 3.0 upgrade. Closed.

PR #11, opened 2021-02-11
Add litentry chain spec on rococo relay chain

chenzongxiong:
For parachain to be listed on polkadot js app, we need to customize our parachain id and parachain name. Otherwise, it will use default chain type **Local**


Satoshi-Kusumoto (commented):
May you explain what's the difference with the origin `get_chain_spec` except these names.
Besides, for a online test net, we  would like to use some preconfigured accounts instead of these test accounts

chenzongxiong:
>  Besides, for a online test net, we would like to use some preconfigured accounts instead of these test accounts

For to connect to rococo relay chain, we need make chain type `ChainType::Live`. Otherwise, we queried the `system.name` on the chain, it will show `Local Testnet` by default, not `Litentry Parachain Collator`, which would pollute the name system in polkadotjs app.  

I agree to configure some accounts for some special purposes. 
However, the rococo relay chain is still a testnet, it should be fine to add these accounts. 
I'd like to introduce kusama and polkadot configuration in the future. In that chain, we need to consider that account carefully.

Also, we need refactor this part source.

Satoshi-Kusumoto (approved):

suinuj (approved):

h4n0 (approved):

Issue #24, opened 2021-03-04
Runtime hooks on_initialize() on_finalize() and offchain_worker() are not executed in every block's lifecycle


h4n0:
Currently, runtime hooks on_initialize() on_finalize() and offchain_worker() are not executed in a deterministic way. For some blocks, on_initialize is executed and for others not. on_finalize and offchain_worker are however never executed. 

Both functions work perfectly on litentry node, however.

This issue is still under investigation.

h4n0:
After consultation from Basti, I found the on_* functions are running correctly and the printed logs that I observed are wrong. It's just the parachain node (NOT collator node) is not executing with native build. Change to native build makes on_* functions work as expected.

offchain_worker() function is not executed because offchain worker functionality is not activated in runtime (`node/service.rs`). This will be fixed with a PR.

PR #26, opened 2021-03-09
Bugfix of offchain worker not properly executed on issue #24

h4n0:
Bugfix of offchain worker not properly executed on issue #24


suinuj (approved):

node/src/service.rs
```diff
@@ -147,6 +147,12 @@ where
 	let telemetry_span = TelemetrySpan::new();
 	let _telemetry_span_entered = telemetry_span.enter();
 
+  if parachain_config.offchain_worker.enabled {
```
geastwood, line 150: indentation seems off.

h4n0: Checked locally, it's tab vs spaces issue. Will write an issue separately about this.

h4n0: #27  editorconfig support needs to be added

ts-tests/tests/utils.ts
```diff
@@ -188,10 +188,11 @@ export async function launchLitentryNode(
     `--ws-port`,
     `9844`,
     `--alice`,
-    `--`,
     `--execution`,
-    //`wasm`,
     `native`,
+    `--`,
```
Satoshi-Kusumoto, line 193: what's the difference with the original one?

PR #30, opened 2021-03-12
Update cumulus

Satoshi-Kusumoto:
To latest cumulus and to Polkadot 0.8.29

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -1,3 +1,6 @@
+use cumulus_client_consensus_relay_chain::{
+	build_relay_chain_consensus, BuildRelayChainConsensusParams,
+};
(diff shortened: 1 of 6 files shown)


node/src/service.rs
```diff
@@ -147,11 +150,11 @@ where
 	let telemetry_span = TelemetrySpan::new();
 	let _telemetry_span_entered = telemetry_span.enter();
 
-  if parachain_config.offchain_worker.enabled {
```
h4n0, line 150: Why commented here? This would cause the offchain worker not activated properly.

h4n0 (changes requested):

h4n0 (approved):

PR #31, opened 2021-03-13
Add remote trigger support for integration test

h4n0:
This commit adds support of remote CI trigger from litentry-pallets


.github/workflows/build_test.yml
```diff
@@ -26,6 +28,8 @@ jobs:
         override: true
     - name: Build relay node and parachain
       run: |
+        cargo update -p pallet-account-linker
```
Satoshi-Kusumoto: So we always test with latest pallet
And the trigger is from where? Intergration test here or in pallet

Satoshi-Kusumoto (dismissed):

h4n0:
This is not needed because #39 has been approved as our strategy of handling dependencies.

PR #51, opened 2021-04-16
Bump dep

Satoshi-Kusumoto:


runtime/src/lib.rs
```diff
@@ -625,7 +710,12 @@ construct_runtime! {
 		Proxy: pallet_proxy::{Pallet, Call, Storage, Event<T>},
 		Recovery: pallet_recovery::{Pallet, Call, Storage, Event<T>},
 		ParachainInfo: parachain_info::{Pallet, Storage, Config},
-		XcmHandler: cumulus_pallet_xcm_handler::{Pallet, Call, Event<T>, Origin},
+
+		// XCM helpers.
+		XcmpQueue: cumulus_pallet_xcmp_queue::{Pallet, Call, Storage, Event<T>},
+		PolkadotXcm: pallet_xcm::{Pallet, Call, Event<T>, Origin},
+		CumulusXcm: cumulus_pallet_xcm::{Pallet, Origin},
+		Spambot: cumulus_ping::{Pallet, Call, Storage, Event<T>} = 99,
```
suinuj, line 718: why just assign this module as value 99?

suinuj (approved):

runtime/src/lib.rs
Satoshi-Kusumoto: Copied from cumulus repo. Need to check

h4n0 (approved):

runtime/src/lib.rs
chenzongxiong: I think `99` can be safely removed. Since the `construct_time` will be expanded as an `enum` along with other functions

refer to 
https://www.notion.so/litentry/Some-notes-db4957f0122f45c2a5d872eabaa9e6cb#ca195cc6d1644c858b0aafa8a30e8346

PR #53, opened 2021-06-09
V0.9.4

Satoshi-Kusumoto:


runtime/src/lib.rs
```diff
@@ -754,12 +755,12 @@ construct_runtime! {
...
 		TransactionPayment: pallet_transaction_payment::{Pallet, Storage},
 		Treasury: pallet_treasury::{Pallet, Call, Storage, Config, Event<T>},
-		Democracy: pallet_democracy::{Pallet, Call, Storage, Config, Event<T>},
+		Democracy: pallet_democracy::{Pallet, Call, Storage, Config<T>, Event<T>},
 		Council: pallet_collective::<Instance1>::{Pallet, Call, Storage, Origin<T>, Event<T>, Config<T>},
 		TechnicalCommittee: pallet_collective::<Instance2>::{Pallet, Call, Storage, Origin<T>, Event<T>, Config<T>},
 		Proxy: pallet_proxy::{Pallet, Call, Storage, Event<T>},
 		Recovery: pallet_recovery::{Pallet, Call, Storage, Event<T>},
-		ParachainSystem: cumulus_pallet_parachain_system::{Pallet, Call, Storage, Inherent, Event<T>, ValidateUnsigned} = 20,
+		ParachainSystem: cumulus_pallet_parachain_system::{Pallet, Call, Config, Storage, Inherent, Event<T>, ValidateUnsigned} = 20,
```
chenzongxiong, line 761: can we remove magic number, like `20`, `21`, `30`

runtime/Cargo.toml
```diff
@@ -11,60 +11,62 @@ serde = { version = "1.0.101", optional = true, features = ["derive"] }
...
-sp-core = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.3" }
-sp-session = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.3" }
-sp-offchain = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.3" }
-sp-block-builder = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.3" }
-sp-transaction-pool = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.3" }
-sp-inherents = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.3" }
-sp-consensus-aura = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.3" }
+sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.4" }
+sp-api = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.4" }
```
chenzongxiong, line 15: I'm curious about the `default-features` flag. What is the difference between `true` and `false`

runtime/src/lib.rs
Satoshi-Kusumoto: We need to figure out what they are. These are from upstream

runtime/Cargo.toml
Satoshi-Kusumoto: Because `default = ['std']` It means to compile without `std` feature

suinuj (approved):

geastwood (commented):
This is all manual work yeh? We have to keep up with the upstream and manually keep track of these changes?

Satoshi-Kusumoto:
> This is all manual work yeh? We have to keep up with the upstream and manually keep track of these changes?

Yes, basically to do better we even need to keep track of those important upstream PRs

PR #61, opened 2021-07-19
bump deps to v0.9.8

Satoshi-Kusumoto:


runtime/src/lib.rs
```diff
@@ -117,8 +116,8 @@ const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
 /// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used
 /// by  Operational  extrinsics.
 const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
-/// We allow for 2 seconds of compute with a 6 second average block time.
-const MAXIMUM_BLOCK_WEIGHT: Weight = 2 * WEIGHT_PER_SECOND;
+/// We allow for .5 seconds of compute with a 12 second average block time.
+const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND / 2;
```
suinuj, line 120: don't understand the update here, why the block time becomes longer, but time for block weight to be less.

Satoshi-Kusumoto:
```
pub const MILLISECS_PER_BLOCK: u64 = 12000;
```
This is already updated and the latter part of the comment was just obsolete

Issue #63, opened 2021-07-29
Integrate NFT pallet into parachain


h4n0:
Deadline: Aug. 4

**Acceptance Criteria**
- [x] NFT pallet can be built with parachain 
- [x] (WONT DO) Integration test should be added for NFT testing (This can be done later for not blocking frontend devs)
- [x] Deployment

chenzongxiong:
@Satoshi-Kusumoto Is the first task finished? If so, please let me know.

PR #64, opened 2021-08-02
add nft pallet

Satoshi-Kusumoto:
Currently we use not the latest version of NFT pallet (deposit upon class creation is missing) since it seems requires some refactor of the code to achieve this.
Part of #63

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -740,6 +740,25 @@ impl pallet_aura::Config for Runtime {
 	type AuthorityId = AuraId;
 }
 
+parameter_types! {
+	pub MaxClassMetadata: u32 = 1024;
+	pub MaxTokenMetadata: u32 = 1024;
+}
+
+impl orml_nft::Config for Runtime {
+	type ClassId = u32;
+	type TokenId = u64;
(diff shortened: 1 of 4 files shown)


runtime/src/lib.rs
```diff
@@ -740,6 +740,25 @@ impl pallet_aura::Config for Runtime {
...
+}
+
+impl orml_nft::Config for Runtime {
+	type ClassId = u32;
+	type TokenId = u64;
+	type ClassData = pallet_nft::ClassData<BlockNumber, Self::ClassId>;
+	type TokenData = pallet_nft::TokenData;
+	type MaxClassMetadata = MaxClassMetadata;
+	type MaxTokenMetadata = MaxTokenMetadata;
+}
```
h4n0, line 755: What is the reason of adding orml nft into runtime?

suinuj: have the same question.

suinuj: need its configuration?

suinuj (approved):

runtime/src/lib.rs
Satoshi-Kusumoto: It is a dependency of pallet NFT

PR #70, opened 2021-08-06
Integration test refactoring

### LINKED ISSUE #59: Refactor source code in ts-tests
The source code in `ts-test` is messy.  We need to refactor it for easier maintenance.

1. putting testing code in a standalone source file
2. reading all sensitive information from environment variables for encrypted files instead of putting them in source code
3. removing useless comment lines.
4. resolving configuration error, e.g. changing parachain id from 1984 to 2022

h4n0:
Refactor integration test code based on #59


ts-tests/tests/utils.ts
```diff
@@ -6,20 +6,25 @@ import { Bytes } from "@polkadot/types";
...
 export const ROCOCO_LOCAL_PATH = `./rococo-local-cfde-real-overseer.json`;
 export const RELAY_NODE_SCRIPT = `./scripts/start-alice-and-bob.sh`;
 export const SPAWNING_TIME = 30000;
 
 // OCW account
-export const OCR_ACCOUNT = "5FEYX9NES9mAJt1Xg4WebmHWywxyeGQK8G3oEBXtyfZrRePX";
+const ocwAccount = DEFAULT_CONFIG.ocw_account;
 
 // Provider is set to localhost for development
 const wsProvider = new WsProvider("ws://localhost:9844");
```
chenzongxiong: Make websocket endpoint configuration

chenzongxiong (changes requested):
Remove `launchRelayNodesAndParachainRegister` from nodejs script. 
Should put in an individual file. 
Only transaction tests need

chenzongxiong (approved):

.gitignore
```diff
@@ -2,10 +2,11 @@
 .idea
 .vscode
 .DS_Store
-/.cargo/config
+.cargo/config
 
 **/node_modules/
-**/para-2022*
+**/para-*
 ts-tests/rococo-local-cfde-real-overseer.json
```
Kailai-Wang: I think this json is not used anymore? Can we remove it from repo

ts-tests/.env.example
```diff
@@ -0,0 +1 @@
+NODE_ENV=<`development` or `staging` or `production`>
```
Kailai-Wang: should this be aligned with the actual switch case?
e.g.
```suggestion
NODE_ENV=<`development` or `staging` or `test` or `ci`>
```

ts-tests/README.md
```diff
@@ -4,16 +4,16 @@ This node.js project aims to test and verify Litentry Runtime as a whole, includ
...
 
-`npm i`
+`yarn`
 
 ## Run
 
 Run tests with command
 
-`npm test`
+`yarn test`
```
Kailai-Wang: I think it will not work: here should be exactly same as the flows in CI settings

ts-tests/scripts/clean-up.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+killall litentry-token-server
+killall polkadot
+killall litentry-collator
```
Kailai-Wang, line 10: maybe remove the logs in /tmp : ideally conditionally -- only when the tests go through.  Not sure if it's feasible though (via $?)

ts-tests/tests/utils.ts
```diff
@@ -1,396 +1,133 @@
...
-  ];
-  const binary = spawn(cmd, args_main_node);
+async function initApiPromise(config: any) {
+    console.log(`Initiating the API (ignore message "Unable to resolve type B..." and "Unknown types found...")`);
+    // Provider is set for parachain node
+    const wsProvider = new WsProvider(config.parachain_ws);
+
+    // Initiate the polkadot API.
+    const api = await ApiPromise.create({
+        provider: wsProvider,
```
Kailai-Wang, line 35: just a understanding problem for me: 
does `await ApiPromise.create` here acutally wait for the accessibility of parachain-ws, or is it just creating an object? I.e. are we making sure that parachains are reachable till now? (`sleep 3` in sh in not the most reliable..)

.github/workflows/build_test.yml
```diff
@@ -38,6 +38,9 @@ jobs:
...
         cd ts-tests
-        npm install 
-        npm test
-
+        echo "NODE_ENV=ci" > .env
+        source ./scripts/start-relay-and-para-chain.sh
+        source ./scripts/start-token-server.sh
+        yarn && yarn register-parachain
+        yarn test
+        source ./scripts/clean-up.sh
```
Kailai-Wang: Great that we separate the initilization of bash script from the core JS logic.
I'd prefer "sh" over "source" though (conventionally, unless there're some exported ENVs which will be transfered across scripts), especially when we have an explicit .env here

chenzongxiong: can you update it. 
Also change `#!/bin/bash` from `#!/bin/sh` in `cleanup.sh`

ts-tests/.env.example
chenzongxiong: Exactly. I add different configurations to our repository. 

Imagine in staging environment, we only need to pull the `ts-tests` repo, and run the following commands

```
echo NODE_ENV=staging > .env
yarn && yarn test
```

As for pulling the `ts-tests`, it can be done by `rundeck` scheduler.

ts-tests/README.md
chenzongxiong: For README.md, I didn't check it carefully. Can you update it?

ts-tests/scripts/clean-up.sh
chenzongxiong: If the tests failed, I think the pipeline will be broken and `clean-up.sh` will never be executed. @h4n0 Is it right?

Redirecting `stdout` to `/tmp/xxxx` can help us debug. And the file size won't increase dramatically since the processes will be killed when `ci-test` is done.

But it's ugly and it may also kill other processes on `ci` server.  I think this should be enhanced.

.gitignore
chenzongxiong: I removed the `rococo-local-cfde-real-overseer.json`, didn't update `.gitignore`. 

I add `.gitignore` in `ts-tests` folder.

ts-tests/tests/utils.ts
chenzongxiong: Yeah. Since `ApiPromise.create` return a promise, not `ApiPromise` handler. 
Check the source code `@polkadot/api/`

.github/workflows/build_test.yml
chenzongxiong: I use `dotenv` to manage `.env`, doesn't affect shell scripts.

ts-tests/scripts/clean-up.sh
Kailai-Wang: I think it's worth a discussion @chenzongxiong  @h4n0 

By default scripts are run as `set -eo pipefail` in github CI, but then the binaries won't be killed if clean-up.sh is not run.

Currently I tried to take full control back to avoid fail early so that:

- it kills the binaries in all cases, regardless of failed or ok tests.
- remove the logs if tests were successful
- keep the logs and print the log-directory if tests failed => but it's rather for local repro-case. In CI the logs are printed anyway.

**But maybe it's better to achieve this using pure CI syntax.** => Todo

And I agree that killall is not the cleanest solution.
Maybe I could have kept the PID of each binary process and only kills them later.

Kailai-Wang (approved):

ts-tests/scripts/start-relay-and-para-chain.sh
```diff
@@ -0,0 +1,26 @@
...
+$POLKADOT_BIN build-spec --chain rococo-local --disable-default-bootnode --raw > $ROCOCO_CHAINSPEC
+
+# generate genesis head and wasm validation files
+$LITENTRY_BIN export-genesis-state --parachain-id $PARACHAIN_ID > para-$PARACHAIN_ID-genesis
+$LITENTRY_BIN export-genesis-wasm > para-$PARACHAIN_ID-wasm
+
+# run alice and bob as relay nodes
+$POLKADOT_BIN --chain $ROCOCO_CHAINSPEC --alice --tmp --port 30333 --ws-port 9944 &> "$1/relay.alice.log" &
+sleep 3
+$POLKADOT_BIN --chain $ROCOCO_CHAINSPEC --bob --tmp --port 30334 --ws-port 9945  &> "$1/relay.bob.log" &
```
chenzongxiong: I'd like to alias a name for `$1`

e.g. 
```
tmp_prefix=$1
```
then
```
$tmp_prefix/relay.alice.log
```

It's much easier for us to understand the source code.

chenzongxiong (approved):

ts-tests/scripts/clean-up.sh
chenzongxiong: Yeah. I agree. I think we can create an issue for this problem

ts-tests/scripts/start-relay-and-para-chain.sh
Kailai-Wang: sure it makes sense. I'll update it

.gitignore
```diff
@@ -2,12 +2,9 @@
...
-/.cargo/config
+.cargo/config
 
 source/
 
 **/node_modules/
-**/para-2022*
-ts-tests/rococo-local-cfde-real-overseer.json
-
-tags.lock
```
h4n0, line 13: Why was tags.lock removed? If is a misoperation, please add this back.

ts-tests/scripts/run-test.sh
```diff
@@ -0,0 +1,17 @@
...
+cd "$basedir"
+
+cd ..
+
+tmpdir=$(mktemp -d /tmp/tmp.XXXXXX)
+
+echo "NODE_ENV=ci" > .env
+./scripts/start-relay-and-para-chain.sh "$tmpdir"
+./scripts/start-token-server.sh "$tmpdir"
+yarn && yarn register-parachain 2>&1 | tee "$tmpdir/register-parachain.log"
```
h4n0, line 15: Are there some specific reasons that npm is replaced by yarn?

ts-tests/scripts/clean-up.sh
h4n0: This is a good point. As my understanding `set -o pipefail` won't terminate the script but just return error after the script runs to the end. 
Best way to implement all these binaries is to spawn them in TS code and they will be killed automatically when the main thread is killed. However due to some limitations we did not achieve this.

ts-tests/.env.example
```diff
@@ -0,0 +1 @@
+NODE_ENV=<`development` or `staging` or `test` or `ci`>
```
h4n0, line 1: How can you guarantee the NODE_ENV on CI server is ci? Should we set it up manually?

chenzongxiong: https://github.com/litentry/litentry-parachain/blob/a1763aff06c6bf22cb4e0609bc664c469ee811e1/ts-tests/scripts/run-test.sh#L12

ts-tests/scripts/run-test.sh
chenzongxiong: `yarn` can fix nested version confliction, `npm` cannot.

Like `pip` vs. `conda`

.gitignore
Kailai-Wang: re-added.

Kailai-Wang (approved):

ts-tests/scripts/clean-up.sh
chenzongxiong: @Kailai-Wang Can you create an issue for me ? I'd like to use `pidfile` instead of `killall`

PR #112, opened 2021-09-13
Introduce docker-build of our parachain

### LINKED ISSUE #79: provide Tooling and APIs for third party users
We need to think through how third party users / normal customers could use our parachains more easily and user-friendly:

- [x] parachain binary compilation (inside docker)
- [x] build and push docker image  (like [what KILT did](https://github.com/KILTprotocol/local-parachain-setup))
- [x] provide docker-compose file to launch local relaychain + parachain networks in one click (e.g. using [parachain-launch)](https://github.com/open-web3-stack/parachain-launch)
- [x] adjust current CI: remove polkadot submodule and use docker to launch our network for testing

Kailai-Wang:
Resolves #79 

This PR
- add scripts to build and push docker-image of our parachain 
- add scripts to launch local dev network using parachain-launch
- adjust GitHub-actions to utilize docker-images for CI tests


.github/workflows/build_and_run_test.yml
```diff
@@ -37,23 +33,41 @@ jobs:
...
+        [ -z "$(docker images --filter=dangling=true -q)" ] || docker rmi -f $(docker images --filter=dangling=true -q)
+        docker images
+        docker push -a litentry/litentry-parachain
+
     - name: Clean up
       if: ${{ always() }}
       run: |
         tmpdir=${{ steps.set-tmpdir.outputs.tmpdir }}
-        ts-tests/scripts/clean-up.sh $tmpdir
+        docker images
```
chenzongxiong: I'd like to put these commands in a `cleaup.sh` script.

.github/workflows/build_and_run_test.yml
```diff
@@ -37,23 +33,41 @@ jobs:
...
+        cd docker/generated
+        docker-compose rm -f -s -v
+        docker volume ls | grep generated_ | sed 's/local *//' | xargs docker volume rm
+        [ -z "$(docker images --filter=dangling=true -q)" ] || docker rmi -f $(docker images --filter=dangling=true -q)
+        docker rmi -f $(docker images litentry/litentry-parachain --format "{{.Repository}}:{{.Tag}}" | grep -v latest)
```
chenzongxiong: Doesn't have to remove image since we can reuse them in this future building. It can save time since we don't have to push the same image everytime

.github/workflows/build_and_run_test.yml
```diff
@@ -37,23 +33,41 @@ jobs:
...
       with:
         name: litentry-parachain-debug.log
         path: ${{ steps.set-tmpdir.outputs.tmpdir }}/**/*
         retention-days: 7
 
+    - name: Push docker image if test passes
+      if: ${{ success() }}
+      run: |
+        [ -z "$(docker images --filter=dangling=true -q)" ] || docker rmi -f $(docker images --filter=dangling=true -q)
+        docker images
```
chenzongxiong, line 57: Should we need this command?

.github/workflows/build_and_run_test.yml
```diff
@@ -15,20 +15,16 @@ jobs:
...
+
+    - name: Build docker image
       run: |
-        cargo build --release
-        cd polkadot
-        cargo build --release
+        docker images
+        cd docker
+        ./build.sh
         cd ..
```
chenzongxiong: Actually, I suspect `cd ..` can also be removed.

.github/workflows/build_and_run_test.yml
```diff
@@ -15,20 +15,16 @@ jobs:
...
-        target: wasm32-unknown-unknown
-        override: true
-    - name: Build relay node and parachain
+
+    - name: Build docker image
       run: |
-        cargo build --release
-        cd polkadot
-        cargo build --release
+        docker images
```
chenzongxiong: I don't know why you need to run `docker images` before `build.sh`. For debugging ?

docker/generated/docker-compose.yml
```diff
@@ -0,0 +1,88 @@
...
+      context: .
+      dockerfile: relaychain.Dockerfile
+    command:
+      - --base-path=/data
+      - --chain=/app/rococo-local.json
+      - --validator
+      - --ws-external
+      - --rpc-external
+      - --rpc-cors=all
+      - --name=bob
```
chenzongxiong, line 48: `--bob` is `validator bob`, so we can remove `--validator` here.

chenzongxiong (changes requested):

docker/generated/docker-compose.yml
Kailai-Wang: > `--bob` is `validator bob`, so we can remove `--validator` here.

this docker-compose.yml is generated automatically
plus I think --bob just makes bob a validator, but --validator starts the node as validator, so they are different args IMHO.

.github/workflows/build_and_run_test.yml
Kailai-Wang: > I don't know why you need to run `docker images` before `build.sh`. For debugging ?

mostly yes, `build.sh` will pull base images and build images that might result in dangling images, I wanted to know stats before it

.github/workflows/build_and_run_test.yml
Kailai-Wang: > Should we need this command?

otherwise it would try to push images with `<none>` tag too :(

.github/workflows/build_and_run_test.yml
Kailai-Wang: > Actually, I suspect `cd ..` can also be removed.

could be..
I think I'll try `working-directory` keyword

.github/workflows/build_and_run_test.yml
Kailai-Wang: > Doesn't have to remove image since we can reuse them in this future building. It can save time since we don't have to push the same image everytime

it actually keeps `litentry/litentry-parachain:latest` for further usage (while removing other tagged versions)

.github/workflows/build_and_run_test.yml
Kailai-Wang: > I'd like to put these commands in a `cleaup.sh` script.

I can do that

.github/workflows/build_and_run_test.yml
Kailai-Wang: I removed it in latest commit

chenzongxiong (approved):

h4n0 (approved):

PR #116, opened 2021-09-27
Add collator-related pallets to runtime and adjust chain specs

### LINKED ISSUE #60: [Genesis] Fix tokenDecimals & tokenSymbol in system.properties
There is no **token-related** information displayed on **subscan** since the response of the ` system. properties` is empty. We need to define the following metadata.

https://litentry.subscan.io/

image

### LINKED ISSUE #93: Add collation related pallets to runtime
- [x] Add collator selection, aura, aura-ext and authorship pallets into runtime
- [x] Chain spec

### LINKED ISSUE #113: Update CI for collation parachain
Since we have collator instead of a single dev node now

Kailai-Wang:
resolves #60
resolves #93 
resolves #113 

This PR:
- adds collator-related pallets to runtime
- categorizes chain specs into dev/staging/prod
- adjusts genesis config
- hardcodes staging accounts/keys
- import properties into chain spec
- adjusts CI so that it passes
- cargo fmt


node/src/command.rs
```diff
@@ -150,27 +152,27 @@ pub fn run() -> Result<()> {
 		Some(Subcommand::BuildSpec(cmd)) => {
 			let runner = cli.create_runner(cmd)?;
 			runner.sync_run(|config| cmd.run(config.chain_spec, config.network))
-		},
+		}
```
Satoshi-Kusumoto, line 155: why these commas are removed?

node/src/command.rs
```diff
@@ -242,8 +244,8 @@ pub fn run() -> Result<()> {
 			}
 
 			Ok(())
-		},
-		Some(Subcommand::Benchmark(cmd)) =>
+		}
+		Some(Subcommand::Benchmark(cmd)) => {
```
Satoshi-Kusumoto, line 248: seems you ran some ftm commands?

Satoshi-Kusumoto (approved):
LGTM
I am just thinking that, do we need a single node dev net?

.github/workflows/build_and_run_test.yml
```diff
@@ -11,14 +11,11 @@ env:
 
 jobs:
   build-test:
-
     runs-on: self-hosted
 
     steps:
     - name: Checkout codes
       uses: actions/checkout@v2
-      with:
```
Satoshi-Kusumoto, line 20: Just wondering why is this removed? A thing that should have been done in the last PR?

Kailai-Wang:
> LGTM
> I am just thinking that, do we need a single node dev net?

I think we do already.
With `docker-compose` you will start a single node dev locally with 2 relaychain nodes + 1 parachain node

.github/workflows/build_and_run_test.yml
Kailai-Wang: > Just wondering why is this removed? A thing that should have been done in the last PR?

A bit long story.
- Last PR removed submodules "polkadot", it was fine to keep that `with` statement as we don't have submodules anyway.
- However in this branch, CI always complains about [this error](https://github.com/litentry/litentry-parachain/actions/runs/1278826837) although no submodules exist in the repo.
- I tried to fix this by removing the `with` statement and .gitmodules files completely.
- The error still persisted, it turns out to be a github bug/unpleasant thing that actions still preserve the previously checked-out stuff which can cause in consistency, despite of `clean` setting to true.
   https://github.com/AutoModality/action-clean
   https://github.community/t/how-to-properly-clean-up-self-hosted-runners/128909/2
- The other solutions didn't work, so eventually the directory was removed manually and the commits weren't reverted.

See all my committed history and related actions :(

node/src/command.rs
Kailai-Wang: > seems you ran some ftm commands?

yes `cargo fmt`.
I mentioned it the in PR description now.

node/src/command.rs
Kailai-Wang: `cargo fmt`

docker/generated/docker-compose.yml
```diff
@@ -74,7 +74,7 @@ services:
       - --force-authoring
       - --execution=wasm
       - --alice
-      - --node-key=c02aab73e4dd1bd1553f0458101c6c227d15b65dae6fe3980f652149d4a59e86
+      - --node-key=cda304f4e14c25bd1444c567511f42b52de1299efec2cd522db97245884b32c0
```
chenzongxiong, line 77: I prefer to removing all generated metadata and generating that info in CI steps.

runtime/src/lib.rs
```diff
@@ -277,6 +277,17 @@ impl pallet_timestamp::Config for Runtime {
 	type WeightInfo = ();
 }
 
+parameter_types! {
+	pub const UncleGenerations: u32 = 0;
```
chenzongxiong, line 281: what's uncle generations? 
will it affect the number of leaves on grafana dashboard ?

docker/generated/docker-compose.yml
Kailai-Wang: > I prefer to removing all generated metadata and generating that info in CI steps.

I prefer that too, which is planned but CI is then required to push those changes as well.
This needs to be done in the future.
I created an issue for it #117 

Or, these files could be generated completely on the fly and never committed into repo. This is easier, but when the user wants to start a dev network locally, he/she has to generate these files first.

h4n0 (approved):

runtime/src/lib.rs
Kailai-Wang: will check them later, it's rather runtime settings which are not addressed in this PR

PR #136, opened 2021-10-08
Add benchmark_metadata

### LINKED ISSUE #135: Benchmark build failed
For new version Benchmark, need implement the benchmark_metadata, otherwise benchmark build be failed.

suinuj:
For the new version Benchmark, need to implement the benchmark_metadata, otherwise, the benchmark build is failed.

resolve #135

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -935,6 +935,21 @@ impl_runtime_apis! {
 
 	#[cfg(feature = "runtime-benchmarks")]
 	impl frame_benchmarking::Benchmark<Block> for Runtime {
+		fn benchmark_metadata(extra: bool) -> (
+			Vec<frame_benchmarking::BenchmarkList>,
+			Vec<frame_support::traits::StorageInfo>,
+		) {
+			use frame_benchmarking::{list_benchmark, Benchmarking, BenchmarkList};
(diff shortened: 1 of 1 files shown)


runtime/src/lib.rs
```diff
@@ -935,6 +935,21 @@ impl_runtime_apis! {
...
+		fn benchmark_metadata(extra: bool) -> (
+			Vec<frame_benchmarking::BenchmarkList>,
+			Vec<frame_support::traits::StorageInfo>,
+		) {
+			use frame_benchmarking::{list_benchmark, Benchmarking, BenchmarkList};
+			use frame_support::traits::StorageInfoTrait;
+			use frame_system_benchmarking::Pallet as SystemBench;
+
+			let mut list = Vec::<BenchmarkList>::new();
+			list_benchmark!(list, extra, frame_system, SystemBench::<Runtime>);
```
Kailai-Wang, line 947: we might have to add the real pallets later..

Kailai-Wang (approved):

runtime/src/lib.rs
suinuj: yes, I have another task to investigate the pallets that should be included in the benchmark list.

PR #149, opened 2021-10-15
Improve CI github actions

### LINKED ISSUE #133: Improvement of current CI/github actions
Some suggestions to improve the current CI, following https://www.notion.so/litentry/Some-revamp-on-CI-85c43e226ba64376b057cbc4cc6d84d6

- [x] use Makefile to simplify executing things from command-line, e.g. like [Acala](https://github.com/AcalaNetwork/Acala/blob/master/Makefile) or [Bifrost](https://github.com/bifrost-finance/bifrost/blob/develop/Makefile)
- [x] allow commits to skip CI with some keywords, e.g.: `if: "!contains(github.event.commits[0].message, '[skip ci]')"`
- [x] push docker-images to hub only for changes (commits) in `dev`, so that `litentry/litentry-parachain:latest` on hub always have the latest stable version of our parachain CI in a separate feature branch will **not** trigger a docker-image push.
- [x] consider to use concurrency to dismiss queued CI (e.g. for a series of commits in a short time, only the last commit should trigger CI)
- [x] ~~if applicable, group dependabot PRs to avoid flooding, see official discussions [here](https://github.com/dependabot/dependabot-core/issues/1190).~~
 Won't do, upstream doesn't have a solution yet and we are not having excessive amount of dependabot PRs => not worth the time.

Kailai-Wang:
resolves #133 

This PR:
- use Makefile to simplify running commands
- only push to docker hub when CI passes and the event is "push to dev"
- allow `[SKIP CI]` keyword in commit message to skip CI
- use `concurrency` keyword to only run CI for the last commit
- add an auto_label_pr GA to help labeling when creating PR
- tidy up: all scripts are located in scripts/ now


geastwood:
This is very interesting, by skiming through the changes, I certainly learned couple tricks. Would be great to spread the knowledge, giving some good resources to read for the team.

@Kailai-Wang @litentry/engineers

Kailai-Wang:
> This is very interesting, by skiming through the changes, I certainly learned couple tricks. Would be great to spread the knowledge, giving some good resources to read for the team.
> 
> @Kailai-Wang @litentry/engineers

sure, will mention these too in the talk next week

.github/workflows/auto_label_pr.yml
```diff
@@ -0,0 +1,22 @@
...
+        uses: andymckay/labeler@master
+        if: github.event.pull_request.draft == true
+        with:
+          remove-labels: 'S1-pleasereview'
+
+      - name: Label PRs
+        uses: andymckay/labeler@master
+        if: github.event.pull_request.draft == false
+        with:
+          add-labels: 'S1-pleasereview'
```
h4n0: Maybe reserve the new line here?

.github/workflows/auto_label_pr.yml
```diff
@@ -0,0 +1,22 @@
+# inspired by polkadot
```
h4n0, line 1: Just a question:
This means if draft is set for new PR, no S1-pleasereview and if draft is not set for new PR, the label S1-pleasereview will be tagged?

Makefile
```diff
@@ -1,82 +1,97 @@
...
 build-benchmark:
 	cargo build --features runtime-benchmarks --release
 
-build-spec:
-	./target/release/litentry-collator build-spec --chain dev --disable-default-bootnode > ./source/local.json
+## test
+
+.PHONY: test-all
+test-all:
+	cargo test
```
h4n0: The test-all here sounds a bit confusing. I expected to run all available tests here. Also, do we need to do cargo test on parachain?

h4n0 (changes requested):

.github/workflows/auto_label_pr.yml
Kailai-Wang: > Just a question: This means if draft is set for new PR, no S1-pleasereview and if draft is not set for new PR, the label S1-pleasereview will be tagged?

yes, or when a PR is converted to draft this will be triggered too.

Makefile
Kailai-Wang: > The test-all here sounds a bit confusing. I expected to run all available tests here. Also, do we need to do cargo test on parachain?

I think the name is indeed a bit confusing, I could change it to `test-cargo-all`
As for integration tests, if they have to be executed in every CI, then they should be included in `test-ci`.
I believe we use to have some parachain tests with pallets.
We could discuss a bit on this.

chenzongxiong (approved):

chenzongxiong:
It looks good to me.  I think that adding additional command `make help` would be more elegant

.github/workflows/auto_label_pr.yml
Kailai-Wang: Adjusted

h4n0 (approved):

PR #162, opened 2021-10-18
generate dev docker-compose files on the fly

### LINKED ISSUE #117: Configure github actions to push generated files automatically
Currently, with the usage of parachain-launch in CI, whenever there's a change that could affect the generated chain-spec (e.g. modifying genesis config, different runtime), we have to manually generate the files in `docker/generated/` and push them to avoid inconsistency.

The ideal procedure is that github actions could generate those files on the fly and push them after CI passes (same time as pushing docker images).

Kailai-Wang:
resolves #117 
The title is self-explanatory


chenzongxiong (commented):
I'd suggest removing `docker/generated-dev/*` folders in git since it generates automatically. And for development, we need to keep the latest environment, even it's unstable

Kailai-Wang:
> I'd suggest removing `docker/generated-dev/*` folders in git since it generates automatically. And for development, we need to keep the latest environment, even it's unstable

I think in this PR it is indeed removed :D

chenzongxiong (approved):

PR #171, opened 2021-10-20
add GA to build docker with features

### LINKED ISSUE #170: Provide docker image with runtime-benchmarks
prerequisite for https://github.com/litentry/research-force/issues/67

We need to provide a docker image which is built with `runtime-benchmarks` feature enabled.

Kailai-Wang:
closes #170 

This PR adds a github action to build docker image with a list of features.
The GA will be triggered manually.


docker/Dockerfile
```diff
@@ -2,11 +2,13 @@ FROM paritytech/ci-linux:production as builder
...
 
 ARG PROFILE=release
+ARG FEATURES
 WORKDIR /litentry-parachain
 
 COPY . /litentry-parachain
 
-RUN cargo build --$PROFILE
+# empty features (--features=) will be ignored
+RUN cargo build --$PROFILE --features=$FEATURES
```
chenzongxiong, line 11: If  `$FEATURES` is empty, then the command reduces to `cargo build --$PROFILE`, right?

.github/workflows/build_docker_with_features.yml
```diff
@@ -0,0 +1,45 @@
+name: Build docker with features
+
+on:
+  workflow_dispatch:
+    inputs:
+      docker_tag:
+        description: The tag for the built docker image
+        required: true
+      features:
+        description: A comma separated list for features for cargo build
```
chenzongxiong: Add reference, see `cargo build --help` for details.

chenzongxiong (approved):

docker/Dockerfile
Kailai-Wang: > If `$FEATURES` is empty, then the command reduces to `cargo build --$PROFILE`, right?

It's actually `cargo build --release --features=`, which is the same as `cargo build --release`: as far as I tested.

PR #181, opened 2021-10-25
Fix parachain launch config

### LINKED ISSUE #180: Fix incorrect parachain-launch config
Most of the startup parameters in the [current parachain-launch config ](https://github.com/litentry/litentry-parachain/blob/dev/docker/parachain-launch-config-dev.yml) was based on the [example config](https://github.com/open-web3-stack/parachain-launch/blob/master/config.yml) from parachain-launch.

It seems to be problematic, particularly the `force-authoring` for the **relaychain**, which caused that the parachain couldn't produce blocks, see e.g. this GA: https://github.com/litentry/litentry-parachain/actions/runs/1378748794

We need to fix this and review other parameters.

Kailai-Wang:
resolves #180 

This PR reviews and fixes the startup options in parachain-launch config.


docker/generated-staging/docker-compose.yml
```diff
@@ -71,14 +72,15 @@ services:
...
+      - --execution=wasm
       - --no-beefy
     environment: {}
     ulimits: *a1
   parachain-2022-0:
     ports:
-      - 9954:9944
+      - 9947:9944
       - 9936:9933
-      - 40333:30333
```
chenzongxiong, line 81: only for dev?

for staging and production, we will use 40333 as default p2p port.

chenzongxiong (approved):

docker/generated-staging/docker-compose.yml
Kailai-Wang: yep only for dev.
Ports in staging and prod are configured in ansible scripts separately.

Issue #150, opened 2021-10-15
Determine pallet (numerical) parameters


Satoshi-Kusumoto:
Check if default values works, if not, give new values. (Need to have brief reasonings)

wangminqi:
https://www.notion.so/litentry/Democracy-Parameter-Research-ea75c3218181408492509b6d7fb197b7

Democracy parameters of other projects

PR #183, opened 2021-10-26
update parameters of Multisig, Balance, Proxy and Vesting

Satoshi-Kusumoto:
According to https://www.notion.so/litentry/Initial-pallets-4fc56ea822484710a2d1028c4cb190a9
Part of #150


Kailai-Wang (approved):
I approve this one first, but it might be better if @suinuj could double check it as he fixed the parameters.

suinuj (approved):

PR #189, opened 2021-10-27
restart relaychain for ci if parachain doesn't produce blocks

### LINKED ISSUE #187: Restart mechanism to fix parachain block production in CI
It seems to be a known issue that starting multiple validators with `parachain-launch` at same time might cause no blocks being produced for parachain, which breaks our CI.

See the issue that was created by Acala:
https://github.com/paritytech/polkadot/issues/3499

and the conversation with xlc:
image

I try to introduce a restart mechanism in CI to restart the relaychain nodes sequentially if no blocks are produced within a certain time.

Kailai-Wang:
resolves #187 

This PR:
- adds a script to restart the relaychain services if no blocks are produced by parachain.
- adds a timeout for docker-build and ci-test

Tested multiple times locally.

### DIFF
--- a/.github/workflows/build_and_run_test.yml
+++ b/.github/workflows/build_and_run_test.yml
@@ -30,12 +30,14 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Build docker image
+      timeout-minutes: 40
       run: |
         ./scripts/build-docker.sh
         echo "============================="
         docker images
 
     - name: Start docker containers and run CI
+      timeout-minutes: 30
       run: |
         make test-ci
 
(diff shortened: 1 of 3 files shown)


h4n0 (approved):

geastwood:
So simply restart to node will be able to solve the issue described in #187 ?

Kailai-Wang:
> So simply restart to node will be able to solve the issue described in #187 ?

that's what I got from Acala and in my tests it actually did the trick.
As far as I understand the root cause lies upstream but it's not prioritized as it mostly happens in tests.

PR #188, opened 2021-10-27
Fee related part 1

wangminqi:
-relocate pallet-payment-interface into parachain/pallets
-implement pallet-payment-interface in parachain runtime

PR #191, opened 2021-10-28
Seperate constants

### LINKED ISSUE #143: Move runtime constants to a separate file
Currently constants, types (e.g Block) are in runtime/lib.rs

Satoshi-Kusumoto:
Move most of the basic constants and configurable constants into two separate files. Generally nothing should be changed.
Closes #143 

@wangminqi fee related things (ex. WeightToFee) should go to runtime/src/constants.rs

### DIFF
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -4,45 +4,14 @@ version = '0.1.0'
 authors = ["Litentry Dev"]
 edition = '2018'
 
-[package.metadata.docs.rs]
-targets = ["x86_64-unknown-linux-gnu"]
-
 [dependencies]
-# External dependencies
-codec = { package = "parity-scale-codec", version = "2.2", default-features = false, features = ["derive", "max-encoded-len"] }
-
 # Substrate dependencies
-sp-consensus-aura = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.11", default-features = false }
-sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.11", default-features = false }
-sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.11", default-features = false }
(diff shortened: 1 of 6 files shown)


wangminqi:
Get it. related code will be relocated after #188

primitives/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = 'primitives'
+version = '0.1.0'
+authors = ["Litentry Dev"]
+edition = '2018'
```
geastwood, line 5: What does this `2018` mean?

runtime/src/lib.rs
```diff
@@ -59,31 +66,9 @@ use xcm_builder::{
...
-pub type BlockNumber = u32;
-
 /// The address format for describing accounts.
 pub type Address = MultiAddress<AccountId, ()>;
 
-/// Block header type as expected by this runtime.
-pub type Header = generic::Header<BlockNumber, BlakeTwo256>;
-
 /// Block type as expected by this runtime.
 pub type Block = generic::Block<Header, UncheckedExtrinsic>;
```
Kailai-Wang, line 73: Will this `Block` conflict with what's defined in `primitives`, or can it be safely deleted?

primitives/Cargo.toml
Satoshi-Kusumoto: Means Rust 2018 edition, the latest one

runtime/src/lib.rs
Satoshi-Kusumoto: In primitives, it is opaque::Block, and if you look at it carefully, with `pub use sp_runtime::OpaqueExtrinsic as UncheckedExtrinsic;`, the `UncheckedExtrinsic` is not the same as the one in runtime/lib.rs. Therefore these are two different `block` (I come firmed this before making this PR)

Kailai-Wang (approved):

h4n0 (approved):

PR #195, opened 2021-10-29
feat: quick implementation of transaction fee

### LINKED ISSUE #145: Fee related stuffs implementation
- [x] has transaction fee to collators
- [x] has block rewards to collators
- [x] has admin interfaces to adjust these two rewards

wangminqi:
quick implementation of transaction fee. This closes #145 .


runtime/src/lib.rs
```diff
@@ -414,12 +423,90 @@ parameter_types! {
 	pub const OperationalFeeMultiplier: u8 = 5;
 }
 
-// TODO: same as rococo-parachain but differnet as parachain-template
+// linear ratio of transaction fee distribution
+struct RatioOf {
+	treasury: u32,
+	author: u32,
+	burned: u32,
+}
```
Satoshi-Kusumoto: Should use 3 consts and assign their value here

runtime/src/lib.rs
```diff
@@ -414,12 +423,90 @@ parameter_types! {
...
+
+// It is recommended to set sum of ratio to 100, yet only decimal loss is concerned.
+impl Default for RatioOf {
+	fn default() -> Self {
+		RatioOf { treasury: 0, author: 0, burned: 100 }
+	}
+}
```
Satoshi-Kusumoto: Then this is not need

Satoshi-Kusumoto (changes requested):
Ref to parachain-common
Move some to constants.rs, DealwithFee to impl.rs
(But for us, the things should go to primitives/

runtime/src/lib.rs
```diff
@@ -66,6 +66,15 @@ use xcm_builder::{
...
 use xcm_executor::{Config, XcmExecutor};
 
+// TODO: MOVE TO COMMON
+// transaction payment pallet related items: TODO, move to common
+use pallet_transaction_payment::{Multiplier, TargetedFeeAdjustment};
+use sp_runtime::{FixedPointNumber, Perquintill};
+use frame_support::traits::{Currency, Imbalance, OnUnbalanced};
+pub type NegativeImbalance<T> = <pallet_balances::Pallet<T> as Currency<
+	<T as frame_system::Config>::AccountId,
+>>::NegativeImbalance;
```
h4n0: fmt missing?

runtime/src/lib.rs
```diff
@@ -414,12 +423,90 @@ parameter_types! {
 	pub const OperationalFeeMultiplier: u8 = 5;
 }
 
-// TODO: same as rococo-parachain but differnet as parachain-template
+// linear ratio of transaction fee distribution
```
h4n0: Capital the first letter to be consistent with the comment style

runtime/src/lib.rs
```diff
@@ -414,12 +423,90 @@ parameter_types! {
...
+struct RatioOf {
```
h4n0: I recommend to use a more detailed name here like TransactionFeeDistributionRatio, and the benefits are:
1. Easy to understand
2. This type is tx fee distribution used only, so if you name it RatioOf it sounds like a general ratio type

runtime/src/lib.rs
```diff
@@ -414,12 +423,90 @@ parameter_types! {
...
+				tips.merge_into(&mut split.1);
+			}
+			use pallet_treasury::Pallet as Treasury;
+			<Treasury<R> as OnUnbalanced<_>>::on_unbalanced(split.0);
+			<ToAuthor<R> as OnUnbalanced<_>>::on_unbalanced(split.1);
+		}
+	}
+}
+
+// transaction payment pallet: TODO, move to common
```
h4n0: ditto. Capital t please

h4n0 (changes requested):

runtime/src/lib.rs
wangminqi: get it

runtime/src/lib.rs
```diff
@@ -7,11 +7,9 @@
...
 pub mod constants;
-pub use constants::{currency::*};
+pub use constants::currency::*;
 
-pub use primitives::*;
-pub use primitives::Index as Index;
-pub use primitives::opaque as opaque;
+pub use primitives::{opaque, Index, *};
 
 use sp_api::impl_runtime_apis;
```
h4n0, line 14: If I remember this correctly, this is explicitly declared on purpose, right? Maybe double check this with @Satoshi-Kusumoto before removing them.

runtime/src/lib.rs
```diff
@@ -68,9 +66,9 @@ use xcm_executor::{Config, XcmExecutor};
 
 // TODO: MOVE TO COMMON
 // transaction payment pallet related items: TODO, move to common
+use frame_support::traits::{Currency, Imbalance, OnUnbalanced};
```
h4n0: I didn't get this TODO. What should be moved to common? Let's try to eliminate TODO comments in our production code.

runtime/src/lib.rs
wangminqi: I think cargo fmt auto-change this...

wangminqi: Let me double check with john

runtime/src/lib.rs
wangminqi: This import is for DealWithFee implemrntation, which should be moved to "common".

By "common", I mean parachain code shared like kusama/polkadot, which we did not have yet

wangminqi: I think we will have a common runtime package for storing these shared code.

primitives/src/lib.rs
```diff
@@ -8,50 +7,50 @@ pub use types::*;
 
 /// Common types of parachains.
 mod types {
-    use::sp_runtime::{
-        traits::{IdentifyAccount, Verify},
-        MultiSignature,
-    };
-    /// Alias to 512-bit hash when used in the context of a transaction signature on the chain.
-    pub type Signature = MultiSignature;
+	use ::sp_runtime::{
```
Satoshi-Kusumoto:
```suggestion
	use sp_runtime::{
```

runtime/src/constants.rs
```diff
@@ -1,14 +1,32 @@
...
-}
\ No newline at end of file
+	use crate::Balance;
+
+	pub const UNIT: Balance = 1_000_000_000_000;
+	pub const DOLLARS: Balance = UNIT;
+	pub const CENTS: Balance = DOLLARS / 100;
+	pub const MILLICENTS: Balance = CENTS / 1_000;
+
+	pub const TREASURY_PROPORTION: u32 = 40u32;
```
Satoshi-Kusumoto, line 12: add a comment for these types

runtime/src/constants.rs
```diff
@@ -1,14 +1,32 @@
...
+		pub treasury: u32,
+		pub author: u32,
+		pub burned: u32,
+	}
+
+	pub const TRANSACTION_PAYMENT_RATIO: TransactionFeeDistributionRatio = TransactionFeeDistributionRatio {
+		treasury: TREASURY_PROPORTION,
+		author: AUTHOR_PROPORTION,
+		burned: BURNED_PROPORTION,
+	};
```
Satoshi-Kusumoto: do not need a struct

Satoshi-Kusumoto (changes requested):
Note the location of DealWithFees and noteAuthorm, should be primitives/impl.rs

geastwood (commented):
Why there are always many formatting changes?

runtime/src/lib.rs
Satoshi-Kusumoto: common = primitives, it exsits

Satoshi-Kusumoto (changes requested):
I don't want to state again
your common = our primitves, pls move related things to the right place
say, DealWithFee should be in impl.rs 
You either clearly state that you reject the suggestion with reasons, or modify according to it..

Satoshi-Kusumoto:
> Why there are always many formatting changes?

because cargo fmt is not in our hook and upstream also does not run it every time

wangminqi:
> I don't want to state again your common = our primitves, pls move related things to the right place say, DealWithFee should be in impl.rs You either clearly state that you reject the suggestion with reasons, or modify according to it..

I have discussed with john. Instead of put transaction fee logic code in primitive, we will create an "impls.rs" under runtime folder.  
This file will be used if we have any self-created code implemented different from standard upstream.

runtime/src/lib.rs
```diff
@@ -26,6 +24,14 @@ use sp_runtime::{
 	ApplyExtrinsicResult,
 };
 
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
```
Satoshi-Kusumoto: I think it's better to put these in impl.rs, because they are specific tests

Satoshi-Kusumoto: therefore it seems that impl should be a folder

runtime/src/mock.rs
```diff
@@ -0,0 +1,221 @@
...
+	type Balance = u64;
+	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
+		smallvec![WeightToFeeCoefficient {
+			degree: 1,
+			coeff_frac: Perbill::zero(),
+			coeff_integer: WEIGHT_TO_FEE.with(|v| *v.borrow()),
+			negative: false,
+		}]
+	}
+}
```
Satoshi-Kusumoto: Another issue is are we going to use this mocked one or a real one here

runtime/src/mock.rs
```diff
@@ -0,0 +1,221 @@
...
+	base_weight: u64,
+	byte_fee: u64,
+	weight_to_fee: u64,
+}
+impl Default for ExtBuilder {
+	fn default() -> Self {
+		Self { balance_factor: 1, base_weight: 0, byte_fee: 1, weight_to_fee: 1 }
+	}
+}
+impl ExtBuilder {
```
Satoshi-Kusumoto: Why no spare lines everywhere

Satoshi-Kusumoto (changes requested):
Generally LGTM

runtime/src/mock.rs
wangminqi: From other projects' usage, Except kusama/polkadot, they simply do not have an unit test for runtime in rust. They use js test script instead.

 A full real mock test will/should target everything. In our meeting last time, we only want to have a transaction fee related test. So personally, I also feel this mock test is not necessary.

runtime/src/lib.rs
wangminqi: ...what about a common folder in runtime? Even though it is our own special implementation different from upstream, it will be shared by our future kusama/polkadot parachain.

runtime/src/mock.rs
wangminqi: added

node/src/command.rs
```diff
@@ -27,14 +27,15 @@ fn load_spec(
 	Ok(match id {
 		"dev" => Box::new(chain_spec::get_chain_spec_dev(para_id)),
 		"staging" => Box::new(chain_spec::get_chain_spec_staging(para_id)),
-		"generate-prod" => Box::new(chain_spec::get_chain_spec_prod(para_id)), // to generate res/chain_spec/prod.json
+		// In order to generate res/chain_spec/prod.json
+		"generate-prod" => Box::new(chain_spec::get_chain_spec_prod(para_id)),
```
Satoshi-Kusumoto, line 31: Do not change

Satoshi-Kusumoto (approved):
LGTM
just a small issue

h4n0 (approved):

h4n0 (approved):

PR #197, opened 2021-10-29
Fix unstable docker build

### LINKED ISSUE #192: Seemingly unstable docker build for parachain
Follow-up of #187 and #190

It's a problem that we've been struggling with: it seems our built parachain docker image is not stable enough and it breaks CI **sometimes** (= not always, even different results on different host-machines).

I tried different combinations to launch the CI network (= 2 relaychain nodes + 1 parachain node):
| Polkadot | Litentry-parachain | CI results | Comments |
| ------------- | ------------- | ------------- | ------------- |
| docker-image | docker-image  | unstable | but locally works |
| downloaded binary | binary from docker-image  | unstable | |
| downloaded binary | locally compiled binary | OK | |

Used polkadot docker-image: `parity/polkadot:v0.9.11`
Used parachain docker-image: `litentry/litentry-parachain:collective` => built on the branch where this problem is first discovered

When CI failed, the symptom is parachain cannot produce blocks, in the logs it's shown that the relaychain-client in parachain-node is keeping connecting and disconnecting to the relaychain-networks.

Is it a problem of base image? We used `paritytech/ci-linux:production` for building and `debian:buster-slim` for packaging, should be same as polkadot.
Or the way we build is wrong?

Docker image building and pushing are **disabled** until we find a robust solution for this.

Kailai-Wang:
resolves #192 

This PR uses `rust:latest` as base image for compilation, which is much more stable than `paritytech/ci-linux:production`.

With this fix the docker build and pushing is again **enabled**.

Tested multiple rounds locally and on CI-host.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,30 +1,46 @@
-FROM paritytech/ci-linux:production as builder
-LABEL description="This is the build stage for Polkadot. Here we create the binary."
+# ==========================
+# stage 1: build
+# https://hub.docker.com/_/rust
+# our host is based on bullseye/sid => rust:latest
+# ==========================
+FROM rust:latest as builder
+WORKDIR /litentry
+COPY . /litentry
 
-ARG PROFILE=release
-ARG FEATURES
-WORKDIR /litentry-parachain
+# always use the toolchain in base image
+# we do not fix toolchain version as it's still evolving rapidly
+RUN rustup default nightly && \
+	rustup target add wasm32-unknown-unknown --toolchain nightly
 
-COPY . /litentry-parachain
+RUN apt-get update && \
+	apt-get dist-upgrade -y -o Dpkg::Options::="--force-confold" && \
+	apt-get install -y cmake pkg-config libssl-dev git clang libclang-dev
 
-# empty features (--features=) will be ignored
-RUN cargo build --$PROFILE --features=$FEATURES
+ARG BUILD_ARGS
 
-# ===== SECOND STAGE ======
+RUN cargo build --release $BUILD_ARGS
 
-FROM debian:buster-slim
-LABEL description="This is the 2nd stage: a very small image where we copy the litentry-collator binary."
-ARG PROFILE=release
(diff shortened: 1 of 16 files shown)


docker/Dockerfile
```diff
@@ -1,30 +1,46 @@
...
+COPY . /litentry
+
+# always use the toolchain in base image
+# we do not fix toolchain version as it's still evolving rapidly
+RUN rustup default nightly && \
+	rustup target add wasm32-unknown-unknown --toolchain nightly
+
+RUN apt-get update && \
+	apt-get dist-upgrade -y -o Dpkg::Options::="--force-confold" && \
+	apt-get install -y cmake pkg-config libssl-dev git clang libclang-dev
```
chenzongxiong, line 17: Should we need `git`?

docker/Dockerfile
```diff
@@ -1,30 +1,46 @@
...
+COPY --from=builder /litentry/target/release/litentry-collator /usr/local/bin
+
+RUN useradd -m -u 1000 -U -s /bin/sh -d /litentry litentry && \
+	mkdir -p /data /litentry/.local/share && \
+	chown -R litentry:litentry /data && \
+	ln -s /data /litentry/.local/share/litentry-collator && \
+# unclutter and minimize the attack surface
+	rm -rf /usr/bin /usr/sbin && \
+# check if executable works in this container
+	/usr/local/bin/litentry-collator --version
```
chenzongxiong, line 39: I think this line can be removed. No need to increase such extra storage.

Kailai-Wang: Which one? Those `/data`? This command itself just check if the compiled version works at all before the docker-image is built.

docker/Dockerfile
Kailai-Wang: I think so, as most dependencies in .TOML are from git repo directly.

chenzongxiong (approved):

h4n0 (approved):

PR #198, opened 2021-10-29
Initial base filter

### LINKED ISSUE #169: Add proper BaseFilter

Satoshi-Kusumoto:
Disallow all pallets except system crucial ones
Because this makes other checks fail, merge after we confirm other tests
Closes #169


ts-tests/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "register-parachain": "ts-node tests/register-parachain.ts",
-    "test": "mocha --exit --sort -r ts-node/register 'tests/**/*.test.ts'"
+    "test": "mocha --exit --sort -r ts-node/register 'tests/base-filter.test.ts'"
```
Satoshi-Kusumoto: @h4n0 maybe we just commented out original one so that we know what to change later (not just you...

h4n0: I was also about to do so but unfortunately json doesn't allow comments. 

An alternative is to have another package.Jason-backup

h4n0: I find a solution for this, which is to add a test-all for the original one. This can be restored as 'test' afterward.

Kailai-Wang (approved):

PR #200, opened 2021-10-29
Add GPLv3 LICENSE

h4n0:
GPLv3 is used as license.
The license notice has been added to each source file, following the advice of GPL official guide: https://www.gnu.org/licenses/gpl-howto.en.html


hanwencheng:
I see that we do not put the copyright comments on all the files in this repo. Any specific reasons?

BTW: I think we also need to put the copyright comments at the top of all the files of other parachain-related repos.

hanwencheng:
For you who wonder the difference between two license, may refer here: https://gist.github.com/Ben-Kaniobi/1556545fd56f8149e803.

Satoshi-Kusumoto (approved):
LGTM

h4n0:
> I see that we do not put the copyright comments on all the files in this repo. Any specific reasons?
> 

Yes, we added our license and copyright notice only in our source files, not in scripts or config files, and this is the common way I think, as our main intellectual property here is the rust source files. 

> BTW: I think we also need to put the copyright comments at the top of all the files of other parachain-related repos.
>

Sure. Our other related repos will be updated soon.

Kailai-Wang (approved):
I approved it.

But shall we put the comment part in a separate file like [Acala](https://github.com/AcalaNetwork/Acala/blob/master/HEADER-GPL3)  or [Bifrost](https://github.com/bifrost-finance/bifrost/blob/develop/HEADER-GPL3) and provide a script to replace them (so that it's easy to change later if need)? 

Low prior though.

Satoshi-Kusumoto:
Closes #204

PR #222, opened 2021-11-07
Misc improvements to GA and scripts

Kailai-Wang:
A few small improvements to github actions and scripts.
I didn't create an issue for this as it's a mixture of many small things including:

- default 'ref' value should be an empty string, as `workflow_dispatch` already runs on the target branch
- more logs about communication relay <> para in dev
- add a small script to bump(change) dependency version when we need to upgrade/downgrade
- cargo update

### DIFF
--- a/.github/workflows/build_wasm.yml
+++ b/.github/workflows/build_wasm.yml
@@ -5,7 +5,7 @@ on:
     inputs:
       ref:
         description: The ref to be used for the repo
-        default: dev
+        default: ''
         required: false
 
 env:
--- a/docker/parachain-launch-config-dev.yml
+++ b/docker/parachain-launch-config-dev.yml
@@ -4,7 +4,7 @@
 #
 # Relaychain Configuration
 relaychain:
-  image: parity/polkadot:v0.9.12 # the docker image to use
+  image: parity/polkadot:latest # the docker image to use
   chain: rococo-local # the chain to use
   runtimeGenesisConfig: # additonal genesis override
     configuration:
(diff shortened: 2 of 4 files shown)


docker/parachain-launch-config-dev.yml
```diff
@@ -4,14 +4,15 @@
 #
 # Relaychain Configuration
 relaychain:
-  image: parity/polkadot:v0.9.12 # the docker image to use
+  image: parity/polkadot:latest # the docker image to use
```
Satoshi-Kusumoto, line 7: Why we use latest?

Satoshi-Kusumoto (approved):
The script is useful

docker/parachain-launch-config-dev.yml
Kailai-Wang: > Why we use latest?

My motivation is that with this we could detect the incompatibility earlier.
Theoretically, there should be no problem connecting an older client to the latest mainnet(polkadot), but if there is, it's a good indication that we should update our client.

.github/workflows/build_wasm.yml
```diff
@@ -5,7 +5,7 @@ on:
     inputs:
       ref:
         description: The ref to be used for the repo
-        default: dev
+        default: ''
```
h4n0, line 8: What does it mean by setting the default as ' '?

Kailai-Wang: It means when not provided, it will stay empty (instead of dev).
Later the actual checkout will then be decided by `github.ref` (running branch)
```
ref: ${{ github.event.inputs.ref || github.ref }}
```

h4n0 (approved):

PR #229, opened 2021-11-11
Add two token migration pallets 

suinuj:


Satoshi-Kusumoto:
As discussed, pls move related pallets to /pallets

PR #230, opened 2021-11-11
add try-runtime support

### LINKED ISSUE #228: Integrate try-runtime support
In order to test `on_runtime_upgrade` hook prior to the real runtime upgrade, we need to integrate the missing try-runtime support, including:

- command-line option for `try-runtime`
- `try-runtime` feature in compilation options
- entrypont for `frame_try_runtime::TryRuntime` in litentry-parachain-runtime

In case we need any GA for this later, a new issue will be created.

Kailai-Wang:
Resolves #228 

This PR adds support to `try-runtime` subcommand.
It's guarded by feature gate.

### DIFF
--- a/.github/workflows/build_docker_with_args.yml
+++ b/.github/workflows/build_docker_with_args.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Build docker image
         run: |
-          ./scripts/build-docker.sh ${{ github.event.inputs.docker_tag }} ${{ github.event.inputs.args }}
+          ./scripts/build-docker.sh ${{ github.event.inputs.docker_tag }} "${{ github.event.inputs.args }}"
           echo "============================="
           docker images
 
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -57,6 +57,14 @@ pub enum Subcommand {
 	/// The custom benchmark subcommmand benchmarking runtime pallets.
 	#[structopt(name = "benchmark", about = "Benchmark runtime pallets.")]
 	Benchmark(frame_benchmarking_cli::BenchmarkCmd),
+
+	/// Try some command against runtime state.
+	#[cfg(feature = "try-runtime")]
+	TryRuntime(try_runtime_cli::TryRuntimeCmd),
+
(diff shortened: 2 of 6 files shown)


.github/workflows/build_docker_with_args.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
       - name: Build docker image
         run: |
-          ./scripts/build-docker.sh ${{ github.event.inputs.docker_tag }} ${{ github.event.inputs.args }}
+          ./scripts/build-docker.sh ${{ github.event.inputs.docker_tag }} "${{ github.event.inputs.args }}"
```
Satoshi-Kusumoto, line 25: because there are multi args now?

node/src/cli.rs
```diff
@@ -57,6 +57,14 @@ pub enum Subcommand {
...
 	#[structopt(name = "benchmark", about = "Benchmark runtime pallets.")]
 	Benchmark(frame_benchmarking_cli::BenchmarkCmd),
+
+	/// Try some command against runtime state.
+	#[cfg(feature = "try-runtime")]
+	TryRuntime(try_runtime_cli::TryRuntimeCmd),
+
+	/// Try some command against runtime state. Note: `try-runtime` feature must be enabled.
+	#[cfg(not(feature = "try-runtime"))]
+	TryRuntime,
```
Satoshi-Kusumoto, line 67: Why we need this one? Seems not that necessary

Satoshi-Kusumoto (approved):
Generally LGTM

.github/workflows/build_docker_with_args.yml
Kailai-Wang: > because there are multi args now?

It wouldn't work if the arg contains space, e.g. `--features try-runtime`, you had to write `--features=try-runtime`.
Now with the quotes both would work.

node/src/cli.rs
Kailai-Wang: > Why we need this one? Seems not that necessary

I copied this from substrate codes..
I think you will get the warning message in `command.rs` if you try to run `try-runtime` while not compiling it with this feature.

PR #232, opened 2021-11-11
Add two pallets for token migration

suinuj:


suinuj:
seems clippy has a bug. if we can skip clippy check

Kailai-Wang:
> seems clippy has a bug. if we can skip clippy check

The error appears to happen when building runtime wasm with unmatching features in clippy.
We should be able to fix it by adding `SKIP_WASM_BUILD=1` before `cargo clippy` in Makefile

Kailai-Wang:
About the pallet integration and filter change:
we'd like to keep `dev` branch as synchronized as the actual parachain state in order to follow the history of filter change and provide a consistent dev codebase for all (that's also polkadot/other projects do), so in this situation there're 2 ways:
1. we keep our changes in `token-bridge` branch and hold back the merge until we decide to include it in the next release
2. we merge to `dev` but keep the runtime version and `baseCallFilter` unchanged.

I prefer option 1 personally..

Kailai-Wang:
Another point is maybe we should add the cargo test as part of our integration test, as more and more (runtime) cargo tests are being added. @h4n0

suinuj:
Agree with @Kailai-Wang , not merge the pr until the next release. use this branch to do upgrade testing.

Kailai-Wang:
The merging strategy was mentioned again last week.

Now option 2 appears to be a good option as we are developing more pallets and merging back and forth later could be a pain. Testing with activated filters should not be a problem once #282 is implemented.

pallets/bridge-transfer/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+authors = ['Litentry Dev']
+edition = '2018'
```
Kailai-Wang: maybe try rust edition 2021

pallets/bridge-transfer/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+	};
+	use frame_system::pallet_prelude::*;
+	pub use pallet_bridge as bridge;
+
+	type ResourceId = bridge::ResourceId;
+
+	type BalanceOf<T> =
+		<<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
+
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
```
Kailai-Wang: not strictly against this, but normally starts from 0

pallets/bridge/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+authors = ['Litentry Dev']
+edition = '2018'
```
Kailai-Wang: see above

runtime/src/lib.rs
```diff
@@ -916,7 +945,9 @@ impl Contains<Call> for BaseCallFilter {
 			call,
 			Call::Sudo(_) |
             // System
-            Call::System(_) | Call::Timestamp(_) | Call::ParachainSystem(_)
+            Call::System(_) | Call::Timestamp(_) | Call::ParachainSystem(_) |
+			// Bridge
+			Call::ChainBridge(_) | Call::BridgeTransfer(_)
```
Kailai-Wang: The extrinsics should not be allowed (for now)

scripts/launch-local-binary.sh
```diff
@@ -89,14 +89,26 @@ $PARACHAIN_BIN export-genesis-wasm --chain dev > para-$PARACHAIN_ID-wasm
...
+
+re="ocal node identity is: (.*)"
+if ! [[ $line =~ $re ]]; then 
+  echo "local node not found for alice in log" 
+  exit 1
+fi
+echo ${BASH_REMATCH[1]}
+
+$POLKADOT_BIN --chain $ROCOCO_CHAINSPEC --bob --tmp --port 30334 --ws-port 9945  --rpc-port 9934 \
+  --bootnodes /ip4/127.0.0.1/tcp/30333/p2p/${BASH_REMATCH[1]} &> "relay.bob.log" &
```
Kailai-Wang, line 103: not sure if it's needed, as the nodes should be able to discover each other as long as they are running on the same host.

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+	#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+	pub enum BridgeEvent {
+		FungibleTransfer(BridgeChainId, DepositNonce, ResourceId, U256, Vec<u8>),
+		NonFungibleTransfer(BridgeChainId, DepositNonce, ResourceId, Vec<u8>, Vec<u8>, Vec<u8>),
+		GenericTransfer(BridgeChainId, DepositNonce, ResourceId, Vec<u8>),
+	}
+
+	impl<A: PartialEq, B: PartialOrd + Default> ProposalVotes<A, B> {
+		/// Attempts to mark the proposal as approve or rejected.
+		/// Returns true if the status changes from active.
```
Kailai-Wang: Maybe typo in comment, here only enum is returned

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+			self.status != ProposalStatus::Initiated
+		}
+
+		/// Returns true if `who` has voted for or against the proposal
+		fn has_voted(&self, who: &A) -> bool {
+			self.votes_for.contains(who) || self.votes_against.contains(who)
+		}
+
+		/// Return true if the expiry time has been reached
+		fn is_expired(&self, now: B) -> bool {
```
Kailai-Wang, line 117: I feel like we don't need `now` as parameter, as it should always be the current block number. (Otherwise the naming `now` is misleading)

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+	pub enum BridgeEvent {
+		FungibleTransfer(BridgeChainId, DepositNonce, ResourceId, U256, Vec<u8>),
+		NonFungibleTransfer(BridgeChainId, DepositNonce, ResourceId, Vec<u8>, Vec<u8>, Vec<u8>),
+		GenericTransfer(BridgeChainId, DepositNonce, ResourceId, Vec<u8>),
+	}
+
+	impl<A: PartialEq, B: PartialOrd + Default> ProposalVotes<A, B> {
+		/// Attempts to mark the proposal as approve or rejected.
+		/// Returns true if the status changes from active.
+		pub fn try_to_complete(&mut self, threshold: u32, total: u32) -> ProposalStatus {
```
Kailai-Wang, line 94: TBH both parameters seem strange to me:
- `threshold` appears to act as both the passingThreshold and totalThreshold
- `total` shouldn't it be calculated from `votes_for.len() + votes_against.len()`? If it's meant to be "maximum possible total votes", then some sanity checks should be performed, e.g. `votes_for.len() + votes_against.len() <= total`

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+			}
+		}
+
+		/// Returns true if the proposal has been rejected or approved, otherwise false.
+		fn is_complete(&self) -> bool {
+			self.status != ProposalStatus::Initiated
+		}
+
+		/// Returns true if `who` has voted for or against the proposal
+		fn has_voted(&self, who: &A) -> bool {
```
Kailai-Wang, line 112: What happens if a voter votes for both `for` and `against`? Do we need to check this?

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+			Self {
+				votes_for: vec![],
+				votes_against: vec![],
+				status: ProposalStatus::Initiated,
+				expiry: BlockNumber::default(),
+			}
+		}
+	}
+
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
```
Kailai-Wang: see above

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+		/// Proposed dispatchable call
+		type Proposal: Parameter
+			+ Dispatchable<Origin = Self::Origin>
+			+ EncodeLike
+			+ GetDispatchInfo;
+		/// The identifier for this chain.
+		/// This must be unique and must not collide with existing IDs within a set of bridged
+		/// chains.
+		#[pallet::constant]
+		type BridgeChainId: Get<BridgeChainId>;
```
Kailai-Wang, line 154: the identical naming `BridgeChainId` might be misleading, it might be better to use `Get<u8>` , or rename to `Get<BridgechainIdType>`

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+	pub type BridgeChainId = u8;
+	pub type DepositNonce = u64;
+	pub type ResourceId = [u8; 32];
+
+	/// Helper function to concatenate a chain ID and some bytes to produce a resource ID.
+	/// The common format is (31 bytes unique ID + 1 byte chain ID).
+	pub fn derive_resource_id(chain: u8, id: &[u8]) -> ResourceId {
+		let mut r_id: ResourceId = [0; 32];
+		r_id[31] = chain; // last byte is chain id
+		let range = if id.len() > 31 { 31 } else { id.len() }; // Use at most 31 bytes
```
Kailai-Wang: I think here a boundary check is missing: `range - 1 - i` can go negative

Kailai-Wang (changes requested):
I've added some suggestions.
I only did a quick skim-through, haven't checked any business logic: probably will do it later.

pallets/bridge/src/lib.rs
suinuj: the minimum value of "range - 1 - i" is zero.

pallets/bridge/src/lib.rs
suinuj: BridgeChainId = u8, so Get<u8> is the same with Get<BridgechainId>. I think BridgechainId is type without any context, since it is started with capital letter. I'd like to unchanged it since it is from upstream.

pallets/bridge/src/lib.rs
suinuj: accepted.

pallets/bridge/src/lib.rs
suinuj: yes, the vote is finalized with the assumption each one just vote once, for or against.

pallets/bridge/src/lib.rs
suinuj: - threshold is dynamic, maybe updated at runtime
- some node may miss the vote, total != votes_for.len() + votes_against.len(), and vote could be finalized before all nodes vote
- votes_for.len() + votes_against.len() <= total not needed.

pallets/bridge/src/lib.rs
suinuj: now is a more general concept than blocknumber, in impl<A: PartialEq, B: PartialOrd + Default> ProposalVotes<A, B> 
type B could be not block number. 
I think pub struct ProposalVotes<AccountId, BlockNumber> could use other name, like identity, timestamp.

pallets/bridge/src/lib.rs
suinuj: accepted.

scripts/launch-local-binary.sh
suinuj: doesn't work in Mac, it is better to have it.

runtime/src/lib.rs
suinuj: will remove it.

pallets/bridge/Cargo.toml
suinuj: accepted.

pallets/bridge-transfer/src/lib.rs
suinuj: accepted.

pallets/bridge-transfer/Cargo.toml
suinuj: accepted.

pallets/bridge/src/lib.rs
Kailai-Wang: > the minimum value of "range - 1 - i" is zero.

-1 I think? range can be 0, i can be 0

scripts/launch-local-binary.sh
Kailai-Wang: strange.. just tried on Mac, it should work. Maybe you have too old polkadot binary? Or firewall settings?

pallets/bridge/src/lib.rs
Kailai-Wang: I'm only checking this fn alone without examining which values the parameters can be. I believe the code will work as expected in the majority of cases though

pallets/bridge/src/lib.rs
Kailai-Wang: Understood

pallets/bridge/src/lib.rs
Kailai-Wang: What I mean is the following codes:
```
pub type BridgeChainId = u8;
type BridgeChainId: Get<BridgeChainId>;
// later in runtime
type BridgeChainId = 1;
```
That's a bit misleading.
Despite not fully agreeing with it, let's not change it to avoid extra merge overhead later

scripts/launch-local-binary.sh
Kailai-Wang: After discussion, the part will be kept so that it works with and without VPN

Kailai-Wang (approved):

PR #233, opened 2021-11-11
Integrate cargo fmt check in GA

### LINKED ISSUE #231: Integrate `cargo fmt` to github action
We'd like to enforce `cargo fmt` check upon commit/PR.
It should fail already upon non-zero exit status, https://github.com/bevyengine/bevy is a bit overkill.

Kailai-Wang:
Resolves #231 

This PR enforces cargo fmt check before starting to compile.
An example which fails on this check is here: https://github.com/litentry/litentry-parachain/actions/runs/1449184808

### DIFF
--- a/.github/workflows/build_and_run_test.yml
+++ b/.github/workflows/build_and_run_test.yml
@@ -19,12 +19,30 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  check-cargo-fmt:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: nightly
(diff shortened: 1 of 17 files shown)


Satoshi-Kusumoto (approved):

h4n0 (approved):
Great to have fmt here! Thanks Kailai!

Just one question: Is there a format config file for this? I would like to apply the formatter to our editors as well (Rust Analyzer support for example) It would be great if this could happen.

Kailai-Wang:
> Great to have fmt here! Thanks Kailai!
> 
> Just one question: Is there a format config file for this? I would like to apply the formatter to our editors as well (Rust Analyzer support for example) It would be great if this could happen.

@h4n0 
a format config is possible, here's [what Acala uses](https://github.com/AcalaNetwork/Acala/blob/master/.rustfmt.toml).
It even has all the license text as header template.

We just need to discuss and unify the config.

PR #244, opened 2021-11-19
add generate_weights.yml template

Kailai-Wang:
According to [this doc](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#:~:text=To%20trigger%20the%20workflow_dispatch%20event%2C%20your%20workflow%20must%20be%20in%20the%20default%20branch), `workflow_dispatch` even can only take effect when it was once-off checked in into default branch.

This PR merely adds the yml template file so that it can be triggered/tested from feature-branch.

### DIFF
--- a/.github/workflows/generate_weights.yml
+++ b/.github/workflows/generate_weights.yml
@@ -1,0 +1,18 @@
+name: Generate weights files
+
+on:
+  workflow_dispatch:
+    inputs:
+      pallets:
+        description: pallets to calculate weights, * for all, or comma listed (e.g. frame-system,pallet-proxy)
+        default: '*'
+        required: true
+
+jobs:
+  build-and-run-benchmark:
+    runs-on: self-hosted
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/generate_weights.yml
```diff
@@ -0,0 +1,18 @@
+name: Generate weights files
+
+on:
+  workflow_dispatch:
+    inputs:
```
h4n0, line 5: Just a question, what are 'pallets' here defined? Is it a self-defined parameter?

Satoshi-Kusumoto (approved):

h4n0 (approved):

.github/workflows/generate_weights.yml
Kailai-Wang: > Just a question, what are 'pallets' here defined? Is it a self-defined parameter?

It's self-defined, see https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs

Issue #243, opened 2021-11-19
Remove filter on dev branch and enable integration test cases


h4n0:
- [x] Remove runtime filter
- [ ] Add cargo test in CI for parachain
- [x] Enable integration test cases for functionalities
- [x] Add release integration test for filter testing

h4n0:
We decide to keep the BaseCallFilter on. Therefore this issue has to be closed.

Details about our new branch rules: https://www.notion.so/litentry/Parachain-GitHub-Repo-Rules-40c2c5c19b3e4f1990b6d19b9453e9f0

PR #257, opened 2021-11-26
Remove filter, and enable tests

h4n0:
This PR solves issue #243 . 

The runtime filter has been removed on the dev branch. At the same time, previous functional ts tests are activated. A new test type "release test" has been added for filter tests. 

For the later release, we shall add the runtime filter back before each release, and run `make test-release-ci-docker ` to verify if the filter has been correctly added. See [here](https://github.com/litentry/litentry-parachain/commit/e97461a4d2d30e3c1a7adb10d88c24254599b8f8#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ec) for details how filter can be added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,10 @@ test-node:
 test-ci-docker: launch-local-docker
 	@./scripts/run-ci-test.sh
 
+.PHONY: test-release-ci--docker ## Run CI tests with docker without clean-up
+test-release-ci-docker: launch-local-docker
+	@./scripts/run-ci-test.sh -release
+
 .PHONY: test-ci-binary ## Run CI tests with binary without clean-up
 test-ci-binary: launch-local-binary
 	@./scripts/run-ci-test.sh
--- a/ts-tests/package.json
+++ b/ts-tests/package.json
@@ -8,8 +8,8 @@
   },
   "scripts": {
     "register-parachain": "ts-node tests/register-parachain.ts",
-    "test": "mocha --exit --sort -r ts-node/register 'tests/base-filter.test.ts'",
-    "test-all": "mocha --exit --sort -r ts-node/register 'tests/**/*.test.ts'"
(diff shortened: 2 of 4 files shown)


ts-tests/package.json
```diff
@@ -8,8 +8,8 @@
   },
   "scripts": {
     "register-parachain": "ts-node tests/register-parachain.ts",
-    "test": "mocha --exit --sort -r ts-node/register 'tests/base-filter.test.ts'",
-    "test-all": "mocha --exit --sort -r ts-node/register 'tests/**/*.test.ts'"
+    "test": "mocha --exit --sort -r ts-node/register 'tests/**/*.test.ts'",
```
Kailai-Wang, line 11: I believe here `base-filter.test.ts` will also be tested for dev branch, which might be unexpected.

Makefile
```diff
@@ -73,6 +73,10 @@ test-node:
 test-ci-docker: launch-local-docker
 	@./scripts/run-ci-test.sh
 
+.PHONY: test-release-ci--docker ## Run CI tests with docker without clean-up
+test-release-ci-docker: launch-local-docker
+	@./scripts/run-ci-test.sh -release
```
Kailai-Wang, line 78: I'd prefer to use `release` here, hyphen might cause some parsing problems on some systems.
Later in `run-ci-test.sh` something like:
```
TYPE=
[ ! -z "$1" ] && TYPE="-$1"
```

Kailai-Wang (changes requested):
I think we are still missing GitHub Action integration here, either `build_and_run_test.yml` or `create_releease_draft.yml` needs to be modified to actually include the base-filter test, and it should be triggered on release branch only.

h4n0:
We decide to keep the BaseCallFilter on. Therefore this PR has to be closed.

Details about our new branch rules: https://www.notion.so/litentry/Parachain-GitHub-Repo-Rules-40c2c5c19b3e4f1990b6d19b9453e9f0

PR #280, opened 2021-12-30
initial implementation of drop3 pallet

### LINKED ISSUE #273: Implement `pallet-drop3` with unittests and benchmarking
- [x] Implementation `pallet-drop3` according to [SW requirements](https://www.notion.so/litentry/Drop3-Software-Requirement-a4f17c322add40db832c0cbe75bad065)
- [x] Unittests
- [x] Optional benchmarking
- [x] Update/deploy the drop3-testnet with this pallet

Kailai-Wang:
resolves #273 

This PR:
- implements `pallet-drop3`, see comments in `src/lib.rs` for notes
- add matching unittest and benchmarking
- add locally generated weights.rs as pallet-benchmarked weights

Todos:
- merge to `drop3_testnet` branch once approved
- deploy the drop3 testing network
- see TODOs in codes


pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+		}
+
+		/// Approve a RewardPool proposal, must be called from admin
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::approve_reward_pool())]
+		pub fn approve_reward_pool(
+			origin: OriginFor<T>,
+			id: T::PoolId,
+		) -> DispatchResultWithPostInfo {
+			let sender = ensure_signed(origin)?;
+			ensure!(sender == Self::admin().unwrap(), Error::<T>::RequireAdmin);
```
suinuj: not use unwrap in runtime to avoid panic.

pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+			ensure!(RewardPools::<T>::contains_key(id), Error::<T>::NoSuchRewardPool);
+
+			RewardPools::<T>::mutate(id, |pool| {
```
suinuj: check if already approved before

pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+			// if CurrentMaxPoolId hasn't reached max, increment and return it
+			if Self::current_max_pool_id() < T::PoolId::max_value() {
+				return CurrentMaxPoolId::<T>::mutate(|id| {
+					*id += 1u64.into();
+					Ok(*id)
+				})
+			}
+
+			// otherwise find the vacant id from the beginning
+			let sorted_ids = Self::get_sorted_pool_ids();
```
suinuj, line 486: u64 is big enough, if necessary to find a used id after reach max value.

pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+		pub fn get_sorted_pool_ids() -> Vec<T::PoolId> {
+			let mut ids = RewardPools::<T>::iter_keys().collect::<Vec<T::PoolId>>();
+			ids.sort();
+			ids
+		}
+
+		#[cfg(test)]
+		// propose a default reward pool, but with given id
+		// mainly used to test get_next_pool_id()
+		pub fn propose_default_reward_pool(id: T::PoolId, should_change_current_max: bool) {
```
suinuj: we can define it in mock file

suinuj (changes requested):

suinuj:
A question about the admin account, if it could be the root account? which one to hold the admin account if we deploy the pallet this way.

pallets/drop3/src/lib.rs
Kailai-Wang: > not use unwrap in runtime to avoid panic.

good point, I changed it to `Some(sender) == Self::admin()`

pallets/drop3/src/lib.rs
Kailai-Wang: > check if already approved before

Maybe I'll keep this.
From code logic, multiple calls of `approve_reward_pool()` should be OK and I just leave the event as it is.
I could have added an if-check inside `mutate`, but the extra DB read in all cases is probably not worth it:
`100% 1 * write` vs `50% 1* read + 50% 1 * (read + write)`

pallets/drop3/src/lib.rs
Kailai-Wang: > u64 is big enough, if necessary to find a used id after reach max value.

Had this thought before, just want to play it safe and cover all cases.
It's unlikely to happen (but still possible as anyone can call `propose_reward_pool` with a minimum transaction fee and reserved balance), but once the id is exhausted, it needs a runtime upgrade just for changing this.

pallets/drop3/src/lib.rs
Kailai-Wang: > we can define it in mock file

Moved.

Kailai-Wang:
> A question about the admin account, if it could be the root account? which one to hold the admin account if we deploy the pallet this way.

The admin could be the root account, but not need to.
```
/// The reason why such an account is needed (other than just using ROOT) is for
/// fast processing of reward proposals, imagine later when sudo is removed
```

suinuj (approved):

pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,508 @@
...
+	/// The current storage version.
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
+
+	#[pallet::pallet]
+	#[pallet::generate_store(pub(super) trait Store)]
+	#[pallet::storage_version(STORAGE_VERSION)]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config: pallet_balances::Config + frame_system::Config {
```
wangminqi: maybe we could add a "type Currency: ReservableCurrency<Self::AccountId>" instead of balance pallet;
As in future we might use multi token or other balance pallet instead?

wangminqi: I see both usage in substrate, but feel a single type make more sense in our condition.

wangminqi (changes requested):
See above; pallet_balance or currency type.

pallets/drop3/src/lib.rs
h4n0: Agree with Minqi, probably `Currency<Self::AccountId> + ReservableCurrency<Self::AccountId>` is more commonly used.

pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,508 @@
...
+		}
+
+		/// Approve a RewardPool proposal, must be called from admin
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::approve_reward_pool())]
+		pub fn approve_reward_pool(
+			origin: OriginFor<T>,
+			id: T::PoolId,
+		) -> DispatchResultWithPostInfo {
+			let sender = ensure_signed(origin)?;
+			ensure!(Some(sender) == Self::admin(), Error::<T>::RequireAdmin);
```
h4n0, line 255: Just from the design perspective, would it be better to use collective and/or root as an alternative approver? By doing this we can later easily remove admin without interfering with the functionalities. Only using admin sounds a bit against the Polkadot philosophy.

pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,508 @@
...
+		/// close the reward pool, remove it from the map
+		/// and unreserve all the remaining token to the pool owner
+		fn unreserve_and_close_reward_pool(id: T::PoolId) -> DispatchResultWithPostInfo {
+			let pool = RewardPools::<T>::take(id).ok_or(Error::<T>::NoSuchRewardPool)?;
+			// we don't care if reserved balance is less than pool.remain
+			let _ = <pallet_balances::Pallet<T> as ReservableCurrency<_>>::unreserve(
+				&pool.owner,
+				pool.remain,
+			);
+			RewardPoolOwners::<T>::remove(pool.id);
```
h4n0, line 459: RewardPool not removed?

pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,508 @@
...
+			let sender = ensure_signed(origin)?;
+			let mut pool = RewardPools::<T>::get(id).ok_or(Error::<T>::NoSuchRewardPool)?;
+			ensure!(
+				Some(sender.clone()) == Self::admin() || sender == pool.owner,
+				Error::<T>::RequireAdminOrRewardPoolOwner
+			);
+			ensure!(pool.approved, Error::<T>::RewardPoolUnapproved);
+
+			pool.started = true;
+			RewardPools::<T>::insert(id, pool);
```
h4n0: Why insert the pool again here? I assume this will overwrite the old one, but wouldn't it be more efficient to just modify the flag itself (only one bit)?

pallets/drop3/src/lib.rs
```diff
@@ -0,0 +1,508 @@
...
+				<pallet_balances::Pallet<T> as ReservableCurrency<_>>::can_reserve(&sender, total),
+				Error::<T>::InsufficientReservedBalance
+			);
+			ensure!(end_at >= start_at, Error::<T>::InvalidProposedBlock);
+
+			let bounded_name: BoundedVec<u8, T::MaximumNameLength> =
+				name.clone().try_into().expect("reward pool name is too long");
+
+			// reserve the owner's balance
+			let _ = <pallet_balances::Pallet<T> as ReservableCurrency<_>>::reserve(&sender, total)?;
```
h4n0: An error will be raised here if insufficient balance happens, so the line 366 might be then redundant. (use map_err to use InsufficientReservedBalance error type here.)

h4n0 (changes requested):

pallets/drop3/src/lib.rs
Kailai-Wang: > Just from the design perspective, would it be better to use collective and/or root as an alternative approver? By doing this we can later easily remove admin without interfering with the functionalities. Only using admin sounds a bit against the Polkadot philosophy.

This is by design. `admin` has the permission to approve/reject proposals, but root and council have the permission to set the admin key: the idea is not to bother council for each approval/rejection when sudo is removed later, as it can be a frequent operation.

pallets/drop3/src/lib.rs
Kailai-Wang: > RewardPool not removed?

It is, see
```
let pool = RewardPools::<T>::take(id).ok_or(Error::<T>::NoSuchRewardPool)?;
```
and the unittests

pallets/drop3/src/lib.rs
Kailai-Wang: > Why insert the pool again here? I assume this will overwrite the old one, but wouldn't it be more efficient to just modify the flag itself (only one bit)?

as the getter only get a copy, so `pool.started = true;` only modifies the copy.
You either have to override it by inserting, or using mutate.

pallets/drop3/src/lib.rs
Kailai-Wang: > maybe we could add a "type Currency: ReservableCurrencySelf::AccountId" instead of balance pallet; As in future we might use multi token or other balance pallet instead?

> Agree with Minqi, probably `Currency<Self::AccountId> + ReservableCurrency<Self::AccountId>` is more commonly used.

Good point, I'll think twice and look forward to changing it

pallets/drop3/src/lib.rs
h4n0: Yes, using mutate would be a better solution in my opinion, as we just modify a flag and we don't have to copy the whole pool twice. How do you think?

pallets/drop3/src/lib.rs
Kailai-Wang: > An error will be raised here if insufficient balance happens, so the line 366 might be then redundant. (use map_err to use InsufficientReservedBalance error type here.)

Indeed, I'll just leave it out.

h4n0 (changes requested):

pallets/drop3/src/lib.rs
Kailai-Wang: Now adjusted.

pallets/drop3/src/lib.rs
Kailai-Wang: > Yes, using mutate would be a better solution in my opinion, as we just modify a flag and we don't have to copy the whole pool twice. How do you think?

It's actually a cleaner solution.
Adjusted using `try_mutate`

h4n0 (approved):

wangminqi (approved):

PR #288, opened 2022-01-07
Implementation of dynamic extrinsic filter

### LINKED ISSUE #282: Discussion and implementation of dynamic extrinsic filtering
This should be considered as one of the generic functionalities.

The motivations are:
- to have a safe mode to cope with emergency, see
   https://www.notion.so/litentry/pallet-transaction_pause-in-Acala-2c65bed945d54d81b8f03b18ad8bdffd
   https://www.notion.so/litentry/maintenance-mode-in-moonbeam-c01cff03852a44c8b30df1dd99533367
- to lift the extrinsic filtering in tests

General idea: https://www.notion.so/litentry/Dynamic-extrinsic-filtering-446715dd0016404983671033cb5ea675

Kailai-Wang:
Resolves #282 

This PR:
- implements `pallet-extrinsic-filter` which allows dynamic extrinsic filtering, see comments in codes
- adds corresponding unittests and benchmarking


pallets/extrinsic-filter/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+	impl<T: Config> Pallet<T> {
+		/// Set the mode
+		/// Weights should be 2 DB writes: 1 for mode and 1 for event
+		#[pallet::weight(2 * T::DbWeight::get().write)]
+		pub fn set_mode(origin: OriginFor<T>, mode: OperationalMode) -> DispatchResultWithPostInfo {
+			T::UpdateOrigin::ensure_origin(origin)?;
+
+			// we don't care which mode we are currently in,
+			// just set to the desired mode
+			Mode::<T>::put(mode);
```
suinuj, line 183: If the mode changed, should we clear all BlockedExtrinsics?

Kailai-Wang: > If the mode changed, should we clear all BlockedExtrinsics?

Good question. We could have implemented it in either way.
Currently this pallet is designed more for security reasons, so it's very conservative: we'd rather have extra (unchanged) blocked extrinsics than having unexpected whitelisted extrinsics.

However I'll write a comment explicitly about this and make a test case. 
Thank you.

suinuj (approved):

runtime/src/lib.rs
```diff
@@ -945,25 +954,41 @@ construct_runtime! {
...
+			Call::Sudo(_) |
+			// System
+			Call::System(_) | Call::Timestamp(_) | Call::ParachainSystem(_) |
+			// ExtrinsicFilter
+			Call::ExtrinsicFilter(_)
+		)
+	}
+}
+
+pub struct NormalModeFilter;
```
h4n0, line 982: If I understand it correctly, we set the current NormalModeFilter the same as SafeModeFilter, and modify this when we gradually add more functionalities here, but the SafeModeFilter stays unchanged. Am I right?

pallets/extrinsic-filter/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+				});
+			} else {
+				return Err(Error::<T>::ExtrinsicNotBlocked.into())
+			}
+			// do not pay the fee upon successful unblock
+			Ok(Pays::No.into())
+		}
+	}
+
+	impl<T: Config> Contains<T::Call> for Pallet<T>
```
h4n0, line 256: This Contains trait is implemented for the ExtrisicFilter pallet, but where is this one used?

runtime/src/lib.rs
Kailai-Wang: > If I understand it correctly, we set the current NormalModeFilter the same as SafeModeFilter, and modify this when we gradually add more functionalities here, but the SafeModeFilter stays unchanged. Am I right?

That's correct.
`SafeModeFilter` contains only the minimum core calls which maintain the parachains running.

As we activate more pallets, `NormalModeFilter` will be augmented to contain those calls.

pallets/extrinsic-filter/src/lib.rs
Kailai-Wang: > This Contains trait is implemented for the ExtrisicFilter pallet, but where is this one used?

In runtime binding

pallets/extrinsic-filter/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+			let allowed_by_mode = match Self::mode() {
+				OperationalMode::Normal => T::NormalModeFilter::contains(call),
+				OperationalMode::Safe => T::SafeModeFilter::contains(call),
+				OperationalMode::Test => T::TestModeFilter::contains(call),
+			};
+			// if allowed_by_mode is true, continue to check BlockedExtrinsics
+			// remember we only have blacklisting, no whitelisting
+			if allowed_by_mode {
+				let CallMetadata { function_name, pallet_name } = call.get_call_metadata();
+				return !BlockedExtrinsics::<T>::contains_key((
```
h4n0, line 270: The logic is a bit complex but basically I think it's correct. 
A minor thing, might be picky, but better to swap these two conditions, as it will be false fast in the case function name is default value.

h4n0 (approved):
LGTM

pallets/extrinsic-filter/src/lib.rs
Kailai-Wang: > The logic is a bit complex but basically I think it's correct. A minor thing, might be picky, but better to swap these two conditions, as it will be false fast in the case function name is default value.

Thank you.
Not sure if any misunderstanding stands here, but I think that's equally fast? Because it's possible to have both `((palletA, fnA), ())` and `((palletA, ()), ())` at the same time, we have to check both conditions anyway as long as neither is false, which is actually unpredictable (you don't know what was blocked).

PR #290, opened 2022-01-12
Runtime integration test

suinuj:


wangminqi (approved):

Kailai-Wang:
Shall we delete the `impls` folder?
Also we should add the cargo test to the Github action, (along with the pallets unittets), but we can do it in a separate PR.

h4n0:
LGTM
I prefer to add cargo test in our CI alongside as well.

h4n0 (approved):

PR #291, opened 2022-01-13
Add multiple runtime and chain specs support 

### LINKED ISSUE #298: Separate polkadot and kusama runtime
This is a preparation for the potential kusama parachain launch.

We need to separate polkadot and kusama runtime and chain specs. To reduce the complexity, the chain-specs should be controlled by command-line args without cargo feature guards (similar to polkadot).

There's no "default" chain-spec too, the user **must** specify a chain type to decide which chain spec will be loaded.

The tasks include:
- add kusama runtime
- refactor the fold structure to include kusama runtime
- refactor chain-spec processing from command-line
- move common codes to `primitives/`

h4n0:
This PR adds litentry kusama runtime and chain spec support, along with the current polkadot runtime and chain spec.

resolves #298


node/Cargo.toml
```diff
@@ -16,86 +16,90 @@ name = 'litentry-collator'
...
 
 [build-dependencies]
-substrate-build-script-utils = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13" }
+substrate-build-script-utils = {git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.13"}
 
 [features]
 default = []
+kitentry-runtime = ["kitentry-parachain-runtime"]
+litentry-runtime = ["litentry-parachain-runtime"]
 runtime-benchmarks = [
```
Kailai-Wang, line 97: we need kitentry features too, same applies to `try-runtime` feature.

Another small thing: shall we define formatting for TOML?

node/src/chain_specs/kitentry.rs
```diff
@@ -0,0 +1,246 @@
...
+	invulnerables: Vec<(AccountId, AuraId)>,
+	candidacy_bond: String,
+	endowed_accounts: Vec<(AccountId, String)>,
+	council: Vec<AccountId>,
+	technical_committee: Vec<AccountId>,
+	boot_nodes: Vec<String>,
+	telemetry_endpoints: Vec<String>,
+}
+
+pub fn get_chain_spec_dev() -> ChainSpec {
```
Kailai-Wang: indeed we have to think about how to deal with dev and staging chain specs.

node/src/command.rs
```diff
@@ -2,7 +2,7 @@
 // This file is part of Litentry.
 //
 // Litentry is free software: you can redistribute it and/or modify
-// it under the terms of the GNU General Public License as published by
```
Kailai-Wang: accidental removal?

node/src/command.rs
```diff
@@ -17,52 +17,95 @@
...
 	service::{new_partial, LitentryParachainRuntimeExecutor},
 };
 use codec::Encode;
 use cumulus_client_service::genesis::generate_genesis_block;
 use cumulus_primitives_core::ParaId;
-use litentry_parachain_runtime::Block;
+#[cfg(feature = "kitentry-runtime")]
+use kitentry_parachain_runtime::{self as parachain_runtime, Block};
+#[cfg(feature = "litentry-runtime")]
+use litentry_parachain_runtime::{self as parachain_runtime, Block};
```
Kailai-Wang: we need to discuss this:
currently I'm inclined to follow manta style, which is also what cumulus/polkadot does.

To guard every command by individual feature could prevent misuse of chain type, but it complicates things too. We would need two binaries, two docker-images etc

Also see https://github.com/paritytech/cumulus/blob/master/polkadot-parachains/src/command.rs#L244-L253

geastwood:
There seems to be a lot of formatting changes.

h4n0:
> There seems to be a lot of formatting changes.

Yes, mainly the cargo toml files. I have fixed them with new commits.

Kailai-Wang (approved):

PR #308, opened 2022-01-18
Update weights for litmus

### LINKED ISSUE #306: Update runtime weights for litmus
The runtime weights under `runtime/litmus/src/weights/` need to be updated

Kailai-Wang:
resolves #306 

This PR:
- recomputes the runtime weights for litmus with the current dependencies
- apply those weights in runtime lib.rs


wangminqi (approved):
Looks good to u

suinuj (approved):

suinuj:
just one question about --chain=generate-litmus, why just use litmus.

Kailai-Wang:
> just one question about --chain=generate-litmus, why just use litmus.

You mean why not just litmus?
The `litmus` chain-spec is loaded from json which is not ready yet and it's generated from `generate-litmus` anyway.
But I do have another comment in codes about `--chain=litmus-dev` vs `--chain=generate-litmus`, I don't think it matters too much, but maybe there's a convention which chain to use. What's your thought on this @suinuj ?

PR #314, opened 2022-01-22
Integrate cargo tests to CI

### LINKED ISSUE #277: Integrate cargo test into CI
Currently, only ts-tests are included in CI.

As we are about to integrate more pallets, therefore we need to assimilate cargo tests into CI as well. This includes:

- [x] rewrite runtime integration tests (rewriting of https://github.com/litentry/litentry-parachain/tree/dev/runtime/src/impls to use `litentry-parachain-runtime` instead of `mock-runtime`)
- [x] integrate integration-tests into CI
- [x] integrate pallet unittests into CI

E.g. integration-tests from Acala:
https://github.com/AcalaNetwork/Acala/tree/master/runtime/integration-tests

Kailai-Wang:
resolves #277 

This PR:
- adjusts folder structure to include both litentry and litmus runtime integration tets
- adds unittests to CI
- adds integration tests to CI
- ~~bump ts-tests dependencies~~ it seems the newer version is incompatible


.github/workflows/build_and_run_test.yml
```diff
@@ -102,9 +102,52 @@ jobs:
...
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          target: wasm32-unknown-unknown
+          default: true
+
+      - name: Run unittests
+        run: cargo test --release -p pallet-* --lib
```
h4n0, line 122: Will this run all unit tests of pallets with the name prefix "pallet-"?

h4n0 (commented):
LGTM. I recommend adding this branch as a PR trigger [here](.github/workflows/build_and_run_test.yml) to activate the newly added code once, and if all tests pass, revert the change and merge it.

h4n0:
The test code refactoring is pretty neat. Really like it!

.github/workflows/build_and_run_test.yml
h4n0: Also I ran this command locally and got an error `zsh: no matches found: pallet-*`, did you run it successfully?

Kailai-Wang: > Also I ran this command locally and got an error `zsh: no matches found: pallet-*`, did you run it successfully?

I believe it comes from the expansion bevaiour of `*` under zsh, maybe try this?
```
setopt nonomatch
```

Anyways on both Linux-bash and MacOS-bash I can get the correct behaviour, where unittests of all pallets will be run.

Kailai-Wang:
> LGTM. I recommend adding this branch as a PR trigger [here](.github/workflows/build_and_run_test.yml) to activate the newly added code once, and if all tests pass, revert the change and merge it.

I didn't quite get it.
I think the effect of this PR has already been tested, e.g. https://github.com/litentry/litentry-parachain/actions/runs/1739492063

h4n0:
> > LGTM. I recommend adding this branch as a PR trigger [here](.github/workflows/build_and_run_test.yml) to activate the newly added code once, and if all tests pass, revert the change and merge it.
> 
> I didn't quite get it. I think the effect of this PR has already been tested, e.g. https://github.com/litentry/litentry-parachain/actions/runs/1739492063

Ah ok. It looks good then. Thanks

.github/workflows/build_and_run_test.yml
h4n0: Yes, it works after switching to bash.

However, it seems that here only the unit tests of modules starting with pallet- prefix are executed. I see your points to separate these with the runtime ones, but by doing this it makes CI execution even longer. Also, there are some packages like primitives which are excluded from running test, and these could be an issue later when we add any packages without the pallet prefix.

Kailai-Wang: > Yes, it works after switching to bash.
> 
> However, it seems that here only the unit tests of modules starting with pallet- prefix are executed. I see your points to separate these with the runtime ones, but by doing this it makes CI execution even longer. Also, there are some packages like primitives which are excluded from running test, and these could be an issue later when we add any packages without the pallet prefix.

Unittests are meant for pallets only, we could provide a list of pallet names for unittests if we have pallets without standard prefix.

Unittests/integration tests/ts-tests are running in parallel. Actually, ts-tests take the most of the time, so CI is not getting longer (or, it gets longer just because we have to do both litmus and litentry ts-tests and these can only be done in sequential because of dependencies in docker, it takes 3 min anyway).

h4n0: It seems it takes 20 min each according to the action run, but I am not sure if it can be boosted if we do unit test and integration test together. 

The current CI time is 52 min already, comparing to the previous time (20min+) it is doubled the time. We may consider to reduce the execution time by offload some non-dependent jobs (clippy, fmt) to github provided hosts, in order to boost the main jobs on our self-hosted machine. What do you think?

Kailai-Wang: > It seems it takes 20 min each according to the action run, but I am not sure if it can be boosted if we do unit test and integration test together.
> 
> The current CI time is 52 min already, comparing to the previous time (20min+) it is doubled the time. We may consider to reduce the execution time by offload some non-dependent jobs (clippy, fmt) to github provided hosts, in order to boost the main jobs on our self-hosted machine. What do you think?

52min is really the exception where our self-hosted CI machine was round-robinned to run CI on both `dev` branch and feature branch. If you check the longest path, it's 26min.
The normal CI process will take 25min, e.g. here: https://github.com/litentry/litentry-parachain/actions/runs/1733110429

Kailai-Wang: we **are** doing unittests and integrations in parallel, and not on our hosts. So we don't care about the resources for it as long as it's not the slowest part in the whole CI.

We could have simply run `cargo test --release`, but the exact failure won't be available on the CI panel (e.g. you don't know if it unittest fails or integration-test fails.

Kailai-Wang: btw `clippy` is run on the github host too

h4n0: Sorry just overlooked the clippy and fmt are running on ubuntu-latest. 
Thanks for the explanation.

h4n0 (approved):

PR #317, opened 2022-01-24
Add PNS pallets

Cupnfish:
There is still some work to be done, and until this is completed, please do not merge.

- [x] Add `benchmark` to determine `weight`
- [x] Determine the initial config (including the official PNS account, pallet level administrator)

and some features need to be improved.

- [ ] better `price oracle` (preferably with relevant features available on the litentry chain)
- [ ] Cross-chain validation
- [ ] Domain auction


wangminqi:
I think there will be a runtime upgrade requirement. As the chain_spec code is actually not used. So whatever initial setting you want to do/import in chain_spec, you need a migration file to replace it.

https://docs.substrate.io/v3/runtime/upgrades/

You may also check this file
https://github.com/litentry/litentry-parachain/blob/backup_parachain_staking_impl/runtime/litentry/src/migration.rs

It should provide an example including remove one pallet, and mimic the genesis build of the replacement pallet parachain staking.

Cupnfish:
Can you help me with the benchmark for github workflow to generate weight? @wangminqi

Kailai-Wang:
Hi @Cupnfish thank you for the PR.
Unfortunately, we are still busy with upgrading to `polkadot-v0.9.16` dependency which is a must for litmus onboarding, so it's very likely that you have to merge `dev` into your branch later and make the adjustments.

Moreover, to integrate the pallet into parachain, we'd probably need to make sure both the integration test and runtime upgrade work. As for the benchmarking, you could use `benchmark-weight-local.sh` to get a locally computed weight first (so that the runtime bounding will syntactically work), later we can run the runtime benchmarking on the remote machine to override the default value.

Cupnfish:
Thanks for the reply, `benchmark-weight-local.sh` is now in use, I will wait for the subsequent update to `polkadot-v0.9.16` before following up on this PR.

Kailai-Wang:
Hi @Cupnfish,

I think you might have misunderstood the weights benchmarking. The runtime weights should actually be done by us later as we will run the benchmarks on the remote machine where we have comparable hardware settings as our collators (so that the generated weights will be as close to the real situation as possible). 

Normally, the pallet writer provides default weights along with the pallet - conventionally as weights.rs in the same folder. You could refer to substrate pallets. It should generally make it possible to have `type WeigthInfo = ();` even without running runtime benchmarks.

It's not a high priority though, the current implementation does compile and we should be able to override it too, so maybe just adapt it if you have time later :)

Cupnfish:
I know what you're talking about, and I'm actually waiting for the remote server to run the benchmark correctly in order to generate the right weight, and I'm submitting the relevant commit to mean that the current benchmark local generation script is fine and won't block running it on the remote server later. @Kailai-Wang

Cupnfish:
About the default `WeigthInfo `implementation, I will make an implementation when I have time, but I don't think it is very necessary.

suinuj:
Litentry implemented an NFT pallet before, which will serve other application like MCP and Drop3, please consider also using it in PNS.

Cupnfish:
> Litentry implemented an NFT pallet before, which will serve other application like MCP and Drop3, please consider also using it in PNS.

Which pallet are you talking about, I don't see `drop3` using the related `nft`, can you give an URL? If there is a good implementation, I will give priority to replace it, after all, the current pns `nft` pallet is a modified `nft` pallet of `orml`.

suinuj:
> > Litentry implemented an NFT pallet before, which will serve other application like MCP and Drop3, please consider also using it in PNS.
> 
> Which pallet are you talking about, I don't see `drop3` using the related `nft`, can you give an URL? If there is a good implementation, I will give priority to replace it, after all, the current pns `nft` pallet is a modified `nft` pallet of `orml`.

https://github.com/litentry/litentry-pallets/tree/dev/pallets/nft for your reference.

Cupnfish:
> > > Litentry implemented an NFT pallet before, which will serve other application like MCP and Drop3, please consider also using it in PNS.
> > 
> > 
> > Which pallet are you talking about, I don't see `drop3` using the related `nft`, can you give an URL? If there is a good implementation, I will give priority to replace it, after all, the current pns `nft` pallet is a modified `nft` pallet of `orml`.
> 
> https://github.com/litentry/litentry-pallets/tree/dev/pallets/nft for your reference.


Thanks for the URL you gave.

I read it and I'm sorry not to be able to use this. 

The current `nft` used by `pns` is magic modified `token id`, i.e. for each domain created, you need to specify its `token id`.（Each domain name and its `token id` are mathematically related.）

This is different from `orml`'s `nft`, so it doesn't replace them.

Cupnfish:
@suinuj @Kailai-Wang 

I used `sudo` to directly `setCode` and this problem appeared.

`1010: Invalid Transaction: Transaction would exhaust the block limits`

Is it due to my `OnRuntimeUpgrade` `weight` not being set enough? Shouldn't I add a `benchmark` to get the `weight` for this part?

I am currently using `sudoUncheckedWeight` for the upgrade, would it be inappropriate in a production environment?

Kailai-Wang:
> @suinuj @Kailai-Wang
> 
> I used `sudo` to directly `setCode` and this problem appeared.
> 
> `1010: Invalid Transaction: Transaction would exhaust the block limits`
> 
> Is it due to my `OnRuntimeUpgrade` `weight` not being set enough? Shouldn't I add a `benchmark` to get the `weight` for this part?
> 
> I am currently using `sudoUncheckedWeight` for the upgrade, would it be inappropriate in a production environment?

Sorry @Cupnfish I missed this message somehow. If you need something urgent next time, you could also ping me on slack.
 
`OnRuntimeUpgrade` will only be executed after the new code is set, so the error happens prior to that.
`sudoUncheckedWeight` should be the right way to go (even in production), do you still see errors with it?

Cupnfish:
> 

Thanks 🙏, everything went well after using `sudoUncheckedWeight` and I think it's ready to be reviewed.

Kailai-Wang:
Obsolete - will reopen once it's relevant again 🙌

PR #323, opened 2022-01-25
Quick fix 0.9.12 13 leftover issue

wangminqi:
This is a quick fix of event printed. As substrate upgrade from 0.9.12 to 0.9.13, transactionPayment implementation does not need to manual generating pallet_balance deposit event, otherwise a duplicated event will be printed.

### DIFF
--- a/runtime/litentry/src/transaction_payment.rs
+++ b/runtime/litentry/src/transaction_payment.rs
@@ -33,10 +33,7 @@ where
 	fn on_nonzero_unbalanced(amount: NegativeImbalance<R>) {
 		let _numeric_amount = amount.peek();
 		let author = <pallet_authorship::Pallet<R>>::author();
-		<pallet_balances::Pallet<R>>::resolve_creating(
-			&author,
-			amount,
-		);
+		<pallet_balances::Pallet<R>>::resolve_creating(&author, amount);
 	}
 }
 
--- a/runtime/litentry/src/transaction_payment.rs
+++ b/runtime/litentry/src/transaction_payment.rs
@@ -31,7 +31,6 @@ where
 	<R as frame_system::Config>::Event: From<pallet_balances::Event<R>>,
 {
 	fn on_nonzero_unbalanced(amount: NegativeImbalance<R>) {
-		let _numeric_amount = amount.peek();
 		let author = <pallet_authorship::Pallet<R>>::author();
 		<pallet_balances::Pallet<R>>::resolve_creating(&author, amount);
 	}
--- a/runtime/litentry/src/transaction_payment.rs
+++ b/runtime/litentry/src/transaction_payment.rs
@@ -37,10 +37,6 @@ where
 			&<pallet_authorship::Pallet<R>>::author(),
 			amount,
 		);
-		<frame_system::Pallet<R>>::deposit_event(pallet_balances::Event::Deposit {
-			who: author,
-			amount: numeric_amount,
-		});
 	}
 }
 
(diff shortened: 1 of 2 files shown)


runtime/litentry/src/transaction_payment.rs
```diff
@@ -31,16 +31,12 @@ where
 	<R as frame_system::Config>::Event: From<pallet_balances::Event<R>>,
 {
 	fn on_nonzero_unbalanced(amount: NegativeImbalance<R>) {
-		let numeric_amount = amount.peek();
-		let author = <pallet_authorship::Pallet<R>>::author();
+		let _numeric_amount = amount.peek();
+		let _author = <pallet_authorship::Pallet<R>>::author();
 		<pallet_balances::Pallet<R>>::resolve_creating(
 			&<pallet_authorship::Pallet<R>>::author(),
```
suinuj: already get the value as _author, why not just use it.

h4n0 (approved):
LGTM.

Just one comment, if these two variables numeric_amount and author are not needed, why not remove them?

runtime/litentry/src/transaction_payment.rs
wangminqi: emmmmm. Yes

runtime/litentry/src/transaction_payment.rs
```diff
@@ -31,16 +31,8 @@ where
...
-		let numeric_amount = amount.peek();
 		let author = <pallet_authorship::Pallet<R>>::author();
-		<pallet_balances::Pallet<R>>::resolve_creating(
-			&<pallet_authorship::Pallet<R>>::author(),
-			amount,
-		);
-		<frame_system::Pallet<R>>::deposit_event(pallet_balances::Event::Deposit {
-			who: author,
-			amount: numeric_amount,
-		});
```
Kailai-Wang, line 43: we don't need any event here, right? As cumulus seems to have one: https://github.com/paritytech/cumulus/blob/master/polkadot-parachains/parachains-common/src/impls.rs#L50-L52

Kailai-Wang:
btw you can use the `.githooks/pre-commit` to make sure your codes are cargo-formatted before pushing.

Otherwise the action page will be flooded..

wangminqi:
> LGTM.
> 
> Just one comment, if these two variables numeric_amount and author are not needed, why not remove them?

Done. Jesus, takes longer than I thought.

wangminqi:
> btw you can use the `.githooks/pre-commit` to make sure your codes are cargo-formatted before pushing.
> 
> Otherwise the action page will be flooded..

I will look into it. Apologize, this time I have something stored locally so I do the manual revise.

runtime/litentry/src/transaction_payment.rs
wangminqi: Yes. I think they are wrong.
https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/currency.rs#L158
https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L1647
Which obviously generating two events of Deposit.

wangminqi: Let me try PR them, hahaha

Kailai-Wang: > Yes. I think they are wrong. https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/currency.rs#L158 https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L1647 Which obviously generating two events of Deposit.

Yea maybe want to trigger such event in any case, but using the Curreny Event should be cleaner.

Issue #336, opened 2022-02-08
Double-check the collation related pallets configuration for litmus


Kailai-Wang:
As previously agreed, we are running the collators ourselves for litmus.

We shall ensure the following two things:
- no entry for the normal user to become a collator, currently it's guarded by `BaseCallFilter` anyway, but we have to examine it after the filter is removed. (e.g. via `MaxCandidates` from `pallet_collator_selection`?) Ideally with integration-test.
- the order of 5 collation-related pallets (Authorship, CollatorSelection, Session, Aura, AuraExt) is desired or doesn't matter, see the discussion screenshot below and the comment in the `v0.9.16` update.
   I did first step research and it appears to be the case.

It would be ideal if we can have documentation about the collator selection process for the session, not mandatory though.

<img width="1100" alt="image" src="https://user-images.githubusercontent.com/7630809/153086431-1d2bb585-960a-4867-8dd6-64e0465f9b56.png">

<img width="1019" alt="image" src="https://user-images.githubusercontent.com/7630809/153086488-f5acfd57-3599-4953-b05d-9471f038ad2d.png">

```
// About the order of these 5 pallets, the comment in cumulus seems to be outdated.
//
// The main thing is Authorship looks for the block author (T::FindAuthor::find_author)
// in its `on_initialize` hook -> Session::find_author, where Session::validators() is enquired.
// Meanwhile Session could modify the validators storage in its `on_initialize` hook. If Session
// comes after Authorship, the changes on validators() will only take effect in the next block.
//
// I assume it's the desired behavior though or it doesn't really matter.
//
// also see the comment above `AllPalletsWithSystem` and
```

Kailai-Wang:
The original substrate commit: 
https://github.com/paritytech/substrate/pull/10043

See the discussion in polkadot companion too:
https://github.com/paritytech/polkadot/pull/4181

It seems that the change of the order (reversed -> normal) actually fixes the problem that the last block author in a session can't get the reward (unless it's elected again after session rotation)

wangminqi:
![image.png](https://images.zenhubusercontent.com/6111125cad96288d72ec18c3/57a9d6cb-2bc0-4147-8e8b-0991451bace3)
![image.png](https://images.zenhubusercontent.com/6111125cad96288d72ec18c3/ae06bd4d-b0b8-49a1-b6fe-7d3ab98c7ab8)
The modification of this commit effects the decl_all_pallets function

[...]

Issue #334, opened 2022-02-08
Consolidate common codes for litmus and litentry in node


Kailai-Wang:
A follow-up of #329 

concretely, these two groups of functions:
- `litmus_parachain_build_import_queue` and `litentry_parachain_build_import_queue`
- `start_litmus_parachain_node` and `start_litentry_parachain_node`

The codes are the same except for the runtime API type. We could have used a single fn with <T> as template.
Now the highest prior but should still be done soon.

Issue #333, opened 2022-02-07
Benchmark adjustment for `polkadot-v0.9.16`


Kailai-Wang:
follow-up issue of #329 

It's created as a separate issue because some potential adjustments about benchmarking are required. After that, the weights should be re-calculated and updated accordingly.

PR #337, opened 2022-02-08
Update dependencies to `polkadot-v0.9.16`

### LINKED ISSUE #329: Update dependencies to `polkadot-v0.9.16`
Quote:
> Please make sure you update to 0.9.16 (your node) so that it produces blocks on Kusama :)

Kailai-Wang:
resolves #329 

This PR:
- updated the dependencies to `polkadot-v0.9.16`
- refactored some of codes
- adjusted the parachain-launch config

TODO:
- #333 
- #334 
- #336

### DIFF
--- a/runtime/litentry/src/lib.rs
+++ b/runtime/litentry/src/lib.rs
@@ -152,7 +152,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 9030,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 0,
+	transaction_version: 1,
 	state_version: 0,
 };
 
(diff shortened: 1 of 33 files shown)


h4n0 (approved):

runtime/litentry/src/lib.rs
```diff
@@ -123,6 +124,12 @@ pub type Executive = frame_executive::Executive<
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
+	// see https://github.com/paritytech/substrate/pull/10043
+	//
+	// With this type the hooks of pallets will be executed
+	// in the order that they are declared in `constrcut_runtime!`.
+	// It was reverse order before.
+	// See the comment before collation related pallets too.
 	AllPalletsWithSystem,
```
wangminqi, line 133: See #336 
I think the safest move for now is stay with AllPalletsWithSystemReversed. I also feel it should do no harm to us though. Let's discuss then decide.

wangminqi (changes requested):

runtime/litentry/src/lib.rs
Kailai-Wang: > See #336 I think the safest move for now is stay with AllPalletsWithSystemReversed. I also feel it should do no harm to us though. Let's discuss then decide.

I was a bit undecided, but after I saw the discussion in the polkadot companion, I'm inclined to have the new `AllPalletsWithSystem` (+ extra confirmation from Basti)

wangminqi (approved):

runtime/litentry/src/lib.rs
Kailai-Wang: As discussed, I'll put a comment in code later that we have to pay attention to the order of inherent calls.

PR #359, opened 2022-02-21
use mint for bridge token import

### LINKED ISSUE #358: Use mint for token imported from bridge
it is better to mint tokens instead of transfer from bridge pallet according to the discussion.

suinuj:
The PR resolves #358, the bridge will use mint for the imported token.

### DIFF
--- a/pallets/bridge-transfer/src/lib.rs
+++ b/pallets/bridge-transfer/src/lib.rs
@@ -100,7 +100,7 @@ pub mod pallet {
 			T::BridgeOrigin::ensure_origin(origin)?;
 
 			let total_issuance = <T as Config>::Currency::total_issuance();
-			if total_issuance + amount > T::MaximumIssuance::get() {
+			if amount > T::MaximumIssuance::get() || total_issuance + amount > T::MaximumIssuance::get() {
 				return Err(Error::<T>::ReachMaximumSupply.into())
 			}
 			if rid == T::NativeTokenResourceId::get() {
--- a/pallets/bridge-transfer/src/lib.rs
+++ b/pallets/bridge-transfer/src/lib.rs
@@ -27,9 +27,8 @@ pub use pallet::*;
 #[frame_support::pallet]
 pub mod pallet {
 	use frame_support::{
-		fail,
 		pallet_prelude::*,
-		traits::{Currency, ExistenceRequirement, StorageVersion},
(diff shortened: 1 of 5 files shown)


wangminqi (approved):

pallets/bridge-transfer/src/lib.rs
```diff
@@ -93,27 +97,20 @@ pub mod pallet {
...
 		) -> DispatchResult {
-			let source = T::BridgeOrigin::ensure_origin(origin)?;
-			// transfer to bridge account from external accounts is not allowed.
-			if source == to {
-				fail!(Error::<T>::InvalidCommand);
-			}
+			T::BridgeOrigin::ensure_origin(origin)?;
 
+			let total_issuance = <T as Config>::Currency::total_issuance();
+			if total_issuance + amount > T::MaximumIssuance::get() {
```
Kailai-Wang: do we need to check overflow here? Like `checked_add`? For example the `issue()` fn in pallet-balances: https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L1453

Or can we remove this check and rely on the error handling of `mint_into()`?

pallets/bridge-transfer/src/tests.rs
```diff
@@ -59,35 +59,30 @@ fn transfer() {
...
+		assert_events(vec![Event::Balances(balances::Event::Deposit {
+			who: RELAYER_A,
 			amount: 10,
 		})]);
 	})
 }
 
 #[test]
-fn transfer_to_holdingaccount() {
+fn exceed_max_supply() {
```
Kailai-Wang, line 91: In this test we only checked the scenario where we transfer one-off balance that exceeds the limit, shall we have another test that verifies the behaviour when we have accumulated balance that exceeds the limit? E.g. the sum of 5 transfers exceeds the max supply.

pallets/bridge-transfer/src/lib.rs
suinuj: agree, overflow check is needed.

pallets/bridge-transfer/src/tests.rs
suinuj: will add this case.

Kailai-Wang (approved):
`MaximumIssuance` can only be updated via runtime upgrade, right? (In case we want to change that)

PR #418, opened 2022-03-28
feat: xcm pahse 1 for sibling chain

wangminqi:


wangminqi:
This update should enable the ability of single side xcm (cross LIT asset to other chain)

Kailai-Wang (approved):
Approved.
This will also be addressed in the `polkadot-v0.9.18` update later.

Kailai-Wang (changes requested):
Oh sec, you are actually mixing up runtime:
`lib.rs` for litentry runtime and `xcm_config` for litmus were modified.

You probably want to change both

runtime/litmus/src/xcm_config.rs
```diff
@@ -149,8 +150,10 @@ impl pallet_xcm::Config for Runtime {
 	// ^ Disable dispatchable execute on the XCM pallet.
 	// Needs to be `Everything` for local testing.
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Nothing;
+	type XcmTeleportFilter = Nothing;
+	// This filter here defines what is allowed for XcmExecutor to handle with TransferReserveAsset
+	// Rule.
+	type XcmReserveTransferFilter = Everything;
```
suinuj, line 156: Just a question here, so the code change is to enable the reserve transfer and disable teleport, right?

wangminqi: Yes. Teleport  means burn/mint for XCM token bridge handling native Litmus asset. Reserve means deposit/withdraw using remote chain's litmus sovereign account.

wangminqi:
> Oh sec, you are actually mixing up runtime: `lib.rs` for litentry runtime and `xcm_config` for litmus were modified.
> 
> You probably want to change both

You are right. Updated.

Kailai-Wang (approved):

PR #429, opened 2022-03-30
Launch rococo parachain

### LINKED ISSUE #346: Launch a rococo parachain
**Update:** we will launch a rococo parachain, with the setup 2 collators + 1 rpc node.

---
Santiago reminded me of the possibility of having a long-lasting parachain testnet on rococo (we would have only 3 days lease period if we don't have a kusama/polkadot parachain):

image

https://docs.substrate.io/tutorials/v3/cumulus/rococo/

I think the pros and cons are palpable:

➕ more public
➕ closer to real relaychains for testing/upgrading
➕ don't have to maintain relaychains
➕ potential XCMP testing

➖ resource and maintenance overhead

Kailai-Wang:
resolves #346 

This PR prepares for the rococo parachain onboarding.
A release will follow.


.github/workflows/issue_stale_bot.yml
```diff
@@ -8,6 +8,7 @@ on:
 permissions:
   contents: read
   issues: write
+  pull-requests: write
```
h4n0, line 11: Why is pull request permission added, but not used?

.github/workflows/issue_stale_bot.yml
```diff
@@ -16,16 +17,16 @@ jobs:
       - uses: actions/stale@v4
         with:
           stale-issue-message: '
-          :exclamation: This issue is stale because it has been open for 20 days with no activity.
+          :exclamation: This issue is stale because it has been open for 60 days with no activity.
           
-          Remove `S2-stale` label or update it, otherwise this issue will be closed in 5 days.
+          Remove `S2-stale` label or update it, otherwise this issue will be closed in 7 days.
```
h4n0, line 22: Would it be possible to comment again 1 day before it's closed? Otherwise, it'll be closed silently if the first stale warning is neglected.

.github/workflows/issue_stale_bot.yml
Kailai-Wang: > Why is pull request permission added, but not used?

I couldn't find the exact failed run, but I saw once it tried to detect the PR status and complained it didn't have the permission - could even be an upstream problem.

.github/workflows/issue_stale_bot.yml
Kailai-Wang: > Would it be possible to comment again 1 day before it's closed? Otherwise, it'll be closed silently if the first stale warning is neglected.

I believe the original action doesn't support this, as the first warning should already be handled correctly 😂 
If we need it we could implement it ourselves though, simply detect this comment and write another comment in 6 days.

h4n0: I see. Just a good-to-have feature. Let's see if we really need it later.

h4n0 (approved):

PR #433, opened 2022-04-04
Enable vest() function call for reward claiming

### LINKED ISSUE #430: Runtime upgrade on Litmus to enable vest function
### Context
Vest functionality needs to be enabled

### Task
- Enable vest functionality in code
- Build a WASM 
- Upgrade this WASM on Litmus

### Acceptance criteria
Litmus is upgraded with the new feature

h4n0:
This PR resolves #430 

The reason of enabling vest() function only is, that the vesting pallets contains other functions like vestedTransfer() which does not require sudo for calling. So we just enable vest() for users to claim their own vested rewards.

### DIFF
--- a/runtime/litmus/src/lib.rs
+++ b/runtime/litmus/src/lib.rs
@@ -864,6 +864,7 @@ impl Contains<Call> for SafeModeFilter {
 pub struct NormalModeFilter;
 impl Contains<Call> for NormalModeFilter {
 	fn contains(call: &Call) -> bool {
+		// Calls that fulfill either matches or match below can pass the filter
 		matches!(
 			call,
 			Call::Sudo(_) |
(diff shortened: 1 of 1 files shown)


Kailai-Wang:
- I believe we could do something similar to [acala](https://github.com/AcalaNetwork/Acala/blob/master/runtime/acala/src/lib.rs#L210-L211), like`is_core_call`, `is_vesting_vest`, that might be more clear
- we should actually bump the runtime version, but I'll do it later together with v0.9.18 update
- clippy fails 😂

h4n0:
> * I believe we could do something similar to [acala](https://github.com/AcalaNetwork/Acala/blob/master/runtime/acala/src/lib.rs#L210-L211), like`is_core_call`, `is_vesting_vest`, that might be more clear
> * we should actually bump the runtime version, but I'll do it later together with v0.9.18 update
> * clippy fails 😂

Yes, I've also checked acala and moonbeam's implementation, and the reason I didn't use that is to have our current code changed as less as possible.

Thanks to clippy, now the change has become one line. (It's interesting to find a linter that really helps)

Kailai-Wang (approved):

PR #453, opened 2022-04-18
enable token bridge

suinuj:
The pr enables token bridge in Litmus, then the user can transfer LIT from Ethereum to Litmus.


Kailai-Wang:
Thanks.
Can we upgrade rococo-runtime too? Maybe we want to test it on rococo beforehand (especially with the frontend), even if not we'll test the runtime upgrade there :)

Kailai-Wang (approved):

PR #460, opened 2022-04-19
XCM implementation

wangminqi:
Update XCM implementation with dev branch.

The following feature is included
(1) Xcm implementation which allow local native token transfer freely. Non-native asset from remote chain is technically supported after proper token register process but not tested.
(2) Asset Manager Pallet, which used for handled remote/native token registering
(3) Unittest of parachain level which mimic the sending of XCM in relay + parachain system. Process after XcmRouter is not finished yet.


Kailai-Wang:
Can you please update the branch to have `dev` changes included? 
I believe orml packages were updated to `v0.9.22` too.

wangminqi:
> Can you please update the branch to have `dev` changes included? I believe orml packages were updated to `v0.9.22` too.

Updated. Although it is annoying that they do not release with an official tag.

pallets/xcm-asset-manager/Cargo.toml
```diff
@@ -0,0 +1,56 @@
+[package]
+name = "pallet-asset-manager"
+authors = [ "Litentry" ]
+edition = "2018"
```
Kailai-Wang: can we have 2021 here? And the authors as `Litentry Dev` -- basically to be consistent with other pallets

primitives/src/lib.rs
```diff
@@ -34,6 +34,15 @@ mod types {
...
 	pub type AccountId = <<Signature as Verify>::Signer as IdentifyAccount>::AccountId;
 
+	/// Signed version of Balance
+	pub type Amount = i128;
+
+	/// AssetId type
+	pub type AssetId = u128;
+
+	/// ParachainID type
+	pub type ParachainId = u32;
```
Kailai-Wang: I don't seem to find where this type is used :/

runtime/litmus/src/tests/mod.rs
```diff
@@ -14,5 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
 
+// mod relay;
```
Kailai-Wang: maybe delete this comment if we don't need it

runtime/litmus/src/tests/setup/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+	weights::{DispatchInfo, PostDispatchInfo, Weight},
+};
+
+pub use crate::{
+	AccountId, AssetManager, Balance, Balances, Call, CumulusXcm, DmpQueue, Event, ParachainSystem,
+	PolkadotXcm, Runtime, System, Tokens, TransactionPayment, Treasury, XTokens, XcmpQueue,
+};
+pub const ALICE: [u8; 32] = [1u8; 32];
+pub const BOB: [u8; 32] = [2u8; 32];
+// use xcm_simulator_example::relay_chain as relay;
```
Kailai-Wang: same as above: maybe delete this comment if we don't need it

runtime/litmus/src/tests/setup/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+pub use sp_std::cell::RefCell;
+pub mod relay;
+pub use pallet_balances::Call as BalancesCall;
+use xcm_simulator::{decl_test_network, decl_test_parachain, decl_test_relay_chain};
+
+pub(crate) fn last_event() -> Event {
+	System::events().pop().expect("Event expected").event
+}
```
Kailai-Wang, line 36: We should move these common functions to some "common" folder one day.. (don't have to change it now)

runtime/litmus/src/tests/setup/relay.rs
```diff
@@ -0,0 +1,153 @@
+use frame_support::{
```
Kailai-Wang, line 17: We are missing licence headers

runtime/litmus/src/tests/xcm_parachain/relay_sproof_builder.rs
```diff
@@ -0,0 +1,185 @@
...
+
+// Cumulus is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+// Litentry modify this file to mimic HRMP Channel info
```
Kailai-Wang, line 21: I'd still suggest to use litentry headers. We can add extra comment stating most of this file is copied from cumulus (if required)

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
+// Copyright 2019-2022 PureStake Inc.
```
Kailai-Wang: We'd better not include other project's licence in our code :D
same for other files

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+// (3) Compatibility to orml_token instead of pallet_assets
+// (4) Code for destroy asset is removed
+
+//! TODO Doc comments for the pallet
+//! # Asset Manager Pallet
+//!
+//! This pallet allows to register new assets if certain conditions are met
+//! The main goal of this pallet is to allow Litmus to register XCM assets
+//! and control the creation of foreign assets
+//! The assumption is we work with AssetTypes, which can then be comperted to AssetIds
```
Kailai-Wang: typos I believe

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+//!
+//! This pallet has four storage items: AssetIdType, which holds a mapping from AssetId->AssetType
+//! AssetIdUnitsPerSecond: an AssetId->u128 mapping that holds how much each AssetId should
+//! be charged per unit of second, in the case such an Asset is received as a XCM asset. Finally,
+//! AssetTypeId holds a mapping from AssetType -> AssetId. ForeignAssetCounter
+//! which holds the counter of foreign assets that have been created so far.
```
Kailai-Wang: Not sure if that's the original comment, but it's a bit confusing and unclear to me 😂 (e.g. with the word `Finally` applied on the third item). Maybe this will be better:
```
This pallet has four storage items:
- AssetIdType
  xxx
- AssetIdUnitsPerSecond
  xxxx
- ...
```

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+//!
+//! This pallet has eight extrinsics: register_foreign_asset, which registers a foreign
+//! asset in this pallet.
+//! set_asset_units_per_second: which sets the unit per second that should be charged for
+//! a particular asset.
+//! change_existing_asset_type: which allows to update the correspondence between AssetId and
+//! AssetType
+//! remove_supported_asset: which removes an asset from the supported assets for fee payment
+//! remove_existing_asset_type: which removes a mapping from a foreign asset to an assetId
```
Kailai-Wang: same as above

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+	pub struct AssetMetadata<Balance> {
+		pub name: Vec<u8>,
+		pub symbol: Vec<u8>,
+		pub decimals: u8,
+		pub minimal_balance: Balance,
+		pub is_frozen: bool,
+	}
+
+	// Defines the trait to obtain a generic AssetType from a generic AssetId and viceversa
+	pub trait AssetTypeGetter<AssetId, AssetType> {
```
Kailai-Wang, line 99: I'm not sure about the necessity of having it as `trait` here as it seems that only this pallet implements it..

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+	pub type AssetIdUnitsPerSecond<T: Config> = StorageMap<_, Blake2_128Concat, T::AssetId, u128>;
+
+	/// Stores the tracker of foreign assets id that have been
+	/// created so far
+	#[pallet::storage]
+	#[pallet::getter(fn foreign_asset_tracker)]
+	pub type ForeignAssetTracker<T: Config> = StorageValue<_, T::AssetId, ValueQuery>;
+
+	/// Litentry: Purestake use AssetType as index. This might lead to security issue if
+	/// MultiLocation Hash changes. Here we use AssetId instead.
```
Kailai-Wang: We don't have to repeat this 😂 maybe it's enough that we mention it once at the beginning

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+		pub fn register_foreign_asset(
+			origin: OriginFor<T>,
+			asset_type: T::ForeignAssetType,
+			metadata: AssetMetadata<BalanceOf<T>>,
+		) -> DispatchResult {
+			T::ForeignAssetModifierOrigin::ensure_origin(origin)?;
+
+			let asset_id = ForeignAssetTracker::<T>::get();
+
+			ensure!(AssetTypeId::<T>::get(&asset_type).is_none(), Error::<T>::AssetAlreadyExists);
```
Kailai-Wang: I believe `contains_key` should be enough..

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+
+			AssetMetadatas::<T>::insert(&asset_id, &metadata);
+			AssetIdType::<T>::insert(&asset_id, &asset_type);
+			AssetTypeId::<T>::insert(&asset_type, &asset_id);
+
+			Self::deposit_event(Event::<T>::ForeignAssetRegistered { asset_id, asset_type });
+			Self::deposit_event(Event::<T>::ForeignAssetMetadataUpdated { asset_id, metadata });
+
+			// Auto increment for Asset cunter
+			ForeignAssetTracker::<T>::put(asset_id + One::one());
```
Kailai-Wang: Shall we consider overflow?  Like `checked_add`

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+			ensure!(AssetTypeId::<T>::get(&asset_type).is_none(), Error::<T>::AssetAlreadyExists);
+
+			AssetMetadatas::<T>::insert(&asset_id, &metadata);
+			AssetIdType::<T>::insert(&asset_id, &asset_type);
+			AssetTypeId::<T>::insert(&asset_type, &asset_id);
+
+			Self::deposit_event(Event::<T>::ForeignAssetRegistered { asset_id, asset_type });
+			Self::deposit_event(Event::<T>::ForeignAssetMetadataUpdated { asset_id, metadata });
+
+			// Auto increment for Asset cunter
```
Kailai-Wang: typo

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn asset_metadatas)]
+	pub type AssetMetadatas<T: Config> =
+		StorageMap<_, Twox64Concat, T::AssetId, AssetMetadata<BalanceOf<T>>, OptionQuery>;
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Register new asset with the asset manager
+		#[pallet::weight(T::WeightInfo::register_foreign_asset())]
+		pub fn register_foreign_asset(
```
Kailai-Wang: As the name suggests `foreign` asset here, should we make an explicit difference between `foreign` asset and `native` asset (LIT)? Alternatively, what if others register "LIT" maliciously using exactly the same asset type?

Shall we reserve one or multiple `AssetId`s for native assets?

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+
+			let asset_id: T::AssetId =
+				AssetTypeId::<T>::get(&asset_type).ok_or(Error::<T>::AssetTypeDoesNotExist)?;
+
+			// Grab supported assets
+			let mut supported_assets = SupportedFeePaymentAssets::<T>::get();
+
+			ensure!(
+				num_assets_weight_hint >= (supported_assets.len() as u32),
+				Error::<T>::TooLowNumAssetsWeightHint
```
Kailai-Wang: I failed to understand why this needs to hold, or why `num_assets_weight_hint` is even needed..

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+			);
+
+			// Only if the asset is not supported we need to push it
+			if let Err(index) = supported_assets.binary_search(&asset_id) {
+				supported_assets.insert(index, asset_id.clone());
```
Kailai-Wang: `supported_assets` looks like a normal vector which contains all the asset_ids who called this `set_asset_units_per_second` fn, do we have a particular reason why it should be kept sorted and inserted only at a specific position? To me a set of assetId seems enough

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+		}
+
+		/// TODO: Weight is no longer related to num_assets_weight_hint; changes here needed
+		/// We do not allow the destroy of asset id so far
+		/// But we will remove all existing related fee settings. So will need re-implement by
+		/// extrinsic call. New default asset type is needed for the assignment of asset_id's
+		/// default asset type mapping
+		#[pallet::weight(T::WeightInfo::remove_asset_type(*num_assets_weight_hint))]
+		#[transactional]
+		pub fn remove_asset_type(
```
Kailai-Wang, line 318: TBH I'm not a fan of the signature of this function. 

As the user of this fn, I wouldn't expect that I have to pass in again the current asset_type and also the "new default" asset_type. I would expect that I provide the asset_id and the extrinsic will remove it from the storage.

Or if it just resets to a "default" asset type, we better call it something like `reset_asset_type`

runtime/litmus/src/lib.rs
```diff
@@ -866,16 +867,19 @@ construct_runtime! {
...
+		Tokens: orml_tokens::{Pallet, Call, Storage, Event<T>} = 55,
 
 		// Litmus pallets
 		ChainBridge: pallet_bridge::{Pallet, Call, Storage, Event<T>} = 60,
 		BridgeTransfer: pallet_bridge_transfer::{Pallet, Call, Event<T>, Storage} = 61,
 		Drop3: pallet_drop3::{Pallet, Call, Storage, Event<T>} = 62,
 		ExtrinsicFilter: pallet_extrinsic_filter::{Pallet, Call, Storage, Event<T>} = 63,
-		IdentityManagement: pallet_identity_management::{Pallet, Call, Storage, Event<T>} = 64,
+		AssetManager: pallet_asset_manager::{Pallet, Call, Storage, Event<T>} = 64,
+		IdentityManagement: pallet_identity_management::{Pallet, Call, Storage, Event<T>} = 65,
```
Kailai-Wang: I believe the index number of `IdentityManagement` is used by @suinuj in TEE tests, so we'd better shift AssetManager after it

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn supported_fee_payment_assets)]
+	pub type SupportedFeePaymentAssets<T: Config> = StorageValue<_, Vec<T::AssetId>, ValueQuery>;
+
+	/// The storages for AssetMetadatas.
+	///
+	/// AssetMetadatas: map AssetId => Option<AssetMetadata>
+	#[pallet::storage]
+	#[pallet::getter(fn asset_metadatas)]
+	pub type AssetMetadatas<T: Config> =
```
Kailai-Wang: Can we have a consistent naming for map? That is, something like AssetIdMetadata (similar to previous map variables)

pallets/xcm-asset-manager/src/weights.rs
```diff
@@ -0,0 +1,202 @@
...
+// --steps
+// 50
+// --repeat
+// 20
+// --template=./benchmarking/frame-weight-template.hbs
+// --record-proof
+// --json-file
+// raw.json
+// --output
+// ./benchmarks/
```
Kailai-Wang, line 43: we probably needs to rerun all these (even though it's just pallet's weight for reference)

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+		#[pallet::weight(T::WeightInfo::relocate_foreign_asset_id())]
+		pub fn relocate_foreign_asset_id(
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
+		) -> DispatchResult {
+			T::ForeignAssetModifierOrigin::ensure_origin(origin)?;
+
+			if asset_id > ForeignAssetTracker::<T>::get() {
+				ForeignAssetTracker::<T>::put(asset_id);
+			}
```
Kailai-Wang: We need an event and/or error here, otherwise we won't know if this call actually changed the state as it always returns `Ok`

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+			AssetIdUnitsPerSecond::<T>::insert(&asset_id, &units_per_second);
+
+			Self::deposit_event(Event::UnitsPerSecondChanged { asset_id, units_per_second });
+			Ok(())
+		}
+
+		/// Add the xcm type mapping for a existing assetId, other assetType still exists if any.
+		/// TODO: Weight is no longer related to num_assets_weight_hint; changes here needed
+		#[pallet::weight(T::WeightInfo::add_asset_type())]
+		pub fn add_asset_type(
```
Kailai-Wang, line 287: It's more like `update_assert_type` to me, the parameter doesn't seem to be consistent either: we are using `asset_id` while other extrinsics use `asset_type` to find the element.

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+			);
+
+			// Insert new asset type info
+			AssetIdType::<T>::insert(&asset_id, &new_asset_type);
+			AssetTypeId::<T>::insert(&new_asset_type, &asset_id);
+
+			// Remove previous asset type info
+			// AssetTypeId::<T>::remove(&previous_asset_type);
+
+			Self::deposit_event(Event::ForeignAssetRegistered {
```
Kailai-Wang: The event name is a bit confusing

Kailai-Wang (changes requested):
Sorry that maybe I've written too many comments, but the PR is too informative for me to go through all the stuff at once -- I only checked the asset-manager part..

Apart from that, there are some general concerns from my side:
- about the licence header, generally it's a bad idea to include other project's header directly (you'll see that in other parachains' source files even if they are copied from cumulus): we only include our headers and write: "inspired by team x or we have referred to xx)
- I'm still inclined to use asset_id as the extrinsic parameter to find everything: maybe @suinuj and @h4n0 can have their opinions too
- it seems we are missing the integration test for litentry: not sure if it's intended or just missing copy/paste 😂

runtime/litmus/src/tests/setup/mod.rs
wangminqi: I add a comment for TODO.

runtime/litmus/src/tests/setup/mod.rs
wangminqi: Removed.

runtime/litmus/src/tests/mod.rs
wangminqi: Removed.

primitives/src/lib.rs
wangminqi: This must be a leftover as somewhere during the test, we stimulated the parachain Id for multilocation.
Removed.

pallets/xcm-asset-manager/Cargo.toml
wangminqi: Changed edition and authors.

runtime/litmus/src/tests/xcm_parachain/relay_sproof_builder.rs
wangminqi: headers changed. Same as "inspired by: " style now.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: changed.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: compert to convert? changed.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: changed.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Sorry, I am a little lazy regarding this part. Document updated.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: This trait is used for xcm_config implementation. `FirstAssetTrader<
		CurrencyId,
		AssetManager,
		XcmFeesToAccount<
			Tokens,
			ConvertedConcreteAssetId<AssetId, Balance, AssetIdMuliLocationConvert, JustTry>,
			AccountId,
			XcmFeesAccount,
		>,
	>,`

wangminqi: So FirstAssetTrader can use any source implemented this trait to convert AssetId and AssetType. I think it is worth to maintain it.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: After changing the header, I think it is better to delete the comment. And let us discuss the AssetId and AssetType later.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: changed.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Does look simpler. I switch all similar usages.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Changed. I added a overflow check and error of overflow accordingly.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Typically there will be three types of asset. 
Foreign asset which origin from other parachain/relaychain.
Local native asset which is LIT.
Local asset which is not LIT but origin/minted on our parachain. (Not implemented yet)

As for LIT AssetType overwrite. This should not be a concern since The "AssetTransactors" implementation use  "LocalAssetTransactor" first, i.e. the mimic LIT will not be used. But it does worth an attention. I added a comment to remind.

100% solution is we should add runtimeUpgrade/genesis build to reserve these AssetTypes of LIT.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: This is for weight charge. There is a binary search below which's weight related to the num_assets. 
This is for weight formula implementation purpose.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: binary_search is quicker and require a sorted array in Vector case.

I personally not sure about using Hashset/Btree type here. Let's discuss this in a meeting.
https://docs.substrate.io/v3/advanced/hash-collections/
https://stackoverflow.com/questions/34305901/advantage-of-binary-search-tree-over-vector-in-c#:~:text=BST%20and%20other%20tree%20structures,memory%20usage%20while%20re%2Dsizing.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: If you are not remove the default asset_type, you can pass nothing about the new default.
If you are removing the default asset_type without passing new default, it will trigger the DefaultAssetTypeRemoved error.
I feel ok about the procedure here. Let's discuss this in a meeting.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Changed. New name AssetIdMetadatas

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Event ForeignAssetTrackerUpdated added if the change does happen.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: a AssetId only mappings to a special AssetType.
Many AssetType can mappings to one same AssetId.
This method is for adding a AssetType mapping to existed AssetId, and using new added AssetType for AssetId's default mapping.

runtime/litmus/src/tests/setup/relay.rs
wangminqi: added

runtime/litmus/src/lib.rs
wangminqi: I have swapped the pallet index accordingly, and save 64 on Moonbase for identity pallet too.
However, if there will be a lot of pallets in future, I suggest we start an individual pallet prefix for such TEE related pallets.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Todo task local /foreign asset multilocation restriction

pallets/xcm-asset-manager/src/lib.rs
Kailai-Wang: I double-checked the code, actually I was talking about the trait `AssetTypeGetter` (not `AssetIdInfoGetter`), which is only implemented by AssetManager, but nothing else.

Anyway it should do no harm to keep it

pallets/xcm-asset-manager/src/lib.rs
Kailai-Wang: I believe current impl has still some flaws, I'll leave my comment inline

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+			AssetIdMetadatas::<T>::insert(&asset_id, &metadata);
+			AssetIdType::<T>::insert(&asset_id, &asset_type);
+			AssetTypeId::<T>::insert(&asset_type, &asset_id);
+
+			Self::deposit_event(Event::<T>::ForeignAssetRegistered { asset_id, asset_type });
+			Self::deposit_event(Event::<T>::ForeignAssetMetadataUpdated { asset_id, metadata });
+
+			asset_id = asset_id.checked_add(&One::one()).ok_or(Error::<T>::AssetIdOverflow)?;
+			// Auto increment for Asset counter
+			ForeignAssetTracker::<T>::put(asset_id);
```
Kailai-Wang: If I'm not mistaken, the old value of `ForeignAssetTracker` was already consumed in line 240. Imagine we get an error in line 251, then `ForeignAssetTracker` will have no value (or default value by getter) as this line won't get executed.

IMO we should use `try_mutate`

We'd better double-check the rest of the code as I might be missing some similar situations.

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+			next_asset_id: T::AssetId,
+		) -> DispatchResult {
+			T::ForeignAssetModifierOrigin::ensure_origin(origin)?;
+			let old_next_asset_id = ForeignAssetTracker::<T>::get();
+			if next_asset_id > old_next_asset_id {
+				Self::deposit_event(Event::<T>::ForeignAssetTrackerUpdated {
+					old_asset_tracker: old_next_asset_id,
+					new_asset_tracker: next_asset_id,
+				});
+				ForeignAssetTracker::<T>::put(next_asset_id);
```
Kailai-Wang: We have the same problem as above, if we don't go into this branch then old value can't be restored.

Btw why would we provide such an extrinsic? 😂 Asset ids shouldn't be allocated freely as the caller's wishes, otherwise I could easily pass in unsigned MAX which will cause all subsequent registration to fail.

Even if not MAX I could pass in numbers like 549548549, all subsequent foreign assets will get some big numbers too, which may be strange.

pallets/xcm-asset-manager/src/lib.rs
Kailai-Wang: Wait a second, do we even need to keep supported assets in separate storage?
Isn't it as simple as:
1. check if the given asset_id exists (= included in the assetIdType map)
2. set the units_per_second if it exists

pallets/xcm-asset-manager/src/lib.rs
Kailai-Wang: I think twice and believe the `add_asset_type` and `remove_asset_type` can be merged into one simple `update_asset_type` function.  

We have two options:
1. The caller provides the `new_asset_type` and an optional `old_asset_type` to be removed. We simply update two storage maps (AssetIdTypes and AssetTypeIds).

Alternatively,

2. As we use `asset_id` as the navigation key, we can only keep 1 to 1 mapping for AssetTypeId, in this case the `new_asset_type` will simply replace the old asset_type. Arguments:
  2.1 the foreign chain can't exhaust all MultiLocations anyway, if it wants a new MultiLocation, simply update it. If it wants to query the asset_id, then please provide the MultiLocation that was registered or lastly updated. 
  2.2 eventually it's still strange to have a 1-to-1 mapping in one direction, and N-to-1 mapping in the other.
  2.3 older MultiLocation could be outdated, it's not logical that I find asset_id A via MultiLocation L1 while A's current MultiLocation is L2. As a simple asset registry we don't have to take too many responsibilities and care about the expiration of MultiLocations.

I think it will cover all the usecases already?

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn supported_fee_payment_assets)]
+	pub type SupportedFeePaymentAssets<T: Config> = StorageValue<_, Vec<T::AssetId>, ValueQuery>;
+
+	/// The storages for AssetIdMetadatas.
+	///
+	/// AssetIdMetadatas: map AssetId => Option<AssetMetadata>
+	#[pallet::storage]
+	#[pallet::getter(fn asset_metadatas)]
+	pub type AssetIdMetadatas<T: Config> =
```
Kailai-Wang: maybe single form? `AssetIdMetadata` -- just to be aligned with other map variables.

Kailai-Wang (changes requested):

runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+	Sibling::from(1).into_account()
+}
+
+fn _sibling_b_account() -> AccountId32 {
+	use sp_runtime::traits::AccountIdConversion;
+	Sibling::from(2).into_account()
+}
+
+#[test]
+fn test_xtokens_recognize_multilocation() {
```
Kailai-Wang, line 57: The network setup and simulation look good to me.

However, the integration test is still missing quite a bit IMHO :(
Current test only verifies that non-supported MultiLocation will be filtered out (`MultiLocationsFilter`), but this has nothing to do with `pallet-asset-manager`.

What we are looking forward to is the test for asset transfer between sibling parachains, e.g. LIT -> xLIT -> LIT: if we can handle it correctly after asset registration, including the fee handling of `units_per_second`

runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+
+	// TODO::This mimic does not finialized yet. Still need mimic XcmRouter
+	// ParaB::execute_with(|| {
+	// 	XcmpQueue::on_idle(0, u64::MAX);
+	// 	assert_eq!(Tokens::free_balance(0, &AccountId::from(BOB)), 999_200_000_000);
+	// })
+}
+
+#[test]
+fn test_pallet_xcm_recognize_multilocation() {
```
Kailai-Wang, line 127: Unfortunatley same problem 😂 

The test runs through even without `asset-manager` part. It's fine if it's meant to be a general integration test for polkadotXCM pallet, but not for `orml-xtokens/orml-tokens`

pallets/xcm-asset-manager/src/lib.rs
wangminqi: make sense. We can delete the supported storage and use units_per_second storage's non-zero to determine if it is supported.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Actually, I do hope to have 1-to-1 mapping plus N-to-1 mapping.
In XCM system, it is necessary to keep compatiable of different multilocation standard.

We can observe many standard so far. ( e.g. <ParachainId(X), Here>, <ParachainId(X), PalletInstance(pallet_balances index)>, <ParachainId(X), GeneralKey(***)>, <ParachainId(X)>)

https://moonriver.subscan.io/event?address=&module=assetmanager&event=assetregistered&startDate=&endDate=&startBlock=&endBlock=&timeType=date&version=1201

The thing I am thinking is, In most circumstances, other parachain change style, we adapt immediately and forget about old style, it is fine. But suppose Parachain A sending a old style of XCM of token on Parachain B to Parachain C (since Parachain A does not update Parachain B's style immediately), Parachain C (that is us) will not be able to recognize the message from A.

And Multilocation compatibility is a common request I think. So N-to-1 mapping I hope it will be worth keeping.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: changed

pallets/xcm-asset-manager/src/lib.rs
wangminqi: I resolve this conversation and continue the correction of in " already consumed" conversation below.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: Originally it is just for an easier bumping of asset_id tracker. Do you think it is worth keeping it under root power? Or just delete it.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: I am not sure if I understand it correctly.

So do you mean asset_id is consumed by deposit_event, so can not be later used?
I think it will not, see the "collator_count" usage of here.
https://github.com/PureStake/moonbeam/blob/master/pallets/parachain-staking/src/lib.rs#L429
I thought use a variable to combine a event struct and use the struct to a function (deposit_event) will not consume the variable.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: And for the next_asset_id consumed case, I reply in another conversation.

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: Actaully, the integration setup of relay+para is not complete. Several bottom implementation of test is still for an automatic execution of XCM. I put this as TODO already.

pallets/xcm-asset-manager/src/weights.rs
wangminqi: file regenerated

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: same as above.  I put this as TODO already.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: storage deleted. After deletion, we do not need the weight hint any more.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: todo: add Register filter.

pallets/xcm-asset-manager/src/lib.rs
h4n0: I have the same feeling as Kailai. The new_default_asset_type is a bit confusing to me.

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+
+			ensure!(AssetIdType::<T>::contains_key(&asset_id), Error::<T>::AssetIdDoesNotExist);
+			ensure!(
+				!AssetTypeId::<T>::contains_key(&new_asset_type),
+				Error::<T>::AssetAlreadyExists
+			);
+
+			// Insert new asset type info
+			AssetIdType::<T>::insert(&asset_id, &new_asset_type);
+			AssetTypeId::<T>::insert(&new_asset_type, &asset_id);
```
h4n0, line 303: Here is a bit misleading IMHO, although the result should be as expected(the first insert updates the value, and the second insert() adds a new k/v pair). Use `mutate` would be much more understandable.

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+	pub type AssetIdType<T: Config> =
+		StorageMap<_, Blake2_128Concat, T::AssetId, T::ForeignAssetType, OptionQuery>;
+
+	/// Reverse mapping of AssetIdType. Mapping from an asset type to an asset id.
+	/// This is mostly used when receiving a multilocation XCM message to retrieve
+	/// the corresponding asset in which tokens should me minted.
+	#[pallet::storage]
+	#[pallet::getter(fn asset_type_id)]
+	pub type AssetTypeId<T: Config> =
+		StorageMap<_, Blake2_128Concat, T::ForeignAssetType, T::AssetId, OptionQuery>;
```
h4n0, line 193: IMO it's not best practice to use 2 maps for this data structure. As we don't have a native bidirectional map container, if Id<>Type is 1<>N, we could use a map of <id, vector<Type>>. A prerequisite of this is, we don't have too many registered ids, otherwise, the lookup of a type would be slow.
The advantage of this solution is less complicated, as the current one is bit too complicated and redundant. We can discuss about this later.

wangminqi: One of advantage by your modification is, if using vector, we can query the existing asset_type of specific asset_id by storage without RPC. 

Disadvantage is the weight of transaction will be larger. Using sortedVector may better the solution.

wangminqi: https://docs.substrate.io/v3/advanced/hash-collections/
https://stackoverflow.com/questions/34305901/advantage-of-binary-search-tree-over-vector-in-c#:~:text=BST%20and%20other%20tree%20structures,memory%20usage%20while%20re%2Dsizing.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: <AssetIdType<T>>::try_mutate(id, |type| {
                type = new_type;
            }).map_err(|_| return Error::<T>::AssetIdDoesNotExist);

Something like this form.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: todo: better wrapper of internal function when modify asset_id and asset_type.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: deleted.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: todo: mutate leftover, not emergency.

pallets/xcm-asset-manager/src/lib.rs
Kailai-Wang: Thanks!

wangminqi: This is now a skipped of default 0 asset id and using try_mutate to auto increment

pallets/xcm-asset-manager/src/lib.rs
h4n0: Would this be added inline the code?

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+//!   charged per unit of second, in the case such an Asset is received as a XCM asset.
+//! - AssetTypeId: A mapping from AssetType -> AssetId.
+//! - ForeignAssetTracker: The counter of foreign assets that have been created so far.
+//! - AssetIdMetadata: An AssetId->AssetMetadata mapping that holds the metadata token info.
+//!
+//! This pallet has six extrinsics:
+//! - register_foreign_asset: Register a foreign asset in this pallet.
+//! - set_asset_units_per_second: Set the unit per second that should be charged for a particular
+//!   asset.
+//! - relocate_foreign_asset_id: Allow to manipulate ForeignAssetTracker value.
```
Kailai-Wang: This comment can be removed

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+//! - update_foreign_asset_metadata: Update the metadata of asset.
+//! - add_asset_type: Add the correspondence between AssetId and AssetType.
+//! - remove_asset_type: Remove the correspondence between AssetId and AssetType. At least one
+//!   relation must exist.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::clone_on_copy)]
+
+// TODO implement
```
Kailai-Wang: This comment can be removed(?) or do we still need something?

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+	/// Stores the tracker of foreign assets id that have been
+	/// created so far
+	#[pallet::storage]
+	#[pallet::getter(fn foreign_asset_tracker)]
+	pub type ForeignAssetTracker<T: Config> = StorageValue<_, T::AssetId, ValueQuery>;
+
+	// // Supported fee asset payments
+	// #[pallet::storage]
+	// #[pallet::getter(fn supported_fee_payment_assets)]
+	// pub type SupportedFeePaymentAssets<T: Config> = StorageValue<_, Vec<T::AssetId>, ValueQuery>;
```
Kailai-Wang: Maybe remove the unneeded code? 😂

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+
+		/// We do not allow the destroy of asset id so far; So at least one AssetTpye should be
+		/// assigned to existing AssetId Both asset_type and potential new_default_asset_type must
+		/// be an existing relation with asset_id
+		/// TODO: Change remove_asset_type with internal function wrapper
+		#[pallet::weight(T::WeightInfo::remove_asset_type())]
+		#[transactional]
+		pub fn remove_asset_type(
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
```
Kailai-Wang: I don't think we need the `asset_id` here, otherwise I can break the function by providing **existing but unmatching** `asset_id` and `asset_type`, which is incorrect but still passes the first two checks.

Kailai-Wang (changes requested):
Sorry but some minor changes needed and a major problem in `remove_asset_type`, I resolved the rest.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: agreed.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: removed

pallets/xcm-asset-manager/src/lib.rs
wangminqi: removed.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: removed

pallets/xcm-asset-manager/src/lib.rs
Kailai-Wang: Thanks, I saw the update.

Shall we also check the `new_default_asset_type` actually maps the same `asset_id` as the asset type to be removed? Right now it only checks existence.

wangminqi: I have corrected the logic accordingly and add an unittest for it.

Kailai-Wang (approved):
Thanks, we might want a different error here but let me just approve it 😂 
https://github.com/litentry/litentry-parachain/blob/727f9c5ae8a5b3d8d689023cb0defd6d6b91995a/pallets/xcm-asset-manager/src/lib.rs#L333

h4n0 (approved):

PR #600, opened 2022-05-25
init identity

suinuj:
The pr just is a mini identity management pallet implementation used to verify the ping/pong between parachain and TEE side chain.


pallets/identity-management/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+homepage = 'https://litentry.com'
+name = 'pallet-identity-management'
+repository = 'https://github.com/litentry/litentry-parachain'
+version = '0.1.0'
+
+[dependencies]
+# third-party dependencies
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "max-encoded-len"] }
+scale-info = { version = "2.1.1", default-features = false, features = ["derive"] }
+log = { version = "0.4.16", default-features = false }
```
Kailai-Wang: Can we have 
```suggestion
log = { version = "0.4", default-features = false }
```
here? The reason is once I got version conflict as other parts of code depend on `0.4` (which will lead to `0.4.17` in Cargo.lock)

runtime/litentry/Cargo.toml
```diff
@@ -84,6 +84,7 @@ pallet-bridge = { path = "../../pallets/bridge", default-features = false }
 pallet-bridge-transfer = { path = "../../pallets/bridge-transfer", default-features = false }
 pallet-drop3 = { path = "../../pallets/drop3", default-features = false }
 pallet-extrinsic-filter = { path = "../../pallets/extrinsic-filter", default-features = false }
+pallet-identity-management = { path = "../../pallets/identity-management", default-features = false }
```
Kailai-Wang: can we move it from litentry-runtime to rococo-runtime?

litmus and rococo are currently the only 2 runtimes that have `tee-dev` features (and thus used for TEE development).

Litentry-parachain will still go through the pallet activation process gradually

Kailai-Wang (changes requested):
The rest looks good to me

Kailai-Wang (approved):
Thank you!

PR #639, opened 2022-06-13
debug: simpler parameter_types

### LINKED ISSUE #615: Replace parameter_types constants with const trait
### Context
see e.g. https://github.com/paritytech/substrate/pull/11254
Using `Constxx<>` can reduce the verbosity when binding constant parameter types.

### Task
Replace the parameter_types constants with the new trait, e.g. `OperationalFeeMultiplier` 
But please consider the readability and maintainability too, e.g. don't hardcode the constant when a constant is used across multiple files or things like `pub const a: u8 = 0`.

### Acceptance criteria
parameter_types constants are replaced.

---

wangminqi:
This closes #615

### DIFF
--- a/pallets/bridge-transfer/src/mock.rs
+++ b/pallets/bridge-transfer/src/mock.rs
@@ -16,7 +16,12 @@
 
 #![cfg(test)]
 
-use frame_support::{ord_parameter_types, parameter_types, weights::Weight, PalletId};
+use frame_support::{
+	ord_parameter_types, parameter_types,
+	traits::{ConstU32, ConstU64},
+	weights::Weight,
+	PalletId,
+};
 use frame_system::{self as system};
 use hex_literal::hex;
 use sp_core::H256;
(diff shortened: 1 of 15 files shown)


Kailai-Wang:
Thanks @wangminqi, generally looks good to me.

I think we're fine to refactor all the `Max...` constants, however, I'm thinking of keeping the following variable as unchanged, what do you think?
- SS58Prefix (we are probably going to move it to some privimits/constants)
- MinimumPeriod
- PreimageMaxSize
those are calculated via arithmetic operations and subject to future changes

wangminqi:
I will keep all mock/test to constU form and recover all parachain code referred above back.

> Thanks @wangminqi, generally looks good to me.
> 
> I think we're fine to refactor all the `Max...` constants, however, I'm thinking of keeping the following variable as unchanged, what do you think?
> 
> * SS58Prefix (we are probably going to move it to some privimits/constants)
> * MinimumPeriod
> * PreimageMaxSize
>   those are calculated via arithmetic operations and subject to future changes

wangminqi:
I feel SS58Prefix should be meaningful to recover back.
MinimumPeriod is relied on SLOT_DURATION, which is already a constant. I feel no need. 
PreimageMaxSize, I am neutral to its recover. Just feel it is a "constant" number that is between worth/not worthy in the constant mod.

Kailai-Wang:
> I feel SS58Prefix should be meaningful to recover back. MinimumPeriod is relied on SLOT_DURATION, which is already a constant. I feel no need. PreimageMaxSize, I am neutral to its recover. Just feel it is a "constant" number that is between worth/not worthy in the constant mod.

Cool, let's just revert `SS58Prefix` and `PreimageMaxSize`

Kailai-Wang (approved):

PR #640, opened 2022-06-13
Ajustment to baseCallFilters

### LINKED ISSUE #635: Allow core calls in any case
### Context
We could _theoretically_ block the core calls (sudo, system, parachainSystem, timestamp) with the extrinsic filter in the current implementation. 

Although no one does that with sudo, this may not be desired. We should explicitly allow those calls in any case to avoid any misoperation.

### Task
- adjust the baseCallFilter to explicitly allow core calls
- write integration tests for the adjustments

### Acceptance criteria
implementation adjusted and test passed

---

### LINKED ISSUE #637: Eliminate magic numbers in transaction_payment integration tests
### Context
The magic numbers in transaction_payment integration tests can cause a lot of confusion, e.g.:

https://github.com/litentry/litentry-parachain/blob/4a20e2de4efad3c31025339e2ff8e72e6cafd4e7/runtime/rococo/src/tests/transaction_payment.rs#L55-L56

These should be replaced by (constant) variables.

https://github.com/litentry/litentry-parachain/blob/4a20e2de4efad3c31025339e2ff8e72e6cafd4e7/runtime/rococo/src/tests/transaction_payment.rs#L78

https://github.com/litentry/litentry-parachain/blob/4a20e2de4efad3c31025339e2ff8e72e6cafd4e7/runtime/rococo/src/tests/transaction_payment.rs#L87-L88

The formula for calculating the fee is also a bit misleading IMO.

In fact, the transaction_payment integration test already [failed](https://github.com/litentry/litentry-parachain/runs/6849631449?check_suite_focus=true) in `polkadot-v0.9.23` update (probably because upstream constant changed).

### Task
Eliminate magic numbers in transaction_payment integration tests and adjust comments how the weights/balances are calculated.

---

Kailai-Wang:
resolves #635 
resolves #637 

This PR:
- explicitly allows core calls in any case
- add integration test for baseCallFilters
- refactor transaction_payment integration-test to remove magic numbers
- activate XTokens

### DIFF
--- a/runtime/litentry/src/lib.rs
+++ b/runtime/litentry/src/lib.rs
@@ -41,7 +41,9 @@ use sp_runtime::{
 use xcm_config::{XcmConfig, XcmOriginToTransactDispatchOrigin};
 
 mod transaction_payment;
-use transaction_payment::{DealWithFees, SlowAdjustingFeeUpdate};
+pub use transaction_payment::{
+	DealWithFees, MinimumMultiplier, SlowAdjustingFeeUpdate, TargetBlockFullness,
+};
 
 #[cfg(test)]
 mod tests;
--- a/runtime/litentry/src/lib.rs
+++ b/runtime/litentry/src/lib.rs
@@ -228,7 +228,7 @@ impl frame_system::Config for Runtime {
 	/// The weight of database operations that the runtime can invoke.
 	type DbWeight = RocksDbWeight;
 	/// The basic call filter to use in dispatchable.
-	type BaseCallFilter = ExtrinsicFilter;
+	type BaseCallFilter = BaseCallFilter;
 	/// Weight information for the extrinsics of this pallet.
 	type SystemWeightInfo = weights::frame_system::WeightInfo<Runtime>;
 	/// Block & extrinsics weights: base values and limits.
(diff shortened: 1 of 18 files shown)


Kailai-Wang:
I'd need your review too @wangminqi , especially the transaction_payment part and XTokens filter part

runtime/litentry/src/lib.rs
```diff
@@ -844,31 +846,36 @@ construct_runtime! {
...
-		matches!(
-			call,
-			Call::Sudo(_) |
-			// System
-			Call::System(_) | Call::Timestamp(_) | Call::ParachainSystem(_) |
-			// ExtrinsicFilter
-			Call::ExtrinsicFilter(_)
-		)
+	fn contains(_call: &Call) -> bool {
+		false
```
wangminqi, line 878: Just to confirm, we deliberate make Safe and Normal Mode Filter same?

runtime/litmus/src/tests/base_call_filter.rs
```diff
@@ -0,0 +1,129 @@
...
+use sp_runtime::traits::Dispatchable;
+type OpaqueCall = WrapperKeepOpaque<<Runtime as frame_system::Config>::Call>;
+
+#[test]
+fn default_mode() {
+	ExtBuilder::default().build().execute_with(|| {
+		assert_eq!(ExtrinsicFilter::mode(), pallet_extrinsic_filter::OperationalMode::Normal);
+	})
+}
+
```
wangminqi, line 32: I suggest we have an extra test that root call bypasses the filter (just in case of future upstream change)

wangminqi (changes requested):

wangminqi (approved):

wangminqi:
I suggest we have an extra unittest that root call bypass the filter. But that will be a light issue.

Kailai-Wang:
> I suggest we have an extra unittest that root call bypass the filter. But that will be a light issue.

I believe `sudo` pallet unittest should already cover that, see
https://github.com/paritytech/substrate/blob/9a140a6190ae1527b89fe77bd426a38489526e44/frame/sudo/src/mock.rs#L115-L120

and the sudo calls in tests.rs still pass

wangminqi:
I mean, something not through sudo, just trigger through origin::root(), like schedule does after democracy. But should be similar. I guess it is fine.

Kailai-Wang:
> I mean, something not through sudo, just trigger through origin::root(), like schedule does after democracy. But should be similar. I guess it is fine.

You're probably talking about `call.dispatch(Origin::root())`, IMO the unittest should be again be covered by substrate, in frame_system for example. If you actually mean integration-test, yea that should be trivial/similar to the unittest body in substrate, but let me add another testcase in the next PR together with 0.9.23 update

Kailai-Wang:
@wangminqi
testcase was added in #649

PR #654, opened 2022-06-19
Relay chain unittest

wangminqi:
Xcm unittest expansion


runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -112,64 +82,64 @@ fn test_xtokens_recognize_multilocation() {
...
 
+	ParaA::execute_with(|| {
+		assert_eq!(
+			Balances::free_balance(&AccountId::from(BOB)),
+			u128::from(500_000_000_000 - UnitWeightCost::get() * 4)
+		);
+		assert_eq!(
+			Balances::free_balance(&sibling_account(2)),
+			500_000_000_000 /* When assets transfer back, the xcm fee is moved to
+			                 * XcmFeesAccount, which is Treasury */
```
Kailai-Wang: shall we check the treasury balance if the fee goes there?

runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -182,33 +152,867 @@ fn test_pallet_xcm_recognize_multilocation() {
...
+		);
+	});
+
+	ParaB::execute_with(|| {
+		assert_eq!(
+			Tokens::free_balance(
+				1, // Asset_id=1. The first registered Token in Para B
+				&AccountId::from(BOB)
+			),
+			u128::from(2_000_000_000_000 - UnitWeightCost::get() * 4)
```
Kailai-Wang: Can we have a comment to explain why `UnitWeightCost::get() * 4` here? As it's not pre-defined anywhere like the other XTokens::transfer

runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -112,64 +82,64 @@ fn test_xtokens_recognize_multilocation() {
...
 	});
 
+	ParaA::execute_with(|| {
+		assert_eq!(
+			Balances::free_balance(&AccountId::from(BOB)),
+			u128::from(500_000_000_000 - UnitWeightCost::get() * 4)
+		);
+		assert_eq!(
+			Balances::free_balance(&sibling_account(2)),
+			500_000_000_000 /* When assets transfer back, the xcm fee is moved to
```
Kailai-Wang: `500_000_000_000/500_000_000_000` might not be the best testcase here, as you can't tell which one is subtractor which one is remainder => maybe 6/4 or 7/3

runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -182,33 +152,867 @@ fn test_pallet_xcm_recognize_multilocation() {
...
+			1_000_000_000_000,
+		);
+
+		// Sending 10 ParaA token after xcm fee to BOB by XTokens::transfer_multiasset
+		assert_ok!(XTokens::transfer_multiasset(
+			Origin::signed(AccountId::from(ALICE)),
+			Box::new(
+				MultiAsset {
+					id: Concrete(para_native_token_multilocation(1)),
+					fun: Fungible((UnitWeightCost::get() * 4 + 10).into())
```
Kailai-Wang: Any specific reason to use this number `UnitWeightCost::get() * 4 + 10`? Can't we use some constant like 0,5 * UNIT?

runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -182,33 +152,867 @@ fn test_pallet_xcm_recognize_multilocation() {
...
+			Balances::free_balance(&sibling_account(2)),
+			1_000_000_000_000 + u128::from(UnitWeightCost::get() * 4 + 10)
+		);
+
+		// Sending 100 ParaA token after xcm fee to BOB by XTokens::transfer_with_fee
+		assert_ok!(XTokens::transfer_with_fee(
+			Origin::signed(AccountId::from(ALICE)),
+			CurrencyId::SelfReserve,
+			100,
+			(UnitWeightCost::get() * 4).into(),
```
Kailai-Wang, line 318: Possible to use a different number as `fee` here? (Same reason for 50/50 above => to increase the randomness of tests)

Kailai-Wang (commented):
Thank you for many testcases, I didn't look into every single detail of them but most of them make sense to me 👍 

Apart from the inline comments, there're a few extra questions/points from my side:
1. Can we replace `AccountId::from(ALICE)` with `alice()` as it's already defined in setup? (same applies to bob)
2. Instead of `1_000_000_000_000`, can we use 1 * UNIT? As it increases the readability
3. The fact that this file `tests/xcm_parachain/mod.rs` actually contains all the xcm-tests is a bit confusing => I might prefer to have a file like `xcm_parachain.rs` and move the relayer_proof_builder to setup
4. I see both the XTokens version and raw version(= polkadot_xcm pallet) are tested, is it necessary? I mean if the XTokens interface is the **only** interface that we should call (as it's a wrapper to underlying xcms), then it's enough to just test xtokens => I might understand it wrongly though
5. As mentioned in the comments, `UnitWeightCost::get() * 4` is really used a lot across the file, maybe we should give it a variable to avoid repeating it. And it should only be used for one purpose (like dest_weight, but not as both dest_weight and fee)
6. Shall we have a testcase where non-registered assets can be handled correctly. I see there's a testcase for non-existing MultiLocation (`parachain(4)`), but I'm not sure if that's enough.

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: > Thank you for many testcases, I didn't look into every single detail of them but most of them make sense to me 👍
> 
> Apart from the inline comments, there're a few extra questions/points from my side:
> 
> 1. Can we replace `AccountId::from(ALICE)` with `alice()` as it's already defined in setup? (same applies to bob)
> 2. Instead of `1_000_000_000_000`, can we use 1 * UNIT? As it increases the readability
> 3. The fact that this file `tests/xcm_parachain/mod.rs` actually contains all the xcm-tests is a bit confusing => I might prefer to have a file like `xcm_parachain.rs` and move the relayer_proof_builder to setup
> 4. I see both the XTokens version and raw version(= polkadot_xcm pallet) are tested, is it necessary? I mean if the XTokens interface is the **only** interface that we should call (as it's a wrapper to underlying xcms), then it's enough to just test xtokens => I might understand it wrongly though
> 5. As mentioned in the comments, `UnitWeightCost::get() * 4` is really used a lot across the file, maybe we should give it a variable to avoid repeating it. And it should only be used for one purpose (like dest_weight, but not as both dest_weight and fee)
> 6. Shall we have a testcase where non-registered assets can be handled correctly. I see there's a testcase for non-existing MultiLocation (`parachain(4)`), but I'm not sure if that's enough.

1. changed
2. make sense, changed
3. I have thought this too. Not only relay_sproof, but also methods like register_channel_info should integrate in setup. Will determine the specific solution in future.
4. I think it will be good to know the risk. If we want to manipulate our soverign account on other chains, we must allow something send xcm commands to certain degree of freedom, which means any xcm can be mimic. You can see the pallet_xcm's test is focus on execute and send methods. And I really think we should maintains these test.
5. There is actually no replacement, so if there are 4=n xcm commands, then it charge n times. It is just most call ending with 4 xcm command combo. I will add some comments around.
6.  That is tricky to do so, as it is otherparachains' freedom to select whatever multilocation they want. We may only test from if they are sibling chain, if they are relaychain etc.

wangminqi:
1. changed
2. make sense, changed
3. I have thought this too. Not only relay_sproof, but also methods like register_channel_info should integrate in setup. Will determine the specific solution in future.
4. I think it will be good to know the risk. If we want to manipulate our soverign account on other chains, we must allow something send xcm commands to certain degree of freedom, which means any xcm can be mimic. You can see the pallet_xcm's test is focus on execute and send methods. And I really think we should maintains these test.
5. There is actually no replacement, so if there are 4=n xcm commands, then it charge n times. It is just most call ending with 4 xcm command combo. I will add some comments around.
6. That is tricky to do so, as it is otherparachains' freedom to select whatever multilocation they want. We may only test from if they are sibling chain, if they are relaychain etc.

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: I add tests acccordingly. One update of my mistake. So non-native asset served as XCM fee will go to treasury, local native token will just get burned.

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: It is linked to the number of XCM commands sending in the end. Maybe no replacement here

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: Changed

runtime/litmus/src/tests/xcm_parachain/mod.rs
Kailai-Wang: I see, thanks.
Then the word `treasury` is a bit misleading as normally it's referred to the treasury account in the treasury pallet 😂

runtime/litmus/src/tests/xcm_parachain/mod.rs
Kailai-Wang: OK, let's add a one-line comment here

Kailai-Wang:
> 1. changed
> 2. make sense, changed
> 3. I have thought this too. Not only relay_sproof, but also methods like register_channel_info should integrate in setup. Will determine the specific solution in future.
> 4. I think it will be good to know the risk. If we want to manipulate our soverign account on other chains, we must allow something send xcm commands to certain degree of freedom, which means any xcm can be mimic. You can see the pallet_xcm's test is focus on execute and send methods. And I really think we should maintains these test.
> 5. There is actually no replacement, so if there are 4=n xcm commands, then it charge n times. It is just most call ending with 4 xcm command combo. I will add some comments around.
> 6. That is tricky to do so, as it is otherparachains' freedom to select whatever multilocation they want. We may only test from if they are sibling chain, if they are relaychain etc.

1. Thanks, but we still have many `AccountId::from(ALICE)` and BOBs 😂  => maybe we can use `find and replace` from the vscode..
2. It's not completely changed either 😂 We have like `500_000_000_000_000_000`
3. OK
4. Got it
5. Then we would need comments (e.g. at the beginning of the file as it's used in multiple locations). However, I mean the places where such a constant isn't necessary, e.g.`dest_weight` in the XTokens::transfer => there we can declare/test whichever weights we want if I understand it correctly. Same argument with one of the inline comments => we want to "randomize" the tests as much as possible
6. I see, it's not about an unknown dest location, but a simple test for an unknown/unregistered currency. Imagine moonriver users try to send tokens to Litentry-parachain with MOVR, which is not yet registered.

wangminqi:
> 1. AccountId::from(ALICE)

sorry, I forget the moonbase and rococo branch modification.

wangminqi:
test for dest_weight not enough should be added.

Kailai-Wang:
and the case where the `dest_weight` exceeds UnitWeight * 4

wangminqi:
Both redundant and insufficient weight parameter input of orml_xtoken is added.
Numbers of currency is in CONSTANT representation now. weight unchanged.

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: changed.

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: changed

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: check treasury account test added

runtime/litmus/src/tests/xcm_parachain/mod.rs
wangminqi: changed.

Kailai-Wang (approved):
Thanks!

PR #677, opened 2022-07-01
Select chains to benchmark in the benchmarking CI

### LINKED ISSUE #647: Select chains to benchmark in the benchmarking CI
### Context
Due to a lack of multi-select box in GHA, we need to combine the options manually.
In the current [.github/workflows/benchmark-runtime-weights.yml](.github/workflows/benchmark-runtime-weights.yml) there's no way to choose which chains to benchmark (you can select one chain each time, or use "all"). It makes it impossible to do rococo + litmus benchmarking (e.g. for pallet-asset-manager which doesn't exist on litentry yet) in a single run.

image

### Task
Adjust the script so that we have a way to combine the chains to benchmark freely. It doesn't matter if the interface is verbose (e.g. we have a true/false option for each chain), as long as it can achieve what we need.

---

ziming-zung:
resolves #647

### DIFF
--- a/.github/workflows/benchmark-runtime-weights.yml
+++ b/.github/workflows/benchmark-runtime-weights.yml
@@ -3,14 +3,21 @@ name: Benchmark runtime weights
 on:
   workflow_dispatch:
     inputs:
-      chain:
-        type: choice
-        description: The chain whose runtime is benchmarked, all = rococo+litmus+litentry
-        options:
-          - all
-          - litmus
-          - litentry
-          - rococo
+      litentry:
+        type: boolean
+        description: The litentry will be benchmarked
+        required: true
+        default: true
+      litmus:
+        type: boolean
+        description: The litmus will be benchmarked
+        required: true
+        default: true
+      rococo:
+        type: boolean
+        description: The rococo will be benchmarked
+        required: true
+        default: true
       pallets:
         description: pallets to benchmark, * for all, or comma listed (e.g. frame-system,pallet-proxy)
         default: "*"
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark-runtime-weights.yml
```diff
@@ -3,14 +3,21 @@ name: Benchmark runtime weights
...
-        type: choice
-        description: The chain whose runtime is benchmarked, all = rococo+litmus+litentry
-        options:
-          - all
-          - litmus
-          - litentry
-          - rococo
+      litentry:
+        type: boolean
+        description: The litentry will be benchmarked
```
Kailai-Wang, line 8: Can we simply write "litentry/litmus/rococo" to be less verbose?

.github/workflows/benchmark-runtime-weights.yml
```diff
@@ -24,6 +31,20 @@ jobs:
...
+          if [ "${{ github.event.inputs.litmus }}" = "true" ]; then
+            chain="$chain litmus"
+          fi
+          if [ "${{ github.event.inputs.litentry }}" = "true" ]; then
+            chain="$chain litentry"
+          fi
+          if [ "${{ github.event.inputs.rococo }}" = "true" ]; then
+            chain="$chain rococo"
+          fi
+          echo "CHAIN=$chain" >> $GITHUB_ENV
```
Kailai-Wang, line 50: I believe we have to exit early if `chain` is empty -- no docker image has to be built if so

.github/workflows/benchmark-runtime-weights.yml
```diff
@@ -42,12 +63,12 @@ jobs:
...
+      - name: Remotely benchmark pallets ${{ github.event.inputs.pallets }} for ${{ env.CHAIN }}
         timeout-minutes: 240
         run: |
           # prepend the asterisk with \ to go through ssh
           arg="${{ github.event.inputs.pallets }}"
-          chain="${{ github.event.inputs.chain }}"
+          chain="${{ env.CHAIN }}"
           if [ "$arg" = "*" ]; then
             arg="\\$arg";
           fi
```
Kailai-Wang, line 78: we do not need `if [ "$chain" = "all" ]` branch below

Kailai-Wang (commented):
Basically looks good, just a few small things (see inline comments)

.github/workflows/benchmark-runtime-weights.yml
ziming-zung: Sorry, I can't find any documentation on github that can define the same type of input in one line

.github/workflows/benchmark-runtime-weights.yml
ziming-zung: correct it

.github/workflows/benchmark-runtime-weights.yml
Kailai-Wang: I simply mean we don't need to write "will be benchmarked" as it's in a benchmarking GHA by itself 😂 
but it's OK

Kailai-Wang (approved):

PR #678, opened 2022-07-01
Add another testcase for `pallet-extrinsic-filter` unittests

Kailai-Wang:
This PR adds another testcase for `pallet-extrinsic-filter` unittests: the blocked extrinsics should keep blocked upon mode switch.


BillyWooo (approved):
looks good for me.

PR #681, opened 2022-07-04
Adjust runtime governance config

### LINKED ISSUE #667: Double-check the runtime parameters and configuration

Kailai-Wang:
resolves #667


runtime/common/src/lib.rs
```diff
@@ -50,3 +55,42 @@ macro_rules! prod_or_fast {
 		}
 	};
 }
+
+/// Instance definition for council and technical committee
+pub type CouncilInstance = pallet_collective::Instance1;
+pub type TechnicalCommitteeInstance = pallet_collective::Instance2;
+pub type CouncilMembershipInstance = pallet_membership::Instance1;
+pub type TechnicalCommitteeMembershipInstance = pallet_membership::Instance2;
```
wangminqi, line 63: This is usage really confused me. Are we allowed to use them in this way? I mean "Instance1" "Instance2", I though they are defined after frame_support::construct_runtime! macro.

runtime/litentry/src/lib.rs
```diff
@@ -463,39 +466,30 @@ impl pallet_democracy::Config for Runtime {
 	type VoteLockingPeriod = EnactmentPeriod; // Same as EnactmentPeriod
 	type MinimumDeposit = MinimumDeposit;
 	/// A straight majority of the council can decide what their next motion is.
-	type ExternalOrigin =
-		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 1, 2>;
+	type ExternalOrigin = EnsureRootOrHalfCouncil;
 	/// A super-majority can have the next scheduled referendum be a straight majority-carries vote.
-	type ExternalMajorityOrigin =
-		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 3, 4>;
+	type ExternalMajorityOrigin = EnsureRootOrTwoThirdsCouncil;
```
wangminqi, line 471: Just confirm it is two third rather than three fourth we would like to have

wangminqi (changes requested):

runtime/litentry/src/lib.rs
Kailai-Wang: > Just confirm it is two third rather than three fourth we would like to have

I know the original one is 3/4, but I just wanted 2/3 for simplicity -- it's every team's choice, some teams use 3/5, some teams use 1/2

wangminqi:
I see quite a lot of changes on non-democracy parts from "EnsureRoot" into "EnsureRootOr***". I am wondering if we should keep it the old way. I feel some of them should be just triggered by democracy root, or we prefer to maintain some special power around?

runtime/common/src/lib.rs
Kailai-Wang: > This is usage really confused me. Are we allowed to use them in this way? I mean "Instance1" "Instance2", I though they are defined after frame_support::construct_runtime! macro.

yea should be possible, see https://github.com/AcalaNetwork/Acala/blob/d6ad56e209b0c4c08736492d3bc8410deb0c6573/runtime/common/src/lib.rs#L204-L213

wangminqi: Just can not help ask even the test turns out fine.

Kailai-Wang:
> I see quite a lot of changes on non-democracy parts from "EnsureRoot" into "EnsureRootOr***". I am wondering if we should keep it the old way. I feel some of them should be just triggered by democracy root, or we prefer to maintain some special power around?

The problem is if we want those actions **only** to be triggered by a referrenda, as it's the only way to dispatch the call as Root later. Involving Council is more efficient but less decentralised. Teams like phala almost use RootOrHalfCouncil all over the place.

wangminqi:
> > I see quite a lot of changes on non-democracy parts from "EnsureRoot" into "EnsureRootOr***". I am wondering if we should keep it the old way. I feel some of them should be just triggered by democracy root, or we prefer to maintain some special power around?
> 
> The problem is if we want those actions **only** to be triggered by a referrenda, as it's the only way to dispatch the call as Root later. Involving Council is more efficient but less decentralised. Teams like phala almost use RootOrHalfCouncil all over the place.

Got it.

runtime/common/src/lib.rs
wangminqi: https://github.com/paritytech/substrate/blob/master/frame/support/src/instances.rs
No worry, I find their hardcode...

Kailai-Wang:
> > > I see quite a lot of changes on non-democracy parts from "EnsureRoot" into "EnsureRootOr***". I am wondering if we should keep it the old way. I feel some of them should be just triggered by democracy root, or we prefer to maintain some special power around?
> > 
> > 
> > The problem is if we want those actions **only** to be triggered by a referrenda, as it's the only way to dispatch the call as Root later. Involving Council is more efficient but less decentralised. Teams like phala almost use RootOrHalfCouncil all over the place.
> 
> Got it.

The only two `EnsureRoot<>` left are:
- `BlacklistOrigin` in democracy
- `ManagerOrigin` in preimage
 
I refer other other teams' implementation to eventually decide to retain them. Plus those two are rarely used.

runtime/litentry/src/lib.rs
```diff
@@ -630,16 +612,16 @@ impl cumulus_pallet_xcmp_queue::Config for Runtime {
 	type ChannelInfo = ParachainSystem;
 	// We use pallet_xcm to confirm the version of xcm
 	type VersionWrapper = PolkadotXcm;
-	type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
-	type ControllerOrigin = EnsureRoot<AccountId>;
+	type ExecuteOverweightOrigin = EnsureRootOrHalfCouncil;
```
wangminqi: The ExecuteOverweightOrigin is for executing an existing XCM message in the queue from a remote chain to be executed without weight check. I only see EnsureRoot implementation from Moonbeam/Acala. And this is a very rarely used and delicate extrinsic setting from my perspective. Maybe we switch it to tech member + root or just root.

runtime/litentry/src/lib.rs
```diff
@@ -630,16 +612,16 @@ impl cumulus_pallet_xcmp_queue::Config for Runtime {
...
+	type ControllerOrigin = EnsureRootOrHalfCouncil;
 	type ControllerOriginConverter = XcmOriginToTransactDispatchOrigin;
 	type WeightInfo = weights::cumulus_pallet_xcmp_queue::WeightInfo<Runtime>;
 }
 
 impl cumulus_pallet_dmp_queue::Config for Runtime {
 	type Event = Event;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
+	type ExecuteOverweightOrigin = EnsureRootOrHalfCouncil;
```
wangminqi: Same for cumulus_pallet_xcmp_queue

runtime/litentry/src/lib.rs
```diff
@@ -381,7 +384,7 @@ impl pallet_scheduler::Config for Runtime {
 	type PalletsOrigin = OriginCaller;
 	type Call = Call;
 	type MaximumWeight = MaximumSchedulerWeight;
-	type ScheduleOrigin = EnsureRoot<AccountId>;
+	type ScheduleOrigin = EnsureRootOrHalfCouncil;
```
wangminqi: Just my personal feeling. So voting through democarcy leads to schedule in the end, which means this origin should be stricter than voting.

And I check other projects, both moonbeam and acala use strict root. We just use root too?

runtime/litentry/src/lib.rs
wangminqi: solved

runtime/litentry/src/lib.rs
```diff
@@ -630,16 +612,16 @@ impl cumulus_pallet_xcmp_queue::Config for Runtime {
 	type ChannelInfo = ParachainSystem;
 	// We use pallet_xcm to confirm the version of xcm
 	type VersionWrapper = PolkadotXcm;
-	type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
-	type ControllerOrigin = EnsureRoot<AccountId>;
+	type ExecuteOverweightOrigin = EnsureRootOrHalfCouncil;
+	type ControllerOrigin = EnsureRootOrHalfCouncil;
```
wangminqi: I feel okay with controllerOrigin. Just it makes more sense if it is tech member + root, as this one is for global disable/enable xcmp channel queue.

wangminqi (changes requested):

runtime/litentry/src/lib.rs
Kailai-Wang: Acala actually uses `EnsureRootOrHalfCouncil`: https://github.com/AcalaNetwork/Acala/blob/d6ad56e209b0c4c08736492d3bc8410deb0c6573/runtime/acala/src/xcm_config.rs#L213
Moonbeam uses only Root.

I'm fine with either, what's your preference on this?

runtime/litentry/src/lib.rs
Kailai-Wang: Sure, let's use Root, thanks!

runtime/litentry/src/lib.rs
Kailai-Wang: Same as `ExecuteOverweightOrigin`, I'm fine with either, simply let me know your preference and I'll change accordingly :)

runtime/litentry/src/lib.rs
wangminqi: Root only then

runtime/litentry/src/lib.rs
wangminqi: root only then.

wangminqi (approved):

PR #688, opened 2022-07-04
Select type to release in the release CI 

ziming-zung:
I'm not sure if these changes will make the code better than before, it seems a bit verbose. Would  `choice-input` be better if release_type had only 3 types(both/client/runtime)?


Kailai-Wang (approved):
Thanks.

It's more verbose in the GHA code, but it's cleaner from the user's experience IMO and that's the goal of coding :)

PR #691, opened 2022-07-05
Use release and production consistently

Kailai-Wang:
This PR tries to use `release` and `production` as cargo profiles across multiple files consistently:
before: dev & prod
now: release & production


BillyWooo (approved):
Looks good for me. Naming is more clearer

PR #700, opened 2022-07-13
Pallet staking replacement

wangminqi:
Replace CollatorSelection with parachain staking


wangminqi:
I would not upgrade it to dev even if it passed.
But let us do the code review before rococo runtime grape test.  
Local runtime upgrade have passed.

wangminqi:
One flaw. So we want to only have collators held by us. Need to discuss this logic.

wangminqi:
> One flaw. So we want to only have collators held by us. Need to discuss this logic.

features updated. Done. We have collator whitelist now.

pallets/parachain-staking/Cargo.toml
```diff
@@ -0,0 +1,49 @@
+[package]
```
Kailai-Wang, line 1: We need to prefix the pallet name as `pallet-*`, otherwise some benchmarking scripts won't work.
Also the author and version adjustment please

pallets/parachain-staking/src/inflation.rs
```diff
@@ -0,0 +1,201 @@
...
+use serde::{Deserialize, Serialize};
+use sp_runtime::{PerThing, Perbill, RuntimeDebug};
+use substrate_fixed::{
+	transcendental::pow as floatpow,
+	types::{I32F32, I64F64},
+};
+
+const SECONDS_PER_YEAR: u32 = 31557600;
+const SECONDS_PER_BLOCK: u32 = 12;
+pub const BLOCKS_PER_YEAR: u32 = SECONDS_PER_YEAR / SECONDS_PER_BLOCK;
```
Kailai-Wang: we could add a YEARS const in `primitives/src/lib.rs` and use it from there

runtime/litentry/src/lib.rs
```diff
@@ -650,24 +652,53 @@ impl pallet_aura::Config for Runtime {
...
-	type MaxInvulnerables = ConstU32<100>;
-	// should be a multiple of session or things will get inconsistent
-	type KickThreshold = Period;
-	type ValidatorId = <Self as frame_system::Config>::AccountId;
-	type ValidatorIdOf = pallet_collator_selection::IdentityCollator;
-	type ValidatorRegistration = Session;
-	type WeightInfo = weights::pallet_collator_selection::WeightInfo<Runtime>;
+	type MonetaryGovernanceOrigin = EnsureRoot<AccountId>;
+	/// Minimum round length is 2 minutes (10 * 12 second block times)
+	type MinBlocksPerRound = ConstU32<10>;
```
Kailai-Wang: Let's make it Const32<{ 2 * MINUTES }> to be more consistent

runtime/litentry/src/migration.rs
```diff
@@ -0,0 +1,197 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Litentry is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
```
Kailai-Wang, line 15: The migration generally looks good! Two small things:
- we should compile it once with `try-runtime` features on to make sure it goes through, I see some potential compilation errors like `<Vec<AccountId>>` in `pre_upgrade`. Eventually we should run try-runtime once with rococo/litmus
- we can add more checks in post_upgrade, like the round count, the number of candidates, total stakes etc

runtime/litentry/src/lib.rs
```diff
@@ -650,24 +652,53 @@ impl pallet_aura::Config for Runtime {
...
+	type DefaultCollatorCommission = DefaultCollatorCommission;
+	type DefaultParachainBondReservePercent = DefaultParachainBondReservePercent;
+	/// Minimum stake required to become a collator
+	type MinCollatorStk = ExistentialDeposit;
+	/// Minimum stake required to be reserved to be a candidate
+	type MinCandidateStk = ExistentialDeposit;
+	/// Minimum stake required to be reserved to be a delegator
+	type MinDelegation = ExistentialDeposit;
+	/// Minimum stake required to be reserved to be a delegator
+	type MinDelegatorStk = ExistentialDeposit;
```
Kailai-Wang: This would definitely need a new config, maybe to have @qiuhaiguo involved.

pallets/parachain-staking/Cargo.toml
wangminqi: done.

pallets/parachain-staking/src/inflation.rs
wangminqi: Would you prefer 365.25 days per year or 365?

wangminqi: 365.25 now

runtime/litentry/src/lib.rs
wangminqi: replace done.

runtime/litentry/src/migration.rs
wangminqi: Compile is fine.
I add two more tests for RoundInfo and initial candidates. 
The test for Total stake is not in try-runtime, which just checks if the collator comes with enough money to support the initial staking condition, and available after the Parachain Staking introduced.

Kailai-Wang (approved):
Approved. Didn't real every single line but the overall code logic is on-point and the unittests look good.

I do have concerns about the runtime configuration, we probably need another PR to tweak runtime parameters, including:
- origins (do we allow the council to manage it, or only root?)
- callFilters (shall we block some fn at the beginning and gradually open it)
- round-related numbers
- inflation-related numbers
- stake/delegation-related numbers

We need to test it thoroughly too:
- collator joining and leaving
- staking reward distribution
- delegations
- inflations
- any corner cases?
- do we need a frontend UI for all the operations for normal users?

I'd suggest @zhizming-zhong to take a look at this PR too and participate in the testing as he had experience with DeFi and tokenomics as such.

wangminqi:
> * o we need a frontend UI for all the operations for normal users?

Thanks. No further Merge will be performed before we fully test the rococo environment and ready the parameters in prod.

Kailai-Wang:
We are at 0.9.26 btw

wangminqi:
I have some modification around bridge transfer, which I think need your review/confirm. @zhizming-zhong

ziming-zung (approved):
Looks good to me.

PR #701, opened 2022-07-14
Extract the common code

### LINKED ISSUE #672: Extract the common code in runtime/ implementations
### Context
Currently we have some repeated code in different `runtime` implementations. E.g. each `runtime/*` has a `constants.rs`, where similar stuff is coded, the integration tests have many common setups too. 

### Task
Extract/separate the common code of different runtime and integration-tests into a dedicated package.
Basically,
`<root>/primitives` should contain the primitives that are used by both client and runtime
`<root>/runtime/common` should contain the common part that used by different runtime implementations.

---

ziming-zung:
resolves #672


Kailai-Wang:
The mode of this file needs to be reverted too
image

runtime/common/src/xcm_impl.rs
```diff
@@ -0,0 +1,491 @@
...
+}
+
+impl<R: BaseRuntimeRequirements> From<Option<MultiLocation>> for CurrencyId<R> {
+	fn from(location: Option<MultiLocation>) -> Self {
+		match location {
+			Some(a)
+				if (a == OldAnchoringSelfReserve::<R>::get()) |
+					(a == NewAnchoringSelfReserve::<R>::get()) =>
+				CurrencyId::<R>::SelfReserve(PhantomData::default()),
+			Some(multi) => CurrencyId::ParachainReserve(Box::new(multi)),
```
Kailai-Wang: I believe here could be further simplified, like:
```suggestion
			Some(multi) => Self::from(multi),
```

runtime/common/src/xcm_impl.rs
```diff
@@ -0,0 +1,491 @@
...
+			if reserve == origin {
+				return true
+			}
+		}
+		false
+	}
+}
+
+/// A set of pallet_config that runtime must implement.
+pub trait BaseRuntimeRequirements:
```
Kailai-Wang: The idea to have a base runtime config constraint is good.

Shall we move `BaseRuntimeRequirements` and `ParaRuntimeRequirements` to a separate file/mod as it's used elsewhere too and not only to `xcm_impl`

runtime/litmus/src/tests/setup/mod.rs
```diff
@@ -1,193 +0,0 @@
-// Copyright 2020-2022 Litentry Technologies GmbH.
```
Kailai-Wang, line 1: Actually this file can be deleted?

runtime/litentry/src/tests/setup.rs
```diff
@@ -15,18 +15,17 @@
...
 
 pub use crate::{
-	AccountId, Balance, Balances, Call, Event, ExtrinsicFilter, MinimumMultiplier, Multisig,
-	Origin, ParachainSystem, Runtime, RuntimeBlockWeights, SlowAdjustingFeeUpdate, System,
-	TargetBlockFullness, TransactionByteFee, TransactionPayment, Treasury, Vesting, UNIT,
+	AccountId, Balance, Balances, Call, Event, ExtrinsicFilter, Multisig, Origin, ParachainSystem,
+	Runtime, System, TransactionByteFee, TransactionPayment, Treasury, Vesting,
 };
+
 pub const ALICE: [u8; 32] = [1u8; 32];
```
Kailai-Wang: I think many of these constants/fns are not needed anymore, like ALICE, alice(), info_from_weight(), ExtBuilder ... => they were moved to common setups already.

Same applies to moonbase and rococo files

runtime/common/src/xcm_impl.rs
```diff
@@ -0,0 +1,491 @@
...
+	SelfReserve,
+	ParachainReserve(Box<MultiLocation>),
+}
+
+// Our currencyId. We distinguish for now between SelfReserve, and Others, defined by their Id.
+#[derive(Clone, Eq, Debug, PartialEq, Encode, Decode, TypeInfo)]
+pub enum CurrencyId<R: BaseRuntimeRequirements> {
+	SelfReserve(PhantomData<R>), // The only parachain native token: LIT
+	ParachainReserve(Box<MultiLocation>), /* Any parachain based asset, including local native minted
+	                              * ones. */
```
Kailai-Wang: Can we format this comment a bit please, e.g. use `// ...` in one line

runtime/common/src/xcm_impl.rs
```diff
@@ -0,0 +1,491 @@
...
+			} else {
+				Err(())
+			}
+		} else {
+			Err(())
+		}
+	}
+}
+
+// implement different types of converters:
```
Kailai-Wang: We can delete these comments if not needed, unless it's worth keeping it in your opinion

runtime/common/src/xcm_impl.rs
```diff
@@ -0,0 +1,491 @@
...
+
+#[derive(Clone, Eq, Debug, PartialEq, Ord, PartialOrd, Encode, Decode, TypeInfo)]
+pub enum CurrencyId4Compare {
+	SelfReserve,
+	ParachainReserve(Box<MultiLocation>),
+}
+
+// Our currencyId. We distinguish for now between SelfReserve, and Others, defined by their Id.
+#[derive(Clone, Eq, Debug, PartialEq, Encode, Decode, TypeInfo)]
+pub enum CurrencyId<R: BaseRuntimeRequirements> {
```
Kailai-Wang, line 239: May I ask why we need `CurrencyId<R>` instead of `CurrencyId`? As the enum members don't seem to use <R> anyway

Kailai-Wang (changes requested):
Despite a few inline comments, I generally like the idea of using macros to refactor out the common code as well as the RuntimeConfig extraction, thanks!

runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -14,1148 +14,96 @@
 // You should have received a copy of the GNU General Public License
 // along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
 
-use super::setup::*;
```
Kailai-Wang, line 17: I believe the whole file can be renamed to `xcm_parachain.rs` to reduce confusion (we don't need an extra mod.rs here)

runtime/litmus/src/tests/xcm_parachain/mod.rs
```diff
@@ -14,1148 +14,96 @@
...
-	.unwrap()
-}
-
-fn sibling_account(x: u32) -> AccountId32 {
-	<LocationToAccountId as xcmConvert<MultiLocation, AccountId32>>::convert(
-		(Parent, Parachain(x)).into(),
-	)
-	.unwrap()
-}
+impl TestXCMRequirements for LitmusTestXCMRequirements {
```
Kailai-Wang: If I try to be picky here:
we can even put these trait binding into macro, as well as the following common test cases, so eventually we'll only have something like in this file
```
run_xcm_tests!(LitmusTestXCMRequirements);
```
or maybe we don't even need `LitmusTestXCMRequirements` if the runtime binding is the same for all chains

runtime/common/src/xcm_impl.rs
ziming-zung: move it to common lib.rs

runtime/litmus/src/tests/setup/mod.rs
ziming-zung: had been deleted.

runtime/litentry/src/tests/setup.rs
ziming-zung: had been deleted.

runtime/common/src/xcm_impl.rs
ziming-zung: had been deleted.

runtime/common/src/xcm_impl.rs
ziming-zung: done.

runtime/litmus/src/tests/xcm_parachain/mod.rs
ziming-zung: Renamed this file to `xcm_parachain.rs`

runtime/litmus/src/tests/xcm_parachain/mod.rs
ziming-zung: Extract common tests into macros, inclue: `base_call_filter` `transaction_payment` `xcm_parachain`

runtime/common/src/xcm_impl.rs
ziming-zung: The type parameter 'R' is used by some traits, like `From<MultiLocation>` `From<CurrencyId<R>>`

Kailai-Wang (approved):
I approve it, as mentioned we still need to revert the generics in CurrencyId

PR #706, opened 2022-07-19
Add start & stop script for aws instance

### LINKED ISSUE #566: Automate the start/stop of AWS instance for parachain benchmarking
### Context
Currently we need to manually start/stop the AWS instance that is used for parachain benchmarking. Ideally it could be started/stopped on-demand (that is when `benchmark-runtime-weights.yml` is run).

An elastic IP is allocated to make sure the public IP won't change for every start/stop cycle. It's not free, we get charged (a small fee) for keeping the IP when the instance is not running.

### Task
Adjust the `benchmark-runtime-weights.yml` so that the AWS instance is started/stopped on-demand from the command line.

### Acceptance criteria
Benchmarking is successfully carried out with the new script.

---

chenzongxiong:
I think this script isn't perfect enough. it doesn't consider

- [x]  the edge of network connections between ci server and remote server. 
- [x] the case of `start-instances` succeeds, however, `stop-instances` fails.
- [x] fetch dynamic ip from restarted instance

I think the second case should included in this case. 
I will refactor the script.

### DIFF
--- a/.github/workflows/benchmark-runtime-weights.yml
+++ b/.github/workflows/benchmark-runtime-weights.yml
@@ -54,17 +54,17 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Build docker image
-        run: |
-          ./scripts/build-docker.sh production runtime-benchmarks --features=runtime-benchmarks
+      # - name: Build docker image
+      #   run: |
+      #     ./scripts/build-docker.sh production runtime-benchmarks --features=runtime-benchmarks
 
-      - name: Push docker image
-        run: |
-          docker push litentry/litentry-parachain:runtime-benchmarks
+      # - name: Push docker image
+      #   run: |
+      #     docker push litentry/litentry-parachain:runtime-benchmarks
 
-      - name: Remove dangling images if any
-        run: |
-          [ -z "$(docker images --filter=dangling=true -q)" ] || docker rmi -f $(docker images --filter=dangling=true -q)
+      # - name: Remove dangling images if any
+      #   run: |
+      #     [ -z "$(docker images --filter=dangling=true -q)" ] || docker rmi -f $(docker images --filter=dangling=true -q)
 
       - name: Start remote instance
         id: start_instance
--- a/.github/workflows/benchmark-runtime-weights.yml
+++ b/.github/workflows/benchmark-runtime-weights.yml
@@ -71,7 +71,11 @@ jobs:
           sleep 1
           ret=`aws ec2 describe-instance-status --instance-ids i-0a3bdc24e56b4b457 | jq '.InstanceStatuses[0].InstanceState.Name'`
           cnt=0
-          while [ "$ret" != "running" || "$cnt" < "5" ]; do ret=`aws ec2 describe-instance-status --instance-ids i-0a3bdc24e56b4b457 | jq '.InstanceStatuses[0].InstanceState.Name'`; cnt=$((cnt + 1)); sleep 1; done
+          while [[ "$ret" != "running" || $cnt -gt 5 ]]; do
+            ret=`aws ec2 describe-instance-status --instance-ids i-0a3bdc24e56b4b457 | jq '.InstanceStatuses[0].InstanceState.Name'`
+            cnt=$((cnt + 1))
+            sleep 1
+          done
           echo "Remote instance running state: $ret, retry count: $cnt"
           ip=`aws ec2 describe-instances --filters 'Name=instance-state-name,Values=running' 'Name=instance-id,Values=i-0a3bdc24e56b4b457' --query 'Reservations[*].Instances[*].[PublicIpAddress]' | jq '.[0][0][0]'`
           echo "Running instances ip address: $ip"
--- a/.github/workflows/benchmark-runtime-weights.yml
+++ b/.github/workflows/benchmark-runtime-weights.yml
@@ -54,17 +54,18 @@ jobs:
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


Kailai-Wang:
I think one point that still needs to be improved is the IP-binding of the benchmark machine.

Currently, we:
- ssh to the host with alias name `parachain-benchmarking` to perform remote tasks
- the ssh config in the CI host is hardcoded with the remote IP, it works because we have fixed IP through elastic binding.

However, we'll get charged if the instance is stopped and elastic IP is still bound.
Therefore we need to dissolve the elastic IP and get the dynamic IP from AWS for each start/stop and use that IP for ssh connection.
Or is there a better alternative? @niteshb3495 might know more about it

niteshb3495:
@Kailai-Wang @chenzongxiong 
I have below understanding of the task. 

1. Rather connecting from CI server (outside of the AWS network) to remote server, can we perform the action (whatever we have dependency on CI server) from Jumphost? This will solve a couple of issues. First, the IP issue, as we can connect from the Jumphost directly on the private IP (private IP do not change once assigned). Hence, starting or stoping the instance won't be an issue. Second, as usual we can have better security. 
2. We can have a simple Jenkins job which will start the instance, perform the action and stop the instance once the work is done. As Jenkins is installed on Jumphost, connectivity won't be an issue.

If we are looking to connect from the CI server, as you mentioned, the IP will change everytime.
There is an option to connect to the instance with the instance ID as well but that would require additional settings and policies to be attached to the instance. More info [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-connect-methods.html#ec2-instance-connect-connecting-ec2-cli).

chenzongxiong:
@niteshb3495 I think cost would be a big problem here if we migrate from CI server to aws since CI server builds the docker image which needs quite a lot resources, which it's expensive in AWS. 

@Kailai-Wang I think `instance-id` is fixed once instance is created (within the same lifecycle https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html). 
If so, we can easily extract `public ip` information from started ip address, then connect to remote benchmarking server.

niteshb3495:
@chenzongxiong - Just for the clarification, do we have any specific reason to use CI server? Can't we run the github actions without self-hosted instance? If so, we can run the required steps directly on the github by using ubuntu OS. 

BTW, on other note, we can fetch the instance IP from the instance ID (by executing the aws command) and can get the public IP even if that changes everytime.

chenzongxiong:
@Kailai-Wang can you explain to nitesh why we need CI server (32G memory, 8threads) ? 
As far as I see, the main reason is that the resources of the server provided by github self is rather limited, which takes a long time to build an docker images (need compile rust binary of substrate, may take hours by using server provided by github action). 

On CI server, the building process takes around 1 hour.

chenzongxiong:
as for private network setting up, self-hosted github action should also be considered.

Kailai-Wang:
we have our self-hosted CI machine mainly for performance reasons. It used to take 20-30min for the docker build while a standard GH runner would take 1h+. Now the polkadot deps get more and more complicated so it takes 40min on our CI host. I believe it would take 1,5 - 2h on a GH runner.

We constantly explore the possibility to accelerate rust compilation inside docker and tried various methods (cargo-chef, GHA cache, buildx..). If we can dramatically reduce the time consumption on a GH runner, I don't mind switching to GH runners, but for now I'll still use self-hosted runner for better speed.

Kailai-Wang:
Yea I think fetching the dynamic public IP would suffice in this use case

niteshb3495:
Understood. 
To get the public IP of the instance from the instance ID, we can execute below command.
```
aws --region AWS-REGION \
ec2 describe-instances \
--filters \
"Name=instance-state-name,Values=running" \
"Name=instance-id,Values=i-00914683ababcba7eb1" \
--query 'Reservations[*].Instances[*].[PublicIpAddress]' \
--output text
```

Happy to help if anything else is required here.

niteshb3495:
Just to clarify, is there any dependency that we have to run/create the docker image on CI server only? Can't we use `parachain-benchmarking` server itself to create an image and do the other work on it?

Kailai-Wang:
> Just to clarify, is there any dependency that we have to run/create the docker image on CI server only? Can't we use `parachain-benchmarking` server itself to create an image and do the other work on it?

There isn't.
It's just the `parachain-benchmarking` server isn't as strong as CI server in terms of CPU/RAM resources. CI mainly requires computational power. Having an AWS instance with the same resource setup is significantly more expensive 😂

chenzongxiong:
@Kailai-Wang the whole pipeline works. can be merged if no further bugs
https://github.com/litentry/litentry-parachain/runs/7417487262?check_suite_focus=true

Kailai-Wang:
Thanks @chenzongxiong, it's looking good.
Shall we set the instance-id also as secret to a) not expose it b) not repeat it everywhere in case we need to change the instance later. Small thing though

You need to update `dev` branch btw

chenzongxiong:
making `instance-id` secret is a good idea. pls added it into `secrets` so that I can update the script later. 
Add a field named `BENCHMARK_INSTANCE_ID` in secrets

Kailai-Wang (approved):
Looks good

chenzongxiong:
@Kailai-Wang feel free to merge it. the bug indeed causes by `secrets`

base on the action results， https://github.com/litentry/litentry-parachain/runs/7427607803?check_suite_focus=true

we can safely merge it now

PR #718, opened 2022-07-20
Update to codecov v3

Kailai-Wang:
We often get "certificate expired" error when uploading to codecov: https://github.com/litentry/litentry-parachain/actions/runs/2706753348

let's see if things get better with v3: https://github.com/codecov/codecov-action/issues/788
This PR doesn't fail CI if upload fails.


Kailai-Wang:
seems that `fail_ci_if_error` didn't work because of [this](https://github.com/codecov/codecov-action/issues/768)

BillyWooo (approved):
looks good for me.

Issue #529, opened 2022-04-22
Verifying an identity


StamfordDigital:
We need to develop an extrinsic which allows a user/extension to **verify an identity**. This extrinsic will be part of our Identity Management Pallet.

Please see further details here: https://www.notion.so/litentry/identity-management-pallet-on-parachain-IMP-e3fad9d10c9547919ebac10799d0b26d#dc519f7595e24ff4a01ebac509855f73

StamfordDigital:
This extrinsic has been completed/deployed. Closing ticket.

Issue #528, opened 2022-04-22
Unlinking an identity


StamfordDigital:
We need to develop an extrinsic which allows a user/extension to **unlink an identity**. This extrinsic will be part of our Identity Management Pallet.

Please see further details here: https://www.notion.so/litentry/identity-management-pallet-on-parachain-IMP-e3fad9d10c9547919ebac10799d0b26d#dd9aee84340a42ba818e5c8ae3b72269

StamfordDigital:
This extrinsic has been completed/deployed. Closing ticket.

Issue #527, opened 2022-04-22
Linking an identity


StamfordDigital:
We need to develop an extrinsic which allows a user/extension to **link an identity**. This extrinsic will be part of our Identity Management Pallet.

Please see further details here: https://www.notion.so/litentry/identity-management-pallet-on-parachain-IMP-e3fad9d10c9547919ebac10799d0b26d#ac410350c1c94f19b1cb9ef10a6b1b81

StamfordDigital:
This extrinsic has been completed/deployed. Closing ticket.

Issue #505, opened 2022-04-22
Linking a Web3 Identity (Identity pallet on parachain)


StamfordDigital:

Kailai-Wang:
This is deprecated / already covered by the implementation.

PR #731, opened 2022-07-26
Add features of link, unlink and verify extrinsics in identity management

### LINKED ISSUE #504: Linking a Web2.0 Identity (Identity Pallet on parachain)

h4n0:
This PR solves #527 #528 #529 , which relate to implementations of link, unlink and verify extrinsics.

Also, it resolves #504 #505 as unit testing


Kailai-Wang (commented):
Generally looks good, thanks.

Could you please also add an extra extrinsic `set_user_shielding_key`? It's used to update the user's shielding key (public key) in TEE storage and it has similar parameters.

Kailai-Wang (approved):
Thanks, looks generally good to me!

PR #762, opened 2022-08-18
Restrict transfer_native and add boundary test

ziming-zung:
This PR:

-  restrict who can call the ```transfer_native```
- add boundary test


Kailai-Wang (approved):
Looks good to me - I approved it.

A small thing:
let's allow everyone to call transfer_native on rococo

Issue #761, opened 2022-08-18
Move `MaximumIssuance` to storage type


Kailai-Wang:
### Context
We'd like to change `MaximumIssuance` from constant to storage type to allow immediate control from privileged origin.

### Task
- [x] change `MaximumIssuance` to StorageValue, with a default value
- [x] a setter for it
- [x] a dedicated origin to set `MaximumIssuance`
- [x] runtime upgrade on rococo and test

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
@wangminqi Does it affect the logic of calculating `ExternalBalances`? If so we should update it within the setter

Kailai-Wang:
All tasks are done, this ticket can be closed.

PR #763, opened 2022-08-19
Move `MaximumIssuance` to storage type

ziming-zung:
resovle #761

### DIFF
--- a/pallets/bridge-transfer/src/lib.rs
+++ b/pallets/bridge-transfer/src/lib.rs
@@ -60,6 +60,9 @@ pub mod pallet {
 		/// The priviledged accounts to call the transfer_native
 		type TransferNativeMembers: SortedMembers<Self::AccountId>;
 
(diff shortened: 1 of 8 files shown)


Kailai-Wang (approved):
Looks good.

Any idea why the ts-test for litentry-runtime failed at the first time?

ziming-zung:
> Looks good.
> 
> Any idea why the ts-test for litentry-runtime failed at the first time?

After sleeping 13 seconds, getting the latest block may miss the previous block

https://github.com/litentry/litentry-parachain/blob/c4de3068d49f72655661cfc0b621d13a11f03ada/ts-tests/tests/bridge.test.ts#L118-L119

I think it needs to be implemented in a different way, eg: listen to new blocks

Kailai-Wang:
I see, so a timing issue.

Let's have another issue for it and get this one merged first

PR #774, opened 2022-08-25
728 introduce taplo for toml formatting

### LINKED ISSUE #728: Introduce taplo for TOML formatting
### Context
It's always beneficial to have a standard format. I tried taplo myself as TOML formatter and it looked good. I suggest integrating it as a standard TOML formatter for our repo.
https://github.com/tamasfe/taplo
https://taplo.tamasfe.dev/cli/introduction.html

### Task
- [x] encourage people to use it to format TOML files before submitting code
- [x] integrate taplo into CI
- [x] optionally adjust pre-commit hooks

---

BillyWooo:
Add TOML formatting tool **taplo** with these configs inside .taplo.toml :
```
align_entries = false
align_comments = false
array_trailing_comma = true
array_auto_expand = true
array_auto_collapse = true
compact_arrays = true
compact_inline_tables = false
compact_entries = false
column_width = 80
indent_tables = false
indent_entries = false
indent_string =  "	"
trailing_newline = true
reorder_keys = true
allowed_blank_lines = 2
crlf = false
```

Resolve #728


.taplo.toml
```diff
@@ -0,0 +1,19 @@
...
+array_trailing_comma = true
+array_auto_expand = true
+array_auto_collapse = true
+compact_arrays = true
+compact_inline_tables = false
+compact_entries = false
+column_width = 80
+indent_tables = false
+indent_entries = false
+indent_string =  "	"
```
Kailai-Wang: I think it's better to use space here, maybe 2 or 4 spaces, but not tabs

.taplo.toml
```diff
@@ -0,0 +1,19 @@
+include = ["**/Cargo.toml"]
+
+[formatting]
+align_entries = false
+align_comments = false
+array_trailing_comma = true
+array_auto_expand = true
```
Kailai-Wang: Maybe false for this, otherwise we'd have something like:
```
scale-info = { version = "2.1.1", default-features = false, features = [
	"derive",
] }
```
which is a bit werid

Kailai-Wang (changes requested):
Generally looking good, I have two small things (and CI has to pass ofc)

.taplo.toml
BillyWooo: 👌

.taplo.toml
BillyWooo: This is a bit hard. **array_auto_expand** controls all arrays. If it is set to *false*, e.g:
```
std = [
	"codec/std",
	"sp-std/std",
	"sp-runtime/std",
	"sp-io/std",
	"sp-core/std",
	"sp-arithmetic/std",
	"teerex-primitives/std",
	"pallet-teerex/std",
	"frame-support/std",
	"frame-system/std",
	"frame-benchmarking/std",
]
```
will become
```
std = ["codec/std", "sp-std/std", "sp-runtime/std", "sp-io/std", "sp-core/std", "sp-arithmetic/std", "teerex-primitives/std", "pallet-teerex/std", "frame-support/std", "frame-system/std", "frame-benchmarking/std"]
```
😢

Kailai-Wang: What about setting `array_auto_collapse` also to false?

Kailai-Wang (approved):
Thanks!

PR #795, opened 2022-08-31
Fix timing issue in tests

ziming-zung:
This PR try to  fix timing issue in tests.
The test sometimes fails, because the test code sleep 13 seconds before fetching the latest block.

https://github.com/litentry/litentry-parachain/blob/abc1ece47d17f4a774b569e46d4f644e0836c11c/ts-tests/tests/bridge.test.ts#L118-L119

may miss the previous block and cause the test to fail.

## Solution
Use ```listen to new block``` instead. 
Reference:  https://polkadot.js.org/docs/api/examples/promise/listen-to-blocks


Kailai-Wang (approved):
Thanks.
Btw why do we still need `await sleep(39)` in the end?

PR #808, opened 2022-09-02
minor update to have better CI control flow

### LINKED ISSUE #802: Fail early in CI
### Context
When any error happens in CI we should fail early and cancel all other running tasks.

Currently it seems that docker continues to build even when clippy fails:  https://github.com/litentry/litentry-parachain/actions/runs/2966854942

---

BillyWooo:
as the topic


Kailai-Wang (approved):
Thanks!

PR #820, opened 2022-09-07
Enable balance transfer for litmus and rococo

Kailai-Wang:
As topic.

This PR:
- activates balances transfer for rococo and litmus
- adjusts the integration-tests for different runtimes
- separates litentry and litmus in ts-tests
- applies prettier setting for formating


wangminqi (commented):
non scripts part looks good to me

Issue #821, opened 2022-09-07
Remove staking migration 


Kailai-Wang:
### Context
The runtime upgrade on Litentry parachain was done.
We need to:

- [x] remove the migration to make sure it doesn't get executed in the next runtime ugprade
- [x] move the migration.rs to somewhere else for code reference (e.g. pallet-staking), but please write proper comment about when/why this migration was used.
- [x] IIRC there're some Todos in the staking pallet after migration (like change public fn back to private)

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
I add another task as I have the impression that something needs to be changed back after the migration

Issue #801, opened 2022-08-31
Remove the staking pallet migration and reposition the code  


Kailai-Wang:
### Context
As the migration was completed for the staking pallet, we want to:

- [x] remove the migration from litentry-runtime (and rococo if it's still there)
- [x] move the migration code somewhere else, e.g. staking pallet -> migration, not sure if we want to add `v1` into it

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
This is done I think

PR #822, opened 2022-09-08
feat: add native token foreign asset type

wangminqi:
For adding gensis build of asset manager and runtime upgrade to correct old code.


wangminqi:
The migration relocated is temporary. There will be an extra pr for #821 and #801 .

pallets/xcm-asset-manager/src/lib.rs
```diff
@@ -145,6 +147,8 @@ pub mod pallet {
 		type Currency: ReservableCurrency<Self::AccountId>;
 
 		type WeightInfo: WeightInfo;
+
+		type NativeTokenForeignAssetType: Get<ForeignAssetTypeOf<Self>>;
```
Kailai-Wang, line 149: It seems this type is never used in the business logic except in the `genesis build`?

runtime/litmus/src/lib.rs
```diff
@@ -125,6 +126,7 @@ pub type Executive = frame_executive::Executive<
 	// it was reverse order before.
 	// See the comment before collation related pallets too.
 	AllPalletsWithSystem,
+	migration::AddNativeTokenForeignAssetTypeForAssetManager<Runtime>,
```
Kailai-Wang, line 135: I don't think we even need the migration, as no storage has ever been set on-chain in the asset manager. Please let me know if I misunderstand it.

pallets/xcm-asset-manager/src/lib.rs
wangminqi: yes. For genesis build only.

runtime/litmus/src/lib.rs
wangminqi: The problem is we miss the genesis build. We need to fix it somehow.

Kailai-Wang: But we missed the train of genesis build already 😂  the migration won't help the genesis build -- if you need a different state, we can directly change the pallet setting and/or send on-chain extrinsics, as the current state for it is pretty much empty anyway

pallets/xcm-asset-manager/src/lib.rs
Kailai-Wang: That doesn't make much sense to me 😂  I would expect something like "you can't register asset type that is equal to NativeTokenForeignAssetType" at least..

wangminqi: What you describe is partially true, we actually can not serialize multilocation, so genesis build can not use NativeTokenForeignAssetType directly.

Kailai-Wang: I mean if it's only used in genesis build, what's the point of having it in the future logic?

runtime/litmus/src/lib.rs
wangminqi: Correct!

wangminqi:
This PR is closed. A new approach will be accepted.

PR #823, opened 2022-09-08
feat: remove Moonbase code

wangminqi:
Remove Moonbase from code


Kailai-Wang (approved):
Looks good, thanks. But please only merge it after the litmus wasm is released.

Issue #778, opened 2022-08-26
Refactor out the fee transfer in `bridge::transfer_fungible`


Kailai-Wang:
### Context
In the current token bridge implementation we transfer the commission fee from the sender account to treasury using `Currency::transfer`:
https://github.com/litentry/litentry-parachain/blob/5249614eb56efb795274af3b04434f68677e9a7c/pallets/bridge/src/lib.rs#L685

While it works, it can create misleading records in the block explorer. An example: https://litmus.statescan.io/extrinsic/838218-2

Maybe we can consider some alternatives, e.g. `burn` + `deposit_into_existing`
The tests shouldn't be broken.

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
I think either @Jinsipang or @zhizming-zhong can take this

Kailai-Wang:
After it's done we should also add integration test for it

Kailai-Wang:
Is this done? @Jinsipang 
Let's use [keyword](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) in PR so that the issue can be auto-linked and auto-closed when the PR gets merged

Marky-Shi:
> Is this done? @Jinsipang Let's use [keyword](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) in PR so that the issue can be auto-linked and auto-closed when the PR gets merged

Yes, this has been completed at the source level, and the tests have all passed

Kailai-Wang:
Thanks, this can be closed now

PR #824, opened 2022-09-08
Fix some events triggered by token-bridge

Marky-Shi:
About issue  #778, Confused transfer events are fired when using token bridges.So I changed the transfer to the treasury in transfer_fungible to the `deposit_into_exiting` method.
However, there are some differences between the underlying implementation of the divesit series of methods and the transfer implementation, which will cause the account balance state before and after the transfer to be inconsistent, so some of the logic in the transfer_fungible is modified to ensure the consistency of the state before and after the transfer.


pallets/bridge-transfer/src/mock.rs
```diff
@@ -170,9 +170,14 @@ pub const ENDOWED_BALANCE: u64 = 100_000_000;
...
 pub fn new_test_ext() -> sp_io::TestExternalities {
 	let bridge_id = PalletId(*b"litry/bg").into_account_truncating();
+	let treasury_account: u64 = 0x8;
 	let mut t = frame_system::GenesisConfig::default().build_storage::<Test>().unwrap();
 	pallet_balances::GenesisConfig::<Test> {
-		balances: vec![(bridge_id, ENDOWED_BALANCE), (RELAYER_A, ENDOWED_BALANCE)],
+		balances: vec![
+			(bridge_id, ENDOWED_BALANCE),
+			(RELAYER_A, ENDOWED_BALANCE),
+			(treasury_account, 10),
```
wangminqi: replace with endowed_balance

pallets/bridge/src/mock.rs
```diff
@@ -120,9 +120,14 @@ pub const TEST_THRESHOLD: u32 = 2;
...
 pub fn new_test_ext() -> sp_io::TestExternalities {
 	let bridge_id = PalletId(*b"litry/bg").into_account_truncating();
+	let treasury_account: u64 = 0x8;
 	let mut t = frame_system::GenesisConfig::default().build_storage::<Test>().unwrap();
 	pallet_balances::GenesisConfig::<Test> {
-		balances: vec![(bridge_id, ENDOWED_BALANCE), (RELAYER_A, ENDOWED_BALANCE)],
+		balances: vec![
+			(bridge_id, ENDOWED_BALANCE),
+			(RELAYER_A, ENDOWED_BALANCE),
+			(treasury_account, 10),
```
wangminqi: replace with endowed_balance

wangminqi (changes requested):

wangminqi (approved):

PR #836, opened 2022-09-14
change the start index of assetManager

wangminqi:


Kailai-Wang:
I think githooks makes sense 😂

wangminqi:
> I think githooks makes sense 😂

e，what？

Kailai-Wang (approved):

Issue #830, opened 2022-09-12
Discard `call_worker` in IMP and its mock


Kailai-Wang:
### Context
Follow-up of https://github.com/litentry/litentry-parachain/issues/828, we'll discard the call_worker forwarding and handle it directly in TEE, see https://github.com/litentry/tee-worker/issues/58 too

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
resolved via #841

Issue #809, opened 2022-09-04
Include web2/web3 verification message in mock-up


Kailai-Wang:
### Context
As a follow-up to #805, we'd like to include the parameters that are needed for the web2/web3 verification messages in the `verify_identity` mockup.

Plain parameters to the TrustedCall are fine. We actually don't need a JSON format here as the whole TrustedCall will be encrypted.

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

PR #841, opened 2022-09-16
Discard call worker

Kailai-Wang:
solves #830 #809 

This PR:
- removes pallet_teerex::call_worker
- adjusts fn parameters
- add validation_data struct for reference


BillyWooo (approved):
LGTM. Also change back to integritee-network/pallets.git, nice.

Issue #794, opened 2022-08-31
Benchmark update


Kailai-Wang:
### Context
For the following pallets:

- [x] bridge
- [x] bridgeTransfer
- [x] parachainStaking

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
I assign this ticket to you both @wangminqi and @Jinsipang. It's not super urgent but we need to have this done before token bridge is onboarded on Litentry.

Marky-Shi:
After adding the benchmark code to the bridge-transfer pallet, running the transfer_native and transfer methods, the test results have been failures, starting with `Bad Origin` because where these two methods are implemented, the origin needs to be verified. After that I tried to add the account constructed in benchmark to the list of allowed to call these two methods, but it is still not feasible, perhaps it is a problem with my coding, I hope you can give some advice, the branch name [issue_794](https://github.com/litentry/litentry-parachain/tree/issue_794) @Kailai-Wang

PR #854, opened 2022-09-23
add benchmark at bridge and bridge transfer pallet

Marky-Shi:
The[ issue 794](https://github.com/litentry/litentry-parachain/issues/794) benchmark update has been completed,and weight files were generated.


pallets/bridge-transfer/src/lib.rs
```diff
@@ -19,10 +19,13 @@
 
 pub use pallet::*;
 
+#[cfg(feature = "runtime-benchmarks")]
+mod benchmarking;
 #[cfg(test)]
 mod mock;
 #[cfg(test)]
 mod tests;
+mod weights;
```
Kailai-Wang: I believe `weights` is actually not used. 
We need to specify the usage of it in extrinsic declaration, something like `[pallet::weight(<T as pallet::Config>::WeightInfo::....`

you could refer to other pallets

Kailai-Wang (changes requested):
I believe there're still some basic problems:

- weights not used
- the runtime weights need benchmarking
- both litmus and litentry runtime weights need to be updated

runtime/litmus/src/lib.rs
```diff
@@ -703,6 +703,7 @@ impl pallet_bridge::Config for Runtime {
 	type Currency = Balances;
 	type ProposalLifetime = ProposalLifetime;
 	type TreasuryAccount = TreasuryAccount;
+	type WeightInfo = ();
```
Kailai-Wang: we can now write `weights::pallet_bridge::WeihtInfo<Runtime>` or something like that

Same applies to other runtimes

runtime/litmus/src/lib.rs
```diff
@@ -738,6 +739,7 @@ impl pallet_bridge_transfer::Config for Runtime {
 	type NativeTokenResourceId = NativeTokenResourceId;
 	type DefaultMaximumIssuance = MaximumIssuance;
 	type ExternalTotalIssuance = ExternalTotalIssuance;
+	type WeightInfo = ();
```
Kailai-Wang: same as above

runtime/litentry/src/lib.rs
```diff
@@ -749,6 +751,11 @@ impl SortedMembers<AccountId> for TechnicalCommitteeProvider {
 	fn add(_: &AccountId) {
 		unimplemented!()
 	}
+	// To ensure that the benchmark code runs through
+	#[cfg(feature = "runtime-benchmarks")]
```
Kailai-Wang, line 755: Not the best way but let's leave as it is

Kailai-Wang (changes requested):

Kailai-Wang (approved):
I approve it first, however, the weight calculation for `acknowledge_proposal` is incorrect.

And we should (almost) **never** modify the generated weights file manually.

I'll create another issue for better readability

PR #868, opened 2022-10-04
Refactor benchmark yml

### LINKED ISSUE #855: Separate image building and benchmarking in GHA
### Context
we have a big single step for building the benchmark image and executing the benchmarks on the remote machine, see `benchmark-runtime-weights.yml`

To increase flexibility, we'd like to:
- [x] add a boolean input parameter to decide if we should build a docker image, or pull it directly from the hub. (Default = build)
- [x] split it into two jobs so that we can restart failed jobs only without starting everything from scratch

---

Kailai-Wang:
resolves #855 

This PR:
- separates docker-build from actual benchmarking
- use system and instance status check for reachability test

### DIFF
--- a/.github/workflows/benchmark-runtime-weights.yml
+++ b/.github/workflows/benchmark-runtime-weights.yml
@@ -3,6 +3,11 @@ name: Benchmark runtime weights
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


BillyWooo (approved):
LGTM

PR #876, opened 2022-10-08
remove Litmus sudo

### LINKED ISSUE #775: Sudo removal on Litmus
### Context
We are aiming to remove sudo on Litmus in two weeks (basically give the bi-directional token bridge some time and observe how well it works)

### Task
- [x] double-check origin control
- [x] remove sudo pallet and code adjustment
- [x] do we need any migrations? A: Yes, remove sudo storage
- [x] balance transfer activation

---

wangminqi:
Remove sudo of Litmus
This close #775


wangminqi:
All ensure root implementations in Litmus are changed into Root+AllCommitte as we discussed. CI test may need to change as Sudo is no longer available. @zhizming-zhong 

cc @Kailai-Wang

node/res/chain_specs/litmus-plain.json
```diff
@@ -169,10 +169,7 @@
       "aura": {
         "authorities": []
       },
-      "auraExt": null,
-      "sudo": {
-        "key": "5ECCQzRn2nUqDesngcoi7FxLhJBXGSnoEmXiMDZXCj94chRP"
-      }
+      "auraExt": null
```
Kailai-Wang: We don't need to remove this

node/src/chain_specs/litmus.rs
```diff
@@ -46,7 +46,6 @@ const DEFAULT_ENDOWED_ACCOUNT_BALANCE: Balance = 1000 * UNIT;
 #[derive(Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 struct GenesisInfo {
-	root_key: AccountId,
```
Kailai-Wang, line 49: Let's keep this as it's read from the JSON file (although never used for Litmus)

Kailai-Wang (commented):
Looks good except for small things that I think are better to keep, thanks. And:

- have you tried `try-runtime`?
- have you got an idea how to adjust tests?

wangminqi:
> Looks good except for small things that I think are better to keep, thanks. And:
> 
> * have you tried `try-runtime`?
> * have you got an idea how to adjust tests?

No, I have not run try-runtime. I plan to do it after the change is finalized.
Yes, I have discussed this with zhiming. The main idea is to build a function that triggers through all committees, it should work exact the same as the old sudo root, so only the minimal change/replacement of ts test will be needed.

node/res/chain_specs/litmus-plain.json
wangminqi: Sure, we will keep it; Just wondering any specific reason to keep it?

node/src/chain_specs/litmus.rs
wangminqi: same above. I will recover this part.

node/res/chain_specs/litmus-plain.json
Kailai-Wang: litmus-plain.json is a 1-to-1 mapping to litmus.json which is the production chain-spec. It's hardcoded JSON and **can't** be changed after the chain launch (I mean the litmus.json), otherwise you get genesis mismatch when starting the nodes.

Kailai-Wang:
> No, I have not run try-runtime. I plan to do it after the change is finalized.

👍 

> Yes, I have discussed this with zhiming. The main idea is to build a function that triggers through all committees, it should work exact the same as the old sudo root, so only the minimal change/replacement of ts test will be needed.

Yes we need to change ts-tests, but I think there's more.

We'll still need sudo in development. Currently we are launching a local dev network through `make launch-docker-litmus`, which will not contain sudo anymore. This is definitely inconvenient (OK we still have `make launch-docker-litentry` but imagine sudo is removed for Litentry too). 

~~`parachain-launch` can reset the sudo key to Alice, but the precondition is we still have to have the sudo pallet. That's probably also one of the reasons why Acala preserved sudo pallet when removing sudo.~~

~~Maybe the best way is to introduce a new runtime with sudo and collator-selection included? Just for local dev purpose. `rococo` could have been such runtime but you'd need to write another migration to revert the staking changes on rococo. What do you think?~~

Nvm, missing of collatorSelection shouldn't be a problem. We should be able to use `rococo-dev` for local network, but let me test it first.

Kailai-Wang:
I have a PR for adding rococo to the local dev env: https://github.com/litentry/litentry-parachain/pull/880

ts-tests/tests/utils.ts
```diff
@@ -67,7 +67,7 @@ export async function initApiPromise(config: any): Promise<ParachainConfig> {
     // Set Eve's balance to 1000000000000000
     const eve_info = await api.query.system.account(eve.address);
     if (eve_info.data.free.lt(new BN(1000000000000000))) {
-        const txSetBalance = api.tx.sudo.sudo(api.tx.balances.setBalance(eve.address, 1000000000000000, 0));
+        const txSetBalance = await sudoWrapper(api, api.tx.balances.setBalance(eve.address, 1000000000000000, 0))
```
Kailai-Wang: I guess this won't work, right?
If we don't need the `if` branch, we can simply delete it.

Kailai-Wang (approved):
One small thing, others look good, thanks.

ts-tests/tests/utils.ts
ziming-zung: i think we should keep it, if delete it, the test will fail in litentry/rococo

Issue #871, opened 2022-10-04
Wrong weight calculation for `acknowledge_proposal`


Kailai-Wang:
### Context
The current weight computation for `pallet_bridge::acknowledge_proposal` is wrong:
- the weight directive is wrongly written
- we should (almost) **never** edit the generated weights file manually
- do we have reason of using 15, 16, 17 for `make_proposal`? We could have simply used empty vec => it doesn't matter too much 

You can refer to the substrate proxy pallet: https://github.com/paritytech/substrate/blob/241b0d0455453499763d0db0b4ea4188012b372f/frame/proxy/src/lib.rs#L194-L202

Please correct it and rerun the weights for this pallet. 

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Marky-Shi:
Thanks for pointing out my mistake, I will go and fix it

Kailai-Wang:
Thanks, please merge the current PR and open a separate PR for this issue

Marky-Shi:
Regarding this problem, I have fixed the benchmark code in the bridge pallet, and after debugging it locally, I submitted a[ pr](https://github.com/litentry/litentry-parachain/pull/877), hoping that you can review it

PR #877, opened 2022-10-09
Issue 871 Wrong weight calculation for acknowledge_proposal

Marky-Shi:
This is about [issue 871](https://github.com/litentry/litentry-parachain/issues/871)'s pr, modifying the incorrect benchmark code in the bridge pallet


Kailai-Wang (changes requested):
Unfortunately I still think it's a bit inaccurate:
- we should use the [auto-close keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to link and close the issue
- normally we just add the call's weight to the extrinsic, I don't think it's necessary to use the it as the parameter (and it will be incorrect as we can have any kind of calls), please check the link in the issue description for reference
- I think you only have to benchmark the `pallet-bridge` again, so instead of using `*` when running GHA

pallets/bridge/src/lib.rs
```diff
@@ -324,7 +324,10 @@ pub mod pallet {
 		/// # <weight>
 		/// - O(1) lookup and insert
 		/// # </weight>
-		#[pallet::weight(<T as Config>::WeightInfo::set_threshold(*threshold as u32))]
+		#[pallet::weight({
+		<T as Config>::WeightInfo::set_threshold()
+		.saturating_add(*threshold as Weight)
```
Kailai-Wang: We don't need this `saturating_add` here

pallets/bridge/src/lib.rs
```diff
@@ -417,9 +420,11 @@ pub mod pallet {
 		/// # <weight>
 		/// - weight of proposed call, regardless of whether execution is performed
 		/// # </weight>
-		#[pallet::weight(
-		<T as Config>::WeightInfo::acknowledge_proposal(call.get_dispatch_info().weight as u32)
-		)]
+		#[pallet::weight({
+		let di = call.get_dispatch_info();
+		< T as Config >::WeightInfo::acknowledge_proposal()
+		.saturating_add(di.weight)
```
Kailai-Wang: let's include `di.class` too

pallets/bridge/src/lib.rs
```diff
@@ -464,7 +469,11 @@ pub mod pallet {
 		/// # <weight>
 		/// - weight of proposed call, regardless of whether execution is performed
 		/// # </weight>
-		#[pallet::weight(<T as Config>::WeightInfo::eval_vote_state(prop.get_dispatch_info().weight as u32))]
+		#[pallet::weight({
+		let di = prop.get_dispatch_info();
+		< T as Config >::WeightInfo::eval_vote_state()
+		.saturating_add(di.weight)
```
Kailai-Wang: same as above

Kailai-Wang (changes requested):

Kailai-Wang (approved):

PR #884, opened 2022-10-11
Add standalone `mock-tee-primitives`

Kailai-Wang:
This is to break the cyclic dependency of the `parachain` repo and the `tee-worker` repo.

This primitive should only be used by mock pallet.

In this way we make sure we don't have blockers when updating the polkadot dependencies.

### DIFF
--- a/mock-tee-primitives/src/identity.rs
+++ b/mock-tee-primitives/src/identity.rs
@@ -1,0 +1,163 @@
+// Copyright 2020-2022 Litentry Technologies GmbH.
+// This file is part of Litentry.
+//
+// Litentry is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Litentry is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


BillyWooo (approved):
LGTM

mock-tee-primitives/src/identity.rs
```diff
@@ -0,0 +1,163 @@
...
+	Substrate(SubstrateNetwork),
+	Evm(EvmNetwork),
+}
+
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum Web2Network {
+	Twitter,
+	Discord,
+	Github,
```
montogeek, line 65: @Kailai-Wang Are you using brand names here or just the name? 
Correct brand name for "Github" is "GitHub".

Kailai-Wang: > @Kailai-Wang Are you using brand names here or just the name? Correct brand name for "Github" is "GitHub".

It's just a name string for GitHub.
Does it make F/E's life easier if we change to `GitHub`?

montogeek: No, it is just me trying to use the correct name, thanks!

PR #894, opened 2022-10-14
Backport from `release-v0.9.10`

Kailai-Wang:
A minor to backport from the `release-v0.9.10` branch:
move the bridge-related calls to the normal operations section


ziming-zung (approved):
Looks good.

PR #897, opened 2022-10-16
Check registered enclave for TEE callback extrinsics

### LINKED ISSUE #895: Use the real TEE account for the IMP extrinsics
### Context
Currently, we restrict in Runtime(rococo/litmus) that only Alice can initiate extrincs to IMP, but actually in the tee-worker, it is initiated by the **enclave account**. We can get the enclave account from `pallet_teerex` storage.

```rust
ord_parameter_types! {
	pub const ALICE: AccountId = sp_runtime::AccountId32::new(hex!["d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"]);
}

impl pallet_identity_management::Config for Runtime {
	type Event = Event;
	type WeightInfo = ();
	// TODO: use the real TEE account
	type TEECallOrigin = EnsureSignedBy<ALICE, AccountId>;
}
```

https://github.com/litentry/litentry-parachain/blob/ecb5622f42/runtime/rococo/src/lib.rs#L856-L865

Kailai-Wang:
resolves #895

### DIFF
--- a/runtime/litmus/src/lib.rs
+++ b/runtime/litmus/src/lib.rs
@@ -26,14 +26,14 @@ extern crate frame_benchmarking;
 use codec::{Decode, Encode, MaxEncodedLen};
 use cumulus_pallet_parachain_system::RelayNumberStrictlyIncreases;
 use frame_support::{
-	construct_runtime, parameter_types,
+	construct_runtime, ord_parameter_types, parameter_types,
 	traits::{
 		ConstU16, ConstU32, ConstU64, ConstU8, Contains, Everything, InstanceFilter, SortedMembers,
 	},
 	weights::{constants::RocksDbWeight, ConstantMultiplier, IdentityFee, Weight},
 	PalletId, RuntimeDebug,
 };
-use frame_system::EnsureRoot;
+use frame_system::{EnsureRoot, EnsureSignedBy};
 use hex_literal::hex;
 
 // for TEE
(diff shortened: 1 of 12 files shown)


Kailai-Wang:
Test with TEE integration test looks good:

image

runtime/litmus/src/lib.rs
```diff
@@ -783,29 +783,22 @@ impl pallet_sidechain::Config for Runtime {
...
+	type TEECallOrigin = EnsureEnclaveSigner<Runtime>;
 }
 
 impl pallet_identity_management_mock::Config for Runtime {
 	type Event = Event;
 	type ManageWhitelistOrigin = EnsureRoot<Self::AccountId>;
 	type MaxVerificationDelay = ConstU32<10>;
-	// TODO: use the real TEE account
-	type TEECallOrigin = EnsureSignedBy<ALICE, AccountId>;
+	type TEECallOrigin = EnsureEnclaveSigner<Runtime>;
```
ziming-zung: Looks good to me. But i  think it would be better for `pallet_identity_management_mock` to still use **Alice**, what do you think?

ziming-zung (approved):

runtime/litmus/src/lib.rs
Kailai-Wang: It's not that relevant TBH as those extrinsics in mock won't be called by anyone anyway 😂 

But yes let me change that to Alice if you find it better

PR #906, opened 2022-10-20
rename `ruleset` into `assertion`

### LINKED ISSUE #903: rename `ruleset` into `assertion`
### Context
We are going to use the terminology `assertion`, instead of `ruleset`

---

BillyWooo:
minor update

resolves #903


pallets/vc-management/src/assertion.rs
```diff
@@ -20,26 +20,26 @@
 // See: https://www.notion.so/litentry/Expected-parameters-in-predefined-rulesets-14f74928aa2b43509167da12a3e75507
 
 use codec::{Decode, Encode, MaxEncodedLen};
-use primitives::Balance;
 use scale_info::TypeInfo;
 use sp_runtime::{traits::ConstU32, BoundedVec};
 
+type Balance = u128;
```
Kailai-Wang: Any reasons not to use primitives::Balance?

runtime/litmus/src/lib.rs
```diff
@@ -783,6 +784,12 @@ impl pallet_sidechain::Config for Runtime {
 	type EarlyBlockProposalLenience = ConstU64<100>;
 }
 
+impl pallet_teeracle::Config for Runtime {
```
Kailai-Wang, line 786: Strange, this should be in `dev` already.

Kailai-Wang (commented):
Conflict needs to be resolved.

I thought some changes were in `dev` already? Like pallet_teeracle stuff, why does it show up again 😂

runtime/litmus/src/lib.rs
BillyWooo: Yes. It should be. Git problem.

pallets/vc-management/src/assertion.rs
BillyWooo: Strange. Didn't noticed this change. Will check.

runtime/litmus/src/lib.rs
Kailai-Wang: One sec, I just noticed it was from `tee-dev` to `dev`. Maybe we shouldn't do it right now, as these two branches diverge. If you need something badly in the `tee-dev` branch (due to incompatible tee-worker development), it's better to submit a PR to `tee-dev` branch directly.

We will merge `dev` to `tee-dev` once the upstream is updated to v0.9.29

Kailai-Wang:
Could you re-do this PR to only include the naming changes and `docker/litmus-parachain-launch-config.tee-dev.yml`? And from a feature branch other than `tee-dev`?

It looks like merging branches into each other might cause history confusion..

BillyWooo:
close for the moment. Nothing need to update.

PR #937, opened 2022-11-07
Use sccache in Dockerfile

### LINKED ISSUE #865: Consider using full GH runner in CI
### Context
We are using self-hosted runner to build docker images in our CI so far, see:
https://github.com/litentry/litentry-parachain/issues/259
https://github.com/litentry/litentry-parachain/issues/393
https://github.com/litentry/litentry-parachain/pull/409

the main reason is self-hosted runner is much faster than GH-runner: ~20m vs 1h
But now self-hosted runner needs 40min as well to build the docker image (mostly because of polkadot update that introduces more compilation deps).

Maybe it's worth checking again how long the compilation on GH-runner would be and switch to it when possible. It will help concurrent CI runs.

---

### LINKED ISSUE #934: Remove `set-output` in GHA files
### Context
Warning from GHA run:
```
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

should be a minor change and low prior

---

### LINKED ISSUE #935: Sccache doesn't seem to work in CI
### Context
I notice that sccache doesn't seem to take effect, an example: https://github.com/litentry/litentry-parachain/actions/runs/3405980852

Even if I re-run the job without changing anything, where I expect the `run-cargo-runtime-tests` to complete quickly, it still did a full compilation.

Strangely, the cargo registry and sccache cache were restored successfully:
```
Cache restored successfully
Cache restored from key: cargo-7cca616f672081f386280a7525ace5640c511a4aebda45281f073c7e8c5e3007
```
```
Cache restored successfully
Cache restored from key: sccache-7cca616f672081f386280a7525ace5640c511a4aebda45281f073c7e8c5e3007
```

It just didn't get applied.

In the sccache stat, it showed:
image

Note the cache hit were all C/C++.

Once it's solved, the next step is to try to use sccache inside docker

---

Kailai-Wang:
resolves #934 
resolves #935 
resolves #865 

This PR did more research and experiments regarding caching in CI.
**Conclusion:**

- sccache in Dockerfile only works with self-hosted runners
- `cargo-chef` + `buildx local/gha cache` could work with github runners but we haven't had the best experience with `cargo-chef` so far.

This PR did some more treaks:
- removed deprecated `set-output`
- improved the cache-key handling
- installed pre-built taplo rather than building it from scratch
- bumped some GHA versions
- tried https://github.com/Swatinem/rust-cache but it didn't work so well. I like `sccache` that displays the exact hits/misses.


BillyWooo (approved):
LGTM

Issue #898, opened 2022-10-17
Update dependencies to `polkadot-v0.9.30`


Kailai-Wang:
### Context
polkadot-v0.9.30 was released, cumulus has the matching branch too.

### Task
- [x] update deps
- [x] make sure compilation/all tests work
- [x] client/runtime update on chains

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
PR merged.
We are looking forward to releasing a new client update for this.

Kailai-Wang:
Clients updated.

Issue #887, opened 2022-10-12
The parachain staking benchmark is running for too long


Marky-Shi:
### Context
[#Action](https://github.com/litentry/litentry-parachain/actions/runs/3155868677/jobs/5134984739)  When using parachain staking benchmark code to generate a weight file, the run time is too long.

### Task
Modifying the staking pallet staking benchmark code improves the runtime without affecting the weight file generation.

Kailai-Wang:
Related: https://github.com/PureStake/moonbeam/issues/1950

Marky-Shi:
https://github.com/litentry/litentry-parachain/actions/runs/3528580789/jobs/5919401779,  This is the result of using the CLI machine to run it. Unfortunately, it's cost 2 hours. It seems that the step and repeat need to be adjusted.

According to the previous test results on the T40 machine, if the step is adjusted to 30, the entire running process will be shortened to 1 hour. If it is run together with other pallet benchmark codes, it will take 2-3 hours to generate the weight file without counting the build-docker time. .

PR #942, opened 2022-11-10
Update v0.9.30

Marky-Shi:
#898  update deps to v0.9.30


runtime/litmus/src/tests/mod.rs
```diff
@@ -24,19 +24,19 @@ pub mod setup {
...
 	runtime_common::run_transaction_payment_tests!();
 }
 
 mod xcm_parachain {
 	use crate::{
 		tests::setup::{
-			Call as RelayCall, Origin as RelayOrigin, ParaA, ParaB, Relay, RelayChainRuntime,
-			TestNet,
+			ParaA, ParaB, Relay, RelayChainRuntime, RuntimeCall as RelayCall,
+			RuntimeOrigin as RelayOrigin, TestNet,
```
wangminqi, line 36: So these name changes should also apply to common/src/tests/setup/relays.rs  ,   right? But no change apply to these files yet. Need double check.

runtime/rococo/src/migration/P9100.rs
```diff
@@ -31,7 +31,7 @@ where
 	<T as frame_system::Config>::Event: From<pallet_parachain_staking::Event<T>>,
 {
 	#[cfg(feature = "try-runtime")]
-	fn pre_upgrade() -> Result<(), &'static str> {
+	fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
```
wangminqi: So these files are not used and only keep for history backup purposes. If no CI error, maybe we should just leave it?

I am open to any solution. minor issue.

wangminqi (changes requested):

runtime/litentry/src/migration/P9100.rs
```diff
@@ -31,7 +31,7 @@ where
 	<T as frame_system::Config>::Event: From<pallet_parachain_staking::Event<T>>,
 {
 	#[cfg(feature = "try-runtime")]
-	fn pre_upgrade() -> Result<(), &'static str> {
```
wangminqi: same above

runtime/litmus/src/tests/mod.rs
Marky-Shi: I compared the moonbeam's code, which is how they implemented it in version 0.9.30
image

wangminqi: check. Looks okay.

wangminqi (approved):

.github/workflows/build-and-test.yml
```diff
@@ -107,6 +107,12 @@ jobs:
           target: wasm32-unknown-unknown
           default: true
 
+      - name: Install protobuf
+        run: |
+          sudo apt-get update && \
+          sudo apt-get upgrade -y && \
+          sudo apt-get install -yq openssl clang libclang-dev cmake  protobuf-compiler
```
Kailai-Wang, line 114: Can you please remove the `upgrade -y` line and only install the needed deps? The step name is "install protobuf" but you are installing a lot more than just protobuf

.github/workflows/build-and-test.yml
```diff
@@ -299,6 +305,12 @@ jobs:
           chmod +x $HOME/.local/bin/sccache
           echo "$HOME/.local/bin" >> $GITHUB_PATH
 
+      - name: Install protobuf
+        run: |
+         sudo apt-get update && \
+         sudo apt-get upgrade -y && \
+         sudo apt-get install -yq openssl clang libclang-dev cmake  protobuf-compiler
```
Kailai-Wang, line 312: Same as above

runtime/litentry/src/weights/pallet_proxy.rs
```diff
@@ -138,20 +138,18 @@ impl<T: frame_system::Config> pallet_proxy::WeightInfo for WeightInfo<T> {
...
 	// Storage: Proxy Proxies (r:1 w:1)
-	/// The range of component `p` is `[1, 31]`.
-	fn anonymous(p: u32, ) -> Weight {
-		Weight::from_ref_time(35_206_000 as u64)
-			// Standard Error: 1_000
-			.saturating_add(Weight::from_ref_time(61_000 as u64).saturating_mul(p as u64))
+	fn create_pure(p: u32, ) -> Weight {
+		Weight::from_ref_time(31_077_000 as u64)
+			// Standard Error: 3_000
+			.saturating_add(Weight::from_ref_time(37_000 as u64).saturating_mul(p as u64))
```
Kailai-Wang, line 144: I see in the runtime setting `()` is used so it probably won't have an impact. But still, this doesn't look like automatically generated

Kailai-Wang (approved):
Principally looks good, only a few small things. Thanks

.github/workflows/build-and-test.yml
Marky-Shi: Thanks for pointing out the deficiencies in the code, for this issue I will open an issue specifically to fix it

.github/workflows/build-and-test.yml
Marky-Shi: Same as above

runtime/litentry/src/weights/pallet_proxy.rs
Marky-Shi: It is to ensure that the compilation passes, so it is copied from the substrate,the weight file of pallet-proxy I will put it in #887  to solve together.

PR #966, opened 2022-11-21
Vc schema

jingleizhang:
This PR:

- add schema_registry storage
- add schema_admin
- add add_schema(), disable_schema() and revoke_schema() with privileged admin


jingleizhang:
related issue https://github.com/litentry/litentry-parachain/issues/1004

pallets/vc-management/src/lib.rs
```diff
@@ -60,30 +65,78 @@ pub mod pallet {
...
+		},
+		// a Schema is issued
+		SchemaIssued {
+			account: T::AccountId,
+			hash: H256,
+			id: Vec<u8>,
+			content: Vec<u8>,
+		},
+		// a Schema is disabled
+		SchemaDisable {
```
Kailai-Wang: A small typo: should be Disable`d`

pallets/vc-management/src/schema.rs
```diff
@@ -0,0 +1,32 @@
...
+#[derive(Clone, Eq, PartialEq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[scale_info(skip_type_params(T))]
+#[codec(mel_bound())]
+pub struct VCSchema<T: Config> {
+	// the schema id, e.g 'http://litentry.com/credentials/schema/2022/twitter/follower'
+	pub id: ContentString,
+	// the schema author
+	pub author: T::AccountId,
+	// schema content
+	pub content: ContentString,
```
Kailai-Wang, line 39: 64 bytes probably not enough?

pallets/vc-management/src/lib.rs
```diff
@@ -165,5 +227,62 @@ pub mod pallet {
 			Self::deposit_event(Event::SomeError { func, error });
 			Ok(Pays::No.into())
 		}
+
+		// Change the schema Admin account
+		#[pallet::weight(195_000_000)]
+		pub fn set_schema_admin(
```
Kailai-Wang, line 252: Will the admin be removed later? (I assume so - if not, the author would always be the admin)

pallets/vc-management/src/lib.rs
```diff
@@ -165,5 +227,62 @@ pub mod pallet {
...
+			Self::deposit_event(Event::SchemaAdminChanged { old_admin: Self::schema_admin() });
+			<SchemaAdmin<T>>::put(new);
+			Ok(Pays::No.into())
+		}
+
+		// It requires the schema Admin account
+		#[pallet::weight(195_000_000)]
+		pub fn add_schema(
+			origin: OriginFor<T>,
+			hash: H256,
```
Kailai-Wang: I have two concerns here:
- how is `hash` calculated? Is it by hashing the content? If so we probably either need to check it, or we simply don't pass it as a parameter
- Is `hash` the best option for the storage key? It's a bit counter-intuitive if two schemas have the same id but differ in content

Kailai-Wang (commented):
The structure looks good, I have a few questions regarding some details though

pallets/vc-management/src/lib.rs
```diff
@@ -96,6 +149,15 @@ pub mod pallet {
 		VCSubjectMismatch,
 		/// The VC is already disabled
 		VCAlreadyDisabled,
+
+		/// Error when the caller account is not the admin
+		RequireSchemaAdmin,
+		/// Schema Id already exists
```
BillyWooo: simply English issue, I think it's normally written like:
```
		/// Schema Id already exists
		SchemaAlreadyExist,
		/// Schema Id does not exist
		SchemaNotExist,
```

pallets/vc-management/src/schema.rs
jingleizhang: It's definitely not enough. If the length is changed to 1024, should be fine for block productions?

pallets/vc-management/src/lib.rs
jingleizhang: Yes, the admin should be removed in a later version, if we open it to external parties.

pallets/vc-management/src/lib.rs
jingleizhang: Thanks! Fixed.

pallets/vc-management/src/lib.rs
jingleizhang: The `hash` here is calculated from content. But an auto-increased ID would be better. Will change the hash to ID.

pallets/vc-management/src/lib.rs
jingleizhang: Thanks, fixed

pallets/vc-management/src/lib.rs
```diff
@@ -165,5 +245,98 @@ pub mod pallet {
...
+				c.status = Status::Disabled;
+				*context = Some(c);
+				Self::deposit_event(Event::SchemaDisabled { account: sender, index });
+				Ok(().into())
+			})
+		}
+
+		// It requires the schema Admin account
+		#[pallet::weight(195_000_000)]
+		pub fn active_schema(
```
Kailai-Wang: I'd suggest renaming it to either `enable` (pair with `disable`), or `activate`

pallets/vc-management/src/lib.rs
```diff
@@ -165,5 +245,98 @@ pub mod pallet {
...
+			index: SchemaIndex,
+		) -> DispatchResultWithPostInfo {
+			let sender = ensure_signed(origin)?;
+			ensure!(Some(sender.clone()) == Self::schema_admin(), Error::<T>::RequireSchemaAdmin);
+
+			let context = SchemaRegistry::<T>::get(index).ok_or(Error::<T>::SchemaNotExists)?;
+			SchemaRegistry::<T>::remove(index);
+			let index = Self::schema_count();
+			if index > 0 {
+				<SchemaCount<T>>::put(index - 1);
```
Kailai-Wang: I don't think this will work :( we don't always remove the last element in `SchemaRegistry`, as a result, we get correct SchemaCount but wrong index when adding a new schema. 

Imagine we have 4 schemas `[1, 4]`, we revoke schema idx = 2, and for the next `add_schema` call we would have idx = 4 which is already occupied.

A simple solution is to always increment the index and do not care if there are indexes revoked.

pallets/vc-management/src/lib.rs
```diff
@@ -165,5 +245,98 @@ pub mod pallet {
...
+			id: Vec<u8>,
+			content: Vec<u8>,
+		) -> DispatchResultWithPostInfo {
+			let sender = ensure_signed(origin)?;
+			ensure!(Some(sender.clone()) == Self::schema_admin(), Error::<T>::RequireSchemaAdmin);
+
+			
+
+			let index = Self::schema_count();
+			<SchemaCount<T>>::put(index + 1);
```
Kailai-Wang: Although unlikely, we probably still need to check overflow here, e.g. `checked_add`, maybe `try_mutate` is better too

Kailai-Wang (changes requested):

pallets/vc-management/src/lib.rs
jingleizhang: get it, is fixing now

jingleizhang:
Have created a new branch `vc-schema-tee`, to fix a LOT of merge conflicts. 

Will close this PR, and open a new one with the latest updates from that branch.

PR #1039, opened 2022-11-30
tee-worker ts-tests

0xverin:
Added some environment variable configurations ,added random shard, and modified startup scripts.


jingleizhang (approved):
I have tested this branch from the beginning on Ubuntu instance, it works fine, the parachain node can produce blocks.

Although there is still a warning message `API/INIT: RPC methods not decorated: mmr_generateHistoricalBatchProof, transaction_unstable_submitAndWatch, transaction_unstable_unwatch` in `tests/register-parachain.ts`. 

Will merge this PR.

PR #1041, opened 2022-11-30
Change the versions of @polkadot

0xverin:


Kailai-Wang:
Possible to merge it into the other PR?
Each CI would take 2h 😂 , after merging into `tee-dev`, it takes another ~2h, and then sync the codebase to the other PR takes another 2h

jingleizhang:
There is a bug in ts-tests/tests/register-parachain.ts that makes the `scripts/launch-local-binary.sh` fail to register parachain, due to the super old version of polkadot-js/api .

We can merge this branch into the other ones, but it eventually have to be merged into the tee-dev.

What do you think @Kailai-Wang ?

Kailai-Wang:
> There is a bug in ts-tests/tests/register-parachain.ts that makes the `scripts/launch-local-binary.sh` fail to register parachain, due to the super old version of polkadot-js/api .
> 
> We can merge this branch into the other ones, but it eventually have to be merged into the tee-dev.
> 
> What do you think @Kailai-Wang ?

Yes please make changes in #1039 directly, otherwise we get too many CI runs in parallel (and they need to be mutually merged..)

jingleizhang:
Understand, now close this PR.  All changes would be included in `ts-tests` branch.

PR #1056, opened 2022-12-04
add multiconfig for launch workers 

### LINKED ISSUE #1000: multiple configs for different envioments
It should include:

- [ ] local / staging / prod envioments
- [ ] kept in TEE enclave, and not hard coded in codes
- [ ] could be selected according to tee-worker's `make` parameters (e.g, similar to `SGX_MODE=SIM`, there may be `CONFIG=LOCAL/STAG/PROD`)

BillyWooo:
as the title.
Launch workers with [local parachain / staging parachain / production parachain]

resolves #1000


jingleizhang:
1. https://github.com/litentry/litentry-parachain/blob/504196026a0719baa9e75a64775a24199e41c41f/tee-worker/litentry/core/data-providers/src/lib.rs#L59-L65

2. https://github.com/litentry/litentry-parachain/blob/504196026a0719baa9e75a64775a24199e41c41f/tee-worker/litentry/core/data-providers/src/twitter_official.rs#L103

3. https://github.com/litentry/litentry-parachain/blob/504196026a0719baa9e75a64775a24199e41c41f/tee-worker/litentry/core/data-providers/src/discord_official.rs#L71

Those hardcodes should be moved to config files.

Kailai-Wang:
As mentioned it might be better to directly extend the `tee-worker/local-setup/github-action-config.json` and add more such JSONs for different envs. Keep it simple.

But please discuss it directly with @jingleizhang

jingleizhang:
@Kailai-Wang  How can we keep those prod config files safe, through our CI tools or platforms(e.g, ansible/Rundeck)?

Kailai-Wang:
> @Kailai-Wang How can we keep those prod config files safe, through our CI tools or platforms(e.g, ansible/Rundeck)?

similar to our collators.
Even in the prod config I don't think there are too many secrets (except API keys maybe?), servers won't be reachable from outside. We can manually copy the files/edit the secrets there, it should be an one-time effort as it's unlikely to change. Alternatively we can use something like git crypt.

Maybe let's discuss it offline instead of in a public thread :)

jingleizhang:
Get it.

tee-worker/litentry/core/data-providers/Cargo.toml
```diff
@@ -14,6 +14,7 @@ thiserror = { version = "1.0.26", optional = true }
 url = { version = "2.0.0", optional = true }
 
 # no_std dependencies
+lazy_static = { version = "1.1.0", features = ["spin_no_std"] }
```
jingleizhang: the latest version is 1.4.0

tee-worker/litentry/core/data-providers/src/lib.rs
```diff
@@ -55,22 +61,67 @@ pub mod twitter_official;
...
-// #[cfg(test)]
-#[cfg(any(test, feature = "mockserver"))]
-pub mod base_url {
-	pub(crate) const TWITTER_OFFICIAL: &str = "http://localhost:9527";
-	pub(crate) const TWITTER_LITENTRY: &str = "http://localhost:9527";
-	pub(crate) const DISCORD_OFFICIAL: &str = "http://localhost:9527";
-	pub(crate) const DISCORD_LITENTRY: &str = "http://localhost:9527";
+lazy_static! {
+	pub static ref G_DATA_PROVIDERS: RwLock<DataProvidersStatic> =
+		RwLock::new(DataProvidersStatic::new());
```
jingleizhang, line 124: Why the RwLock is needed here?

tee-worker/scripts/launch_local_worker.sh
```diff
@@ -1,80 +1,126 @@
...
+--node-port ${NODE_PORT} \
+--node-url ${NODE_URL} \
+--trusted-external-address wss://${WORKER_ENDPOINT} \
+--trusted-worker-port ${trusted_worker_port} \
+--untrusted-external-address ws://${WORKER_ENDPOINT} \
+--untrusted-http-port ${untrusted_http_port} \
+--untrusted-worker-port ${untrusted_worker_port} \
+--twitter-official-url ${TWITTER_OFFICIAL_URL} \
+--twitter-litentry-url ${TWITTER_LITENTRY_URL} \
+--twitter-auth-token ${TWITTER_AUTH_TOKEN} \
```
jingleizhang: Those values should be directly loaded from config files, and should not be passed in as command-line arguments. For example, if we want to add a new config field, we don't need to change the launch script again.

tee-worker/litentry/core/data-providers/src/lib.rs
BillyWooo: to avoid race condition and `unsafe` code block.

tee-worker/service/src/cli.yml
```diff
@@ -80,9 +80,14 @@ args:
...
     - clean-reset:
-          long: clean-reset
-          short: c
-          help: Cleans and purges any previous state and key files and generates them anew before starting.
+        long: clean-reset
+        short: c
+        help: Cleans and purges any previous state and key files and generates them anew before starting.
+    - running-mode:
+        long: running-mode
+        help: Litentry TEE service running mode <dev/stage/prod>
```
Kailai-Wang: I think staging is the conventional word 😂, same for `worker_config_*`

tee-worker/scripts/launch.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
Kailai-Wang, line 1: when shall we run `launch.sh` and when `launch.py`?

tee-worker/local-setup/github-action-config-rococo.json
```diff
@@ -15,7 +15,9 @@
         "-u",
         "wss://parachain-eu-0.staging.litentry.io",
         "-p",
-        "443"
+        "443",
+        "running-mode",
+        "stage"
```
Kailai-Wang: I think `-u` and `-m` can be confusing/conflicting: can I use `-u wss://parachain-eu-0.staging.litentry.io` and `-m dev` at the same time? Again it depends what are `dev/staging/prod` supposed to do

tee-worker/local-setup/worker_config_dev.json
```diff
@@ -0,0 +1,8 @@
+{
```
Kailai-Wang, line 1: I don't see a difference between dev/staging/prod, this needs a clear definition: what exactly should differ in these different configs? -- needs to be evaluated/defined by who opens this issue (maybe @jingleizhang)

Kailai-Wang (commented):
Overall it looks better, thanks.

The biggest concern is still what should be defined in `dev/staging/prod` configs -- I'll forward this to @jingleizhang

tee-worker/local-setup/worker_config_dev.json
BillyWooo: Yes. For the moment. These three files are almost dummy placeholders. Need to define/clarify later.

tee-worker/local-setup/github-action-config-rococo.json
BillyWooo: I think there is no shortcut for 'running-mode'. Always need to type the full name. I also feel confusing about these parameter shortcuts. That's the reason NO shortcut provided here.

tee-worker/scripts/launch.sh
BillyWooo: Both are fine. `launch.sh` would be more preferred for local launch & test. `launch.py` + `config.json` would have more flexibility on ports parameter inputs (`launch.sh` will just increase the port number between different worker. `config.json` can give any port number).

tee-worker/service/src/cli.yml
BillyWooo: 🤣

jingleizhang (approved):
At least we have those prod/staging/dev config files so that we can work with the same application code in different environments. 

All parameters should be moved into the config file, and only one parameter that represents the env needs to be kept, actually. But for now, we're just using the config to save some no-more hard-coded information. At least it helps us keep some api keys safe.

Please correct me if I'm wrong: `prod` stands for `Litentry`, `staging` means `parachain-eu-0.staging.litentry.io`, and `dev` means `rococo`. 

I think we could change those names to something more clear if we think it's necessary. 😂

PR #1057, opened 2022-12-04
Add more testcases for IMP mock

Kailai-Wang:
This PR adds two more test cases for IMP mock, both originated from problems that F/E colleagues encountered:
- link two (different) twitter handles for one account should work
- wrong web3 validation message should throw `UnexpectedMessage` error 

This PR fixes the file checker error in `tee-worker-ci.yml`


BillyWooo (approved):
LGTM

Issue #1073, opened 2022-12-07
More flexible challenge code generation in tests


Kailai-Wang:
### Context

The challenge code is always hardcoded when `mockserver` feature is activated: https://github.com/litentry/litentry-parachain/blob/c485920fbdc980fcf0a8aac1ace9011c6606c8d1/tee-worker/app-libs/stf/src/helpers.rs#L134-L141

This makes it impossible to test e.g. https://github.com/litentry/litentry-parachain/issues/981, where we want to explicitly check the update of challenge code.

We need to either:
- remove the hardcoded challenge code, and make mockserver to read the challenge code live from the chain state
or
- still use hardcoded but make it different each time, e.g. calculated from some outside inputs (env/command line/....)

And generally I'm not a fan of`mockserver` feature as it requires a re-compilation and makes the test switch between dev/staging hard. Can we somehow remove this feature and use unified codebase? E.g. fully controlled by command line or startup args for worker.

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

PR #1074, opened 2022-12-07
Emit new challenge code for each `create_identity`

### LINKED ISSUE #981: [identityManagement] Support generating new challenge codes for non-verified Identities.
### Context
Pallet: identityManagement[Mock] 

Consider the following scenario
1. A user request `identityManagement.linkIdentity('Twitter', 'my-handle')` -> `ChallengeCodeGenerated` event is emitted.
1. User request again `identityManagement.linkIdentity('Twitter', 'my-handle')`.

**Received:**
The transaction fails with `IdentityAlreadyExist`.

**Desired:**
On point 2, generate a new challenge code. i.e., a `ChallengeCodeGenerated` event is emitted.

## How this useful

- Better Developer Experience: It makes testing easier as you can re-use the same `web_type` and `handle` to get the same results while not minding spending tokens.
- Better User Experience: Users won't need to pay fees for `unlinkIdentity` in case they lose their challengeCode, and retrieving through an indexer is not an option for them.
- Better Consumer Experience: as a dApp developer, I would like to leave ChallengeCodes retrieval from an archive node for later.

Kailai-Wang:
resolves #981 

More test cases for IMP mock and adjustment in sgx-runtime pallets.
some leftover renamings too

TODO: #1073

### DIFF
--- a/pallets/identity-management-mock/src/lib.rs
+++ b/pallets/identity-management-mock/src/lib.rs
@@ -171,8 +171,8 @@ pub mod pallet {
 		ShieldingKeyDecryptionFailed,
 		/// unexpected decoded type
 		WrongDecodedType,
-		/// identity already exists when creating an identity
-		IdentityAlreadyExist,
+		/// identity already verified when creating an identity
+		IdentityAlreadyVerified,
 		/// identity not exist when removing an identity
 		IdentityNotExist,
 		/// no shielding key for a given AccountId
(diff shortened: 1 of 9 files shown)


pallets/identity-management-mock/src/lib.rs
```diff
@@ -310,14 +310,17 @@ pub mod pallet {
 				},
 			};
 
-			ensure!(
-				!IDGraphs::<T>::contains_key(&who, &identity),
-				Error::<T>::IdentityAlreadyExist
-			);
+			if let Some(c) = IDGraphs::<T>::get(&who, &identity) {
```
ziming-zung, line 313: I think we should do the same changes for the `identity-managment-pallet`in the tee-worker, include renamings

https://github.com/litentry/litentry-parachain/blob/c485920fbdc980fcf0a8aac1ace9011c6606c8d1/tee-worker/litentry/pallets/identity-management/src/identity_context.rs#L28-L37

https://github.com/litentry/litentry-parachain/blob/c485920fbdc980fcf0a8aac1ace9011c6606c8d1/tee-worker/litentry/pallets/identity-management/src/lib.rs#L184-L204

and so on....

ziming-zung (changes requested):

pallets/identity-management-mock/src/lib.rs
Kailai-Wang: I think it's already done in this PR? 😂 
https://github.com/litentry/litentry-parachain/pull/1074/files#diff-0673e79aa46adac84174d588afcb0816746db60d4b560b4cdd01097951b322b0

ziming-zung: My fault, i should be ok.😂

ziming-zung (approved):
Looks good.

PR #1043, opened 2022-11-30
update Polkadot v0.9.32

Marky-Shi:
update parachain deps to polkadot-v0.9.32

PR #1083, opened 2022-12-12
update dependencies from fork pallets to origin integritee pallets v0.9.32

BillyWooo:
as title


Kailai-Wang:
We probably don't need to update the weights (again), as it was already done in https://github.com/litentry/litentry-parachain/pull/1043

Marky-Shi (approved):

Kailai-Wang:
btw I suggest pinning the revision in the Cargo.toml as we are using the `master` branch -- not 100% needed, but nice-to-have

Kailai-Wang (approved):
Thanks!

PR #1087, opened 2022-12-13
Add test local pallets benchmarking shell

### LINKED ISSUE #986: optimize the benchmark  scripts
### Context
optimize parachain's benchmark  scripts 

reference:
https://github.com/PureStake/moonbeam/blob/master/scripts/benchmarking.sh

Marky-Shi:
reloves #986 
this pr add a script that can run specified pallet benchmarkings and gen weight file at weights dir 

image

image

### DIFF
--- a/scripts/local-pallets-benchmark.sh
+++ b/scripts/local-pallets-benchmark.sh
@@ -24,8 +24,7 @@ function help {
     echo "  ${0}                       " "list all benchmarks and provide a selection to choose from"
     echo "  ${0} --check               " "list all benchmarks and provide a selection to choose from, runs in 'check' mode (reduced steps and repetitions)"
     echo "  ${0} foo bar               " "run a benchmark for pallet 'foo' and benchmark 'bar'"
-    echo "  ${0} foo bar --all         " "run a benchmark for all pallets"
-    echo "  ${0} foo bar --all --check " "run a benchmark for all pallets in 'check' mode (reduced steps and repetitions)"
+    echo "  ${0} foo bar all           " "run the pallet all benchmark method"
 }
 
 WEIGHTS_PATH="$(pwd)/weights"
@@ -36,7 +35,14 @@ fi
 
 CHAIN_TYPE="--chain=${1}-dev"
 PALLET=${2//-/_}
-EXTRINSIC=${3//-/_}
+
+EXTRINSIC=
+
+if [[ ${3} == 'all' ]];then
+ EXTRINSIC=*
+else
+  EXTRINSIC=${3//-/_}
+fi
 
 function choose_and_bench {
     readarray -t options < <(${BINARY} benchmark pallet --list $CHAIN_TYPE | sed 1d)
--- a/scripts/local-pallets-benchmark.sh
+++ b/scripts/local-pallets-benchmark.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+
+# This script can be used for running moonbeam's benchmarks.
+#
+# The moonbeam binary is required to be compiled with --features=runtime-benchmarks
+# in release mode.
+
+set -e
+
+BINARY="./target/release/litentry-collator"
+
+
+if [[ ! -f "${BINARY}" ]]; then
+    echo "binary '${BINARY}' does not exist."
+    echo "ensure that the moonbeam binary is compiled with 'make build-node-benchmarks ' nd in release mode."
+    exit 1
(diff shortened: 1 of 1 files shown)


scripts/local-pallets-benchmark.sh
```diff
@@ -0,0 +1,106 @@
...
+# in release mode.
+
+set -e
+
+BINARY="./target/release/litentry-collator"
+
+
+if [[ ! -f "${BINARY}" ]]; then
+    echo "binary '${BINARY}' does not exist."
+    echo "ensure that the litentry binary is compiled with 'make build-node-benchmarks ' nd in release mode."
```
jingleizhang: nd -> and

Marky-Shi: yeah has been fixed

jingleizhang (approved):
LGTM

scripts/local-pallets-benchmark.sh
```diff
@@ -0,0 +1,113 @@
...
+    echo "  ${0} litentry|litmuts|rococo pallet_name benchmark_method  [--check]"
+    echo ""
+    echo "EXAMPLES:"
+    echo "  ${0}                       " "list all benchmarks and provide a selection to choose from"
+    echo "  ${0} --check               " "list all benchmarks and provide a selection to choose from, runs in 'check' mode (reduced steps and repetitions)"
+    echo "  ${0} foo bar               " "run a benchmark for pallet 'foo' and benchmark 'bar'"
+    echo "  ${0} foo bar all           " "run the pallet all benchmark method"
+}
+
+WEIGHTS_PATH="$(pwd)/weights"
```
Kailai-Wang, line 30: This is guaranteed to be wrong: `pwd` can be anything, you need to use absolute path or `cd` to a known folder before this

scripts/local-pallets-benchmark.sh
```diff
@@ -0,0 +1,113 @@
...
+if [ ! -d $WEIGHTS_PATH ];then
+  mkdir -p  weights/
+
+fi
+
+CHAIN_TYPE="--chain=${1}-dev"
```
Kailai-Wang, line 36: It contradicts with what's written in `help`
`./scripts/local-pallets-benchmark.sh --check` give you the wrong chain-type

scripts/local-pallets-benchmark.sh
```diff
@@ -0,0 +1,113 @@
...
+              $TEMPLATE \
+              $OUTPUT
+
+}
+
+if [[ "${@}" =~ "--help" ]]; then
+    help
+else
+    CHECK=0
+    if [[ "${@}" =~ "--check" ]]; then
```
Kailai-Wang, line 96: I think the handling of `--check` is wrongly implemented.

tried:
```
./scripts/local-pallets-benchmark.sh rococo --check
# select 66
# got error 
Error: Input("No benchmarks found which match your input.")
```

tried `./scripts/local-pallets-benchmark.sh rococo pallet-proxy --check`, same error:
```
pallet  pallet-proxy-weights.rs
benchmarking 'pallet-proxy::' --check=1, writing results to '--output=/home/kai/litentry-parachain/weights/pallet-proxy_weights.rs'
will run other pallet (pallet-proxy) benchmark code
2022-12-13 12:57:03 💥 keeping old session because of empty collator set!
2022-12-13 12:57:03 💥 keeping old session because of empty collator set!
Error: Input("No benchmarks found which match your input.")
```

tried `./scripts/local-pallets-benchmark.sh rococo pallet-proxy add_proxy --check`, select `76) pallet_membership, swap_member`, it seems `add_proxy` is benchmarked, but the generated weight file is under `Created file: "/home/kai/litentry-parachain/weights/pallet_membership_weights.rs"`

It's a mess

Kailai-Wang (commented):
TBH I don't think we need such a script.

We have `scripts/benchmark-weight-local.sh` which will do the job already, this script has an extra "selection" process, which, if you really want, can be integrated into the existing script, but not a new one.

Secondly, please make sure to run/test the script yourselves before committing - at least some of the classic args.

scripts/local-pallets-benchmark.sh
Kailai-Wang: the only thing that worked is benchmarking the whole pallet, e.g.
```
./scripts/local-pallets-benchmark.sh rococo pallet-extrinsic-filter
```

PR #1099, opened 2022-12-16
Allow IDHCS to create identity on users' behalf

### LINKED ISSUE #847: Allow IDHCS to trigger `create_identity` on user's behalf
### Context
Following the discussion with @kaylawangnow, we'd like to allow IDHCS to send `link_identity` to request the challenge code on user's behalf. This would increase the user's experience a lot as the user only has to sign one tx to verify the identity.

We have to consider the backwards compatibility, so we need to add an origin filter to allow either:
- IDHCS's address, or
- the user him/herself

---

Kailai-Wang:
resolves #847 

This PR:
- adds a `Delegatee` list to manage authorised delegatees who can act on users' behalf -- currently only `create_identity`, can be extended later
- removes caller whitelists from IMP mock
- adjusts the origin check of `create_identity` and the counterpart in TEE
- add unittests 
- tee-worker/ts-tests prettier


BillyWooo (approved):
👍   LGTM

PR #1108, opened 2022-12-20
enclave sign interface

### LINKED ISSUE #1078: enclave can sign any given data within stf
input:  any given data in Vec\<u8\> format

sign method: ed25519

output: signature and enclave's public key

zTgx:
input: any given data in Vec<u8> format

sign method: ed25519

output: signature and enclave's public key


Kailai-Wang:
Maybe I misunderstand it but don't we have the `sign` and `verify` methods for `sp_core::ed25519::Pair` already?

PR #1112, opened 2022-12-21
1004 manage vc schemas

jingleizhang:


pallets/vc-management/src/lib.rs
```diff
@@ -165,5 +239,101 @@ pub mod pallet {
...
+			let index = Self::schema_index();
+			let new_index = index.checked_add(1u64).ok_or(Error::<T>::SchemaIndexOverFlow);
+			if new_index.is_ok() {
+				<SchemaRegistryIndex<T>>::put(new_index.as_ref().unwrap());
+			}
+			SchemaRegistry::<T>::insert(
+				index,
+				VCSchema::<T>::new(id.clone(), sender.clone(), content.clone()),
+			);
+			Self::deposit_event(Event::SchemaIssued { account: sender, shard, index });
```
Kailai-Wang: A bit problem in this logic: if `new_index` is somehow not ok, it will still insert a new entry with key = `index`, which overwrites the existing one and emits the event. 

We need to stop it whenever an error happens, maybe use `?` here

Kailai-Wang: If in doubts we should add a test case for it

runtime/rococo/src/lib.rs
```diff
@@ -888,6 +888,7 @@ impl pallet_identity_management_mock::Config for Runtime {
 impl pallet_vc_management::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type TEECallOrigin = EnsureEnclaveSigner<Runtime>;
+	type SetAdminOrigin = EnsureSignedBy<ALICE, AccountId>;
```
Kailai-Wang: I suggest using `EnsureRootOrHalfCouncil` too, as rococo is a public network. So everyone has access to `ALICE`

Kailai-Wang (changes requested):
I found some small problems

And let's use auto-close keyword to link to an issue :)

runtime/rococo/src/lib.rs
jingleizhang: fixed

pallets/vc-management/src/lib.rs
jingleizhang: fixed

Kailai-Wang (approved):
Approved, thanks!

Maybe we can add a comment later that `SchemaRegistryIndex` actually stores the "next available index".

PR #1115, opened 2022-12-22
add teerex pallet

zTgx:
- [x]  add pallets: sidechain/teeracle/teerex
- [x]  add upstream_commit
- [x]  add script: generate_teerex_upstream_patch


Cargo.toml
```diff
@@ -8,6 +8,11 @@ members = [
   'pallets/identity-management',
   'pallets/identity-management-mock',
   'pallets/vc-management',
+  'pallets/tee/primitives/teerex',
+  'pallets/tee/sidechain',
+  'pallets/tee/teeracle',
+  'pallets/tee/teerex',
+  'pallets/tee/test-utils',
```
Kailai-Wang: IMO putting them directly under `pallets/` would be better, reasons:
- to be consistent with other pallets
- these are not tee-specific, some extrinsics should be called from worker, but some are callable by users

The primitives should be put in different pallets, or in the root `primitives` if anything is common enough.

.github/workflows/parachain-ci.yml
```diff
@@ -240,7 +240,7 @@ jobs:
           default: true
 
       - name: Run unittests
-        run: cargo test --locked --release -p pallet-* --lib
+        run: cargo test --features=skip-ias-check --locked --release -p pallet-* --lib
```
Kailai-Wang, line 243: this is really strange, running `pallet-sidechain` alone is fine but it fails with wildcard.

Can you try to investigate what happened? Is it related to the file structure (they are in subdirectories of `pallets`/)

Kailai-Wang (commented):
Thanks! I've raised two questions

.github/workflows/parachain-ci.yml
Kailai-Wang: Do we still need this feature? Or do we actually need two lines, one with `skip-ias-check` feature, one without

Kailai-Wang (approved):
Looks good, thanks. 

I left a small comment though. Apart from that, the shell script needs adjustment too due to the changed directory structure.

.github/workflows/parachain-ci.yml
zTgx: Yes, we need.
> enable `skip-ias-check` to run `identity-management` unit tests and 
> disable to run  `pallet-sidechain` 's unit tests

Marky-Shi (commented):
This doesn‘t seem to be the result of executing the benchmark, and needs to be regenerated via benchmark-code.

I ran the benchmark code locally and some errors were encountered, I'll go and try to fix it.

Starting benchmark: pallet_teerex::register_enclave
2022-12-27 10:08:57 panicked at 'Benchmark does not allow the `skip-ias-check` flag.'

zTgx:
> This doesn‘t seem to be the result of executing the benchmark, and needs to be regenerated via benchmark-code.
> 
> I ran the benchmark code locally and some errors were encountered, I'll go and try to fix it.
> 
> Starting benchmark: pallet_teerex::register_enclave 2022-12-27 10:08:57 panicked at 'Benchmark does not allow the `skip-ias-check` flag.'

run benchnark needs to disable `tee-dev` feature

Marky-Shi:
> > This doesn‘t seem to be the result of executing the benchmark, and needs to be regenerated via benchmark-code.
> > I ran the benchmark code locally and some errors were encountered, I'll go and try to fix it.
> > Starting benchmark: pallet_teerex::register_enclave 2022-12-27 10:08:57 panicked at 'Benchmark does not allow the `skip-ias-check` flag.'
> 
> run benchnark needs to disable `tee-dev` feature

By default, tee-dev features are enabled, but it's not certain what the effect of canceling this feature will be. But for now, the existence of tee-dev affects the generation of teerex pallet weight files.

Marky-Shi:
> > > This doesn‘t seem to be the result of executing the benchmark, and needs to be regenerated via benchmark-code.
> > > I ran the benchmark code locally and some errors were encountered, I'll go and try to fix it.
> > > Starting benchmark: pallet_teerex::register_enclave 2022-12-27 10:08:57 panicked at 'Benchmark does not allow the `skip-ias-check` flag.'
> > 
> > 
> > run benchnark needs to disable `tee-dev` feature
> 
> By default, tee-dev features are enabled, but it's not certain what the effect of canceling this feature will be. But for now, the existence of tee-dev affects the generation of teerex pallet weight files.

make build-node-benchmarks 
./scripts/benchmark-weight-local.sh rococo pallet-teerex pallet
./scripts/benchmark-weight-local.sh rococo pallet-teerex runtime

Issue #985, opened 2022-11-28
Merge tee-worker to parachain


Kailai-Wang:
### Context
After internal discussion we'd like to merge the [tee-worker](https://github.com/litentry/tee-worker) repo to this repo to make it a mono-repo.

Advantages are:
- easier commit sync and PR review
- manage issues/PRs in one place
- better version control
- simplified release process

The merge needs to be done at `tee-dev` branch first, and CI needs to be adjusted to cover different cases.
This needs to be done ASAP as it's the base for the subsequent development

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
Closed via #987

PR #1116, opened 2022-12-22
Use local branch to generate artefacts

Kailai-Wang:
A leftover of #985 

This PR allows the TEE pallets call on rococo too.


ziming-zung (approved):
LGTM

PR #1117, opened 2022-12-22
Write test for resuming worker

### LINKED ISSUE #1114: Write tests for #999 (resume tee-worker)
### Context
We still miss (integration) test for #999 to make sure our changes do fix it: the worker can be resumed and continue to produce blocks, and the chain state is kept unchanged.

---

ziming-zung:
This PR

- resolves  #1114


tee-worker/ts-tests/utils.ts
```diff
@@ -2,8 +2,8 @@ import './config';
 import WebSocketAsPromised = require('websocket-as-promised');
 import WebSocket = require('ws');
 import Options from 'websocket-as-promised/types/options';
-import { ApiPromise, Keyring, WsProvider } from '@polkadot/api';
-import { StorageKey, Vec } from '@polkadot/types';
+import {ApiPromise, Keyring, WsProvider} from '@polkadot/api';
```
Kailai-Wang, line 5: We need to be consistent about the formatting 😂 I simply used prettier, the config is the same as ts-tests in parachain. It's OK if we use something else, just have to unify it

tee-worker/service/src/main.rs
```diff
@@ -193,8 +193,10 @@ fn main() {
 
 	#[cfg(feature = "mockserver")]
 	thread::spawn(move || {
-		info!("*** Starting mock server");
-		lc_mock_server::run();
+		if !config.disable_mock_server {
+			info!("*** Starting mock server");
+			lc_mock_server::run();
+		}
```
Kailai-Wang, line 209: May I ask why we need `disable_mock_server`? It's guarded by `mockserver` feature and in CI we use the `sidechain` image for testing resume-worker

Kailai-Wang (commented):
Some comments/questions inline.

The idea looks good, thanks. I'd like to ask for more testcases when possible:

- kill both workers and resume
- launch 2 workers, kill worker0 and resume (because worker0 is the primary validateer)
- we probably need to wait until sidechain produces blocks (so `new_block.number` > `old_block.number`) , to avoid the cases that the worker can resume but can't produce new blocks
- Not sure how difficult it is, ideally we can test the state is preserved. E.g. we set balance before killing workers, after resuming the balance is persisted

Kailai-Wang:
CI needs to be fixed ofc 😂

tee-worker/service/src/main.rs
ziming-zung: I usually use `mockserver` for local compilation. So if I start 2 workers locally, I need this option to disable starting mockserver.

ziming-zung:
> Some comments/questions inline.
> 
> The idea looks good, thanks. I'd like to ask for more testcases when possible:
> 
> * kill both workers and resume
> * launch 2 workers, kill worker0 and resume (because worker0 is the primary validateer)
> * we probably need to wait until sidechain produces blocks (so `new_block.number` > `old_block.number`) , to avoid the cases that the worker can resume but can't produce new blocks
> * Not sure how difficult it is, ideally we can test the state is preserved. E.g. we set balance before killing workers, after 

Currently the test code is already waiting for the sidechain to produce blocks.
For now, it's not easy to get the storage state of the sidechain, so personally I  prefer to use `block_hash`.
I'm writing more test cases, but some problems don't seem to be fixed soon. :frowning_face:

Kailai-Wang:
> > Some comments/questions inline.
> > The idea looks good, thanks. I'd like to ask for more testcases when possible:
> > 
> > * kill both workers and resume
> > * launch 2 workers, kill worker0 and resume (because worker0 is the primary validateer)
> > * we probably need to wait until sidechain produces blocks (so `new_block.number` > `old_block.number`) , to avoid the cases that the worker can resume but can't produce new blocks
> > * Not sure how difficult it is, ideally we can test the state is preserved. E.g. we set balance before killing workers, after
> 
> Currently the test code is already waiting for the sidechain to produce blocks. For now, it's not easy to get the storage state of the sidechain, so personally I prefer to use `block_hash`. I'm writing more test cases, but some problems don't seem to be fixed soon. ☹️

It's OK, let's do the rest in another issue/PR.

Kailai-Wang (approved):
Looks good, thanks. Please merge it once CI finishes :)

PR #1135, opened 2022-12-30
restruct Identity with enum(fix #1051)

### LINKED ISSUE #1051: Restructure or add sanity check of Identity
### Context
F/E colleagues have discovered a problem that any combination of `web_type` and `handle` is allowed, e.g. you can use `Twitter` and `Address32`, which is actually wrong.

It's better to prevent this from parachain side, we need to either:
- do sanity checks and throw an error
- force it by restructuring. e.g. something like `Identity::Twitter(PlainString)`

Better ideas?

---

DogLi:
restruct `Identiy` with enum to avoid the free combination of network type and handle type


tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -15,6 +15,7 @@
 // along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
 
 #![cfg_attr(not(feature = "std"), no_std)]
+extern crate core;
```
Kailai-Wang: I believe this is redundant?

mock-tee-primitives/src/identity.rs
```diff
@@ -67,97 +81,128 @@ pub enum Web2Network {
...
 
 #[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct Identity {
-	pub web_type: IdentityWebType,
-	pub handle: IdentityHandle,
+pub enum Identity {
+	Substrate(SubstrateIdentity),
+	Evm(EvmIdentity),
+	Web2(Web2Identity),
```
Kailai-Wang: I'm basically fine with it, how does this new structure sound to you?  @zhizming-zhong @Verin1005 @jonalvarezz 

Btw I'm just thinking if we should simplify it even more to flatten out the `SubstrateIdentity`/`EvmIdentity`/`Web2Identity`, so something like this:
```
	Substrate(SubstrateNetwork, Address32),
	Evm(EvmNetwork, Address20),
	Web2(Web2Network, IdentityString),
```

Kailai-Wang (approved):
Basically looks good to me, thanks. 

But about the new `Identity` enum structure, I'd like to hear opinions from F/E colleagues.

edit: CI should be fixed though

tee-worker/litentry/primitives/src/lib.rs
DogLi: it's auto-added by the IDE, I'll fix this.

mock-tee-primitives/src/identity.rs
ziming-zung: I prefer the flatter one

jingleizhang: agree, the flatter version is more simpler

jonalvarezz: Thank you, Kai.

What matters the most to clients is how easy it is to manipulate and traverse the objects.

Do I understand correctly that the following will be the corresponding Polkadot.js API types?

For `Substrate(SubstrateNetwork, Address32)`
```js
Identity: {
  Substrate: [SubstrateNetworkObject, '0x1234....']
}
```

For `Substrate(SubstrateNetworkIdentity)`
```js
Identity: {
  Substrate: {
    network: '...',
    address: '0x1234....'
  }
}
```

If so, we would prefer `Identity.Substrate.network` over `Identity.Substrate[1]`. Object properties over array indexes, in general.

Kailai-Wang: I actually think they should have the same polkadot-js types (i.e. without array indexes), but please correct me if I'm wrong @Verin1005 @zhizming-zhong

ziming-zung: Yes, if the order of their fields is the same, they have the same value after encoding

jingleizhang (approved):
LGTM

mock-tee-primitives/src/identity.rs
jonalvarezz: If that's the case, then there is not much difference from our side :)

The plain version looks easier to follow, though 👍

PR #1140, opened 2023-01-02
Support wrapped substrate signature verification

### LINKED ISSUE #1137: `identityManagement.verifyIdentity` doesn't accept signatures generated from Browser Extensions (Polkadot.js)
### Context
`identityManagement.verifyIdentity` validates the received Substrate signatures using `sp_io::crypto::sr25519_verify`
https://github.com/litentry/litentry-parachain/blob/37390508a87b79c09a1cba26488b09ebc45756eb/pallets/identity-management-mock/src/lib.rs#L603

`sp_io::crypto::sr25519_verify` works fine for local accounts, e.g. Alice, Bob, or those created using Keyring, but it fails at verifying signatures from Polkadot.js or any Browser Extension – main use case for dApps.

In other words, signatures created using https://polkadot.js.org/apps/#/signing will fail in `sp_io::crypto::sr25519_verify`.

Kailai and I were able to confirm that signatures from extensions wrap the data inside `<Bytes>...</Bytes>` as reported [here](https://substrate.stackexchange.com/questions/1810/transaction-fails-with-account-from-an-extension-but-ok-with-local-account). 

### Task
Even though we could find a workaround for this issue from a dApp, the pallets should consider extension signatures valid. Here is a Rust test supporting this use case: https://github.com/polkadot-js/wasm/blob/1916a4e10cd7fc151edbfbd95a9dafc724920110/packages/wasm-crypto/src/rs/sr25519.rs#L252-L269

What do you think?

---

Kailai-Wang:
resolves #1137 

Supports signature for both raw and wrapped message when verifying substrate signature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1112,9 +1112,9 @@ dependencies = [
 
 [[package]]
 name = "comfy-table"
-version = "6.1.3"
+version = "6.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e621e7e86c46fd8a14c32c6ae3cb95656621b4743a27d0cffedb831d46e7ad21"
+checksum = "6e7b787b0dc42e8111badfdbe4c3059158ccb2db8780352fa1b01e8ccf45cc4d"
 dependencies = [
  "strum",
  "strum_macros",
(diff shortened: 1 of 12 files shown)


ziming-zung (approved):
Looks good

jingleizhang (commented):
The changes are good. But should we migrate the changes to Identity Management Pallet as well?

Kailai-Wang:
> The changes are good. But should we migrate the changes to Identity Management Pallet as well?

It is implemented in [tee-worker/litentry/core/identity-verification/src/web3/mod.rs](https://github.com/litentry/litentry-parachain/pull/1140/files#diff-2fe275c827f03e32775c21ff04de288aa561826fbf0e27968d639bb01b6f62f8), otherwise the ts-test won't pass :)

PR #1151, opened 2023-01-04
Fix wrong IDGraph state in IMP mock

### LINKED ISSUE #1148: The returned IdGraph is not reflecting the latest state
Hi folks, we found the following bug. Is it expected?

### Steps to reproduce

1. Link an Identity and complete both `createIdentity` and `verifyIdentity` transactions.
2. Inspect the `idGraph` returned as part of the `IdentityManagement[Mock].IdentityCreated` event

### Result

The `idGraph` includes the newly added Identity, but its property `isVerified`  comes as `false`

Here is a screenshot for reference.
Screen Shot 2023-01-04 at 09 09 48

Notice that the added identity (`0x3cc0d57dd514...`) comes with `isVerified` false.

### Expected:

The new identity is in the idGraph and is set to `isVerified == true`

### More info

Consequent events containing `idGraph` reflect the correct value. 

Here is a screenshot of a subsequent challenge code request for another Identity:

Screen Shot 2023-01-04 at 09 10 43

Additionally, I noticed that running `tee-worker/ts-tests` fail at the end by asserting precisely if `isVerified == true`

Kailai-Wang:
resolves #1148 

The real pallet works correctly, we have assertions in ts-tests to reassure it: https://github.com/litentry/litentry-parachain/blob/5c57a53e33909fb103138e1ebc3691a9743ea154/tee-worker/ts-tests/identity.test.ts#L276


jingleizhang (approved):
LGTM

Issue #1028, opened 2022-09-06
Build Assertion 8: Total EVM and Substrate Transactions: User has over X number of transactions


StamfordDigital:
Build **Assertion 8: Total EVM and Substrate Transactions: User has over X number of transactions**
- One or more EVM and Substrate address(es)/account(s) must be provided and passed to the Parachain from the F/E using the `params` extrinsic param to verify this
- The address(es)/account(s) provided must be linked and verified against the `idhub-address` which made the extrinsic call
- The TEE will call The Graph to determine how many **transactions** the user has carried out historically over for the EVM and Substrate address(es)/account(s) provided.
- The TEE will validate if the user has had over **100**, **1000**, and **10,000** transactions (tx)
- The JSON payload of the VC will illustrate which transaction levels they have validated against successfully.

PR #1168, opened 2023-01-10
Assertion 8&10

### LINKED ISSUE #1026: Build Assertion 10: BTC Hodler: The user held Wrapped BTC before a specific date/year
Build **Assertion 10: BTC Hodler: The user held Wrapped BTC before a specific date/year**
- One or more EVM Address(es)/account(s) must be provided and passed to the Parachain from the F/E using the `params` extrinsic param to verify this
- EVM address(es)/account(s) provided must be linked and verified against the `idhub-address` which made the extrinsic call
- The TEE will call The Graph to determine how much **Wrapper BTC** the user has held historically over for the EVM address(es)/account(s) provided.
- The TEE will validate if the user has held **Wrapped BTC** before against all the required years. A **single BTC Hodler VC** will be generated if we find that the user has indeed held BTC before X date/year.
- The JSON payload of the VC will illustrate which date they have held BTC before 2011, 2015, and 2020.

ziming-zung:
This PR
resolves: #1026 #1028


tee-worker/litentry/core/data-providers/src/graphql.rs
```diff
@@ -212,4 +293,21 @@ mod tests {
...
+				VerifiedCredentialsNetwork::Kusama,
+				VerifiedCredentialsNetwork::Polkadot,
+			],
+		};
+		let mut client = GraphQLClient::new();
+		if let Ok(r) = client.query_total_transactions(query) {
+			assert!(r.get(0).unwrap().total_transactions >= 41)
+		} else {
+			assert!(false)
+		}
```
DogLi: how about this?
```
let r = client.query_total_transactions(query);
assert!(r.is_ok());
assert!(r.unwrap().get(0).unwrap().total_transactions > 41);
```

tee-worker/litentry/core/data-providers/src/graphql.rs
```diff
@@ -179,12 +216,56 @@ impl GraphQLClient {
...
+						if let Ok(obj) = serde_json::from_value::<TotalTxsStruct>(x.clone()) {
+							if result.contains_key(&obj.address) {
+								result.get_mut(&obj.address).and_then(|origin| {
+									origin.total_transactions += obj.total_transactions;
+									None::<u8>
+								});
+							} else {
+								result.insert(obj.address.clone(), obj.clone());
+							}
+						}
```
DogLi: why returns a `None::<u8>`?
```
if let Some(origin) = result.get_mut(&obj.address) {
    origin.total_transactions += obj.total_transactions;
} else {
    result.insert(obj.address.clone(), obj.clone());
}
``` 

or

```
let mut result: HashMap<String, u64> = HashMap::new();

for x in element {
    if let Ok(obj) = serde_json::from_value::<TotalTxsStruct>(x.clone()) {
        result[obj.address] = result.get(&obj.address).unwrap_or_default() + obj.total_trasanctions;    
    }
}

let result = result.into_iter().map(|(k, v)| TotalTxsStruct {...}).collect();
```

tee-worker/litentry/core/data-providers/src/graphql.rs
```diff
@@ -179,12 +216,56 @@ impl GraphQLClient {
...
+				});
+			}
+			None::<u8>
+		});
+		if !result.is_empty() {
+			// Ok(result.iter().map(|(_, v)| v.clone()).collect::<Vec<TotalTxsStruct>>())
+			Ok(result.values().cloned().collect::<Vec<TotalTxsStruct>>())
+		} else {
+			Err(Error::GraphQLError("Invalid GraphQL response".to_string()))
+		}
```
DogLi, line 257: Maybe "Empty GraphQL response" would be better. 

if the total txs of a address is `0`, will the GraphQL return nothing or a `TotalTxStruct` with zero `total_transactions`? if it returns nothing, I think it's ok that `result.is_empty()`

tee-worker/litentry/core/data-providers/src/graphql.rs
ziming-zung: Maybe we should also assert that the list is not empty.

tee-worker/litentry/core/data-providers/src/graphql.rs
ziming-zung: Maybe ~http request get an error(status != 200) or~ parses the json error, but I ignore these errors in the method, so I think maybe the empty_graphql_response is not accurate. :sweat_smile:

tee-worker/litentry/core/data-providers/src/graphql.rs
ziming-zung: Yes, you are right. I forgot to write code like this. Too many `and_then` 😅

ziming-zung:
I don't know why `cargo test` failed in tee-worker. It's no problem locally. Who can help me to run `cargo test --release` locally. :expressionless:

tee-worker/litentry/core/assertion-build/src/a8.rs
```diff
@@ -0,0 +1,66 @@
...
+			_ => {
+				debug!("ignore identity: {:?}", identity);
+				None
+			},
+		};
+		if let Some(query) = query {
+			if let Ok(result) = client.query_total_transactions(query) {
+				result.iter().for_each(|v| {
+					total_txs += v.total_transactions;
+				})
```
DogLi: or 
`total_txs = result.iter().map(|v| v.total_transactions).sum();`

tee-worker/litentry/core/data-providers/src/graphql.rs
DogLi: so it will not be an error when the client `get_with` response status != 200?

```
self
			.client
			.get_with::<String, QLResponse>(path, query.as_slice())
```

ziming-zung: It should only have parsing json errors

zTgx:
> I don't know why `cargo test` failed in tee-worker. It's no problem locally. Who can help me to run `cargo test --release` locally. expressionless

I pulled your branch, run `cargo test --release` in tee-worker dir. outpus 1 failed once. when i re-run it, all cases passed.

```shell
---- per_shard_slot_worker_tests::slot_timings_are_correct_with_multiple_shards stdout ----
thread 'per_shard_slot_worker_tests::slot_timings_are_correct_with_multiple_shards' panicked at 'First shard end time, expected: 1673353002653, actual: 1673353002670', sidechain/consensus/slots/src/per_shard_slot_worker_tests.rs:46:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Kailai-Wang:
It really seems to be a server problem, CI passed too

ziming-zung:
> It really seems to be a server problem, CI passed too

Maybe the unit test send request to mock server ,not real one.
If the unit test fail, rerunning the task will take a lot of time

Kailai-Wang:
> > It really seems to be a server problem, CI passed too
> 
> Maybe the unit test send request to mock server ,not real one

Does it? Then it needs to be investigated 😂  a mock server shouldn't have such problem.

ziming-zung:
> > > It really seems to be a server problem, CI passed too
> > 
> > 
> > Maybe the unit test send request to mock server ,not real one
> 
> Does it? Then it needs to be investigated 😂 a mock server shouldn't have such problem.

Sorry, I mean we could modify the test and make it send request to mock server :sweat_smile:

Kailai-Wang:
Oh sorry, yes for unittest we could use a mockserver. GraphQL server should go to integration-test / e2e test

But let's do it in another PR if we go for it.

Kailai-Wang (approved):
Thanks, I believe some control logic still needs to be adjusted when the VC generation kicks in, but that's something for later

PR #1172, opened 2023-01-10
1138 refinement of generate upstream patch

### LINKED ISSUE #1138: refinement generate upstream patch
### Context
There are two `generate_upstream_patch` now:
- one for pallets: teeracle, teerex, sidechain related.
- one for tee-worker

### Task
The file structure/naming/usage of scripts/result are a bit confusing now. Need some cleanup.

### Acceptance criteria
Both `generate_upstream_patch` can work clearly and independently.

---

BillyWooo:
As title:
pallets upstream generator: 
- As the whole pallets-repo is splitted and only part of the repo is used, many manual resolve steps are still necessary.
worker upstream generator:
- Patch won't go to wrong folder/file any more. Just need to resolve conflict as usual.

resolves #1138


Marky-Shi (approved):

scripts/generate_integritee_pallets_upstream_patch.sh
```diff
@@ -8,16 +8,18 @@ cleanup() {
...
+# From upstream repo (https://github.com/integritee-network/pallets),
+# only 'teerex', 'teeracle', 'sidechain' and 'primitives' are taken in.
 
-UPSTREAM="https://github.com/integritee-network/pallets.git"
-ROOTDIR=$(git rev-parse --show-toplevel)
+# The patch will be generated under ./pallets/upstream.patch
+
+UPSTREAM="https://github.com/integritee-network/pallets"
+ROOTDIR0=$(git rev-parse --show-toplevel)
+ROOTDIR="$ROOTDIR0/pallets"
```
Kailai-Wang: TBH the names are confusing - normally you only have one `ROOTDIR`, it's hard to differentiate between `ROOTDIR` and `ROOTDIR0`.

If you want a different working directory, maybe simply `PALLETS_DIR` or so

.gitignore
```diff
@@ -16,4 +16,6 @@ tags.lock
 
 recipe.json
 
-gethdata
\ No newline at end of file
+gethdata
+
+*.patch
```
Kailai-Wang, line 21: not sure if this will work, do we need `**/upstream.patch`?

scripts/generate_integritee_pallets_upstream_patch.sh
```diff
@@ -44,15 +47,22 @@ cd worker
...
+git rev-parse --short HEAD > "$ROOTDIR/upstream_commit"
 
 echo "======================================================================="
 echo "upstream_commit is updated."
 echo "be sure to fetch the upstream to update the hashes of files."
-echo "upstream.patch is generated, to apply it, run:"
-echo '  # git am -3 --include=test-utils --include=teerex --include=teeracle --include=sidechain --include=primitives < upstream.patch'
+echo ""
+echo "upstream.patch is generated, to apply it, RUN FROM $ROOTDIR0:"
+echo "  git am -3 --directory=pallets < pallets/upstream.patch"
```
Kailai-Wang: will it find the `primitives/` and apply the patch even in the wrong dir?

Kailai-Wang (commented):
Thank you! I have a few comments tho.

Btw, does it make sense to merge two scripts into one? Something like
```
generate_upstream_patch.sh <pallet|worker> <working-dir> <upstream-hash>
arg2 and arg3 are optional
``

Because I see many duplicated commands

File: .gitignore (hunk 2 above)
Comment (BillyWooo) [reply .gitignore:21]: it will work.

File: scripts/generate_integritee_pallets_upstream_patch.sh (hunk 3 above)
Comment (BillyWooo) [reply scripts/generate_integritee_pallets_upstream_patch.sh]: Yes. It will find. Actually I tried several ways/directories back and forth. Either way needs manually check and even move files around. To set '--directory=pallets' is the least effort.

File: scripts/generate_integritee_pallets_upstream_patch.sh (hunk 1 above)
Comment (BillyWooo) [reply scripts/generate_integritee_pallets_upstream_patch.sh]: Sure. I can rename them.

**Comment by BillyWooo** [comment]
> Thank you! I have a few comments tho.
> 
> Btw, does it make sense to merge two scripts into one? Something like
> 
> ```
> generate_upstream_patch.sh <pallet|worker> <working-dir> <upstream-hash>
> arg2 and arg3 are optional
> ``
> 
> Because I see many duplicated commands
> ```

I just thought it may be too much for one commit. But I think it worth to try it out.

**Review by Kailai-Wang — CHANGES_REQUESTED** [review]
There're quite some typos/syntax errors 😂 
I strongly recommend running the script locally once before we re-review the logic

File: scripts/generate_upstream_patch.sh [hunk 4]
```diff
@@ -0,0 +1,138 @@
...
+
+while getopts ":p:w:h:" opt; do
+	case $opt in
+		p)
+			has_pallets=true
+			pallets_commit=$OPTARG
+			;;
+		w)
+			HAS_WORKER=true
+			WORKER_COMMIT=$OPTARG
```
Kailai-Wang: Mixed big and small letters seem confusing, is it intended or wrong?

I tried to run `./scripts/generate_upstream_patch.sh -p`, then both pallets and worker patches were generated. Is it intended?

scripts/generate_upstream_patch.sh
```diff
@@ -0,0 +1,138 @@
...
+	then
+		echo "  git am -3 --exclude=tee-worker/Cargo.lock --exclude=tee-worker/enclave-runtime/Cargo.lock --directory=tee-worker < tee-worker/upstream_worker.patch"
+	fi
+	echo ""
+	echo "after that, please:"
+	echo "- pay special attention: "
+	if $HAS_PALLETS == "true"
+	then
+		echo "  * ALL changes/conflicts from pallets/upstream.patch should ONLY apply into:"
+		echo "    - pallets/(sidechain, teeracle, teerex, test-utils)"
```
Kailai-Wang: and parentchain

scripts/generate_upstream_patch.sh
```diff
@@ -0,0 +1,138 @@
...
+  echo "		-p specify the tag|branch|commit-hash for upstream pallets"
+  echo "		-w specify the tag|branch|commit-hash for upstream worker"
+}
+
+# This function generates a patch for the diffs between commit-A and commit-B
+# of the upstream repo, where
+# commit-A: the commit recorded in ./<TARGET_DIR>/upstream_commit
+# commit-B: the HEAD commit of upstream master or a given commit
+#
+# A patch will be generated under ./<TARGET_DIR>/upstream.patch
```
Kailai-Wang, line 32: should be `upstream_pallets.patch` or `upstream_worker.patch`

scripts/generate_upstream_patch.sh
```diff
@@ -0,0 +1,138 @@
...
+generate_upstream_patch() {
+	local TARGET_DIR=$1
+	local UPSTREAM_NAME=$2
+	local UPSTREAM_URL=$3
```
Kailai-Wang: Actually these three args can be consolidated into one, `worker` or `pallets`, only TARGET_DIR needs to be adjusted. 

If that's the case, we don't need `getopts` handling either:
```
generate_upstream_patch.sh <pallets|worker> [upstream-hash]
```
is simple enough.

But I'm also fine if we leave as it is

scripts/generate_upstream_patch.sh
BillyWooo: -p must follow a parameter; otherwise it will be dropped off.

scripts/generate_upstream_patch.sh
BillyWooo: maybe `getopts` is still needed. Here it is designed to handle: `pallets only`, `worker only`, `pallets and worker`.

Kailai-Wang (approved):
Approved, thanks!

Let's check the $OPTARGS later: e.g. `$0 -p` should be the same as `$0 -p master`

PR #1176, opened 2023-01-10
add Credentials A1 to TEE

jingleizhang:
Just a simple example to show the Credential workflow:

- user requests `Assertion::A1` through Parachain `request_vc`
- Credential generates an `unsigned_credential` from json templates
- A1 generates assertions into the credential
- TEE passes the encrypted credential back to Parachain

There are some important following tasks:

- [x] unsigned_credential needs to store the issuer's attestation metadata
- [ ] and needs to be signed
- [x] unsigned_credential is modified between stf-task sender and receiver, which needs to be supported somehow
- [x] make the workflow of Credential be clear and clean


tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -574,9 +579,50 @@ where
...
+								node_metadata_repo
+									.get_from_metadata(|m| m.vc_some_error_call_indexes())??,
+								"get_user_shielding_key".as_bytes(),
+								"error".as_bytes(),
+							)));
+						}
+					},
+					Err(error) => {
+						debug!("credential to_json() got error {:?}", error);
+
```
DogLi: I think `error!` should be better.

jingleizhang: Correct, thanks!

tee-worker/litentry/core/credentials/src/credentials.rs
```diff
@@ -0,0 +1,269 @@
+// Copyright 2020-2022 Litentry Technologies GmbH.
```
Kailai-Wang: The file structure `credentials/src/credentials.rs` seems strange: we have a crate and file with the same name. Can we move the content of this file to `lib.rs`?

tee-worker/litentry/core/credentials/src/credentials.rs
```diff
@@ -0,0 +1,269 @@
...
+				let raw = include_str!("templates/a1.json");
+				let credential: Credential = Credential::from_template(raw, who)?;
+				Ok(credential)
+			},
+			_ => Err(Error::UnsupportedAssertion),
+		}
+	}
+
+	pub fn add_assertion_a1(&mut self, _web2_cnt: i32, _web3_cnt: i32) {
+		self.credential_subject.assertions = "\"or\": [{\"src\": \"$web2_account_cnt\", \"op\": \">\", \"dsc\": \"0\",},{\"src\": \"$web3_account_cnt\", \"op\": \">\", \"dsc\": \"0\",}]".to_string();
```
Kailai-Wang: maybe let's use raw string literals for better readability, something like this:
```
format!(r#"{{"type": "type1", "type2": {}}}"#, var1);
```

tee-worker/litentry/core/credentials/src/credentials.rs
```diff
@@ -0,0 +1,269 @@
...
+
+pub fn now() -> String {
+	let ts = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+	let naive = NaiveDateTime::from_timestamp_opt(ts.as_secs() as i64, ts.subsec_nanos()).unwrap();
+	let datenow_time = DateTime::<FixedOffset>::from_utc(naive, FixedOffset::east_opt(0).unwrap());
+	datenow_time.to_rfc3339_opts(SecondsFormat::Secs, true)
+}
+
+#[derive(Serialize, Deserialize, Encode, Decode, Clone, Debug, PartialEq, Eq, TypeInfo)]
+pub enum ProofType {
```
Kailai-Wang: We'll need quite some comments in this file, as it's the core part of the VC format.

E.g. explanations of `ProofType`, `CredentialType`: what do they exactly mean and what are the possible values? Same applies to other important structs/fields.

pallets/vc-management/src/lib.rs
```diff
@@ -44,8 +44,8 @@ pub use assertion::*;
 mod schema;
 pub use schema::*;
 
-// VCID type in the registry, maybe we want a "did:...." format?
-pub type VCID = u64;
+// hash of a VC, computed via blake2_256
+pub type VCHASH = H256;
```
Kailai-Wang: Maybe I need to get this, but after the user gets a VC, how can he verify the VC hash without a unique ID? He can only tell the hash of the VC exists on chain, but not necessarily **match** the VC he gets, isn't it 🤔

Kailai-Wang (commented):
I haven't finished reviewing it, but maybe let's double-check some questions first 🤔

tee-worker/litentry/core/credentials/src/credentials.rs
jingleizhang: for sure, will refactor this model and add comments.

tee-worker/litentry/core/credentials/src/credentials.rs
```diff
@@ -0,0 +1,269 @@
...
+	string::{String, ToString},
+	time::{SystemTime, UNIX_EPOCH},
+	vec::Vec,
+};
+
+pub const PROOF_PURPOSE: &str = "assertionMethod";
+pub const MAX_CREDENTIAL_SIZE: usize = 2048;
+
+pub fn now() -> String {
+	let ts = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
```
BillyWooo: Will this part of code execute inside TEE? I think "SystemTime::now()" is suspicious to run inside TEE.

jingleizhang: - It is not possible to have a trusted and precise timestamp inside TEE, but it is feasible to measure the **Duration** between two timestamps from the support of teaclave sgx_tstd https://github.com/apache/incubator-teaclave-sgx-sdk/blob/master/sgx_tstd/src/time.rs

- The "SystemTime::now()" could run inside TEE enclave, the scenario is here https://github.com/litentry/litentry-parachain/blob/250d02a9affddaeda4d629283a35006bc3841387/tee-worker/enclave-runtime/src/top_pool_execution.rs#L94

pallets/vc-management/src/lib.rs
jingleizhang: Anyway, the VCID and VCHASH are both needed in the VCMP. Fixing this right now.

tee-worker/litentry/core/credentials/src/credentials.rs
jingleizhang: added

jingleizhang:
cc @Kailai-Wang @BillyWooo @DogLi, all reviews are fixed.

tee-worker/app-libs/stf/Cargo.toml
```diff
@@ -39,6 +39,7 @@ sp-core = { default-features = false, features = ["full_crypto"], git = "https:/
 sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.36" }
 
 # litentry
+lc-credentials-tee = { path = "../../litentry/core/credentials", default-features = false }
```
Kailai-Wang: Sorry I'm a bit picky here, why with `-tee` suffix? It's under tee-worker already and there's no other confusion

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
```diff
@@ -65,17 +101,31 @@ where
...
-			Assertion::A1 => lc_assertion_build::a1::build(self.req.vec_identity.to_vec()),
+			Assertion::A1 => lc_assertion_build::a1::build(
+				self.req.vec_identity.clone(),
+				self.req.credential.clone(),
+			)
+			.map(|credential| {
+				let encoded_callback = TrustedCallFork::build_assertion_runtime(
+					self.context.enclave_signer.get_enclave_account().unwrap(),
+					self.req.who.clone(),
+					Box::new(credential),
```
Kailai-Wang: A few things that I don't understand:
- why do we need `TrustedCallFork`? If you need to change something from the `encoded_callback`, you could decode it, extract any necessary parameters and rebuild the encoded callback.
- this would call `build_assertion_runtime` synchronously - but it's another thread, we don't have any sync mechanism to shared data (e.g. sp_io storages). I think it's best to submit the callback back to the top pool. 
- maybe we don't need to submit it back to the top pool at all. Can't we simply send the result to parachain? If you need the shielding key you could simply pass it in as request parameters.

Please feel free to share your thoughts too @zhizming-zhong

tee-worker/litentry/core/stf-task/sender/src/lib.rs
```diff
@@ -98,6 +98,7 @@ pub struct AssertionBuildRequest {
 	pub who: AccountId,
 	pub assertion: Assertion,
 	pub vec_identity: BoundedVec<Identity, MaxIdentityLength>,
+	pub credential: Credential,
```
Kailai-Wang: Question: why do we need to pass `Credential` around? Can't we generate it on the stf-task handler side? (do we have enough information?)

Kailai-Wang (changes requested):
I haven't reviewed every line but only looked through the core logic.

I assume some VC fields will be improved over time, but it's important that we have a starting skeleton 👍 

I have some major concerns though, mainly regarding the extra `credential` field and the post-building workflow. Let's discuss it offline if you feel necessary.

tee-worker/app-libs/stf/Cargo.toml
jingleizhang: get it, its `-tee` suffix could be removed to avoid confusion, as it is already under tee-worker

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
jingleizhang: `TrustedCallFork` is kind of a hack to the `encoded_callback`.

Generating a credential needs the following steps: 
1. Generating an unsigned credential from templates  <-- `TrustedCallSigned build_assertion_preflight`
2. Check the assertion and get its built return value  <-- `lc-assertion-build` processed by stf_task receiver
3. Add Issuer's enclave metadata (more might be added if needed) <-- `TrustedCallSigned build_assertion_runtime`
4. Sign the credential by enclave <-- `TrustedCallSigned build_assertion_runtime`
5. Pass the signed credential back to parachain <-- `TrustedCallSigned build_assertion_runtime`

The major difficulty is `unsigned_credential` is needed to be shared between stf-task sender and receiver, but there isn't a sync mechanism or a runtime pallet for it to share/store the value.

 Let's discuss it offline to figure out how to make the workflow more simple and clear.

tee-worker/litentry/core/stf-task/sender/src/lib.rs
jingleizhang: The reason here is quite relative to the question above. 

It might be possible:

- pass all needed data to stf-task handler
- simply send the result to parachain, avoid submitting it back to the top pool

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
ziming-zung: I think the biggest problem is that crate `stf` can't be imported into `stf_task_receiver`, so that there will be a circular dependency problem

jingleizhang:
cc @Kailai-Wang @BillyWooo @zhizming-zhong , the refactoring of assertion building workflow is ready.

Kailai-Wang (approved):
Overall looks good.

We'd need a ts-test to test the validity of the logic.

PR #1186, opened 2023-01-16
cache `list-workers` result for use later

zTgx:
as the title.
cached `list-workers` command results in cli shell for use later.


BillyWooo:
What's the reason to cache it?

zTgx:
> What's the reason to cache it?

it's not a big deal to request `list-workers` twice to parse mrenclave, but I think we can parse it from the first time `list-workers` called result, no need to request it again.

Kailai-Wang (approved):

PR #1193, opened 2023-01-16
Check extrinsic status in `indirect_calls_executor`

### LINKED ISSUE #1092: Check extrinsic status in `indirect_calls_executor`
### Context
As far as I experimented, the `IndirectCallsExecutor` simply iterates over all extrinsics in a parentchain block and handles them, without checking the extrinsic status.

I intentionally misconfigured the parameter so that the extrinsic failed and threw out an error, but this extrinsic was still parsed. This seems incorrect.

---

ziming-zung:
This PR:
resolves #1092 

~Currently it only checks if `create_identity` was successful.~
~If we think this solution is okay, I will continue to modify it later~


tee-worker/core-primitives/node-api/api-client-types/src/lib.rs
```diff
@@ -42,6 +45,31 @@ pub type ParentchainExtrinsicParamsBuilder = PlainTipExtrinsicParamsBuilder;
 pub type ParentchainUncheckedExtrinsic<Call> =
 	UncheckedExtrinsicV4<Call, SubstrateDefaultSignedExtra<PlainTip>>;
 
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct ParentchainUncheckedExtrinsicWithStatus<Call: Decode> {
+	pub xt: UncheckedExtrinsicV4<Call, SubstrateDefaultSignedExtra<PlainTip>>,
+	pub status: bool,
+}
```
DogLi: here we defined `ParentchainUncheckedExtrinsic<Call>`, use it.

ziming-zung: I move it to `itp_types`. Maybe it is more reasonable

scripts/pre-commit.sh
```diff
@@ -0,0 +1,30 @@
...
+# It is recommended to execute the script before commit
+# which will help us to reduce test/fmt/clippy failures in CI
+
+start=$(date +%s)
+
+make fmt
+taplo fmt
+
+root_dir=$(git rev-parse --show-toplevel)
+CARGO_TARGET_DIR=${root_dir}/tmp
```
Kailai-Wang: This is not a general setting :) Actually we have the pre-commit hooks for formatting already

tee-worker/core-primitives/types/Cargo.toml
```diff
@@ -17,6 +17,9 @@ serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 # local dependencies
 itp-sgx-runtime-primitives = { path = "../../core-primitives/sgx-runtime-primitives", default-features = false }
 
+# scs
+substrate-api-client = { default-features = false, git = "https://github.com/scs/substrate-api-client.git", branch = "polkadot-v0.9.36" }
```
Kailai-Wang, line 21: Let's not forget to add `/std` in std features

tee-worker/core/parentchain/indirect-calls-executor/src/lib.rs
```diff
@@ -338,6 +345,30 @@ mod test {
 		assert_eq!(1, top_pool_author.pending_tops(shard_id()).unwrap().len());
 	}
 
+	#[test]
+	fn indirect_call_can_not_be_added_to_pool() {
```
Kailai-Wang: I think a better name would be `failed_indirect_call_is_skipped' or `failed_indirect_call_can_not_be_added_to_pool`

tee-worker/service/src/parentchain_handler.rs
```diff
@@ -142,6 +146,49 @@ where
...
+								"block:{:?}, extrinsic index: {:?}, success: {:?}",
+								&signed_block.block.header.number, i, success,
+							);
+						}
+
+						// `indirect_calls_executor` should know the if tx was successful.
+						// if we change the block type `sp_runtime::generic::Block` or `sp_runtime::generic::SignedBlock`,
+						// this change will affect many structs or files, like block_importer, block_import_dispatcher, consensus and so on.
+						// `OpaqueExtrinsic` (Vec<u8>) encoded value contains tx status, and decode Vec<u8> in `indirect_calls_executor`,
+						// this solution is probably the least change
```
Kailai-Wang: Maybe let's move this comment to line 149 for better readability, it also helps to explain why we try to get the events of parentchain blocks

tee-worker/service/src/parentchain_handler.rs
```diff
@@ -142,6 +146,49 @@ where
...
+							.filter(|event| {
+								if let Ok(event) = event {
+									event.pallet_name() == "System"
+										&& event.variant_name() == "ExtrinsicSuccess"
+										&& event.phase() == Phase::ApplyExtrinsic(i as u32)
+								} else {
+									false
+								}
+							})
+							.count() > 0;
```
Kailai-Wang, line 180: Additional question, is `count()` always 1 if the extrinsic succeeds? Any cases where count() can be > 1?

Kailai-Wang (commented):
Looks good overall, thanks. I have left comments for several small things.

A general question though: instead of inserting the `status` to the opaque vec, what prevents us from simply dropping the failed extrinsics when importing paentchain blocks? (handling in `parentchain_handler.rs`)

scripts/pre-commit.sh
zhizming-zhong: > Looks good overall, thanks. I have left comments for several small things.
> 
> A general question though: instead of inserting the `status` to the opaque vec, what prevents us from simply dropping the failed extrinsics when importing paentchain blocks? (handling in `parentchain_handler.rs`)

Dropping failed extrinsics might not be a good solution. In certain cases, we may need to know which extrinsics failed or need to count which extrinsics failed. What do you think?

ziming-zung:
> Looks good overall, thanks. I have left comments for several small things.
> 
> A general question though: instead of inserting the `status` to the opaque vec, what prevents us from simply dropping the failed extrinsics when importing paentchain blocks? (handling in `parentchain_handler.rs`)

Dropping failed extrinsics might not be a good solution. In certain cases, we may need to know which extrinsics failed or need to count which extrinsics failed. What do you think?

scripts/pre-commit.sh
ziming-zung: Actually, this script just is a patch to the original pre-commit (clippy, unit test)

Kailai-Wang:
> > Looks good overall, thanks. I have left comments for several small things.
> > A general question though: instead of inserting the `status` to the opaque vec, what prevents us from simply dropping the failed extrinsics when importing paentchain blocks? (handling in `parentchain_handler.rs`)
> 
> Dropping failed extrinsics might not be a good solution. In certain cases, we may need to know which extrinsics failed or need to count which extrinsics failed. What do you think?

I can't think of such use cases 😂  at least not in the sidechain. It seems we only care about the successfully executed extrinsics. Shall we ask integritee if there's any case?

Kailai-Wang (approved):
Looks good thanks :)

PR #1200, opened 2023-01-17
Remove GH cache after PR is merged

### LINKED ISSUE #938: Remove cache when a branch gets deleted
### Context
Github will start to evict caches once [the total size is over 10G](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).

We have these 2 caches:
```
cargo-<branch-name>-<hash>
sccache-<branch-name>-<hash>
```
See https://github.com/litentry/litentry-parachain/actions/caches

Caches will be inevitably generated during PR creation and merging. When a PR gets closed and the branch gets deleted, we'd like to automatically delete the corresponding caches to make as much room as possible

---

BillyWooo:
This action workflow will remove the PR branch related caches after the PR is closed.

resolves #938

### DIFF
--- a/.github/workflows/rm_gh_cache_after_pr_closed.yml
+++ b/.github/workflows/rm_gh_cache_after_pr_closed.yml
@@ -1,0 +1,44 @@
+name: Cleanup caches after PR closed
+on:
+  pull_request:
+    # branches:
+    #   - dev
+    types:
+      - closed
+  workflow_dispatch:
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
+
+      - name: Cleanup
+        run: |
+          gh extension install actions/gh-actions-cache
+
+          REPO=${{ github.repository }}
+          BRANCH=${{ github.head_ref }}
+
+          cacheKeysForPR=$(gh actions-cache list -R $REPO | cut -f 1 )
+
+          ## Setting this to not fail the workflow while deleting cache keys.
+          set +e
+          echo "Deleting caches after PR closed for branch=$BRANCH ..."
+          for cacheKey in $cacheKeysForPR
+          do
+              if [[ $cacheKey == *"$BRANCH"* ]]
+              then
+                echo "  Delete cache: $cacheKey"
+                gh actions-cache delete $cacheKey -R $REPO --confirm
+              fi
+          done
+          echo "Done"
+        env:
+          GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/rm_gh_cache_after_pr_closed.yml
+++ b/.github/workflows/rm_gh_cache_after_pr_closed.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   cleanup:
     runs-on: ubuntu-latest
+    permissions: write-all
     steps:
       - name: Check out code
         uses: actions/checkout@v3
--- a/.github/workflows/rm-gh-cache-after-pr-closed.yml
+++ b/.github/workflows/rm-gh-cache-after-pr-closed.yml
@@ -1,11 +1,8 @@
 name: Cleanup caches after PR closed
 on:
   pull_request:
-    # branches:
-    #   - dev
     types:
       - closed
-  workflow_dispatch:
 
 jobs:
   cleanup:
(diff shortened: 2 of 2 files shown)


.github/workflows/rm_gh_cache_after_pr_closed.yml
```diff
@@ -0,0 +1,45 @@
+name: Cleanup caches after PR closed
```
Kailai-Wang: Let's rename the file with hyphens: `x-y-z.yml` (to be consistent with other files)

.github/workflows/rm_gh_cache_after_pr_closed.yml
```diff
@@ -0,0 +1,45 @@
...
+on:
+  pull_request:
+    # branches:
+    #   - dev
```
Kailai-Wang: So this action can be triggered on PRs towards any branch, is that intended? If so let's remove the commented code

.github/workflows/rm_gh_cache_after_pr_closed.yml
```diff
@@ -0,0 +1,45 @@
...
+    types:
+      - closed
+  workflow_dispatch:
```
Kailai-Wang: What's the branch if it's triggered by `workflow_dispatch` (I assume it's null), then run logic will be broken.

.github/workflows/rm_gh_cache_after_pr_closed.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    permissions: write-all
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
```
Kailai-Wang: Why do we need this step?

.github/workflows/rm_gh_cache_after_pr_closed.yml
```diff
@@ -0,0 +1,45 @@
...
+          BRANCH=${{ github.head_ref }}
+
+          cacheKeysForPR=$(gh actions-cache list -R $REPO | cut -f 1 )
+
+          ## Setting this to not fail the workflow while deleting cache keys.
+          set +e
+          echo "Deleting caches after PR closed for branch=$BRANCH ..."
+          for cacheKey in $cacheKeysForPR
+          do
+              if [[ $cacheKey == *"$BRANCH"* ]]
```
Kailai-Wang: Can we simplify it by `grep` in line 30?

Kailai-Wang (changes requested):
Have some questions. Btw is this GHA tested somewhere? If so we can put the link here

BillyWooo:
> Have some questions. Btw is this GHA tested somewhere? If so we can put the link here

Yes. Tested here: https://github.com/BillyWooo/litentry-parachain/actions/runs/3942679464/jobs/6746609073
Also here: https://github.com/BillyWooo/litentry-parachain/actions/runs/3947330343/jobs/6756029090

.github/workflows/rm_gh_cache_after_pr_closed.yml
BillyWooo: Oh, yes.

.github/workflows/rm_gh_cache_after_pr_closed.yml
BillyWooo: Yes. Intended. Will remove.

.github/workflows/rm_gh_cache_after_pr_closed.yml
BillyWooo: forgot to remove. was used for testing.

.github/workflows/rm_gh_cache_after_pr_closed.yml
BillyWooo: Yes.

.github/workflows/rm_gh_cache_after_pr_closed.yml
BillyWooo: Yes. Should be removed. Not necessary.

Kailai-Wang (approved):
Thank you, we can test it in real scenarios

Kailai-Wang:
I think we still need a small adjustment for this one:
```
cacheKeysForPR=$(gh actions-cache list -R $REPO | cut -f 1 | grep "$BRANCH")
```
it's possible that `grep` finds nothing because the target branch didn't produce any caches, like this run:
https://github.com/litentry/litentry-parachain/actions/runs/3948688455

but the exit code is 1

I'll submit a minor fix for this

PR #1201, opened 2023-01-17
Remove stale files

BillyWooo:
These files originally created to test https request long time ago. Not needed any more.


Kailai-Wang (commented):
Thank you. What about `https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/scripts/https_test.sh` ?

Kailai-Wang (approved):
Thank you

Issue #1199, opened 2023-01-17
Refactor `indirect_calls_executor`


ziming-zung:
### Context
Currently, there are a lot of if-else codes in `indirect_calls_executor` and most of them are repetitive codes. So we can do some simple refactoring to extract some code with the same functionality to trait.

PR #1202, opened 2023-01-18
Refactor `indirect_call_executor`

ziming-zung:
This PR resloves #1199


tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs
```diff
@@ -247,179 +246,43 @@ impl<ShieldingKeyRepository, StfEnclaveSigner, TopPoolAuthor, NodeMetadataProvid
...
+				&request_vc,
+			];
+			for executor in executors {
+				match executor.decode_and_execute(self, &mut encoded_xt_opaque.as_slice()) {
+					Ok(ExecutionStatus::Success) | Ok(ExecutionStatus::Skip) => break,
+					Ok(ExecutionStatus::NextExecutor) => continue,
+					Err(e) => {
+						log::error!("fail to execute indirect_call. due to {:?} ", e);
+						// continue or return error???
+						continue
```
DogLi: return error, I think. because we decode first, then execute, the error is the execute error. if the decode fails, it will continue.

tee-worker/core/parentchain/indirect-calls-executor/src/lib.rs
```diff
@@ -33,5 +33,11 @@ pub mod sgx_reexport_prelude {
...
 pub mod error;
 pub mod indirect_calls_executor;
-
+pub mod litentry;
 pub use indirect_calls_executor::*;
+
+pub enum ExecutionStatus {
+	Success,
+	NextExecutor,
+	Skip,
```
DogLi: I don't see the situation when using the `Skip`. It can be removed, if we need it, add it again.

tee-worker/core/parentchain/indirect-calls-executor/src/litentry/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+	}
+}
+
+pub(crate) trait DecorateExecutor<
+	ShieldingKeyRepository,
+	StfEnclaveSigner,
+	TopPoolAuthor,
+	NodeMetadataProvider,
+>
+{
```
DogLi: I think the `DecorateExecutor` is not necessary. use the `decode_and_execute` function in the trait `Executor` directly, or rename `Executor` to `DecorateExecutor`

tee-worker/core/parentchain/indirect-calls-executor/src/litentry/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+		if let Ok(xt) = self.decode(input) {
+			if self.is_target_call(xt.function.clone(), context.node_meta_data_provider.as_ref()) {
+				self.execute(context, xt)
+			} else {
+				Ok(ExecutionStatus::NextExecutor)
+			}
+		} else {
+			Ok(ExecutionStatus::NextExecutor)
+		}
+	}
```
DogLi: in fact, we don't need the `ExecutionStatus`, just move the logic to the `for` loop, because it's so simple.

tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs
Kailai-Wang: Yea I think we should stop the processing in this arm: the decode and `is_target_call` errors are already handled above

tee-worker/core/parentchain/indirect-calls-executor/src/litentry/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+		input: &mut &[u8],
+	) -> Result<ExecutionStatus, Error> {
+		if let Ok(xt) = self.decode(input) {
+			if self.is_target_call(xt.function.clone(), context.node_meta_data_provider.as_ref()) {
+				self.execute(context, xt)
+			} else {
+				Ok(ExecutionStatus::NextExecutor)
+			}
+		} else {
+			Ok(ExecutionStatus::NextExecutor)
```
Kailai-Wang: Should we try the next executor if decoding fails? I.e. is this decoding executor-specific?

tee-worker/core/parentchain/indirect-calls-executor/src/litentry/mod.rs
Kailai-Wang: That makes sense. Can we merge that? Unless we have a reason

tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs
Kailai-Wang: Let's write comments here no matter which way we go

Kailai-Wang (commented):
Thanks, good ideas. Apart from some inline comments, I have two general points:

- Shall we go one step further and refactor the integritee part using Executor too? We can remove the macros then. => I expect integritee to be happy to accept it as PR
- Any possibility to reduce the repetition in each executor implementation? I mean we have to write for each executor, maybe use macros?
```
impl<ShieldingKeyRepository, StfEnclaveSigner, TopPoolAuthor, NodeMetadataProvider>
	Executor<ShieldingKeyRepository, StfEnclaveSigner, TopPoolAuthor, NodeMetadataProvider>
	for CreateIdentity
where
	ShieldingKeyRepository: AccessKey,
	<ShieldingKeyRepository as AccessKey>::KeyType: ShieldingCryptoDecrypt<Error = itp_sgx_crypto::Error>
		+ ShieldingCryptoEncrypt<Error = itp_sgx_crypto::Error>,
	StfEnclaveSigner: StfEnclaveSigning,
	TopPoolAuthor: AuthorApi<H256, H256> + Send + Sync + 'static,
	NodeMetadataProvider: AccessNodeMetadata,
	NodeMetadataProvider::MetadataType: IMPCallIndexes + TeerexCallIndexes + VCMPCallIndexes,
```

tee-worker/core/parentchain/indirect-calls-executor/src/litentry/mod.rs
ziming-zung: After discussing with @DogLi, we decided to keep the trait `DecorateExecutor`. Since there is a associated type `Call` in trait `Executor`, `Executor` can't be directly inserted into Vector in `indriect_call_executor`, which requires `DecorateExecutor`. 
Now I just put the implementation of the function `decode_and_execute` into `DecorateExecutor`, it makes the code more reasonable.

tee-worker/core/parentchain/indirect-calls-executor/src/litentry/mod.rs
ziming-zung: Same as above.

tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs
ziming-zung: After my thinking, I also agree to return error. Because the logic of the original code needs to be keep the same, when encountering an error, exit the function

tee-worker/core/parentchain/indirect-calls-executor/src/litentry/mod.rs
ziming-zung: It should try the next one. Function logic keep the same as the original one. This means it trying to decode until it is correct ~`function/call` type is correct

ziming-zung:
> Thanks, good ideas. Apart from some inline comments, I have two general points:
> 
> * Shall we go one step further and refactor the integritee part using Executor too? We can remove the macros then. => I expect integritee to be happy to accept it as PR
> * Any possibility to reduce the repetition in each executor implementation? I mean we have to write for each executor, maybe use macros?
> 
> ```
> impl<ShieldingKeyRepository, StfEnclaveSigner, TopPoolAuthor, NodeMetadataProvider>
> 	Executor<ShieldingKeyRepository, StfEnclaveSigner, TopPoolAuthor, NodeMetadataProvider>
> 	for CreateIdentity
> where
> 	ShieldingKeyRepository: AccessKey,
> 	<ShieldingKeyRepository as AccessKey>::KeyType: ShieldingCryptoDecrypt<Error = itp_sgx_crypto::Error>
> 		+ ShieldingCryptoEncrypt<Error = itp_sgx_crypto::Error>,
> 	StfEnclaveSigner: StfEnclaveSigning,
> 	TopPoolAuthor: AuthorApi<H256, H256> + Send + Sync + 'static,
> 	NodeMetadataProvider: AccessNodeMetadata,
> 	NodeMetadataProvider::MetadataType: IMPCallIndexes + TeerexCallIndexes + VCMPCallIndexes,
> ```

* Sure, no problem
* Personally, I'm not a fan of macros. Even though it looks verbose now, it's easier to read and understand. Of course, if we  need macros, I can change to this way. In addition, the IDE's support for macros is not very good

tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs
ziming-zung: So strange, I found different handling about error  in this function. See:
https://github.com/litentry/litentry-parachain/blob/2371278fe9e014c4111865480b5f86d22511896a/tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs#L224-L233

https://github.com/litentry/litentry-parachain/blob/2371278fe9e014c4111865480b5f86d22511896a/tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs#L255-L261

Maybe we should take the same handling as the first one.

Kailai-Wang:
Let's leave it as it is, I was thinking of something like trait wrapper, but maybe not so easy to do it :(

Once the integritee part is refactored, I'm happy to merge it

tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs
Kailai-Wang: Do you mean it should have error'ed out or not continued with the next `if`? I actually think our handling is correct while integritee's is wrong.

Btw I saw they are pushing successfully processed calls to `executed_shielding_calls`, maybe we should do the same? Although these are just used to compute a merkle hash

tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs
```diff
@@ -143,7 +147,11 @@ where
 		Ok(())
 	}
 
-	fn submit_trusted_call(&self, shard: ShardIdentifier, encrypted_trusted_call: Vec<u8>) {
+	pub(crate) fn submit_trusted_call(
+		&self,
+		shard: ShardIdentifier,
+		encrypted_trusted_call: Vec<u8>,
```
BillyWooo: maybe remove the comma `,`. Not necessary change.

tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls_executor.rs
BillyWooo: I think here it shouldn't return error. The code Zhiming cited is logic from us, which is not appropriate according to INTEGRITEE's intention.

Kailai-Wang: Sorry maybe I didn't make myself clear, I actually meant `break` here. We should stop processing this extrinsic (= don't try the next executor) and continue with the next extrinsic. When all extrinsics are processed, we call `create_processed_parentchain_block_call` to report to the parachain.

And yes, we shouldn't error out using the `?`: `create_processed_parentchain_block_call` needs to be called in any case.

DogLi:
LGFM

Kailai-Wang:
Almost done. I have two extra small suggestions:
- I think we should use the same handling as `shield_funds`, it should be one of the executors and we need to push successfully processed extrinsics to `executed_shielding_calls` -> we may need to rename this. But, it's strange that call_worker doesn't get pushed in integritee-worker, which I think is wrong.
- What about merging the content of  `indirect_calls_executor.rs` to `lib.rs`? We have the parent folder name as `indirect_calls_executor` already. We'll submit this change to upstream too

Kailai-Wang (approved):
Thanks! Please feel free to merge it when the content of `indirect_calls_executor.rs` is moved to `lib.rs`

PR #1211, opened 2023-01-19
1206 worker config to enclave

### LINKED ISSUE #1206: `worker-config` doesn't work in `enclave-runtime`
### Context
Currently, we can specify `worker-config` via the flag `--running-mode`. See: https://github.com/litentry/litentry-parachain/blob/2371278fe9e014c4111865480b5f86d22511896a/tee-worker/service/src/main.rs#L136-L180

But I found that the settings about `xxx_url` didn't work when i run the test.

After digging into the code, I think the root cause is that `service`(untrusted env) and `enclave-runtime`(trusted env)  don't share memory with each other. This means that **the variables initialized in the service are not actually initialized in the enclave-runtime** (`stf_task` run on the `enclave-runtime`).

### Solution
Initialize via `enclave-api`. like: 

https://github.com/litentry/litentry-parachain/blob/2371278fe9e014c4111865480b5f86d22511896a/tee-worker/service/src/main.rs#L483-L486

https://github.com/litentry/litentry-parachain/blob/2371278fe9e014c4111865480b5f86d22511896a/tee-worker/core-primitives/enclave-api/src/enclave_base.rs#L159-L168

BillyWooo:
initialize global variables inside enclave


tee-worker/service/src/main.rs
```diff
@@ -586,8 +521,57 @@ fn start_worker<E, T, D, InitializationHandler, WorkerModeProvider>(
...
+
+		stf_configs.push(twitter_official_url.into_bytes());
+		stf_configs.push(twitter_litentry_url.into_bytes());
+		stf_configs.push(twitter_auth_token.into_bytes());
+		stf_configs.push(discord_official_url.into_bytes());
+		stf_configs.push(discord_litentry_url.into_bytes());
+		stf_configs.push(discord_auth_token.into_bytes());
+		stf_configs.push(graphql_url.into_bytes());
+		stf_configs.push(graphql_auth_key.into_bytes());
+	}
```
ziming-zung, line 585: I think we can put these `xxx_url` into a struct, and then we encode it into bytes, decode it in enclave.
See: 
https://github.com/litentry/litentry-parachain/blob/2371278fe9e014c4111865480b5f86d22511896a/tee-worker/core-primitives/enclave-api/src/sidechain.rs#L38-L61

https://github.com/litentry/litentry-parachain/blob/0ecd109b82f755ab59170045228ec541c69e2c35/tee-worker/enclave-runtime/src/lib.rs#L339-L354

BillyWooo: Thanks. Good reference. Will update.

ziming-zung (approved):
Thanks. Look good

Kailai-Wang:
Thanks, feel free to merge it when CI passes

PR #1212, opened 2023-01-19
Run runtime upgrade simulation conditionally

### LINKED ISSUE #1167: Run runtime upgrade simulation conditionally
### Context
The current https://github.com/litentry/litentry-parachain/blob/dev/.github/workflows/simulate-runtime-upgrade.yml will be triggered whenever a release is published.

We should only run the simulation when:
- a new runtime wasm is included in the release
- the new runtime version is greater than the on-chain version

The change can be done either in GHA or in shell script.

---

Kailai-Wang:
resolves #1167 

Only do it when there's an existing wasm and the released runtime version > on-chain runtime version.

### DIFF
--- a/.github/workflows/simulate-runtime-upgrade.yml
+++ b/.github/workflows/simulate-runtime-upgrade.yml
@@ -11,6 +11,7 @@ on:
         required: true
 
 env:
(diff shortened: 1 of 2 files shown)


jingleizhang (approved):
LGTM

PR #1222, opened 2023-01-25
Expose the enclave's shielding key

### LINKED ISSUE #1221: Expose the TEE shielding key in parachain storage
### Context

As topic. It was somehow not addressed in #1155 - maybe we forgot it.
We need this very soon as it otherwise blocks IDH from switching to the real IMP pallet.

---

Kailai-Wang:
resolves #1221 

The shielding key (JSON serialised bytes) is added to the EnclaveRegistry upon enclave registration. The struct is kept the same as the direct RPC call response.


pallets/teerex/src/lib.rs
```diff
@@ -109,7 +109,7 @@ pub mod pallet {
 		T::Moment::saturated_from::<u64>(172_800_000) // default 48h
 	}
 	#[pallet::storage]
-	#[pallet::getter(fn heartbeat_timeout_storage)]
+	#[pallet::getter(fn heartbeat_timeout)]
```
BillyWooo, line 112: thanks. good catch.

BillyWooo (approved):
LGTM

PR #1232, opened 2023-01-28
Replace dockerize with healthcheck

### LINKED ISSUE #1068: Try to replace `dockerize` with `healthcheck` in tee-worker CI
### Context
https://docs.docker.com/compose/compose-file/#healthcheck

It's better to use more standard way for waiting handling

---

Kailai-Wang:
resolves #1068 

- node uses `nc` to check if the port is open
- workers use `curl` to check if it's initialised

The parachain image doesn't have shell, so we have to quote the test command with `CMD`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -46,6 +46,9 @@ ARG PROFILE
 
 COPY --from=builder /litentry/target/$PROFILE/litentry-collator /usr/local/bin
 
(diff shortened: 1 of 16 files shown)


jingleizhang:
LGTM. The `/is_initialized` should be useful to ts-test as well.

jingleizhang (approved):

Issue #1010, opened 2022-10-17
Evaluate the exposure of trusted RPC port in production env


Kailai-Wang:
In order for privileged users to query the storage via RPC, integritee has built-in support for RPC requests of trusted state. We should evaluate:

- is it safe to keep it open in production env?
- if so, which TrustedGetter should we have?

Kailai-Wang:
The pros and cons are obvious IMO:

++ you get the synchronous responses for your trustedGetter queries - very F/E friendly
--  the worker IP/port gets exposed, attackers might try ddos to bring the service down => maybe we'll have some non-validateer nodes

jingleizhang:
> the worker IP/port gets exposed, attackers might try ddos to bring the service down

We can put an AWS LoadBalance or even an Nginx reverse proxy in front of those workers to protect them.

jingleizhang:
Does TrustedGetter have any auth mechanism? For example, does the user still need to sign a transaction to call the rpc?

Kailai-Wang:
> Does TrustedGetter have any auth mechanism? For example, does the user still need to sign a transaction to call the rpc?

It does. It requires a signature from the user to prove its identity

jingleizhang:
Get it. I think making the trusted RPC port be exposed to F/E is acceptable.

Kailai-Wang:
This is done, we have a decision

PR #1247, opened 2023-01-31
Remove feature `mockserver`

### LINKED ISSUE #1073: More flexible challenge code generation in tests

### LINKED ISSUE #1205: Remove feature `mockserver`
As title.

When i am working on this issue, i think i should split the implementation into the following sub-issues:

- [x] #1073 
- [x] #1206 
- [x] #1207

ziming-zung:
This PR
resolves #1205 
resolves #1073


tee-worker/service/src/config.rs
```diff
@@ -61,7 +62,8 @@ pub struct Config {
 	pub running_mode: String,
```
BillyWooo, line 62: Here, these three commands all come from Litentry. They can be put together.

tee-worker/local-setup/worker-config-local.json
```diff
@@ -0,0 +1,10 @@
+{
```
BillyWooo: why add one more mode here? Is 'dev/staging/prod' not enough ?

tee-worker/service/src/config.rs
```diff
@@ -61,7 +62,8 @@ pub struct Config {
...
 
 	/// Litentry
-	pub disable_mock_server: bool,
+	pub mock_server: bool,
```
BillyWooo: Maybe 'mock_server_enable' or something similar would be more intuitive.

tee-worker/local-setup/worker-config-local.json
jingleizhang: `dev` equals `local`, so this config file should not be added.

zhizming-zhong: I misunderstand, I see tdf in the dev  configuration, so add one. I will delete it later

ziming-zung: Hmmmm, I think the `dev` configuration is different from the `local` configuration, and the `local` configuration is more used in ts-test. We can rename `local` to `mock`, which looks more intuitive

tee-worker/service/src/config.rs
ziming-zung: had adjusted.

tee-worker/service/src/config.rs
ziming-zung: Had adjusted.

tee-worker/local-setup/worker-config-local.json
BillyWooo: Hmmm, also make sense. Let's rename it to `mock`.

tee-worker/service/src/config.rs
BillyWooo: It would be better to follow the document format to add some description as comment. Because it will appear at the command line help menu.

BillyWooo (commented):
Overall it looks good to me. 
Please also update tee-worker/scripts/launch.sh accordingly.

tee-worker/litentry/core/challenge-code-cache/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+pub struct ChallengeCodeCache {
+	codes: RwLock<HashMap<Vec<u8>, Vec<u8>>>,
+	enable: RwLock<bool>,
+}
+
+impl ChallengeCodeCache {
+	pub fn insert_challenge_code(&self, identity: Identity, code: ChallengeCode) {
+		if self.enable.read().map_or(false, |r| *r) {
+			if let Ok(mut codes_lock) = self.codes.write() {
+				log::debug!("cahce challenge_code: {:?}, code:{:?}", identity, code);
```
Kailai-Wang: Typo: `cache`

tee-worker/litentry/core/challenge-code-cache/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	pub fn enable(&self) {
+		if let Ok(w) = self.enable.write() {
+			let mut enable_lock = w;
+			*enable_lock = true;
+		}
+	}
+
+	pub fn is_enable(&self) -> bool {
```
Kailai-Wang: normally should be `is_enabled`

Kailai-Wang (approved):
Looks good, thanks.

Actually I'm thinking this way of retrieving challenge code (mock-server thread -> ffi -> enclave -> global cache) might be an overkill, as we do have the trusted/untrusted RPC query to retrieve the state of the sidechain storage. Maybe it's enough to simply add another RPC method.

tee-worker/local-setup/worker-config-local.json
ziming-zung: Done, pls check.

ziming-zung:
> Looks good, thanks.
> 
> Actually I'm thinking this way of retrieving challenge code (mock-server thread -> ffi -> enclave -> global cache) might be an overkill, as we do have the trusted/untrusted RPC query to retrieve the state of the sidechain storage. Maybe it's enough to simply add another RPC method.

I also want to implement in this way at first. There is a point to consider here, whether there are any potential risks in exposing storage value via RPC. see: #1010

Kailai-Wang:
They will be used for development only, no RPC will be exposed in production - even if there're such requirements, we have to consider it prudently

Kailai-Wang:
Another question:

Can we enable the mock server by default in `launch.py`? It's used in dev anyway

zhizming-zhong:
> 

Done

ziming-zung:
> Another question:
> 
> Can we enable the mock server by default in `launch.py`? It's used in dev anyway

Done

ziming-zung:
> They will be used for development only, no RPC will be exposed in production - even if there're such requirements, we have to consider it prudently

Let's implement it in the next PR

Kailai-Wang:
All good, thanks. Feel free to merge it once CI passes.

PR #1248, opened 2023-01-31
Add scheduled(allowed) MRENCLAVE

### LINKED ISSUE #1177: Enclave update [1] - add `update_enclave` implementation
It's the preparation of the enclave update:
- we should hold a list of allowed enclaves to track the MRENCLAVE history
- add an extrinsic `update_enclave(new_mrenclave, scheduled_block_number)` to schedule the enclave update
- do we need to re-generate the ecc signing account? If so we need to fund it somewhere

Kailai-Wang:
resolves #1177 

This PR uses a `StorageMap` sidechain_blocknumber -> scheduled_mrenclave to keep track of allowed MRENCLAVEs.

The `skip-scheduled-enclave-check` feature is activated as default when building parachain images - not to break any existing setups/tests.

### DIFF
--- a/pallets/teerex/src/benchmarking.rs
+++ b/pallets/teerex/src/benchmarking.rs
@@ -38,6 +38,13 @@ fn ensure_not_skipping_ra_check() {
 	};
 }
 
+fn ensure_not_skipping_scheduled_enclave_check() {
+	#[cfg(not(test))]
+	if cfg!(feature = "skip-scheduled-enclave-check") {
+		panic!("Benchmark does not allow the `skip-scheduled-enclave-check` flag.");
+	};
+}
+
 fn generate_accounts<T: Config>(amount: u32) -> Vec<T::AccountId> {
 	(0..amount).map(|n| account("dummy name", n, n)).collect()
 }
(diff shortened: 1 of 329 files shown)


pallets/teerex/src/lib.rs
```diff
@@ -108,6 +110,23 @@ pub mod pallet {
...
+	// sidechain_block_number -> expected MREnclave
+	//
+	// about the first time enclave registration:
+	// prior to `register_enclave` this map needs to be populated with (0, expected-mrenclave),
+	// otherwise the registration will fail
+	//
+	// Theorectically we could always push the enclave in `register_enclave`, but the problem is
+	// anyone could try to register it as long as the enclave is remotely attested:
+	// see https://github.com/litentry/litentry-parachain/issues/1163
+	// so we need an "enclave whitelist" anyway
```
BillyWooo, line 125: Overall, it LGTM. Just some question: how should it (0, expected-mrenclave) be populated?

pallets/teerex/src/lib.rs
```diff
@@ -186,6 +205,15 @@ pub mod pallet {
...
 			}
 
+			// TODO: imagine this fn is not called for the first time (e.g. when worker restarts),
+			//       should we check the current sidechain_blocknumber >= registered
+			// sidechain_blocknumber?
+			#[cfg(not(feature = "skip-scheduled-enclave-check"))]
+			ensure!(
+				ScheduledEnclave::<T>::iter_values().any(|m| m == enclave.mr_enclave),
+				Error::<T>::EnclaveNotInSchedule
+			);
```
DogLi, line 215: how about skip if the ScheduledEnclave is empty?
Actually, we need to set the current MrEnclave, in order to let the tee-worker know whether it needs to upgrade or not.

pallets/teerex/src/benchmarking.rs
```diff
@@ -65,9 +72,21 @@ benchmarks! {
...
+		// we need different parameters, unfortunately - since the way to calculate
+		// MRENCLAVE differs depending on if `skip-ias-check` feature is present.
+		Teerex::<T>::update_scheduled_enclave(
+			RawOrigin::Root.into(),
+			0u32,
+			#[cfg(feature = "skip-ias-check")]
+			<[u8; 32]>::decode(&mut TEST4_SETUP.cert).unwrap_or_default(),
+			#[cfg(not(feature = "skip-ias-check"))]
+			TEST4_MRENCLAVE,
+		).unwrap();
```
DogLi: using `MREnclave` instead of `[u8; 32]` is better.

pallets/teerex/src/lib.rs
Kailai-Wang: As discussed yesterday, we do need to set the allowed MRENCLAVE, otherwise the registration will fail.

DogLi:
LGTM

BillyWooo (approved):
LGTM

pallets/teerex/src/lib.rs
Kailai-Wang: Via polkadot-js UI :)

PR #1262, opened 2023-02-02
Fixed a bug caused by the issue-1137 test flow

0xverin:
optimized #1137 
Fixed a bug where running test-identity twice would timeout.


ziming-zung (approved):
Looks good.

PR #1276, opened 2023-02-05
Use events in `some_error`

### LINKED ISSUE #1275: Replace `Error` with `event` for `some_error`
### Context
Pre-task of  #1272 - so far `some_error` emits the pallet errors, e.g.: https://github.com/litentry/litentry-parachain/blob/ce0ff808b44c204149da4c395734e0dc10fbabe9/pallets/identity-management/src/lib.rs#L277-L299

But `Error` can't carry along extra parameters, therefore we'd like to use Events here - hope it also makes the F/E's life a bit easier (but not too much :D) because you don't have to look up Errors according to the module and indexes.

---

Kailai-Wang:
resolves #1275


ziming-zung (approved):
Looks good. Thanks.

PR #1280, opened 2023-02-06
Improve running mode configs

### LINKED ISSUE #1269: Improve the usage of the running mode config file
### Context
After discussion, we'd like to:

- merge different running modes (`mock`, `dev`, `staging`, `prod`) into one single config file: `running-mode-config.json`
- introduce an env for each config field (e.g. TWITTER_OFFICIAL_URL), use it when it's present
- if no env is present, fallback to the values in `running-mode-config.json`
- (optional) add an extra arg to use a specified running-mode config file 

---

ziming-zung:
This PR resolves #1269


tee-worker/service/src/main.rs
```diff
@@ -872,3 +828,47 @@ fn check_we_are_primary_validateer(
 		node_api.enclave_count(Some(*register_enclave_xt_header.parent_hash()))?;
 	Ok(enclave_count_of_previous_block == 0)
 }
+
+fn data_provider(config: &Config) -> DataProvidersStatic {
+	let build_in_mode = vec!["dev", "mock", "prod", "staging"];
```
Kailai-Wang: maybe you mean `built_in_mode`?

tee-worker/service/src/main.rs
```diff
@@ -872,3 +828,47 @@ fn check_we_are_primary_validateer(
...
+	let build_in_config: Value =
+		serde_json::from_slice(include_bytes!("running-mode-config.json")).unwrap();
+
+	let mut data_provider_config = if build_in_mode.contains(&config.running_mode.as_str()) {
```
Kailai-Wang: So if I understand it correctly, `running-mode` should be either one of the built-in modes, or a file path?

tee-worker/service/src/main.rs
```diff
@@ -872,3 +828,47 @@ fn check_we_are_primary_validateer(
...
+	} else {
+		let file_path = config.running_mode.as_str();
+		let mut file = File::open(file_path)
+			.map_err(|e| format!("{:?}, file:{}", e, file_path))
+			.unwrap();
+		let mut data = String::new();
+		file.read_to_string(&mut data).unwrap();
+		serde_json::from_str::<DataProvidersStatic>(data.as_str()).unwrap()
+	};
+	if let Ok(v) = env::var("TWITTER_OFFICIAL_URL") {
```
Kailai-Wang, line 849: A small question is if all `env`s are set but the running-mode is incorrectly configured, should we still allow this? Current logic is we would panic out before checking the env.

Not a big deal though

Kailai-Wang (approved):
Looks good, thanks, just a few small questions from my side

tee-worker/service/src/main.rs
ziming-zung: Yes, it shoud be `built_in_mode` :rofl:

tee-worker/service/src/main.rs
ziming-zung: Yes

tee-worker/service/src/main.rs
ziming-zung: I think `built_in_config` or config from file_path should always be correct. :sweat_smile:

PR #1281, opened 2023-02-06
Propagate the indirect handling errors to parachain

### LINKED ISSUE #1229: Don't fail silently if decryption fails in `parentchain-indirect-calls-executor`
### Context

IDH tries to use a fake TEE shielding key to encrypt the indirect calls, they expected to get an error but nothing happened.

I think it's because the indirect-calls-executor silently ignores the errors: when `decode_and_execute` fails to decode, we simply go to the next executor, and we don't send out any errors even when all executors are exhausted:

https://github.com/litentry/litentry-parachain/blob/abfa4c06c7a5c3f0437d11745c9b4dd9ac1bc11e/tee-worker/core/parentchain/indirect-calls-executor/src/executor/mod.rs#L136-L138

Basically, we have to differentiate between
- "decoding errors", where the call can't be decoded into supported call type, and
- "execution errors" due to e.g. decryption failure, or execution failure

---

Kailai-Wang:
resolves #1229 

Acutally I'm not sure if it's the best way to map/populate the errors, suggestions are welcome.

We still need ts-tests for this @Verin1005 
manual tests look good:
image


ziming-zung (approved):
Looks good. :thumbsup:

Verin1005:
I have some questions  here. Under what circumstances did the error event occur(CreateIdentityHandlingFailed)? I used the wrong identity and could obtain it, I used the wrong errorAseKey but could not obtain it.CreateIdentityHandlingFailed just handling errors identity information?

PR #1283, opened 2023-02-06
fix assertions field type to Vec

zTgx:
as the title.


tee-worker/litentry/core/credentials/src/lib.rs
```diff
@@ -358,8 +362,7 @@ impl Credential {
 				let timestamp = AssertionLogic::new_item("$timestamp", Op::GreaterEq, &from_date);
 
 				let assertion = AssertionLogic::new_and().add_item(lit_amounts).add_item(timestamp);
-
-				credential.credential_subject.assertions = assertion;
+				credential.credential_subject.set_assertion(assertion);
```
Kailai-Wang: I think `push` is better? Maybe we don't need `set_assertion` as `push` is native for `Vec`

Kailai-Wang (commented):
Thanks, a small thing:

tee-worker/litentry/core/credentials/src/lib.rs
zTgx: Yes, it's redundant. fields `assertions` and `values`  is a one-to-one relationship. will refactor `set_assertions` and `set_values` both.

jingleizhang (approved):
Looks good!

PR #1286, opened 2023-02-07
Query tee strorage via RPC

### LINKED ISSUE #1253: Query the storage via RPC
### Context
Currently, there is a RPC in tee-worker to query the storage value, which is called `TrustedGetter`. But it can only query specific storage value ,  and we need a general RPC to query all storage value, which should only work in a specific stage. Now it seems that there are 3 benefits:

* ts-test
* the implementation of getting challenge-code more lightweight
* state migration?

ziming-zung:
This PR resolves #1253


jingleizhang:
How to test this feature in cmd, could you offer an example?

ziming-zung:
> How to test this feature in cmd, could you offer an example?

First, you should get the mrenclave/shard from cli:
`
./integritee-service mrenclave
`

And then pass the mrenclave to subcommand `query-storage`
`
./integritee-cli trusted --mrenclave 8SzhRktGX4p9ETHogYLzCiGn4jJy5J9hHVBb6w6izRrP query-storage Sudo Key
`

Actually, I want to write the command in one line, like:
`
./integritee-service mrenclave | xargs  ./integritee-cli trusted --mrenclave $1 query-storage Sudo Key
`
But it seems there is a problem. :thinking:

Kailai-Wang:
Yes an example as code comments would be beneficial.

Theoretically if this works, we should be able to replace the `balance` and `nonce` commands with the storage query? This is a good chance to test the functionality and usability too

tee-worker/cli/src/trusted_base_cli/commands/query_storage.rs
```diff
@@ -0,0 +1,73 @@
...
+	module: String,
+
+	storage: String,
+}
+impl QueryStorageCommand {
+	pub(crate) fn run(&self, cli: &Cli, trusted_args: &TrustedCli) {
+		let direct_api = get_worker_api_direct(cli);
+
+		let key = storage_value_key(self.module.as_str(), self.storage.as_str());
+		let key = format!("0x{}", hex::encode(key));
```
DogLi: Taking the `format` into function would be better.

tee-worker/cli/src/trusted_base_cli/commands/query_storage.rs
```diff
@@ -0,0 +1,73 @@
...
+						},
+						DirectRequestStatus::TrustedOperationStatus(status) => {
+							warn!("request status is: {:?}", status);
+							if let Ok(value) = Hash::decode(&mut return_value.value.as_slice()) {
+								warn!("Trusted call {:?} is {:?}", value, status);
+							}
+						},
+					}
+				};
+			},
```
DogLi: I think the response value should be returned as `Result<Vec<u8>>` instead of just printing a `String` format log. The return value can always be a `Vec<u8>`, but decoding to a `String` may fail.

tee-worker/cli/src/trusted_base_cli/commands/query_storage.rs
```diff
@@ -0,0 +1,73 @@
...
+use itp_utils::FromHexPrefixed;
+use log::{debug, error, warn};
+use my_node_runtime::Hash;
+
+use std::sync::mpsc::channel;
+
+#[derive(Parser)]
+pub struct QueryStorageCommand {
+	/// Storage Key
+	module: String,
```
DogLi: is the `/// Storage Key` comment here right?

tee-worker/cli/src/trusted_base_cli/commands/query_storage.rs
```diff
@@ -0,0 +1,73 @@
...
+
+	storage: String,
+}
+impl QueryStorageCommand {
+	pub(crate) fn run(&self, cli: &Cli, trusted_args: &TrustedCli) {
```
DogLi: maybe rename the function name to `query`?

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -143,6 +160,52 @@ where
...
 	});
 
+	if let Some(state) = state {
+		// state_getStorage
+		let state_get_storage = "state_getStorage";
+		io.add_sync_method(state_get_storage, move |params: Params| {
+			// jsonrpc_core::serde::Deserialize<'de>` is not implemented for `H256`
+			// can not parse ShardIdentifier
+			// TODO: hanld error
+			match params.parse::<Vec<String>>() {
```
DogLi: because we send `vec![trusted_args.mrenclave.clone(), key]`, here use `[String; 2]` instread of `Vec<String>` can be better. so we can parse the key and shard_id conveniently.

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -143,6 +160,52 @@ where
 		Ok(json!(json_value))
 	});
 
+	if let Some(state) = state {
+		// state_getStorage
+		let state_get_storage = "state_getStorage";
+		io.add_sync_method(state_get_storage, move |params: Params| {
+			// jsonrpc_core::serde::Deserialize<'de>` is not implemented for `H256`
+			// can not parse ShardIdentifier
+			// TODO: hanld error
```
DogLi: should fix the `TODO`, just create another function `invalid_params` to return the error response.

tee-worker/cli/src/trusted_base_cli/commands/query_storage.rs
BillyWooo: Looks like `run` is a name for consistency.

BillyWooo:
> > How to test this feature in cmd, could you offer an example?
> 
> First, you should get the mrenclave/shard from cli: `./integritee-service mrenclave`
> 
> And then pass the mrenclave to subcommand `query-storage` `./integritee-cli trusted --mrenclave 8SzhRktGX4p9ETHogYLzCiGn4jJy5J9hHVBb6w6izRrP query-storage Sudo Key`
> 
> Actually, I want to write the command in one line, like: `./integritee-service mrenclave | xargs ./integritee-cli trusted --mrenclave $1 query-storage Sudo Key` But it seems there is a problem. 🤔

Is that practical to have a cli test for it? Or just extend some existed test?

Kailai-Wang:
I turned this PR into draft to not trigger CI, please feel free to make it "ready for review" when you update it @zhizming-zhong

ziming-zung:
> > > How to test this feature in cmd, could you offer an example?
> > 
> > 
> > First, you should get the mrenclave/shard from cli: `./integritee-service mrenclave`
> > And then pass the mrenclave to subcommand `query-storage` `./integritee-cli trusted --mrenclave 8SzhRktGX4p9ETHogYLzCiGn4jJy5J9hHVBb6w6izRrP query-storage Sudo Key`
> > Actually, I want to write the command in one line, like: `./integritee-service mrenclave | xargs ./integritee-cli trusted --mrenclave $1 query-storage Sudo Key` But it seems there is a problem. 🤔
> 
> Is that practical to have a cli test for it? Or just extend some existed test?

I prefer to extend/update some existed test

tee-worker/cli/demo_sidechain.sh
```diff
@@ -110,7 +110,9 @@ echo ""
 ${CLIENTWORKER1} trusted --mrenclave ${MRENCLAVE} --direct set-balance ${ICGACCOUNTBOB} 0
 
 echo "Get balance of Alice's incognito account (on worker 1)"
-${CLIENTWORKER1} trusted --mrenclave ${MRENCLAVE} balance ${ICGACCOUNTALICE}
+# ${CLIENTWORKER1} trusted --mrenclave ${MRENCLAVE} balance ${ICGACCOUNTALICE}
+# ICGACCOUNTALICE's public key is 0x50503350955afe8a107d6f115dc253eb5d75a3fe37a90b373db26cc12e3c6661
+${CLIENTWORKER1} trusted --mrenclave ${MRENCLAVE} get-storage System Account 0x50503350955afe8a107d6f115dc253eb5d75a3fe37a90b373db26cc12e3c6661
```
Kailai-Wang: Is it possible to replace all the getter commands (`balance`, `nonce`) with `get-storage`? Just comment out the original command for reference

tee-worker/cli/src/trusted_base_cli/commands/get_storage.rs
```diff
@@ -0,0 +1,154 @@
...
+///    Map Storage: ./integritee-cli trusted --mrenclave $mrenclave get-storage System Account 0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d
+#[derive(Parser)]
+pub struct GetStorageCommand {
+	/// Pallet Name
+	pallet_name: String,
+
+	/// Storage Name
+	storage_name: String,
+
+	/// Stroage Key
```
Kailai-Wang, line 30: Do you have an example of multiple keys? Like `DoubleMapStorage`?

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -143,6 +154,51 @@ where
 		Ok(json!(json_value))
 	});
 
+	if let Some(state) = state {
```
Kailai-Wang: Maybe it's better to have the `if let` inside the handler? I.e. we should always have this rpc method available, just return empty/error when no state is present by the time of requesting

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -143,6 +154,51 @@ where
...
+						Err(msg) => {
+							let error_msg = format!("decode shard failure due to: {}", msg);
+							return Ok(json!(compute_hex_encoded_return_error(error_msg.as_str())))
+						},
+					};
+					match state.load_cloned(&shard) {
+						Ok((state, _hash)) => {
+							// Get storage by key hash
+							let value = state.get(key_hash.as_slice()).cloned().unwrap_or_default();
+							warn!("query storage value:{:?}", &value);
```
Kailai-Wang: `debug!` here :)

tee-worker/cli/src/trusted_base_cli/commands/get_storage.rs
```diff
@@ -0,0 +1,154 @@
...
+		let val =
+			DecodedValueThunk::decode_with_metadata(&mut bytes, return_ty_id, &metadata).unwrap();
+
+		Some(val.to_value().unwrap())
+	} else {
+		None
+	}
+}
+
+fn send_request(
```
Kailai-Wang: I think this is not so well named, maybe `get_storage_value_internal` or `send_get_storage_request`

tee-worker/cli/src/trusted_base_cli/commands/get_storage.rs
```diff
@@ -0,0 +1,154 @@
...
+			})
+			.collect();
+		if let Some(v) = get_storage_value(
+			direct_api,
+			mrenclave,
+			self.pallet_name.as_str(),
+			self.storage_name.as_str(),
+			storage_entry_keys,
+		) {
+			println!("{}", v);
```
Kailai-Wang, line 44: I'm wondering what's the result of the non-number value? Hex-encoded strings? E.g. the identity in the IMT pallet

Kailai-Wang (commented):
Generally looks good, thanks! I only have several small questions. Btw I doubt if the upstream would merge it as it depends on `subxt`, which does more or less similar things to `substrate-api-client` 😂

tee-worker/cli/src/trusted_base_cli/commands/get_storage.rs
ziming-zung: It shoule be a JSON-String. See the shell script.

tee-worker/cli/demo_sidechain.sh
ziming-zung: Done, pls check.

ziming-zung:
> Generally looks good, thanks! I only have several small questions. Btw I doubt if the upstream would merge it as it depends on `subxt`, which does more or less similar things to `substrate-api-client` 😂

Finally, i remove `subxt` from dependencies, and implement it in a new way. :dizzy_face:. It's harder than i thought 😮‍💨

Kailai-Wang (approved):
Thanks! Good job removing `subxt` dependency 👍 

I updated the script a bit to fix CI

Issue #1124, opened 2022-12-27
Stop both two worker then resume(ts-test)


ziming-zung:
### Context
related to #999 
When starting 2 workers and waiting for them to produce blocks, then kill the worker process. Wait for some time (6 sec), try to resume worker0 and worker1. It seems worker0 or worker1 can not be resumed. 

Ts-Test code:

```typescript 
step("Two workers(stop 2 workers --> resume 2 workers)", async function () {
        // launch worker0
        let {
            process: worker0,
            shard: shard,
        } = await launchWorker("worker0", binary_dir, worker0_dir, commands.worker0.commands.first_launch, true)
        global_shard = shard
        let worker0_conn = await initWorkerConnection(`ws://localhost:${commands.worker0.untrusted_ws_port}`)
        const {
            number: old_block_number,
            hash: old_block_hash
        } = await waitWorkerProducingBlock(worker0_conn, global_shard, 4)


        // 2 workers were actually launched
        // first launch worker1
        let {
            process: worker1
        } = await launchWorker("worker1", binary_dir, worker1_dir, commands.worker1.commands.first_launch, true)
        let worker1_conn = await initWorkerConnection(`ws://localhost:${commands.worker1.untrusted_ws_port}`)
        await waitWorkerProducingBlock(worker1_conn, global_shard, 3)
        console.log("=========== worker0/worker1 launch successfully ==================")

        await sleep(5)
        await killWorker(worker0)
        await sleep(5)
        await killWorker(worker1)
        console.log("=========== worker0/worker1 stopped ==================")


        // resume worker0
        let {
            process: r_worker0
        } = await launchWorker("worker0", binary_dir, worker0_dir, commands.worker0.commands.resume, false)
        await worker0_conn.open()
        await sleep(6)

        // resume worker1
        let {
            process: r_worker1
        } = await launchWorker("worker1", binary_dir, worker1_dir, commands.worker1.commands.resume, false)
        await worker1_conn.open()

        await waitWorkerProducingBlock(worker0_conn, global_shard, 5)
        await waitWorkerProducingBlock(worker1_conn, global_shard, 5)


        const block_at_w0 = await getBlockHash(worker0_conn, old_block_number, global_shard)
        const block_at_w1 = await getBlockHash(worker1_conn, old_block_number, global_shard)
        // assert.equal(block_at_w0.result!.hash, old_block_hash, "worker0 block_hash is not same")
[...]

PR #1287, opened 2023-02-07
Function `search_peers` should not exit on error

ziming-zung:
As mentionded in https://github.com/litentry/litentry-parachain/issues/1124#issuecomment-1367690264 or question box, there is a bug in function `search_peers`. We can fix it simply: as title.


Kailai-Wang (approved):
Looks good, thanks. While syncing `dev` branch, could you please also add inline comment about the background? E.g. paste the link to the problem (issue comment).

PR #1288, opened 2023-02-07
Vc assertion ts-test

0xverin:
1.assertion ts-test add A2 A3 A4 A7 A8 A10 
2.add vcRegistry


tee-worker/cli/ts_test_vc.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -o pipefail
+
+cd /ts-tests
+yarn install
+yarn run test-vc:staging
```
Kailai-Wang: Why not merge it into `ts_test.sh`? Do we want to parallelize it? If so may be can still use the same `ts_test.sh` but different args.

Btw `ts_test_vc.sh` won't run because we miss the matching yml entry

tee-worker/ts-tests/indirect_calls.ts
```diff
@@ -59,14 +57,18 @@ export async function createIdentity(
...
 
-    const tx =context.substrate.tx.identityManagement
-        .createIdentity(context.shard, signer.address, `0x${ciphertext}`, null)
-    
+    const tx = context.substrate.tx.identityManagement.createIdentity(
+        context.shard,
+        signer.address,
+        `0x${ciphertext}`,
+        null
+    );
```
Kailai-Wang, line 65: These format changes are strange -- did it follow `.prettierrc` or use other formatters?

tee-worker/ts-tests/vc.test.ts
```diff
@@ -25,9 +23,13 @@ describeLitentry('VC test', async (context) => {
...
-            A1: assertion.A1,
-        });
-        assert(vc !== u8aToHex());
+        for (const key in assertion) {
+            const eventData = await requestVC(context, context.defaultSigner[0], aesKey, true, context.shard, {
+                [key]: assertion[key as keyof Assertion],
+            });
+            assert(eventData![0] == u8aToHex(context.defaultSigner[0].addressRaw) && eventData![1] && eventData![2]);
+            const registry = (await context.substrate.query.vcManagement.vcRegistry(eventData![1])) as any;
+            assert.equal(registry.toHuman()!['status'], 'Active');
```
Kailai-Wang, line 29: I think there're much more contents to assert 😂 , including the general asserts (like JSON format validation, metadata validation,  signature validation, hash validation) and specific asserts

but maybe you want to do it in a separate PR
cc @zhizming-zhong

tee-worker/ts-tests/vc.test.ts
```diff
@@ -4,19 +4,17 @@ import { requestVC, setUserShieldingKey } from './indirect_calls';
...
     A4: [10, 'A4'],
-    A5: ['A5', 'A5'],
-    A6: 'A6',
+    // A5: ['A5', 'A5'],
+    // A6: 'A6',
     A7: [10, 10],
-    A8: [10],
-    A9: 'A9',
+    A8: 'A8',
+    // A9: 'A9',
```
Kailai-Wang: Any reasons to comment out these entries? Generally we shouldn't have commented code - either leave them or remove them - git has history.

Kailai-Wang (changes requested):
Have left some questions - please feel free to review on it too @zhizming-zhong

tee-worker/cli/ts_test_vc.sh
0xverin: I was a little unfamiliar with CI configuration, and I thought that adding a new file would allow both tasks to work together, which would reduce the wait time. I'll fix it right away.

tee-worker/ts-tests/indirect_calls.ts
0xverin: I use the format configured locally.

tee-worker/ts-tests/vc.test.ts
0xverin: we do here temporarily is to confirm the effectiveness of vc, and other functions such as whether vc is correct will be added later. At this point,  just running the requestVC process.

Kailai-Wang (approved):
Thanks, we can merge it as long as CI passes

jingleizhang (approved):
LGTM, VC assertions and on-chain status checking are covered. 

Currently @zTgx is working on VC verifications in ts-test, that includes:

- enclave attestation related works on both Parachain and VC
- Issuer signature validation

Issue #815, opened 2022-09-06
Build Ruleset 1: Link and verify a Web2.0 and Web3 account which will generate a Verifiable Credential (VC)


StamfordDigital:
Build **Ruleset 1: Link and verify a Web2.0 and Web3 account** which will generate a **Verifiable Credential (VC)** if the following validates:
    - Check the IDHub address has linked at least one Web2.0 and at least one Web3 account
    - Check that the Web2.0 and Web3 are both verified
    - If both checks pass a VC should be generated and returned to the F/E.

PR #1289, opened 2023-02-07
update VC-A1

zTgx:
Update VC-A1 assertions.

According to [Build Ruleset 1](the https://github.com/litentry/litentry-parachain/issues/815), 
* Check the IDHub address has linked at least one Web2.0 and at least one Web3 account

no need to show the real web2/web3 account numbers.


jingleizhang (approved):
Looks good, thanks!

PR #1291, opened 2023-02-08
add disableVc&revokeVc

0xverin:
as the title
The test flow:disableVc,check index and registry's status(should be Disabled)=>revokeVc,check index and registry(should be null)


0xverin:
If I go to disableVc after revokeVc and no error is returned, I think event should be added here to tell the client what specific error occurs.
cc @ @zTgx @jingleizhang

zTgx:
> If I go to disableVc after revokeVc and no error is returned, I think event should be added here to tell the client what specific error occurs. cc @ @zTgx @jingleizhang

I think It's necessary to give users a response when they request a call. Maybe we can add an event here.

Kailai-Wang (approved):
Looks good thanks. Yes please fix the CI :) 

btw if you split it into two scripts, it's better to rename the old one to something like `identity_test`, because `vc_test` is also ts_test

Kailai-Wang:
Btw `.github/workflows/tee-worker-ci.yml` seems reformatted :( there are a lot of indention changes @Verin1005

0xverin:
> Btw `.github/workflows/tee-worker-ci.yml` seems reformatted :( there are a lot of indention changes @Verin1005
I set my local prettier,it automatically format when I open the file.

PR #1293, opened 2023-02-08
Enclave update [2] - adjust key provisioning(fix #1178)

### LINKED ISSUE #1178: Enclave update [2] - adjust key provisioning
Currently encryption key will only be shared with other enclaves with the same MRENCLAVE -> we want to loosen the restriction.

The shielding key and state key stay unchanged during the update. We allow the workers to provision the keys to any other workers that run an enclave in the MRENCLAVE list, so it’s possible that worker0 with an older enclave sends the keys to worker1 with a newer enclave, as long as both enclaves are in the whitelist.

DogLi:
Before syncing the state key, check the current MrEnclave matches the scheduled enclaves.

### DIFF
--- a/pallets/teerex/src/lib.rs
+++ b/pallets/teerex/src/lib.rs
@@ -403,8 +403,6 @@ pub mod pallet {
 		EmptyEnclaveRegistry,
 		/// Can not found the desired scheduled enclave.
 		ScheduledEnclaveNotExist,
-		/// Enclave already stored in scheduled list
-		ScheduledEnclaveAlreadyExist,
 		/// Enclave not in the scheduled list, therefore unexpected.
 		EnclaveNotInSchedule,
 	}
--- a/pallets/teerex/src/lib.rs
+++ b/pallets/teerex/src/lib.rs
@@ -377,7 +377,7 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			T::EnclaveAdminOrigin::ensure_origin(origin)?;
 			ensure!(
-				!ScheduledEnclaveIndex::<T>::contains_key(&mr_enclave),
+				!ScheduledEnclaveIndex::<T>::contains_key(mr_enclave),
 				Error::<T>::ScheduledEnclaveAlreadyExist
 			);
 			ScheduledEnclave::<T>::insert(sidechain_block_number, mr_enclave);
--- a/pallets/teerex/src/lib.rs
+++ b/pallets/teerex/src/lib.rs
@@ -127,6 +127,21 @@ pub mod pallet {
 	#[pallet::getter(fn scheduled_enclave)]
 	pub type ScheduledEnclave<T: Config> = StorageMap<_, Blake2_128Concat, u32, MREnclave>;
 
+	/// store the last MrEnclave Index, the biggest one
+	/// Watch out: we start indexing with 1 instead of zero in order to
+	/// avoid ambiguity between Null and 0.
+	#[pallet::storage]
+	#[pallet::getter(fn scheduled_enclave_count)]
+	pub type ScheduledEnclaveCount<T: Config> = StorageValue<_, u64, ValueQuery>;
+
+	/// storage map of the enclave_index <-> MrEnclave, starts from 1
+	/// Watch out: we start indexing with 1 instead of zero in order to
+	/// avoid ambiguity between Null and 0.
+	#[pallet::storage]
+	#[pallet::getter(fn scheduled_enclave_index)]
+	pub type ScheduledEnclaveRegistry<T: Config> =
+		StorageMap<_, Blake2_128Concat, u64, MREnclave, ValueQuery>;
+
 	#[pallet::storage]
 	#[pallet::getter(fn heartbeat_timeout)]
 	pub type HeartbeatTimeout<T: Config> =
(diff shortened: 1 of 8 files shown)


pallets/teerex/src/lib.rs
```diff
@@ -355,8 +376,18 @@ pub mod pallet {
 			mr_enclave: MREnclave,
 		) -> DispatchResultWithPostInfo {
 			T::EnclaveAdminOrigin::ensure_origin(origin)?;
+			ensure!(
+				!ScheduledEnclaveIndex::<T>::contains_key(mr_enclave),
+				Error::<T>::ScheduledEnclaveAlreadyExist
+			);
```
Kailai-Wang: I also think of such before, but in some corner cases we could (theoretically) have duplicate MRENCLAVEs, like if a newer enclave has problems and we want to revert to the old one: MR1 => MR2 => MR1, with different sidechain block numbers. It's a rare case but possible

pallets/teerex/src/lib.rs
```diff
@@ -371,7 +402,12 @@ pub mod pallet {
 				ScheduledEnclave::<T>::contains_key(sidechain_block_number),
 				Error::<T>::ScheduledEnclaveNotExist
 			);
+			let mr_enclave = <ScheduledEnclave<T>>::get(sidechain_block_number)
+				.ok_or(Error::<T>::ScheduledEnclaveNotExist)?;
```
Kailai-Wang: With this I believe you can actually remove the above `ensure!`

pallets/teerex/src/lib.rs
```diff
@@ -355,8 +376,18 @@ pub mod pallet {
...
+				Error::<T>::ScheduledEnclaveAlreadyExist
+			);
 			ScheduledEnclave::<T>::insert(sidechain_block_number, mr_enclave);
 			Self::deposit_event(Event::UpdatedScheduledEnclave(sidechain_block_number, mr_enclave));
+			let enclave_idx = Self::scheduled_enclave_count()
+				.checked_add(1)
+				.ok_or("[Teerex]: Overflow adding new MrEnclave to registry")?;
+			<ScheduledEnclaveRegistry<T>>::insert(enclave_idx, mr_enclave);
+			<ScheduledEnclaveIndex<T>>::insert(mr_enclave, enclave_idx);
+			<ScheduledEnclaveCount<T>>::put(enclave_idx);
```
Kailai-Wang: As the enclave_idx is always increasing, we might have a problem when we want to override an existing schedule, like `update_scheduled_enclave(10, MR1)`, then `update_scheduled_enclave(10, MR2)`. It's not so uncommon to do so

pallets/teerex/src/lib.rs
```diff
@@ -371,7 +402,12 @@ pub mod pallet {
 				ScheduledEnclave::<T>::contains_key(sidechain_block_number),
 				Error::<T>::ScheduledEnclaveNotExist
 			);
+			let mr_enclave = <ScheduledEnclave<T>>::get(sidechain_block_number)
+				.ok_or(Error::<T>::ScheduledEnclaveNotExist)?;
+			let removed_index = <ScheduledEnclaveIndex<T>>::get(mr_enclave);
+			// remove
 			ScheduledEnclave::<T>::remove(sidechain_block_number);
+			ScheduledEnclaveRegistry::<T>::remove(removed_index);
```
Kailai-Wang: Shall we update `ScheduledEnclaveIndex` and `ScheduledEnclaveCount` too?

Kailai-Wang (commented):
Thanks for the PR!

Maybe I miss something, but I don't quite get the benefits of having those additional storages - why can't we simply check if the worker-enclave is contained in the `ScheduledEnclave` values? 

We could have a lenient check first, or if we want it more strictly, we can check additionally that current sidechain block numbers, so that we can provide keys to "future" MREnclaves (= whose registered sidechain_block_number >= current sidechain_block_number)

Apart from this I've left some questions about the code logic

pallets/teerex/src/lib.rs
DogLi: sure, I'll revert it.

tee-worker/service/src/sync_state.rs
```diff
@@ -86,10 +86,17 @@ async fn get_enclave_url_of_first_registered<NodeApi: PalletTeerexApi, EnclaveAp
 	enclave_api: &EnclaveApi,
 ) -> Result<String> {
 	let self_mr_enclave = enclave_api.get_mrenclave()?;
+	let all_scheduled_mr_enclaves = node_api.all_schedule_mr_enclaves(None)?;
+	if all_scheduled_mr_enclaves.is_empty() {
+		return Err(Error::NoParachainScheduledMrEnclaves)
+	}
```
Kailai-Wang, line 92: `contains` should be enough I guess?

tee-worker/service/src/tests/mock.rs
```diff
@@ -42,13 +43,20 @@ impl PalletTeerexApi for TestNodeApi {
 		Ok(enclaves())
 	}
 
+	fn all_schedule_mr_enclaves(&self, _at_block: Option<Hash>) -> ApiResult<Vec<MrEnclave>> {
+		let enclaves = enclaves();
+		let mr_enclaves = enclaves.into_iter().map(|e| e.mr_enclave).collect();
```
Kailai-Wang: maybe we need to de-duplicate it here, there can be repeated MRENCLAVEs.

Kailai-Wang (approved):
Thanks, I approved it - I only have two small questions

tee-worker/service/src/sync_state.rs
DogLi: yes, but if there is no scheduled enclave in the parachain, the error should be different from the `InvalidEnclave` I think.

PR #1294, opened 2023-02-08
update VC-A4/7/10

zTgx:
* update VC-A4/7/10 payload
* update template assertions in jsonfile
* refactor generate_unsigned_credential
* fix A4 typos


BillyWooo (commented):
In general, the code looks good. I got one question: why introduced `pub const ASSERTION_FROM_DATE: [&str; 7]` , instead of timestamp parameter?

jingleizhang (approved):
LGTM, thanks!

jingleizhang:
> In general, the code looks good. I got one question: why introduced `pub const ASSERTION_FROM_DATE: [&str; 7]` , instead of timestamp parameter?

They need a yearly range of `from_date` and `to_date` to calculate the score: https://www.notion.so/web3builders/Score-calculation-7200acc657b2461b841278d72fd6f584

PR #1297, opened 2023-02-08
A quick fix of CI

Kailai-Wang:
Includes:

- take skipped job in `needs` into consideration when defining dependencies (see comment in `ci.yml`)
- prefer runtime volume mounts over build time `COPY`, so that modifying the scripts doesn't need to rebuild tee image


BillyWooo (approved):
LGTM.

PR #1299, opened 2023-02-09
fix minor ci improve

BillyWooo:
as title

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,11 +172,6 @@ jobs:
     needs:
       - fmt
       - check-file-change
-    if: >
-      needs.check-file-change.outputs.parachain_src == 'true' ||
-      needs.check-file-change.outputs.parachain_test == 'true' ||
-      needs.check-file-change.outputs.tee_src == 'true' ||
-      needs.check-file-change.outputs.tee_test == 'true'
     steps:
       - uses: actions/checkout@v3
 
@@ -211,10 +206,6 @@ jobs:
     needs:
       - fmt
       - check-file-change
-    if: >
-      needs.check-file-change.outputs.parachain_src == 'true' ||
-      needs.check-file-change.outputs.tee_src == 'true' ||
-      needs.check-file-change.outputs.tee_test == 'true'
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,6 @@ jobs:
         working-directory: ./tee-worker/enclave-runtime
         run: |
           cargo fmt --all -- --check
-          # TODO: no taplo check in enclave-runtime
 
       - name: Fail early
         if: failure()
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -17,9 +17,34 @@ concurrency:
...
+          filters: .github/file-filter.yml
+          list-files: shell
+
   test:
     name: Coverage Report
     runs-on: [ubuntu-latest]
+    needs:
+      - check-file-change
+    if: >
+      needs.check-file-change.outputs.parachain_src == 'true' ||
```
Kailai-Wang: I think it only covers the parachain pallets, maybe checking parachain part is enough

.github/workflows/ci.yml
```diff
@@ -207,6 +211,10 @@ jobs:
     needs:
       - fmt
       - check-file-change
+    if: >
+      needs.check-file-change.outputs.parachain_src == 'true' ||
+      needs.check-file-change.outputs.tee_src == 'true' ||
+      needs.check-file-change.outputs.tee_test == 'true'
```
Kailai-Wang: This is not needed, this step should always be executed, the building itself however not (check the sub-steps below)

.github/workflows/ci.yml
```diff
@@ -140,7 +139,7 @@ jobs:
       - fmt
       - check-file-change
     container: "integritee/integritee-dev:0.1.10"
-    if: needs.check-file-change.outputs.parachain_src == 'true'
+    if: needs.check-file-change.outputs.tee_src == 'true'
```
Kailai-Wang, line 142: thanks!

.github/workflows/ci.yml
```diff
@@ -173,6 +172,11 @@ jobs:
     needs:
       - fmt
       - check-file-change
+    if: >
+      needs.check-file-change.outputs.parachain_src == 'true' ||
+      needs.check-file-change.outputs.parachain_test == 'true' ||
+      needs.check-file-change.outputs.tee_src == 'true' ||
+      needs.check-file-change.outputs.tee_test == 'true'
```
Kailai-Wang: same as `tee-build`, we don't need the `if` actually

Kailai-Wang (changes requested):

.github/workflows/ci.yml
BillyWooo: Actually I think it is good to have. Because the build/pull step is going to be used in test. But test also depends on these conditions. If the test is skipped, the build/pull here can also be skipped.

.github/workflows/coverage.yml
BillyWooo: Ah, yes. You are right. I was confused by the low coverage numbers. Because of quite many new pallets introduced.
BTW, we need to add more test code to have a better coverage.

.github/workflows/ci.yml
BillyWooo: come to alignment after offline discussion.

.github/workflows/ci.yml
BillyWooo: come to alignment after offline discussion.

Kailai-Wang (approved):

PR #1300, opened 2023-02-09
Add VC-A11 ETH hodler

zTgx:
as title.

### DIFF
--- a/tee-worker/litentry/core/assertion-build/src/a11.rs
+++ b/tee-worker/litentry/core/assertion-build/src/a11.rs
@@ -28,7 +28,7 @@ use lc_data_providers::graphql::{
 	GraphQLClient, VerifiedCredentialsIsHodlerIn, VerifiedCredentialsNetwork,
 };
 use litentry_primitives::{
-	Assertion, EvmNetwork, Identity, ParentchainBlockNumber, ASSERTION_FROM_DATE,
+	Assertion, Balance, EvmNetwork, Identity, ParentchainBlockNumber, ASSERTION_FROM_DATE,
 };
 use log::*;
 use std::{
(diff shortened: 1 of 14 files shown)


tee-worker/litentry/core/assertion-build/src/a11.rs
```diff
@@ -0,0 +1,97 @@
...
+			credential_unsigned.credential_subject.values.push(flag);
+
+			return Ok(credential_unsigned)
+		},
+		Err(e) => {
+			error!("Generate unsigned credential failed {:?}", e);
+		},
+	}
+
+	Err(Error::Assertion7Failed)
```
Kailai-Wang: Here should be `Assertion11Failed`

tee-worker/litentry/core/assertion-build/src/a11.rs
```diff
@@ -0,0 +1,97 @@
...
+use std::{
+	str::from_utf8,
+	string::{String, ToString},
+	vec,
+	vec::Vec,
+};
+
+pub fn build(
+	identities: Vec<Identity>,
+	min_balance: u128,
```
Kailai-Wang: Shall we use the type `Balance` here?

Kailai-Wang (commented):
Thanks, two small things:

tee-worker/litentry/core/assertion-build/src/a11.rs
zTgx: Thanks, fixed.

tee-worker/litentry/core/assertion-build/src/a11.rs
zTgx: Yes, Balance is better.

jingleizhang (approved):
LGTM

PR #1302, opened 2023-02-09
No panic is AES encyption

### LINKED ISSUE #1226: Error handling of AES encryption
### Context
We use a basic `unwrap` here when encrypting the data returned to the parachain:

https://github.com/litentry/litentry-parachain/blob/abfa4c06c7a5c3f0437d11745c9b4dd9ac1bc11e/tee-worker/litentry/primitives/src/lib.rs#L53-L61

This means we simply panic when something wrong happens. It's better to give explicit errors here (and generally anything that might block the frontend so that they don't wait forever), also due to #1227 

---

Kailai-Wang:
reoslves #1226 

A minor PR to return default results instead of panic in AES encryption within the enclave, see the comment in code:

> // Returns the default if any error happens
// We don't propagate the error to upper level as this function is used in too many places,
// it's too verbose to handle them all and pass back to the parentchain as events.
// We rely on the parentchain event consumers to handle them correctly (and they kind of
// have to, because they'll find all fields are 0)


ziming-zung (approved):
LGTM

Issue #1257, opened 2023-02-02
ts-test sends extrinsic in parallel


ziming-zung:
### Context
Currently with more and more ts-test, the test takes more time and  ts-test sends extrinsic to parentchain one by one. so I think we can improve ts-test to make it sends transactions in parallel.

Kailai-Wang:
What's the status of this one? (I'm asking as it's labeled with high priority)

PR #1305, opened 2023-02-10
error case test

0xverin:
resolves #1281
I think the errorCiphertext is the same effect as the error TeeShieldingKey,right?The effect of my test is the same.


Kailai-Wang (changes requested):
Thanks for the PR

I'm fine with the test logic but we can improve several things IMO:
- revert the accidentally formatted files => if we want, we should do it in a separate PR with a deterministic formatter
- I don't think we need the extra error-test yml, this can be merged into identity-test. If you check the time consumption, you'll the actual tests take only 2 min. And these are indeed identity-test related
- TBH I see many duplicate code. Can we refactor it to have a `listenEvent` where we pass in the desired event? The code logic should be very similar. Similarly, we don't need to have a dedicated `setErrorUserShieldingKey`, we can simply reuse `setUserShieldingKey` but pass in the fake keys. => this can be done in another PR

0xverin:
> 

At the beginning, I tested all error case in `identity.test.ts`. I was worried that there would be more and more testing work here, so I added an` error.test.ts`. and `listenEvent` only had one method at the beginning, without adding `listenErrorevent`. But considering that error.test is added, so it is add it accordingly.After consideration, there is no need to add error.test, because  #1257 will solve the  testing time is too long caused by the increasing error case of tests.

Kailai-Wang:
I mean even if we use `error.test.ts`, more and more error test cases will be added in this file, and eventually we'll have the same problem anyway :(

Kailai-Wang (approved):
Looks good, thanks! We probably want to refactor `Listen*Event` though - will have another issue for it

PR #1307, opened 2023-02-10
bugfix twitter verification, add assertion a6, refactoring a2&a3

jingleizhang:
as the title:

- bugfix twitter verification
- add assertion a6
- refactoring a2&a3
- a5 is unavailable until Twitter's new API policy is confirmed


tee-worker/litentry/core/assertion-build/src/a2.rs
```diff
@@ -35,55 +35,54 @@ use std::vec::Vec;
...
+		if let Identity::Web2 { network, address } = identity {
 			if matches!(network, Web2Network::Discord) {
 				discord_cnt += 1;
-				if let Ok(response) = client.check_join(guild_id.to_vec(), handler.to_vec()) {
+				if let Ok(response) = client.check_join(guild_id.to_vec(), address.to_vec()) {
 					if response.data {
 						has_joined = true;
-						break
+
+						//Assign role "ID-Hubber" to each discord account
```
Kailai-Wang, line 56: I have a general question: it seems an action item (assignment) is performed during the VC build, is it correct? I thought VC should only check against certain state (if some condition is true or false), and the action is the consequence of consuming VC. 

Assigning a discord role and requesting a VC should be independent I assume, but according to the code logic, it seems that the user can never be assigned the IDHubber role without requesting a VC - which seems strange.

This discussion is not so urgently needed though - we can decide it later

zTgx (approved):
looks good.

PR #1310, opened 2023-02-11
some minor refactor

BillyWooo:
some refactor about test

PR #1321, opened 2023-02-14
Enclave update [4] - state migration

### LINKED ISSUE #1180: Enclave update [4] - state migration and copy-over
After the worker starts with the new MRENCLAVE, the first thing to do is state migration - if any, and copy the last state just before the update to the new shard folder, which will be served as the base state for the incoming STF.

We need this before the shard value has changed due to different MRENCALVE, or is there a way to keep shard unchanged?

- [x] #1304

ziming-zung:
This PR resolves #1180


tee-worker/local-setup/py/helpers.py
```diff
@@ -45,10 +45,6 @@ def setup_working_dir(source_dir: str, target_dir: str):
         else:
             print(f'{source} does not exist. Did you run make?')
 
-def copy_config_files(source_dir: str, target_dir: str):
-    files_to_copy = ['worker-config-dev.json', 'worker-config-mock.json', 'worker-config-prod.json', 'worker-config-staging.json']
-    [shutil.copy(f'{source_dir}/{f}', f'{target_dir}/{f}') for f in files_to_copy]
-
```
BillyWooo: nice catches. Thanks. It would be better to make a separate PR. But keep it as it is.

BillyWooo (approved):
In general, it looks good for me. I assume there will be a new test once all enclave update issues are covered.

tee-worker/local-setup/py/helpers.py
Kailai-Wang: This will be covered in https://github.com/litentry/litentry-parachain/pull/1310, but yea let's just leave it

PR #1324, opened 2023-02-14
Use !failure() to cover skipped jobs

Kailai-Wang:
as title.

After I found the [latest dev run](https://github.com/litentry/litentry-parachain/actions/runs/4173415293) still didn't push the images, I realised `success()` is evaluated as `false` whenever there're skipped jobs.


BillyWooo (approved):
LGTM

PR #1327, opened 2023-02-14
Expose account as cleartext in the returned events

### LINKED ISSUE #1313: Do not encrypt `account` in returning events of IMP
### Context
Currently we encrypt the `account` in the returning events of IMP, e.g.
```
IdentityCreated { account: AesOutput, identity: AesOutput, id_graph: AesOutput },
```

To be more consistent with VCMP and better indexing, IDH colleagues propose to make it public. Generally it should be OK as long as the identity is not leaked - that is, we consider it OK to expose `some user A has successfully created an identity`, as long as the details of the identity remain covert.

---

### LINKED ISSUE #1314: Refactor `listenEvent` in tee ts-tests
### Context
We have some duplicate code in listening to different events (error-events, specific events) in tee ts-tests, can we have a more general `listenEvent` with the desired event as a parameter?

---

Kailai-Wang:
resolves #1313 
fixes maybe #1314 too

This PR:
- changes the account from `AesOutput` to `T::AcccountId`
- adds a simple `SgxParentchainTypeConverter` to covert sidechain types to parentchain types - it can be extended later
- wasn't meant to solve two issues in one PR but has to refactor `listenEvent` as the original one doesn't meet the requirements for mixed ciphertext/cleartext in the returned events
- adds a bit more logs to list the events in each block in ts-tests

cc @Verin1005 @zhizming-zhong especially for the ts-test part


tee-worker/ts-tests/utils.ts
```diff
@@ -141,161 +144,47 @@ export async function sendTxUntilInBlock(api: ApiPromise, tx: SubmittableExtrins
...
-                }
+            console.log(`\n--------- block #${header.number}, hash ${header.hash} ---------\n`);
+            const apiAt = await api.at(header.hash);
+
+            const records: EventRecord[] = (await apiAt.query.system.events()) as any;
+            records.forEach((e, i) => {
+                const s = (e.toHuman() as any).event.section;
+                const m = (e.toHuman() as any).event.method;
+                const d = e.event.data;
+                console.log(`Event[${i}]: ${s}.${m} ${d}`);
```
ziming-zung, line 182: Why not change these code to 
```ts
const records = (await apiAt.query.system.events()) as Vec<EventRecord>;
records.forEach((e, i) => {
    const s = e.event.section;
    const m = e.event.method;
    const d = e.event.data;
    console.log(`Event[${i}]: ${s}.${m} ${d}`);
});
```

tee-worker/ts-tests/utils.ts
```diff
@@ -32,6 +32,9 @@ const crypto = require('crypto');
 // TODO add self signed certificate ??
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
 
+// maximum number of blocks that we wait in listening events before we timeout
+const listenTimeoutInBlock = 15;
```
ziming-zung: When I use [litentry-runtime](https://github.com/litentry/integritee-node/tree/litentry-runtime) as the parent chain, 15 blocks may not be enough, so calculate `listenTimeoutInBlock` by `slot_duration` in the original code

ziming-zung (commented):
Basically it looks good to me , just two doubts from my side.

tee-worker/ts-tests/utils.ts
0xverin: Me too.

0xverin:
This pr needs to be merged first, and my pr #1328  needs to be changed, I still use the old listen*events. cc @jingleizhang cc @zhizming-zhong cc @Kailai-Wang

tee-worker/ts-tests/utils.ts
```diff
@@ -141,161 +144,47 @@ export async function sendTxUntilInBlock(api: ApiPromise, tx: SubmittableExtrins
...
+// We can listen to multiple `methods` as long as they are emitted in the same block.
+// The event consumer should do the decryption optionaly as it's event specific
+//
+// TODO: occassionally multiple events for an extrinsic are not included in the same block,
+//       e.g. `create_identity` => `IdentityCreated`, `ChallengeCodeGenerated`
+//       this is because the extrinsics are submitted asynchronously and in rare cases these two
+//       extrinsics are included in the different parentchain blocks
+// Solutions:
+//  1. (pallet change) use one single extrinsic to emit both events, if they should always be triggered on pair
+//  2. (ts-test change) only resolve this promise when both events are received, but not necessarily in the same block
```
ziming-zung, line 162: I think we can simply implement solution 2 like this:
```ts
export async function listenEvent(api: ApiPromise, section: string, methods: string[]) {
    return new Promise<Event[]>(async (resolve, reject) => {
        let startBlock = 0;
        let events: EventRecord[] = []
        const unsubscribe = await api.rpc.chain.subscribeNewHeads(async (header) => {
            const currentBlockNumber = header.number.toNumber();
            if (startBlock == 0) startBlock = currentBlockNumber;
            if (currentBlockNumber > startBlock + listenTimeoutInBlock) {
                reject('timeout');
                return;
            }
            console.log(`\n--------- block #${header.number}, hash ${header.hash} ---------\n`);
            const apiAt = await api.at(header.hash);

            const records = (await apiAt.query.system.events()) as Vec<EventRecord>;
            records.forEach((e, i) => {
                const s = e.event.section;
                const m = e.event.method;
                const d = e.event.data;
                console.log(`Event[${i}]: ${s}.${m} ${d}`);
            });
            const filtered_events = records.filter(
                ({ phase, event }) =>
                    phase.isApplyExtrinsic && section === event.section && methods.includes(event.method)
            );            
            events.push(...filtered_events);
            if (events.length == methods.length) {
                resolve(events.map((e) => e.event));
                unsubscribe();
                return;
            }
        });
    });
}
```

Kailai-Wang: Thanks. This however only makes sure the length is equal, but not the content, so theoretically it can be two same events 😂 (if a block contains that).

Anyway I think both option 1 and 2 are needed - they can serve different purposes. I'll leave it to the next pr.

tee-worker/ts-tests/utils.ts
Kailai-Wang: Thanks, changed. It didn't work out in my first try, therefore I changed it - but now it seems working again, strange 😂

tee-worker/ts-tests/utils.ts
Kailai-Wang: I see, reverted.

ziming-zung (approved):
:thumbsup:

tee-worker/ts-tests/utils.ts
0xverin:
```
export async function listenEvent(api: ApiPromise, section: string, methods: string[], txLen: number) {
    return new Promise<Event[]>(async (resolve, reject) => {
        let startBlock = 0;
        let events: EventRecord[] = []
        const unsubscribe = await api.rpc.chain.subscribeNewHeads(async (header) => {
            const currentBlockNumber = header.number.toNumber();
            if (startBlock == 0) startBlock = currentBlockNumber;
            const timeout = await getListenTimeoutInBlocks(api);
            if (currentBlockNumber > startBlock + timeout) {
                reject('timeout');
                return;
            }
            console.log(`\n--------- block #${header.number}, hash ${header.hash} ---------\n`);
            const apiAt = await api.at(header.hash);

            const records: EventRecord[] = (await apiAt.query.system.events()) as any;
            records.forEach((e, i) => {
                const s = e.event.section;
                const m = e.event.method;
                const d = e.event.data;
                console.log(`Event[${i}]: ${s}.${m} ${d}`);
            });
            const filtered_events = records.filter(
                ({ phase, event }) =>
                    phase.isApplyExtrinsic && section === event.section && methods.includes(event.method)
            );
            events.push(...filtered_events)
            if (events.length == txLen) {
                resolve(events.map((e) => e.event));
                unsubscribe();
                return;
            }
        });
    });
}
```
I think it be better to write it like above(I've tested it a few times ).I see this [CI](https://github.com/litentry/litentry-parachain/actions/runs/4441086313/jobs/7797592061?pr=1477) failed,maybe trigger this, I remember this pr and `TODO`. txLen is [txs.length](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/indirect_calls.ts#L49). cc @BillyWooo

PR #1368, opened 2023-02-24
Sequentialise workflow runs on dev branch

### LINKED ISSUE #1359: CI for dev branch should be queued to be executed.
### Context
As new PR policy allows more flexible merge operation, such scenario would happen: previous CI runners for dev branch have not yet finished, another new PR has been merge, which triggers new CI runners for dev and cancels the previous one. Shown as the screenshot: [job-2](https://github.com/litentry/litentry-parachain/actions/runs/4240729543) cancels [job-1](https://github.com/litentry/litentry-parachain/actions/runs/4240537947).  While job-2 still hold/upload old litentry-parachain image, job-1 is canceled by job-2.

### Task
Fix the CI config. 

### Acceptance criteria
CI runner for dev branch should be queued to be executed.

image

---

Kailai-Wang:
resolves #1359 

This should be enough, tests are welcome.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,8 +47,8 @@ env:
 
 concurrency:
   # see https://stackoverflow.com/questions/74117321/if-condition-in-concurrency-in-gha
-  # it guarantees:
-  # - for push in dev: all triggered CIs will be run, none is cancelled
+  # along with the `sequentialise` job below, it guarantees:
+  # - for push in dev: all triggered CIs will be run sequentially, none is cancelled
   # - for PR: later triggered CIs will cancel previous runs, maximum one CI will be run
   group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
   cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


BillyWooo:
Will the following CI (which also runs against 'dev' branch) be queued to be executed?

Kailai-Wang:
> Will the following CI (which also runs against 'dev' branch) be queued to be executed?

queued.

But there's a corner case when multiple jobs are pending:

job 1 is running
job 2 is triggered => it sees 1 is running => 2 in pending state
job 3 is triggered => 2 will be cancelled => 3 in pending state

See https://docs.github.com/en/actions/using-jobs/using-concurrency

I think with current GHAs it's not easy to achieve a waiting chain without cancelling the intermediate jobs.

Kailai-Wang:
Or do we actually want it to be executed?

Kailai-Wang:
I could make sure there's no jobs skipped on `dev`, but there's another corner case I can think of:

job 1 is triggered => the commit changed both `parachain` and `tee` src, so both docker images will be built
job2 is triggered => the commit only contains changes in `tee` src, so only tee image needs to be re-built.

Then we'll have a mismatch of parachain and tee images regardless of the completion order of job 1 and 2:
- job 1 finishes first => when job 2 finishes, we will have the old parachain image and new tee image
- job 2 finishes first => when job 1 finishes, we will have the new parachain image and old tee image

I think we still need to conditionally push the image (= only push it when it's actually rebuilt)

edit: actually such problem can happen more often: as long as job 2 finishes later than job 1 and both jobs rebuild images => we only have image from job 1.

Kailai-Wang:
I think it's ready now. I use a combination of `concurrency` and `mktcode/consecutive-workflow-action` to make sure (hopefully):
 - for push in dev: all triggered CIs will be run sequentially, none is cancelled
 - for PR: later triggered CIs will cancel previous runs, maximum one CI will be run

BillyWooo (approved):
awesome

Issue #1274, opened 2023-02-05
StfError in trusted_call fails silently


jingleizhang:
https://github.com/litentry/litentry-parachain/blob/ce0ff808b44c204149da4c395734e0dc10fbabe9/tee-worker/app-libs/stf/src/trusted_call_litentry.rs#L125-L126

The trusted_call doesn't return to Parachain when an error happens, as the StfError is just emitted into the worker log.

Parachain needs to know when there is a trusted_call error.

Kailai-Wang:
Good catch, I think in general the error handling in trusted_call needs to be improved, even `ensure!` doesn't propagate the error properly.

Issue #1064, opened 2022-12-06
make returned error value be clear and specific 


jingleizhang:
Identity Management Pallet only has `SomeError` right now which is not quite clear.

Those returned error values should be more clear and specific, and more friendly to FE and users.

reference error codes: https://docs.litentry.com/products/my-crypto-profile/developer-documentation/api-references

Kailai-Wang:
related to https://github.com/litentry/litentry-parachain/issues/1008

PR #1372, opened 2023-02-24
bugfix: convert stf_error to imp_error

jingleizhang:
as the title

### DIFF
--- a/tee-worker/app-libs/stf/src/trusted_call.rs
+++ b/tee-worker/app-libs/stf/src/trusted_call.rs
@@ -42,6 +42,7 @@ use litentry_primitives::{
 	UserShieldingKeyType, ValidationData,
 };
 use log::*;
+use parentchain_primitives::{ErrorString, IMPError};
 use sp_io::hashing::blake2_256;
 use sp_runtime::{traits::Verify, MultiAddress};
 use std::{format, prelude::v1::*, sync::Arc};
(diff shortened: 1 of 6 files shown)


tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -672,6 +667,14 @@ where
...
 }
 
+pub(crate) fn convert_stf_error(e: StfError) -> IMPError {
+	match e {
+		StfError::Dispatch(s) =>
+			IMPError::StfError(ErrorString::truncate_from(s.as_bytes().to_vec())),
+		_ => IMPError::StfError(ErrorString::truncate_from(format!("{:?}", e).as_bytes().to_vec())),
+	}
+}
+
```
Kailai-Wang: This converter is not needed I think.

`StfError` implements `Display` and thus `to_string`, we can simply use `e.to_string()` => for `StfError::Dispatch` too.

I suggest we just implement:
```
impl StfError {
  pub fn to_imp_error(&self) ....
}
```

What do you think?

Kailai-Wang (approved):
Thanks for spotting this, looks good - I only have one small suggestion.

It is related to https://github.com/litentry/litentry-parachain/issues/1064 -- meaning not all error-handling cases were covered.

We need to have ts-tests too - at least for some of these StfErrors - let's have an issue for that @jingleizhang @Verin1005

Kailai-Wang:
related to #1274 too?

jingleizhang:
a following issue for ts-test is created: https://github.com/litentry/litentry-parachain/issues/1374

tee-worker/app-libs/stf/src/trusted_call.rs
jingleizhang: a following issue for ts-test is created: https://github.com/litentry/litentry-parachain/issues/1374

jingleizhang: fixed

jingleizhang:
It is related to https://github.com/litentry/litentry-parachain/issues/1274, but this issue doesn't resolve it completely.

PR #1375, opened 2023-02-26
Improve docker rebuild strategy in CI

Kailai-Wang:
This PR:

- adds inputs to workflow_dispatch to allow forcibly docker image rebuild
- use `set-condition` to extend the file-filter with the input parameters
- remove hooks - the test **job** will always be executed, but the test **step** will be skipped if needed
- only push the image if it's rebuilt

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,20 +6,26 @@ name: General CI
 # - unit tests
 # - integration tests
 
-# the tee-worker part is a modified version of tee-worker/.github/workflows/build_and_test.yml
+# Some notes:
+#
+# [1] the tee-worker part is a modified version of tee-worker/.github/workflows/build_and_test.yml
 # with extra triggering control.
 #
-# the original file (`tee-worker/.github/workflows/build_and_test.yml`) is kept to sync
+# [2] the original file (`tee-worker/.github/workflows/build_and_test.yml`) is kept to sync
 # upstream changes, therefore we need to manually apply the changes to this file.
 
-# please beware that if a job in `needs` is skipped, its dependent job will also be skipped,
+# [3] please beware that if a job in `needs` is skipped, its dependent job will also be skipped,
 # see https://github.com/actions/runner/issues/491
 #
 # jobs that will always be executed:
 # - fmt
-# - check-file-change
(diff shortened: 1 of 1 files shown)


BillyWooo (commented):
1. Why don't put `if: needs.set-condition.outputs.run_tee_test == 'true'` at `tee-test` job-level, instead of step-level ?
2. Why need `push_docker`? e.g. in case of neither of `rebuild_parachain` or `rebuild_tee` is true?

Kailai-Wang:
> 1. Why don't put `if: needs.set-condition.outputs.run_tee_test == 'true'` at `tee-test` job-level, instead of step-level ?
> 2. Why need `push_docker`? e.g. in case of neither of `rebuild_parachain` or `rebuild_tee` is true?

1. because otherwise it's impossible to configure the status check that covers all cases without the hook, see https://github.com/orgs/community/discussions/26822

2. `push-docker` is only for workflow_dispatch, otherwise there won't be a way to forcibly push the image - and yes, to use it, one of them has to be rebuilt, see https://github.com/litentry/litentry-parachain/actions/runs/4274158316

BillyWooo:
> > 1. Why don't put `if: needs.set-condition.outputs.run_tee_test == 'true'` at `tee-test` job-level, instead of step-level ?
> > 2. Why need `push_docker`? e.g. in case of neither of `rebuild_parachain` or `rebuild_tee` is true?
> 
> 1. because otherwise it's impossible to configure the status check that covers all cases without the hook, see https://github.com/orgs/community/discussions/26822
> 2. `push-docker` is only for workflow_dispatch, otherwise there won't be a way to forcibly push the image - and yes, to use it, one of them has to be rebuilt, see https://github.com/litentry/litentry-parachain/actions/runs/4274158316

2. So I think `push-docker` input from workflow_dispatch is not needed any more. Because its value can be propagated out: `push-docker =  rebuild-parachain-docker || rebuild-tee-docker`

Kailai-Wang:
It can't, because by default a docker-push can only be triggered by a commit on `dev` branch. You don't want it to be pushed automatically on some random feature branch

BillyWooo (approved):
LGTM

PR #1380, opened 2023-02-27
Launch up script update

BillyWooo:
Optimize the python launch script to wait for worker to start.

### DIFF
--- a/tee-worker/local-setup/launch.py
+++ b/tee-worker/local-setup/launch.py
@@ -61,15 +61,25 @@ def main(processes, config_path, parachain_type):
         print('Starting litentry-parachain done')
         print('----------------------------------------')
 
-    UNTRUSTED_HTTP_PORT=4545
     c = pycurl.Curl()
     worker_i = 0
     worker_num = len(config["workers"])
     for w_conf in config["workers"]:
         processes.append(run_worker(w_conf, worker_i))
         print()
-        url = 'http://localhost:' + str(UNTRUSTED_HTTP_PORT+worker_i) + '/is_initialized'
+
+        idx = 0
+        if ( "-h" in w_conf["flags"] ):
+            idx = w_conf["flags"].index("-h") + 1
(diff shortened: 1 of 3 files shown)


tee-worker/local-setup/launch.py
```diff
@@ -58,19 +61,36 @@ def main(processes, config_path, parachain_type):
...
+                    break
+                if counter >= 60:
+                    print("Worker initialization error. Exit")
+                    return 0
+                counter = counter + 1
+
+        worker_i += 1
+
+    c.close()
+    print("Worker(s) started!")
```
Kailai-Wang, line 105: I think you need to print "workers started" until the last worker is started (currently it looks for `localhost:4545` which is the first worker)

So when I see this message, worker2 is not fully started, actually

tee-worker/local-setup/launch.py
```diff
@@ -61,14 +61,15 @@ def main(processes, config_path, parachain_type):
...
+    UNTRUSTED_HTTP_PORT=4545
     c = pycurl.Curl()
-    c.setopt(pycurl.URL, 'http://localhost:4545/is_initialized')
-
     worker_i = 0
     worker_num = len(config["workers"])
     for w_conf in config["workers"]:
         processes.append(run_worker(w_conf, worker_i))
         print()
+        url = 'http://localhost:' + str(UNTRUSTED_HTTP_PORT+worker_i) + '/is_initialized'
```
Kailai-Wang: Actually the port should come from parsing the worker.config (I could use something other than 4545 due to e.g. port conflict), but let's leave it - it's for local use anyway

Kailai-Wang (approved):

tee-worker/local-setup/launch.py
BillyWooo: Updated.

tee-worker/local-setup/launch.py
BillyWooo: Fixed.

PR #1385, opened 2023-02-28
Add `NativeTokenMinted` event

### LINKED ISSUE #1373: Add a substrate event for successful ETH => parachain transfer
### Context

There's no such event right now, making it a bit difficult to identify if a user really receives LIT on the parachain. We could search for the vote-related events in the block explorer, which is close but still not direct enough.

A minor change, the name can be something like `BridgeTransfer.Minted`

https://github.com/litentry/litentry-parachain/blob/25bf6e0e8137209a0cc1c60cc2c89729ddc83fc6/pallets/bridge-transfer/src/lib.rs#L160-L184

---

Kailai-Wang:
resolves #1373 

A minor PR to add a `NativeTokenMinted` event for easier tracking of successful token transfer from ETH => parachain.


BillyWooo (approved):
LGTM

PR #1386, opened 2023-03-01
Get scheduled mr_enclaves from parachain.(fix #1383)

### LINKED ISSUE #1383: Enclave Update [3-1] get scheduled enclaves from parachain in tee-worker
sync the parachain blocks, parse the extrinsics to get the scheduled enclave, then store it to a constraint file.

#1295

DogLi:


Kailai-Wang:
I think the `fix #1383` won't link the issue automatically if it's written in the title - we still need to use auto-close keyword.

tee-worker/core-primitives/sgx/enclave-scheduled/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+	fn sync_to_static_file(&self) -> IOResult<()> {
+		let s = serde_json::to_string(&self).map_err(|e| Error::new(ErrorKind::InvalidInput, e))?;
+		seal(s.as_bytes(), SCHEDULED_ENCLAVE_FILE)
+	}
+
+	fn add_scheduled_enclave(&mut self, scheduled_enclave: ScheduledEnclaveInfo) -> IOResult<()> {
+		self.scheduled_enclaves.push(scheduled_enclave);
+		self.sync_to_static_file()?;
+		Ok(())
+	}
```
Kailai-Wang: I'm not sure if this will work.

Remember `update_scheduled_enclave` can **update** the enclave instead of **adding** a new enclave. I think we have two options here:
- use `map` instead of `vec` for storage, and replay whatever happens upstream here
- copy the upstream storage `ScheduledEnclave` over here - it's always up-to-date (triggered by every block import if it's not costly, or by extrinsic parsing)

tee-worker/core-primitives/sgx/enclave-scheduled/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+		sidechain_block_number: SidechainBlockNumber,
+	) -> IOResult<()> {
+		self.scheduled_enclaves
+			.retain(|info| info.sidechain_block_number != sidechain_block_number);
+		self.sync_to_static_file()?;
+		Ok(())
+	}
+
+	/// get the next scheduled enclave
+	fn get_next_scheduled_enclave(
```
Kailai-Wang, line 124: This is not yet used?

We need to sort the collection by `sidechain_number`, otherwise we can get the wrong one

jingleizhang (approved):
Looks good.

PR #1369, opened 2023-02-24
add VC issuer attestation

zTgx:
* update `SgxEnclaveMetadata`
* add issuer attestation
* fix vc signature
* add vc logs
* update vcPubkey to vec
* Remove unused codes
* merged ts-tests pr: https://github.com/litentry/litentry-parachain/pull/1328

PR #1389, opened 2023-03-01
Refactor enum Assertion

### LINKED ISSUE #1387: refactor enum Assertion
### Context

Two `enum Assertion` exists right now, we need to refactor to one, and move it to `primitives/core/src/` where `tee-worker`  report from.
case 1:
https://github.com/litentry/litentry-parachain/blob/1347ff356996b3db7158711c5fef915e724d6a72/pallets/vc-management/src/assertion.rs#L33-L46

case 2:
https://github.com/litentry/litentry-parachain/blob/1347ff356996b3db7158711c5fef915e724d6a72/tee-worker/litentry/primitives/src/assertion.rs#L41-L54

outofxxx:
resolved #1387 

* remove parachain_core_primitives
* fix wildcard imports


pallets/sidechain/src/tests.rs
```diff
@@ -208,6 +208,7 @@ fn register_enclave(signer_pub_key: &MrSigner, cert: &[u8], expected_enclave_cou
 		cert.to_vec(),
 		URL.to_vec(),
 		None,
+		None,
```
BillyWooo: this is a fix for another PR?

tee-worker/core-primitives/types/Cargo.toml
```diff
@@ -29,6 +29,7 @@ sp-std = { default-features = false, git = "https://github.com/paritytech/substr
 
 # litentry
 litentry-primitives = { path = "../../litentry/primitives", default-features = false }
+parentchain-primitives = { package = "core-primitives", path = "../../../primitives/core", default-features = false }
```
Kailai-Wang: I think this can be chaotic because we have a crate:
=> depends on `parentchain-primitives` directly
=> depends on `litentry-primitives` that depend on `parentchain-primitives`

Can we make it linearly dependent? Like external crates only depend on `litentry-primitives`. `litentry-primitives` will re-export anything that it needs from `parentchain-primitives`

There're more places than this one

Kailai-Wang (commented):
Thank you! I only have one question regarding linear dependency

pallets/sidechain/src/tests.rs
outofxxx: It was fix in https://github.com/litentry/litentry-parachain/pull/1369, I will revert this.

BillyWooo:
Hi @zhouhuitian , I think `litentry-primitives` includes `parachain-core-primitives`. Do you think so ?

tee-worker/core-primitives/sgx-runtime-primitives/src/types.rs
```diff
@@ -21,7 +21,7 @@ use sp_runtime::{
 	MultiSignature, OpaqueExtrinsic,
 };
 
-use parentchain_primitives::AccountId as ParentchainAccountId;
+use litentry_primitives::AccountId as ParentchainAccountId;
```
Kailai-Wang: I think it's better to set this alias inside `litentry_primitives`, like:
```
pub use parentchain_primitives::AccountId as ParentchainAccountId
```
similar to BlockNumber.
We use `Parentchan*` to differentiate it from those in sidechains.

tee-worker/litentry/core/assertion-build/src/a1.rs
```diff
@@ -25,9 +25,9 @@ use itp_stf_primitives::types::ShardIdentifier;
 use itp_types::AccountId;
 use lc_credentials::Credential;
 use lc_stf_task_sender::MaxIdentityLength;
-use litentry_primitives::{Assertion, Identity, ParentchainBlockNumber};
+use litentry_primitives::{Identity, ParentchainBlockNumber};
 use log::*;
-use parachain_core_primitives::VCMPError;
+use parachain_core_primitives::{Assertion, VCMPError};
```
Kailai-Wang: We shouldn't have both `litentry_primitives` and `parachain_core_primitives`, we should only depend on `litentry_primitives`.

If you need `Assertion`, you should re-import it in the `litentry_primitives`: `pub use parentchain_primitives::Assertion` and use it here: `use litentry_primitives::Assertion`

Same applies to other assertion-build files

tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -49,10 +48,11 @@ pub mod sgx_reexport_prelude {
 
 use rand::Rng;
 
-pub use assertion::*;
 pub use enclave_quote::*;
 pub use validation_data::*;
 
+pub use parentchain_primitives::*;
```
Kailai-Wang: Let's try to avoid importing `*` but only those we actually need. Plus it shadows line 28. I'm just concerned that we expose more than what we need, and use the wrong primitives.

Kailai-Wang (changes requested):
Still some places where we should only depend on `litentry_primitives` (if I'm not mistaken)

jingleizhang (approved):
Now looks good!

Kailai-Wang (approved):

PR #1396, opened 2023-03-03
added integritee restart workflow

niteshb3495:


.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - name: restarting the service
+        if: "${{ github.event.inputs.action == 'restart' }}"
+        run: |
+          pkill -9 integritee-service
+          cd /home/ubuntu/litentry-parachain/tee-worker
+          ./local-setup/launch.py local-setup/github-action-config.json
+
+      - name: stoping the service
+        if: "${{ github.event.inputs.action == 'restart' }}"
```
jingleizhang: The action here should be stop?

.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: |
+          pkill -9 integritee-service
+
+      - name: starting the service
+        if: "${{ github.event.inputs.action == 'restart' }}"
```
jingleizhang: The action here should be start?

.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  action:
+    runs-on: self-hosted #tee-staging (162.19.84.27)
+
+    steps:
+      - name: restarting the service
+        if: "${{ github.event.inputs.action == 'restart' }}"
+        run: |
+          pkill -9 integritee-service
+          cd /home/ubuntu/litentry-parachain/tee-worker
```
jingleizhang: the base directory should be `/home/ubuntu/repo`?

.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,37 @@
...
+        required: true
+        default: restart
+        options: 
+        - restart
+        - stop
+        - start
+
+jobs:
+  action:
+    runs-on: self-hosted #tee-staging (162.19.84.27)
```
Kailai-Wang: Is this self-hosted runner configured? Otherwise it won't work out of the box.

Alternatively we could use GH runner but pass in the target host IP and ssh keys as secrets (I prefer this way)

.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,37 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      action:
+        type: choice
+        description: Select the action
+        required: true
+        default: restart
+        options:
```
Kailai-Wang: `choice` is better if we only allow one action each time.

maybe we only need two options:
- stop only
- stop and start

.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  action:
+    runs-on: self-hosted #tee-staging (162.19.84.27)
+
+    steps:
+      - name: restarting the service
+        if: "${{ github.event.inputs.action == 'restart' }}"
+        run: |
+          pkill -9 integritee-service
```
Kailai-Wang: What about the parachain?

Kailai-Wang (changes requested):

.github/workflows/staging-sgx-restart.yml
Kailai-Wang: I think the python script only responds to `SIGINT` and `SIGTERM`, so you need to send those signals, not `SIGKILL`

BillyWooo (commented):
@niteshb3495  Just as we talked in the call. 
Kill python
kill integritee-service
kill docker container

.github/workflows/staging-sgx-restart.yml
BillyWooo: Maybe we can connect the `SIGKILL` too?

Kailai-Wang:
> @niteshb3495 Just as we talked in the call. Kill python kill integritee-service kill docker container

I thought sending the right signals to `launch.py` is enough?

.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,28 @@
+name: Staging SGX parachain
+
+on:
+  workflow_dispatch:
+    inputs:
+      action:
+        type: choice
+        description: Select the action
+        required: true
+        default: stop
```
Kailai-Wang: I'm not sure if `choice` has the `required` and `default` property

.github/workflows/staging-sgx-restart.yml
Kailai-Wang: Let's remove the IP in file :)

.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,28 @@
...
+        - stop
+        - start
+
+jobs:
+  action:
+    runs-on: self-hosted #tee-staging (162.19.84.27)
+
+    steps:
+      - name: restarting the service
+        if: "${{ github.event.inputs.action == 'stop' }}"
```
Kailai-Wang: the name and the action is contradictory :(

.github/workflows/staging-sgx-restart.yml
```diff
@@ -0,0 +1,28 @@
...
+        run: |
+          sudo systemctl stop integritee.service
+
+      - name: stoping the service
+        if: "${{ github.event.inputs.action == 'start' }}"
+        run: |
+          sudo systemctl start integritee.service
```
Kailai-Wang: Shall we use `restart` here? Otherwise I have to run this GHA twice (first stop, then start)

Kailai-Wang: Maybe we should also collect the logs whenever we do a full reset (restart)

.github/workflows/staging-sgx-restart.yml
niteshb3495: Yes, self-hosted runner is already configured. For testing, it's a dev server. 
Adding IP and secret is an excellent way to have, but GHA comes with the agent based service which is a more secure way to use. We only need to set up the agent. The runs-on parameter value (self-hosted) will be the same everywhere. To avoid confusion, we can give the comment as to where (name of the server, in this case, tee-dev) the GH agent is set up, for reference. 

IP removed.

.github/workflows/staging-sgx-restart.yml
niteshb3495: added restart

not sure what you mean by `Maybe we should also collect the logs whenever we do a full reset (restart)`

Kailai-Wang: will the logs be overwritten? Maybe we want to keep the previous logs before restarting :)

niteshb3495: I tested the restart by adding stop and start actions respectively. 
But in some cases, as I faced just now while running the [workflow] (https://github.com/litentry/litentry-parachain/actions/runs/4342133882/jobs/7582587026), if the service is not running on the host, the workflow will fail.
Adding stop, start and restart all three options in case, the service is not already running to stop.

BillyWooo: for the logs, maybe it's easy to handle inside python script when do the 'graceful_exit'

Kailai-Wang: > for the logs, maybe it's easy to handle inside python script when do the 'graceful_exit'

Yes I mean we need to upload this logs somewhere, as this one is triggered by GHA. 

Whoever ran this GHA doesn't necessarily have the permission to log on to the host to check the logs (e.g. IDH colleagues)

PR #1401, opened 2023-03-06
Fix assertion build panicked due to Uft8Error

### LINKED ISSUE #1397: assertion build panicked due to Uft8Error
Error logs:

`<unnamed>' panicked at 'called Result::unwrap() on an Err value: Utf8Error { valid_up_to: 2, error_len: Some(1) }', /home/ubuntu/repo/litentry-parachain/tee-worker/litentry/core/assertion-build/src/a4.rs:[2023-03-02T15:21:08Z INFO  integritee_service::ocall_bridge::sidechain_ocall] Enclave produced sidechain blocks: [17032]`

`fatal runtime error: failed to initiate panic, error
thread 'thread '<unnamed><unnamed>' panicked at '' panicked at 'called Result::unwrap() on an Err value: Sgx(SGX_ERROR_ENCLAVE_CRASHED)called Result::unwrap() on an Err value: Sgx(SGX_ERROR_ENCLAVE_CRASHED)', ', service/src/main.rsservice/src/main.rs::638499::8118
[2023-03-02T15:21:16Z ERROR integritee_service] Parentchain block syncing terminated with a failure: EnclaveApi(Sgx(SGX_ERROR_ENCLAVE_CRASHED))`

outofxxx:
resolved #1397

### DIFF
--- a/tee-worker/litentry/core/assertion-build/src/a11.rs
+++ b/tee-worker/litentry/core/assertion-build/src/a11.rs
@@ -60,30 +60,33 @@ pub fn build(
 
 		if let Identity::Evm { network, address } = id {
 			if matches!(network, EvmNetwork::Ethereum) {
-				if let Ok(addr) = from_utf8(address.as_ref()) {
-					let addresses = vec![addr.to_string()];
+				match from_utf8(address.as_ref()) {
+					Ok(addr) => {
+						let addresses = vec![addr.to_string()];
 
-					for (index, from_date) in ASSERTION_FROM_DATE.iter().enumerate() {
-						// if found is true, no need to check it continually
-						if found {
-							from_date_index = index + 1;
-							break
-						}
+						for (index, from_date) in ASSERTION_FROM_DATE.iter().enumerate() {
+							// if found is true, no need to check it continually
+							if found {
+								from_date_index = index + 1;
+								break
+							}
 
-						let credentials = VerifiedCredentialsIsHodlerIn::new(
-							addresses.clone(),
-							from_date.to_string(),
-							VerifiedCredentialsNetwork::Ethereum,
-							String::from(""),
-							q_min_balance,
-						);
-						let is_hodler_out = client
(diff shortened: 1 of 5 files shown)


tee-worker/litentry/core/assertion-build/src/a11.rs
```diff
@@ -60,30 +60,31 @@ pub fn build(
 
 		if let Identity::Evm { network, address } = id {
 			if matches!(network, EvmNetwork::Ethereum) {
-				let address = from_utf8(address.as_ref()).unwrap().to_string();
-				let addresses = vec![address];
+				if let Ok(addr) = from_utf8(address.as_ref()) {
```
Kailai-Wang: Do we need to pass it to parentchain instead of ignoring it? @jingleizhang 

I see something in https://github.com/litentry/litentry-parachain/blob/656c1985e6c682f34ae5d4a7dbc1ad0653d67acd/primitives/core/src/error.rs#L64-L65

Is this supposed to be used?

BillyWooo: Agree. I think we should propagate it out.

jingleizhang (commented):
The Uft8Error should be propagated to parent chain.

tee-worker/litentry/core/assertion-build/src/a11.rs
outofxxx: In those contexts(A4/A7/A10/A11) no need to propagate `ParseError` to parachain:
1. those payload are integer(Balance), return `ParseError` will be confusing.
2. in `from_utf8(address.as_ref())`, `address` is from `IDGraph`
3. we can add logs, if parse error.

jingleizhang: agree, some of ParseErrors are totally internal, and don't need to be propagated out

jingleizhang (approved):
Only the UTF8Error A8 needs to be propagated to parentchain as vcManagement.ParseError. LGTM.

PR #1404, opened 2023-03-06
Handle all StfErrors

### LINKED ISSUE #1274: StfError in trusted_call fails silently

Kailai-Wang:
resolves #1274

This PR tries to:
- handle all potential StfErrors to ensure they are propagated to the parachain.
- refactor the error handling a bit to avoid duplicate code

**Another important note:**

IMO we need to guarantee the stf execution always returns Ok() with extrinsic status `success`, otherwise it will be removed silently from the top pool, which might cause a state mismatch during sidechain block import for other workers. 
Please check the comments in `tee-worker/app-libs/stf/src/trusted_call.rs`, discussions are welcome.

### DIFF
--- a/tee-worker/app-libs/stf/src/trusted_call.rs
+++ b/tee-worker/app-libs/stf/src/trusted_call.rs
@@ -30,9 +30,9 @@ use codec::{Decode, Encode};
 use frame_support::{ensure, traits::UnfilteredDispatchable};
 pub use ita_sgx_runtime::{Balance, ConvertAccountId, Index, SgxParentchainTypeConverter};
 use itp_node_api::metadata::{
-	pallet_imp::IMPCallIndexes, pallet_teerex::TeerexCallIndexes, provider::AccessNodeMetadata,
+	pallet_imp::IMPCallIndexes, pallet_system::SystemSs58Prefix, pallet_teerex::TeerexCallIndexes,
+	pallet_vcmp::VCMPCallIndexes, provider::AccessNodeMetadata,
 };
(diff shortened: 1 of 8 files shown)


tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -219,6 +218,24 @@ where
...
+	//
+	// The execution will most likely fail again. However, the state could have been changed already by applying
+	// the state diff from the imported sidechain block. This could cause an inconsistent/mismatching state,
+	// for example, the nonce. See the nonce handling below: we increased the nonce no matter the STF is executed
+	// successfully or not.
+	//
+	// This is probably the reason why the nonce-handling test in `demo_shielding_unshielding.sh` sometimes fails.
+	//
+	// for now we should always return Ok(()) for this function and propagate the exe, at least for
+	// litentry STFs. I believe this is the right way to go, but it still needs more discussions.
```
BillyWooo, line 238: Since we can't r/w storage async, e.g. IDGraphs, etc., I think this is the right way to go. 👍

BillyWooo (approved):
LGTM. Wait for the CI.

jingleizhang:
Insightful thought for identifying the state inconsistent/mismatching issue.

PR #1405, opened 2023-03-06
added integritee stop, start, restart instructions

niteshb3495:
new branch and stop, start, restart changes.


.github/workflows/dev-integritee-restart.yml
```diff
@@ -8,6 +8,7 @@ on:
         description: Select the action
         options: 
         - stop
+        - start
```
BillyWooo: So we still need it?

niteshb3495: yes, I tried adding only restart. but if the service is not running on the server, the restart fails at "stop" action. 
thought to have "start" action as well if the service is not running already.

Kailai-Wang: I do think `systemctl  restart` can be used to start a stopped service.

You can simply ignore the error when stopping the service, either redirect the errors or something like
`sudo systemctl stop integritee.service || true` so that GHA can continue to run

Kailai-Wang (changes requested):

Kailai-Wang:
Another question, how do you update the binaries/docker images when something changed?

Imagine we have pushed a fix into code, how do you re-deploy it with latest changes?

niteshb3495:
> Another question, how do you update the binaries/docker images when something changed?
> 
> Imagine we have pushed a fix into code, how do you re-deploy it with latest changes?

I believe that is been handled in the code, @BillyWooo

.github/workflows/dev-integritee-restart.yml
niteshb3495: Thank you for the input. I am not aware of this option. Let me check and make the changes.

niteshb3495: I have made the changes. Adding `|| true` works fine. 
Thank you for the suggestion, Kai.

BillyWooo:
> > Another question, how do you update the binaries/docker images when something changed?
> > Imagine we have pushed a fix into code, how do you re-deploy it with latest changes?
> 
> I believe that is been handled in the code, @BillyWooo

Yes. For the moment, we need to take care of update docker image & worker binary manually. Because we want the restart process to be as fast as possible. @Kailai-Wang

BillyWooo (approved):
LGTM

Kailai-Wang (approved):

PR #1416, opened 2023-03-07
Check mrenclave and scheduled height before produce blocks(fix #1295)

### LINKED ISSUE #1295: Enclave update [3-2] Check mrenclave and scheduled height before produce blocks
before producing the new block, compare the current height with the scheduled height and make sure the enclave is matched.

DogLi:
1. after sidechain import block, set the block suspender if `block_height >= scheduled block height`
2. if `block_height == scheduled block height`, migration

resolve #1295


Kailai-Wang:
Looks like the TEE tests failed.

jingleizhang:
It was closed because it was wrongly connected to another PR...

Kailai-Wang:
Yea this PR https://github.com/litentry/litentry-parachain/pull/1462 wasn't linked to any issue.

**It's always a good practice to link the issue that will be fixed by a PR**

tee-worker/sidechain/consensus/common/src/block_production_suspension.rs
```diff
@@ -18,75 +18,84 @@
...
+	fn suspend_for_production(&self) -> Result<()>;
 
 	/// Resume block sidechain block production.
-	fn resume(&self) -> Result<()>;
-}
-
-/// Trait to query if sidechain block production is suspended.
-pub trait IsBlockProductionSuspended {
-	fn is_suspended(&self) -> Result<bool>;
+	fn resume_for_production(&self) -> Result<()>;
```
Kailai-Wang: I like the original names TBH 😂 
The trait is already with the name `..Production`, so I'm OK if the member function doesn't have a `production` keyword.

I'm fine with just `suspend_for_sync` and `resume` - if you think `suspend_for_sync` is inaccurate, we could simply use `suspend`

tee-worker/sidechain/consensus/aura/src/block_importer.rs
```diff
@@ -236,9 +245,40 @@ impl<
...
+			let block_suspender = self.block_suspender();
+			let scheduled_mr_enclave = scheduled_enclave.mr_enclave;
+			let scheduled_block_number = scheduled_enclave.sidechain_block_number;
+			let current_mr_enclave = block_suspender
+				.current_mr_enclave()
+				.map_err(|_| ConsensusError::GetMrEnclaveFailed)?;
+			if current_mr_enclave != scheduled_mr_enclave {
+				if scheduled_block_number <= block_number {
+					block_suspender
+						.suspend_for_production()
```
Kailai-Wang: It's suspended here but I don't find any place where it gets resumed, is that correct?

tee-worker/sidechain/consensus/aura/src/block_importer.rs
```diff
@@ -236,9 +245,40 @@ impl<
 	where
 		F: FnOnce(&Self::SidechainState) -> Result<SignedSidechainBlock, ConsensusError>,
 	{
-		self.state_handler
+		let sidechain_block = self
+			.state_handler
 			.execute_on_current(shard, |state, _| verifying_function(state))
-			.map_err(|e| ConsensusError::Other(format!("{:?}", e).into()))?
+			.map_err(|e| ConsensusError::Other(format!("{:?}", e).into()))??;
+		let block_number = sidechain_block.block().header().block_number();
```
Kailai-Wang: the `sidechain_block.block()` is the sidechain block that is going to be imported - I find it questionable: I thought it should be the block that is going to be produced.

Or more in general: why put the logic in `verify_import`?

tee-worker/sidechain/consensus/aura/src/block_importer.rs
```diff
@@ -236,9 +245,40 @@ impl<
...
+		if let Some(scheduled_enclave) = scheduled_enclaves.get_next_scheduled_enclave(block_number)
+		{
+			let block_suspender = self.block_suspender();
+			let scheduled_mr_enclave = scheduled_enclave.mr_enclave;
+			let scheduled_block_number = scheduled_enclave.sidechain_block_number;
+			let current_mr_enclave = block_suspender
+				.current_mr_enclave()
+				.map_err(|_| ConsensusError::GetMrEnclaveFailed)?;
+			if current_mr_enclave != scheduled_mr_enclave {
+				if scheduled_block_number <= block_number {
```
Kailai-Wang: This if-condition is always false if I'm not mistaken.

The `scheduled_block_number` is calculated from `get_next_scheduled_enclave` where the returned block number is greater than the `block_number`:
```
		self.scheduled_enclaves
			.iter()
			.filter_map(
				|(k, v)| {
					if k > &current_side_chain_number {
						Some(v.clone())
					} else {
						None
					}
				},
```

tee-worker/sidechain/consensus/aura/src/block_importer.rs
```diff
@@ -236,9 +245,40 @@ impl<
...
+					block_suspender
+						.suspend_for_production()
+						.map_err(|_| ConsensusError::SetBlockSuspenderFailed)?;
+					warn!("need to update, reason: enclave is outdated");
+				}
+				if scheduled_block_number == block_number {
```
Kailai-Wang: As mentioned above, we probably will never go into this branch.

Kailai-Wang (changes requested):
I didn't go very details - only skimmed through the logic in `tee-worker/sidechain/consensus/aura/src/block_importer.rs`.

I'm not sure why the check logic happens inside a block importer rather than the block proposer. Maybe the argument is we need to import all missing blocks before producing a block but I'm not sure if there's any subtle synchronisation issue. 

The scheduled enclave is something we get from parentchain syncing, not from sidechain state. In that sense I think we don't have to rely on the (imported) sidechain block to do this.

Additionally, is `get_next_scheduled_enclave` correctly implemented?
I think we need such functionality, given a sidechain_block_number, the function should return the expected MRENCALVE. For example, by syncing the parentchain blocks we know that:

|sidechain_block_number|expected_MRENCLAVE|
|-|-|
|0|A|
|20|B|
|25|C|

so:
for input 19 - it should return A => It means we are going to generate block number 19 and we have to check if our MRENCLAVE is equal to A
for input 20 -> it should return B
for input 26 -> it should return C

and so on.

Kailai-Wang:
And yes - having a feature branch in the same repo would be beneficial for the reviewers to check the code locally :)

Kailai-Wang:
Superseded by #1579

Issue #1346, opened 2023-02-17
Resume-worker test times out


Kailai-Wang:
### Context
It happened a few times already, see e.g. https://github.com/litentry/litentry-parachain/actions/runs/4202077590

The logs are a bit spammy so it's hard to find the really useful logs

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

BillyWooo:
Keep notice on #1124

Kailai-Wang:
https://github.com/litentry/litentry-parachain/actions/runs/4651018408

[logs-lit-resume-worker.zip](https://github.com/litentry/litentry-parachain/files/11186114/logs-lit-resume-worker.zip)

arieljbirnbaum:
After #1846, the failing scenarios here seem to always be the two-worker ones; mostly the "kill both workers and restart both" one. Since multiworker mode isn't reliably working at the moment, it's unlikely that we can stabilize the test script.

For now, we should drop the multiworker scenarios. PR #2041 .

Do we have an issue (or a list thereof) to track the work needed for enabling multiworker mode? We should add a note there to reimplement these tests at that moment; I also want to leave a link to that issue from the test script to track the debt.

PR #1419, opened 2023-03-08
websocket reconnect parachain once disconnect(fix #997)

### LINKED ISSUE #997: Can't reconnect to the parentchain once disconnected
I assume this is a different problem other than litentry/litentry-parachain#999, once the worker node lost connection with the parentchain (e.g. due to an unstable network), it can't reconnect.

Maybe related: https://github.com/integritee-network/worker/issues/813

DogLi:
reconnect after the ws disconnect.

testing is in progress

resolve #997

### DIFF
--- a/tee-worker/core/rpc-client/src/ws_client.rs
+++ b/tee-worker/core/rpc-client/src/ws_client.rs
@@ -24,6 +24,8 @@ use log::*;
 use openssl::ssl::{SslConnector, SslMethod, SslStream, SslVerifyMode};
 use parking_lot::Mutex;
 use std::sync::{mpsc::Sender as MpscSender, Arc};
+use std::thread::sleep;
+use std::time::Duration;
 use url::{self};
 use ws::{connect, util::TcpStream, CloseCode, Handler, Handshake, Message, Result, Sender};
 
@@ -66,14 +68,14 @@ impl WsClientControl {
 }
 
 #[derive(Clone)]
-pub struct WsClient {
+pub struct WsClient<'a> {
 	web_socket: Sender,
 	request: String,
-	result: MpscSender<String>,
+	result: &'a MpscSender<String>,
 	do_watch: bool,
 }
 
-impl WsClient {
+impl<'a> WsClient<'a> {
 	/// Connect a web-socket client for multiple request/responses.
 	///
 	/// Control over the connection is done using the provided client control.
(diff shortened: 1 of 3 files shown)


DogLi:
need to restest issue #1124 #1346

tee-worker/core/rpc-client/src/ws_client.rs
```diff
@@ -82,37 +86,43 @@ impl WsClient {
...
 		control: Arc<WsClientControl>,
-	) -> Result<()> {
-		debug!("Connecting web-socket connection with watch");
-		connect(url.to_string(), |out| {
-			control.subscribe_sender(out.clone()).expect("Failed sender subscription");
-			WsClient::new(out, request.to_string(), result.clone(), true)
-		})
+	) {
+		debug!("Connecting web-socket connection with watch to server {url}");
+		loop {
```
Kailai-Wang: I get the idea, but I think we shouldn't put infinite loop in a primitive function of client API. If we need to reconnect, we should handle it in the caller scope.

Btw this function is only used by `DirectClient` which sends requests to the worker-ws via CLI - actually I don't think we need automatic reconnection there

tee-worker/service/src/main.rs
```diff
@@ -694,25 +690,32 @@ fn start_worker<E, T, D, InitializationHandler, WorkerModeProvider>(
...
-		})
-		.unwrap();
-
-	println!("[+] Subscribed to events. waiting...");
-	let timeout = Duration::from_secs(600);
 	loop {
-		if let Ok(events_str) = receiver.recv_timeout(timeout) {
-			let event_bytes = Vec::from_hex(events_str).unwrap();
-			let events = Events::new(metadata.clone(), Default::default(), event_bytes);
+		let (sender, receiver) = channel();
```
Kailai-Wang, line 694: Same as above - I think we are fine with disruption without reconnection, it won't break the loop anyway and it just stops printing events which is not a big deal. We can add logs when disconnected though.

What do you think?

tee-worker/service/src/main.rs
```diff
@@ -787,27 +790,39 @@ fn send_extrinsic(
...
-		}?;
-
-		println!(
-			"[+] Received finalized header update ({}), syncing parent chain...",
-			new_header.number
-		);
-
-		last_synced_header = parentchain_handler.sync_parentchain(last_synced_header)?;
+		let (sender, receiver) = channel();
+		if let Err(e) = parentchain_handler.parentchain_api().subscribe_finalized_heads(sender) {
```
Kailai-Wang, line 796: It should work, thanks. A few Q:

- could it somehow spin off unlimited ws-connection (or background thread)? Is the graceful shutdown already handled?
- we only need some tests for it
- do you think it should be applied to upstream as well?

tee-worker/service/src/main.rs
DogLi: Actually,  if just add add log when disconnected, it will not sync Event, is it ok? if ok, then just print error log. 

For me, I think  it's better to reconnect automatically, or the operator/user needs to always check the log and restart the program to sync Event.

In a normal situation, the websocket should always keep alive. I think the disconnect is always an error situation which is not expected.

tee-worker/service/src/main.rs
DogLi: Actually, this api calls the `substrate-client-api` crate, and in the substrate client api, it spawns a ws connect in a thread. there is no graceful shutdown.

I think it's no need to apply to the upstream, upstream should keep simple, just make reconnect when we use it.

DogLi: if we want to graceful shutdown, just break the `receiver` loop. when the sender sends data, it will fail, the thread will over with an error, then websocket connect will be shutdown.

tee-worker/service/src/main.rs
Kailai-Wang: The events are not used anywhere else except for printing if I read the code correctly.

I'm always concerned about unjoined/unterminated threads - imagine `event_subscriber` thread can't connect to parachain, will it get terminated? If it's always live, then we'll have more and more detached threads.

I'm OK to apply the automatic reconnection only if we can manage the lifecycle of the thread (e.g. we guarantee there's maximum one such subscriber thread at any time)

tee-worker/service/src/main.rs
jingleizhang: btw, the `graceful shutdown` is needed. Just created a new issue https://github.com/litentry/litentry-parachain/issues/1449.

DogLi:
test steps:

1. first start integritee-node
`integritee-node --base-path=/tmp/node_data --rpc-cors=all --ws-external --dev --ws-port 9955`

2. start tee-worker
`./local-setup/launch.py /tmp/test.json remote` where the test.json is:

```json
{
  "workers": [
    {
      "source": "bin",
      "flags": [
        "--clean-reset",
        "-P",
        "2005",
        "-w",
        "2006",
        "-r",
        "3448",
        "-h",
        "4550",
	"--node-port",
	"9955",
        "--running-mode",
        "mock",
        "--enable-mock-server"
      ],
      "subcommand_flags": [
        "--skip-ra",
        "--dev"
      ]
    }
  ]
}
```
3. restart integritee node to see if the tee-worker works ok or not.

jingleizhang:
Tested. But should the second cmd be `./local-setup/launch.py /tmp/test.json remote` ?

jingleizhang (approved):
LGTM

DogLi:
> Tested. But should the second cmd be `./local-setup/launch.py /tmp/test.json remote` ?

yes

PR #1421, opened 2023-03-08
tee imp errors&extrinsic in parallel

### LINKED ISSUE #1257: ts-test sends extrinsic in parallel

### LINKED ISSUE #1374: ts identity test should cover TEE IMP errors
Currently ts-test `test-identity` covers all happy paths in TEE Identity Management Pallet. 

In this issue, ts-test should make TEE IMP [errors](https://github.com/litentry/litentry-parachain/blob/7681fdea23c39d63e16fb4a9cebac3193b722e95/tee-worker/litentry/pallets/identity-management/src/lib.rs#L93) tested with specific Parachain event returned accordingly.

eg:

- verify two same identities to one account, get `IdentityAlreadyVerified` error
- verify an identity to an account but it isn't created before, get `IdentityNotExist` error
- remove a nonexistent identity from an account, get `IdentityNotExist` error 
- remove a challenge code before the code is set, get `ChallengeCodeNotExist` error
- etc

0xverin:
resolves #1257
resolves #1374

### DIFF
--- a/tee-worker/ts-tests/identity.test.ts
+++ b/tee-worker/ts-tests/identity.test.ts
@@ -14,7 +14,7 @@ import {
 import { ethers } from 'ethers';
 import { HexString } from '@polkadot/util/types';
 import { KeyringPair } from '@polkadot/keyring/types';
-import { createErrorIdentity, setErrorUserShieldingKey } from './indirect_error_calls';
+import { createErrorIdentity, setErrorUserShieldingKey, verifyErrorIdentity } from './indirect_error_calls';
 
 const twitterIdentity = <LitentryIdentity>{
     Web2: <Web2Identity>{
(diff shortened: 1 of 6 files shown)


0xverin:
Does [this](https://github.com/litentry/litentry-parachain/blob/1374-tee-imp-errors/tee-worker/ts-tests/identity.test.ts#L290) cover `remove a challenge code before the code is set` case? I use an account that has never been `createIdentity` and then removeIdentity,I think the account don't set challenge code before.

Kailai-Wang (approved):
Looks fine, thanks.

If we have more test-cases than written in #1374 we will raise a separate PR.

For `code not found` there's no such `ChallengeCodeNotExist` as `StfError`, will address it later so that we just need to check the enum other than the internal string message.

PR #1378, opened 2023-02-27
1152 batch calls

BillyWooo:
Add batch_all support for current IMP extrinsics:
- SetUserShieldingKey
- CreateIdentity
- VerifyIdentity
- RemoveIdentity

Add ts-test
 

Resolves #1152

PR #1428, opened 2023-03-10
Revert "1152 batch calls"

jingleizhang:
Reverts litentry/litentry-parachain#1378


Kailai-Wang (approved):
Let's commit it first to not block any other PRs

PR #1443, opened 2023-03-13
Add port check prior to parachain launch

### LINKED ISSUE #1424: Add pre-launch check before starting local networks
### Context

We can have multiple people working on the same server.

It's better to have some pre-launch checks (e.g. ports, processes) so that we don't accidentally break others' work if they are working on the same host. 

---

Kailai-Wang:
resolves #1424 

Please let me know if it doesn't suffice


BillyWooo (approved):
LGTM

PR #1454, opened 2023-03-14
1441 validate json schema

### LINKED ISSUE #1441: VC: validation in ts-test according to the JSON Schema
The generic VC JSON Schema is [here](https://www.notion.so/web3builders/VC-Verification-e21fcb4cd9004672ac25b44609c8a86b?pvs=4#b9fa04eda3ac4299b69822092025cf26).

The validation in ts-test needs to be added.

0xverin:
as the title
resolves #1441

### DIFF
--- a/tee-worker/ts-tests/indirect_error_calls.ts
+++ b/tee-worker/ts-tests/indirect_error_calls.ts
@@ -140,7 +140,7 @@ export async function removeErrorIdentities(
         });
     }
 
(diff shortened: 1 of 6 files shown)


jingleizhang (approved):
Looks good! The issue of `missing await` we discussed on daily is fixed too.

If the outputs of `vcManagement.VCIssued` are clear text, it could be better for checking.
image

Let's improve this in the later PRs.

PR #185, opened 2021-10-26
Add bootnode telemtry to genesis

Kailai-Wang:
resolves https://github.com/litentry/research-force/issues/75

Staging deployment to be followed.

PR #1456, opened 2023-03-14
build_assertion error propagate to parachain

### LINKED ISSUE #1392: build_assertion error doesn't propagate to parachain
### Context

Repro-case:

- set user shielding key for "//Bob"
- request vc (e.g. A1) for "//Alice"

In tee-worker I see:
```
[2023-03-02T09:29:30Z DEBUG ita_stf::trusted_call_litentry] who "d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d", assertion A1
[2023-03-02T09:29:30Z ERROR ita_stf::trusted_call_litentry] user shielding key is missing, "d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d", A1
[2023-03-02T09:29:30Z ERROR itp_stf_executor::executor] Stf execute failed: AssertionBuildFail
```

which is good.
However, such error never arrives in parachain.

When I check the code, we never properly handle the errors:
https://github.com/litentry/litentry-parachain/blob/19951d385f70f3cb676653e22008638c4abdc234/tee-worker/app-libs/stf/src/trusted_call.rs#L601

Please also have ts-tests for this

---

outofxxx:
resolved #1392 

* remove `VCNotExists` from Event
* ts-test will check extrinsic' status, so remove it.

### DIFF
--- a/pallets/vc-management/src/lib.rs
+++ b/pallets/vc-management/src/lib.rs
@@ -93,7 +93,6 @@ pub mod pallet {
 		VCDisabled { index: VCIndex },
 		// a VC is revoked on chain
 		VCRevoked { index: VCIndex },
-		VCNotExist { index: VCIndex },
 		// event that should be triggered by TEECallOrigin
 		// a VC is just issued
 		VCIssued { account: T::AccountId, index: VCIndex, vc: AesOutput },
--- a/pallets/vc-management/src/lib.rs
+++ b/pallets/vc-management/src/lib.rs
@@ -167,25 +167,13 @@ pub mod pallet {
 		pub fn disable_vc(origin: OriginFor<T>, index: VCIndex) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
 
-			if !VCRegistry::<T>::contains_key(index) {
-				Self::deposit_event(Event::VCNotExist { index });
-				return Ok(().into())
-			}
-
 			VCRegistry::<T>::try_mutate(index, |context| {
-				match context.take() {
-					Some(mut c) => {
-						ensure!(who == c.subject, Error::<T>::VCSubjectMismatch);
-						ensure!(c.status == Status::Active, Error::<T>::VCAlreadyDisabled);
-						c.status = Status::Disabled;
-						*context = Some(c);
-						Self::deposit_event(Event::VCDisabled { index });
-					},
-					None => {
-						Self::deposit_event(Event::VCNotExist { index });
-					},
-				}
-
+				let mut c = context.take().ok_or(Error::<T>::VCNotExist)?;
+				ensure!(who == c.subject, Error::<T>::VCSubjectMismatch);
+				ensure!(c.status == Status::Active, Error::<T>::VCAlreadyDisabled);
+				c.status = Status::Disabled;
+				*context = Some(c);
+				Self::deposit_event(Event::VCDisabled { index });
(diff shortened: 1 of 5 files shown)


pallets/vc-management/src/lib.rs
```diff
@@ -166,6 +166,12 @@ pub mod pallet {
 		#[pallet::weight(195_000_000)]
 		pub fn disable_vc(origin: OriginFor<T>, index: VCIndex) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
+
+			if !VCRegistry::<T>::contains_key(index) {
+				Self::deposit_event(Event::VCNotExist { index });
+				return Ok(().into())
+			}
+
```
Kailai-Wang: I don't think we need this, `try_mute` below already does the job.

Btw I think we should emit an error instead of the event in line #185 - F/E will be able to detect this as it's synchronous on-chain logic

pallets/vc-management/src/lib.rs
```diff
@@ -189,6 +195,11 @@ pub mod pallet {
 		pub fn revoke_vc(origin: OriginFor<T>, index: VCIndex) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
 
+			if !VCRegistry::<T>::contains_key(index) {
+				Self::deposit_event(Event::VCNotExist { index });
+				return Ok(().into())
+			}
+
```
Kailai-Wang: same as above - `ok_or` below does the job already.

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
```diff
@@ -175,7 +177,11 @@ where
 			credential.add_proof(&sig, credential.issuance_block_number, &enclave_account);
 
 			if credential.validate().is_err() {
-				error!("failed to validate credential");
+				let error_msg = "failed to validate credential";
+				error!("	[BuildAssertion] : {}", error_msg);
+
+				self.on_failure(to_vcmp_error(error_msg));
```
Kailai-Wang, line 183: calling `on_failure` within `on_success` sounds strange, can we have better handling? (ignore this if we can't)

Kailai-Wang (changes requested):

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
outofxxx: At first, I wanted to write another method to deal with such errors, but later I felt a little redundant because I felt that `on_failure` needs to handle all errors from `on_process` and `on_success`, so this method is not written, but directly call `on_failure` method in `on_success`. Please correct me if I understand this wrong.

Kailai-Wang: Thanks - if we don't have better ways to handle it, let's just leave it as it is right now :)

pallets/vc-management/src/lib.rs
0xverin: https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-get-the-decoded-enum-for-an-extrinsicfailed-event

Kailai-Wang (approved):
Looks good to me, thanks

PR #1462, opened 2023-03-15
sleep for for a while before the event websocket reconnect(fix #1461)

### LINKED ISSUE #1461: need to sleep for a while before the event websocket reconnection
the event websocket reconnect in a loop, need to sleep for a while before reconnect.

#997

DogLi:


jingleizhang (approved):
Thanks!

Kailai-Wang:
Thanks! Shall we add a comment (either in the code or in the related issue) about why we need to do that?

And let's link the issue that will be fixed by this PR

jingleizhang:
When the connection breaks, the event logs(`[!] event websocket disconnected, try to connect again`) are too many, so that  it sleeps 1 sec here.

PR #1463, opened 2023-03-15
build_assertion error case test

### LINKED ISSUE #1392: build_assertion error doesn't propagate to parachain

0xverin:
resolves #1392


tee-worker/ts-tests/identity.test.ts
```diff
@@ -310,7 +315,9 @@ describeLitentry('Test Identity', (context) => {
 
     step('remove prime identity NOT allowed', async function () {
         // create substrate identity
-        const [resp_substrate] = (await createIdentities(context, context.defaultSigner[0], aesKey, true, [substrateIdentity])) as IdentityGenericEvent[];
+        const [resp_substrate] = (await createIdentities(context, context.defaultSigner[0], aesKey, true, [
+            substrateIdentity,
+        ])) as IdentityGenericEvent[];
```
Kailai-Wang: please feel free to adjust `.prettierrc` if you believe the old one has better readability

tee-worker/ts-tests/identity.test.ts
```diff
@@ -389,6 +393,38 @@ describeLitentry('Test Identity', (context) => {
...
+
+        resp_not_created_identities.map((item: any) => {
+            const result = item.toHuman().data.reason;
+            assert(
+                result.search('IdentityNotExist') !== -1,
+                'remove twitter should fail with reason `IdentityNotExist`'
+            );
+        });
+    });
+
```
Kailai-Wang: 1. This step seems identical with the previous one, we should have comment at least about why we repeat it
2. even though we repeat it - we should have a different step name
3. we should avoid duplicate code on step-level
4. we should avoid duplicate code inside a step: I don't see any difference between `const resp_not_exist_identities...` and `const resp_not_created_identities...` except for the signer. Can't we use a function to wrap it up?

tee-worker/ts-tests/indirect_error_calls.ts
```diff
@@ -149,3 +150,79 @@ export async function removeErrorIdentities(
...
+    context: IntegrationTestContext,
+    signer: KeyringPair,
+    listening: boolean,
+    indexList: HexString[]
+): Promise<string[] | undefined> {
+    let txs: TransactionSubmit[] = [];
+
+    for (let k = 0; k < indexList.length; k++) {
+        const tx = context.substrate.tx.vcManagement.revokeVc(indexList[k]);
+        const nonce = await context.substrate.rpc.system.accountNextIndex(signer.address);
```
Kailai-Wang: A small improvement: I think we can move it out of the loop - we'll get the same result each time anyway.

tee-worker/ts-tests/indirect_error_calls.ts
```diff
@@ -149,3 +150,79 @@ export async function removeErrorIdentities(
...
+    context: IntegrationTestContext,
+    signer: KeyringPair,
+    listening: boolean,
+    indexList: HexString[]
+): Promise<string[] | undefined> {
+    let txs: TransactionSubmit[] = [];
+
+    for (let k = 0; k < indexList.length; k++) {
+        const tx = context.substrate.tx.vcManagement.disableVc(indexList[k]);
+        const nonce = await context.substrate.rpc.system.accountNextIndex(signer.address);
```
Kailai-Wang: see below

tee-worker/ts-tests/vc.test.ts
Kailai-Wang, line 1: One of my general feelings here is (you might have other opinions):

We want to cover as many test cases as possible, but we don't want to over-test them. For example, we want to test the error handling of the case where a user requests a VC without setting the shielding key. This is a good test case, but we don't have to repeat it for all VC types that we have - because there isn't any difference in terms of such error handling for different assertion enums (if we have one someday, then we should test it).

So it doesn't make much difference **you test A1 only** vs **you test A1 - A11**, one VC type is enough. Making it a vector of VCs impairs the code readability and prolongs the whole test time unnecessarily.

The same applies to the **identity test**.

Kailai-Wang (changes requested):
I get the idea behind the tests, but I have a few comments about the coding (style).

tee-worker/ts-tests/identity.test.ts
0xverin: The difference between the two removeErrorIdentities is that there is no challengeCode for `context.defaultSigner[2]`,so for the convenience of testing, I directly change the signer and repeat it.

tee-worker/ts-tests/vc.test.ts
0xverin: My idea is to select a random assertion to trigger the specified error, I don't know the difference between them, I worry that it will not cover all case(maybe A1 trigger,but A2 not trigger)

tee-worker/ts-tests/identity.test.ts
Kailai-Wang: Yes I know the signer is different :) I mean we should point it out in the comment and try to remove duplicate code.

tee-worker/ts-tests/vc.test.ts
Kailai-Wang: > select a random assertion

But if I understand the code correctly, we iterate over all assertions here (instead of choosing one randomly). And yes there's no difference in terms of that specific error handling, we can ask eric or zhouhui if we are unsure.

I'm mainly concerned about the execution time - it probably won't be such a problem if a step takes 5s. Imagine we have 100 assertions later, we can't afford to test them all for **a generic handling** that applies to all assertion types (we should still do individual test if something differs)

0xverin: I have confirmed with @zhouhuitian that only one test is needed, and all the procedures are the same,so only use A1 to triggers the error events.

tee-worker/ts-tests/identity.test.ts
```diff
@@ -308,8 +291,8 @@ describeLitentry('Test Identity', (context) => {
         assertIdentityRemoved(context.defaultSigner[1], substrate_extension_identity_removed);
     });
 
-    step('remove prime identity NOT allowed', async function () {
-        // create substrate identity
+     step('remove prime identity NOT allowed', async function () {
+           // create substrate identity
```
Kailai-Wang: Indention problem?

tee-worker/ts-tests/identity.test.ts
```diff
@@ -355,57 +338,49 @@ describeLitentry('Test Identity', (context) => {
         const events = await listenEvent(context.substrate, 'identityManagement', ['StfError']);
         expect(events.length).to.be.equal(1);
         const result = events[0].method as string;
-    });
+     });
```
Kailai-Wang: Indention

tee-worker/ts-tests/identity.test.ts
```diff
@@ -355,57 +338,49 @@ describeLitentry('Test Identity', (context) => {
...
 
     step('remove error identities', async function () {
-        //remove a nonexistent identity from an account
-        const resp_not_exist_identities = (await removeErrorIdentities(context, context.defaultSigner[0], true, [
-            twitterIdentity,
-            ethereumIdentity,
-            substrateIdentity,
-        ])) as string[];
+        const identites = [twitterIdentity, ethereumIdentity, substrateIdentity];
```
Kailai-Wang: typo: `identities`

tee-worker/ts-tests/indirect_error_calls.ts
```diff
@@ -149,3 +155,73 @@ export async function removeErrorIdentities(
...
+        const key = keys[index];
+        const tx = context.substrate.tx.vcManagement.requestVc(mrEnclave, {
+            [key]: assertion[key as keyof Assertion],
+        });
+        let newNonce = nonce.toNumber() + index;
+        txs.push({ tx, nonce: newNonce });
+    }
+
+   
+
```
Kailai-Wang: too many line breakers :(

tee-worker/ts-tests/vc.test.ts
```diff
@@ -17,6 +19,10 @@ const assertion = <Assertion>{
     A10: [10],
     A11: [10],
 };
+
+
+// it doesn't make much difference test A1 only vs test A1 - A11, one VC type is enough.
+//So only use A1 to trigger the wrong event
```
Kailai-Wang: extra line breakers

comments: first line with space, second line without

Kailai-Wang (approved):
Looks better, thanks.

There're many formatting problems though, please fix it before committing.

Btw I don't think these .ts files went through prettier.

Kailai-Wang:
And please resolve any conversation that is already addressed.

0xverin:
I finally found where my local formatter configuration was wrong, my formatter was conflicting.:(

PR #1478, opened 2023-03-16
IDGraph adjustment

### LINKED ISSUE #1442: What happens when an IDGraph bloats up?
### Context

We included the encrypted IDGraph in the returning event in https://github.com/litentry/litentry-parachain/pull/1080

But what if a user links too many identities and the IDGraph grows into a huge object? Maybe you can't even submit extrinsic anymore.

We need to:
- test when it starts to be problematic
- devise a plan to fix it

@jonalvarezz maybe we have to discuss the necessity of having such IDGraph in the returning event again.

---

Kailai-Wang:
resolves #1442 

This PR:
- caps the maximum number of identities in returned IDGraph at 20 (tested by unittest in IMT)
- only returns IDGraph for `verify_identity`
- prettier ts
- replace `context.substrate` with `context.api` as it's more conventional and `substrate` is a bit ambiguous

TODO: real ts-tests for linking 20+ identities => mainly to see how large the IDGraph would be.


BillyWooo (approved):
LGTM.  CI should pass.

PR #1479, opened 2023-03-16
1471 check shielding key inside pallet

### LINKED ISSUE #1471: Move the check logic of user-shielding-key-set into pallet
### Context

Imagine a user calls `create_identity` without setting up the shielding key, F/E will get the error "InvallidUserShieldingKey", but the identities are already created in the sidechain

This is because we execute the pallet logic first and don't check the shielding key until we want to send back the result later:
https://github.com/litentry/litentry-parachain/blob/a6b7e2ed7fe56c65b12791db092e0516d3e0bc49/tee-worker/app-libs/stf/src/trusted_call.rs#L488-L498

This is obviously incorrect. We should move the shielding key check either into the pallet or before the pallet call. I prefer the former solution.

related:
- https://github.com/litentry/litentry-parachain/issues/1223 (this is a state mismatch)
- https://github.com/litentry/litentry-parachain/issues/1250 (ts-test should also check the actual storage of the sidechain)

---

BillyWooo:
as title, move the shielding key checking inside pallet extrinsics. Otherwise it would be a bug to check shielding key after extrinsic executed.

resolved #1471

### DIFF
--- a/tee-worker/app-libs/stf/src/trusted_call.rs
+++ b/tee-worker/app-libs/stf/src/trusted_call.rs
@@ -508,11 +508,7 @@ where
 								aes_encrypt_default(&key, &code.encode()),
 							)));
 						} else {
-							add_call_from_imp_error(
-								calls,
-								node_metadata_repo,
-								IMPError::InvalidUserShieldingKey,
-							);
+							error!("Can't create identity: InvalidUserShieldingKey");
 						}
 					},
 					Err(e) => {
@@ -544,11 +540,7 @@ where
 								aes_encrypt_default(&key, &identity.encode()),
 							)));
 						} else {
-							add_call_from_imp_error(
-								calls,
-								node_metadata_repo,
-								IMPError::InvalidUserShieldingKey,
-							);
+							error!("Can't remove identity: InvalidUserShieldingKey");
 						}
 					},
 					Err(e) => {
@@ -608,11 +600,7 @@ where
 								aes_encrypt_default(&key, &id_graph.encode()),
 							)));
 						} else {
-							add_call_from_imp_error(
-								calls,
-								node_metadata_repo,
-								IMPError::InvalidUserShieldingKey,
-							);
+							error!("Can't verify identity: InvalidUserShieldingKey");
 						}
 					},
 					Err(e) => {
--- a/tee-worker/app-libs/stf/src/trusted_call.rs
+++ b/tee-worker/app-libs/stf/src/trusted_call.rs
@@ -495,19 +495,24 @@ where
 				) {
 					Ok(code) => {
 						debug!("create_identity_runtime {} OK", account_id_to_string(&who));
-						// For sure to get key here. It's already checked in pallet:
-						// ita_sgx_runtime::IdentityManagementCall::<Runtime>::create_identity
-						let key = IdentityManagement::user_shielding_keys(&who).unwrap();
-						let id_graph =
-							ita_sgx_runtime::pallet_imt::Pallet::<Runtime>::get_id_graph(&who);
-						calls.push(OpaqueCall::from_tuple(&(
-							node_metadata_repo
-								.get_from_metadata(|m| m.identity_created_call_indexes())??,
-							SgxParentchainTypeConverter::convert(who),
-							aes_encrypt_default(&key, &identity.encode()),
-							aes_encrypt_default(&key, &code.encode()),
-							aes_encrypt_default(&key, &id_graph.encode()),
-						)));
+						if let Some(key) = IdentityManagement::user_shielding_keys(&who) {
+							let id_graph =
+								ita_sgx_runtime::pallet_imt::Pallet::<Runtime>::get_id_graph(&who);
+							calls.push(OpaqueCall::from_tuple(&(
+								node_metadata_repo
+									.get_from_metadata(|m| m.identity_created_call_indexes())??,
+								SgxParentchainTypeConverter::convert(who),
+								aes_encrypt_default(&key, &identity.encode()),
+								aes_encrypt_default(&key, &code.encode()),
+								aes_encrypt_default(&key, &id_graph.encode()),
+							)));
+						} else {
+							add_call_from_imp_error(
+								calls,
+								node_metadata_repo,
+								IMPError::InvalidUserShieldingKey,
+							);
+						}
 					},
 					Err(e) => {
 						debug!(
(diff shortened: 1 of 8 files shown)


tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -495,24 +495,19 @@ where
...
-						} else {
-							add_call_from_imp_error(
-								calls,
-								node_metadata_repo,
-								IMPError::InvalidUserShieldingKey,
-							);
-						}
+						// For sure to get key here. It's already checked in pallet:
+						// ita_sgx_runtime::IdentityManagementCall::<Runtime>::create_identity
+						let key = IdentityManagement::user_shielding_keys(&who).unwrap();
```
Kailai-Wang: Even if we are pretty sure in the runtime that the key should exist, we should still avoid using `unwrap ()` (you never know where could go wrong - e.g. sidechain corruption). 

No panic in enclave is/will be a must.

tee-worker/ts-tests/identity.test.ts
Kailai-Wang, line 1: It's a bit surprising that we didn't have test cases for the missing shielding key case.

We should have at least one case for it - or 3 in a batch.

Kailai-Wang (changes requested):

tee-worker/app-libs/stf/src/trusted_call.rs
BillyWooo: add check back.

tee-worker/ts-tests/identity.test.ts
BillyWooo: add new test

tee-worker/litentry/primitives/src/identity.rs
```diff
@@ -84,6 +85,18 @@ impl SubstrateNetwork {
...
+			Self::TestNet => 13,
+		}
+	}
+
+	pub fn get_network(prefix: u16) -> Self {
+		match prefix {
+			0 => Self::Polkadot,
+			2 => Self::Kusama,
+			31 => Self::Litentry,
+			131 => Self::Litmus,
```
Kailai-Wang, line 97: We need to use `LitentryRococo` for this one: we are using rococo for our tests and initial launch.

Whether we will have the protocol on Litmus is unsettled.

But I also understand it might be unpleasant to have the same prefix for Litmus and rococo => I'll check if we can modify this in the SS58Registry.

tee-worker/litentry/primitives/src/identity.rs
```diff
@@ -84,6 +85,18 @@ impl SubstrateNetwork {
 			Self::Litentry => 31,
 			Self::Litmus => 131,
 			Self::Khala => 30,
+			Self::TestNet => 13,
+		}
+	}
+
+	pub fn get_network(prefix: u16) -> Self {
```
Kailai-Wang: I think either `new` or `from_ss58_prefix` would be better as a name :)

tee-worker/ts-tests/identity.test.ts
```diff
@@ -331,7 +343,8 @@ describeLitentry('Test Identity', (context) => {
         const substratePrimeIdentity = <LitentryIdentity>{
             Substrate: <SubstrateIdentity>{
                 address: `0x${Buffer.from(context.defaultSigner[0].publicKey).toString('hex')}`,
-                network: 'Litentry',
+                // When testing with integritee-node, change network to: TestNet
+                network: 'Litmus',
```
Kailai-Wang, line 347: See above.

If Integritee-node has a prefix - we can integrate it too.

What's better is we have a helper function in ts-test that gets the network from substrate node metadata (ss58Prefix) - similar to what we did in rust.

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -200,6 +202,8 @@ pub mod pallet {
 			parent_ss58_prefix: u16,
 		) -> DispatchResult {
 			T::ManageOrigin::ensure_origin(origin)?;
+			ensure!(Self::user_shielding_keys(&who).is_some(), Error::<T>::InvalidUserShieldingKey);
```
Kailai-Wang, line 205: Maybe we can save some error checking here: https://github.com/litentry/litentry-parachain/blob/32927fe47122c88e088c052f6ba7d3c08398d852/tee-worker/app-libs/stf/src/trusted_call.rs#L511-L514

e.g. just print the error logs instead.

tee-worker/litentry/pallets/identity-management/src/tests.rs
```diff
@@ -46,6 +46,13 @@ fn set_user_shielding_key_works() {
 #[test]
 fn create_identity_works() {
 	new_test_ext().execute_with(|| {
+		let shielding_key: UserShieldingKeyType = [0u8; USER_SHIELDING_KEY_LEN];
+		assert_ok!(IMT::set_user_shielding_key(
+			RuntimeOrigin::signed(ALICE),
```
Kailai-Wang: ALICE happens to work here because it's the same as the `ManagerOrigin` - maybe we should use either `One` here or modify the ManagerOrigin in `mock.rs` also to `ALICE`

tee-worker/litentry/pallets/identity-management/src/tests.rs
```diff
@@ -46,6 +46,13 @@ fn set_user_shielding_key_works() {
...
+			BOB,
+			shielding_key.clone()
+		));
+
```
Kailai-Wang: If this needs to be done for every case, we can add it to `new_test_ext` in `mock.rs` to avoid duplicate code

tee-worker/litentry/pallets/identity-management/src/tests.rs
Kailai-Wang, line 1: Do we have a unit test case for creating identity without setting user shielding key first? I mean unit test, not e2e (ts) test

Kailai-Wang (changes requested):
A few small complaints :(

tee-worker/litentry/primitives/src/identity.rs
BillyWooo: Yeah. Let's keep a note for this point. 'Cause I also found out the value I got when debugging is 131 => Litmus. But I debug with rococo.

tee-worker/ts-tests/identity.test.ts
BillyWooo: The value is `13` when testing with integritee-node. Not sure whether this is their registered  ss58Prefix. Didn't check.  To have such a helper function is a nice idea. Will try.

tee-worker/litentry/pallets/identity-management/src/lib.rs
BillyWooo: 👌

tee-worker/litentry/pallets/identity-management/src/tests.rs
BillyWooo: Just added one.

tee-worker/litentry/pallets/identity-management/src/tests.rs
BillyWooo: 👌

tee-worker/litentry/pallets/identity-management/src/tests.rs
BillyWooo: not applicable any more.

tee-worker/litentry/pallets/identity-management/src/mock.rs
```diff
@@ -119,12 +125,26 @@ pub fn bob_web3_identity() -> Identity {
 	Identity::Substrate { network: SubstrateNetwork::Litentry, address: bob_key_hex.into() }
 }
 
+pub fn new_test_ext_wo_shielding_key() -> sp_io::TestExternalities {
```
Kailai-Wang: maybe we can simply add a boolean param to new_test_ext: `new_test_ext(set_shielding_key: bool)` to avoid duplicated code

tee-worker/litentry/primitives/src/identity.rs
Kailai-Wang: We can leave it now, but let's bear in mind the test in another `prime_id` test might fail because it expects the prime_id to be of rococo network type

Kailai-Wang (approved):
Thanks, a small thing tho

tee-worker/litentry/pallets/identity-management/src/mock.rs
BillyWooo: 👍  should be.

PR #1507, opened 2023-03-22
1250 check sidechain storage

### LINKED ISSUE #1250: Check sidechain storage value in ts-test
### Context
Currently, we only check events from parachain  in ts-test,  which is not enough. We also need to check whether the storage in sidechain has changed. This ensures that sidechain storage is consistent with parentchain storage

### LINKED ISSUE #1470: Use clear names for the signer
### Context

The current array like defaultSigner is very confusing. When I read something like: `context.defaultSigner[0], context.defaultSigner[1]...` it's hard to figure out which signer it is.

Would it make sense to change it to map-like structure? Like `defaultSigner['alice']`. Maybe it's even better to have something like `signer_alice()` or `get_signer('//Alice')` as it's constructed (globally) from the known account anyway - we don't have to pass them around all the time as `context` members.

@Verin1005 what do you think?

---

0xverin:
as the title;
btw resolves #1470;

There are quite a lot of modifications this pr, which includes changes to the directory structure of the code. These changes are made to the following files:
  `common/call.ts` (rpc call)
  `common/helpers.ts` (format and setup)
  `common/transactions.ts` (transactions utils)
  `common/utils` (some functional functions)

I have formatted all the ts code.

Regarding the content of this PR, the main code changes are located [here](https://github.com/litentry/litentry-parachain/blob/1250-check-sidechain-storage/tee-worker/ts-tests/common/utils.ts#L308).
Usage [example](https://github.com/litentry/litentry-parachain/blob/1250-check-sidechain-storage/tee-worker/ts-tests/identity.test.ts#L132)

The main points of the test case are as follows:

1. Before calling setShieldingKey, query the shieldingKey of sidechain storage. The return value should be an empty hex (`0x`). After calling setShieldingKey, query the shieldingKey of sidechain storage again, and the value should be equal to aesKey (tested in both vc-test and identity-test).

2.  Similarly, idgraph and challengecode are queried before and after the steps of createIdentity, verifyIdentity, and removeIdentity (only in identity-test).


tee-worker/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
BillyWooo: Is it necessary to commit this lock file?

0xverin: no, misoperation.

tee-worker/ts-tests/common/call.ts
```diff
@@ -0,0 +1,41 @@
...
+export async function getMetadata(wsClient: WebSocketAsPromised, api: ApiPromise): Promise<any> {
+    let request = { jsonrpc: '2.0', method: 'state_getMetadata', params: [], id: 1 };
+    let respJSON = await sendRequest(wsClient, request, api);
+    const registry = new TypeRegistry();
+    const metadata = new Metadata(registry, respJSON.value);
+    registry.setMetadata(metadata);
+    return metadata;
+}
+
+export async function getSideChainStroage(
```
Kailai-Wang, line 31: typo: `..Storage`

tee-worker/ts-tests/common/utils.ts
```diff
@@ -397,3 +304,145 @@ export async function checkFailReason(
...
+                theKeyItemAppliedHasher = identity(theKeyItemEncoded);
+            } else {
+                throw new Error(`The hasher ${theHasher} is not support.`);
+            }
+            storageKey = u8aConcat(storageKey, theKeyItemAppliedHasher);
+        }
+    }
+    return storageKey;
+}
+export async function buildStorageData(
```
Kailai-Wang, line 358: We could have better namings IMO: this one should be `buildStorageKey`, and the previous one is something like `buildStorageKeyHelper` or `buildStorageKeyInternal`

tee-worker/ts-tests/common/utils.ts
```diff
@@ -397,3 +304,145 @@ export async function checkFailReason(
...
+        storageKey = buildStorageKey(metadata, prefix, method, key, hashers, input);
+        valueType = metadata.registry.createLookupType(value);
+    } else {
+        throw new Error('Only support plain and map type');
+    }
+    console.debug(`storage key: ${u8aToHex(storageKey)}`);
+    return u8aToHex(storageKey);
+}
+
+export async function checkUserShieldingKeys(
```
Kailai-Wang, line 387: `get..` might be better than `check...` as we are getting the value of a certain storage key, same below

tee-worker/ts-tests/common/type-definitions.ts
```diff
@@ -184,9 +184,9 @@ export type IntegrationTestContext = {
     api: ApiPromise;
     teeShieldingKey: KeyObject;
     mrEnclave: HexString;
-    defaultSigner: KeyringPair[];
-    //@todo add type
     ethersWallet: any;
+    substrateWallet: any,
+    metaData: Metadata
```
Kailai-Wang, line 189: I think `metadata` is a single word :)

tee-worker/ts-tests/identity.test.ts
```diff
@@ -123,67 +124,123 @@ const discordValidationData = <LitentryValidationData>{
 
 describeLitentry('Test Identity', (context) => {
     const aesKey = '0x22fc82db5b606998ad45099b7978b5b4f9dd4ea6017e57370ac56141caaabd12';
+
     const errorAseKey = '0xError';
     const errorCiphertext = '0xError';
     var signature_ethereum;
     var signature_substrate;
+    step('check user sidechain storage before create', async function () {
+        const encode = context.api.createType('LitentryIdentity', twitterIdentity).toHex();
```
Kailai-Wang, line 133: I think `encode` is not a good name - something like `identity_hex` would be better, or even `identity` when it's local. Same below

tee-worker/ts-tests/common/utils.ts
```diff
@@ -397,3 +304,145 @@ export async function checkFailReason(
...
+    context: IntegrationTestContext,
+    pallet: string,
+    method: string,
```
Kailai-Wang, line 390: We can hardcode the `pallet` and `method` names - they are unlikely to change, only the `address` can verify (i.e. as per which address I want to check).

It will reduce a lot of duplicate code. Same for challengeCode and idGraphs.

This function is specific to our code anyway.

tee-worker/ts-tests/identity.test.ts
```diff
@@ -247,26 +321,42 @@ describeLitentry('Test Identity', (context) => {
...
-        assertIdentityVerified(context.defaultSigner[0], ethereum_identity_verified);
-        assertIdentityVerified(context.defaultSigner[0], substrate_identity_verified);
+        assertIdentityVerified(context.substrateWallet.alice, twitter_identity_verified);
+        assertIdentityVerified(context.substrateWallet.alice, ethereum_identity_verified);
+        assertIdentityVerified(context.substrateWallet.alice, substrate_identity_verified);
         //Bob
-        assertIdentityVerified(context.defaultSigner[1], substrate_extension_identity_verified);
+        assertIdentityVerified(context.substrateWallet.bob, substrate_extension_identity_verified);
     });
+    step('check IDGraph after createIdentity', async function () {
```
Kailai-Wang, line 338: We also need to check the challenge code is removed after the identity is verified :)

tee-worker/ts-tests/common/utils.ts
```diff
@@ -397,3 +304,145 @@ export async function checkFailReason(
...
+        jsonrpc: '2.0',
+        method: 'state_getStorage',
+        params: [base58mrEnclave, storageKey],
+        id: 1,
+    };
+    let resp = await sendRequest(context.tee, request, context.api);
+
+    return resp.value;
+}
+export async function checkUserChallengeCode(
```
Kailai-Wang, line 407: A very small naming issue:
Why it has `User` in the name? If it has to match the storage method name, it should be without `User` and with `s`: `checkChallengeCodes`. If `User` simply means "user's", then the IDGraph should be named similarly, like `checkUserIDGraphs`

tee-worker/ts-tests/common/utils.ts
```diff
@@ -397,3 +304,145 @@ export async function checkFailReason(
...
+    let resp = await sendRequest(context.tee, request, context.api);
+    return resp.value;
+}
+
+export async function checkIDGraph(
+    context: IntegrationTestContext,
+    pallet: string,
+    method: string,
+    address: HexString,
+    identity: HexString
```
Kailai-Wang, line 433: A more advanced feature is to make `identity` optional, very often we need to query the `Vec<(LitentryIdentity, IdentityContext)>` of a given `address`.

It makes the verification as well, as it can:
- verify the length of IDGraph
- verify the element in batch

tee-worker/ts-tests/identity.test.ts
```diff
@@ -123,67 +124,123 @@ const discordValidationData = <LitentryValidationData>{
...
+            context,
+            'IdentityManagement',
+            'UserShieldingKeys',
+            u8aToHex(context.substrateWallet.alice.addressRaw)
+        );
+        console.log('resp_shieldingKey', resp_shieldingKey);
+        assert.equal(resp_shieldingKey, aesKey, 'resp_shieldingKey should be equal aesKey after set');
+    });
+
+    step('check idgraph from sidechain storage before create', async function () {
```
Kailai-Wang, line 188: We need to check here the prime_id already exists in the IDGraph
cc @BillyWooo

Kailai-Wang (approved):
Well done, thanks! 👍 

I mainly checked the utils and identity-test.

I approved it - some follow-up chores though, see comments. But please use another 1-2 PRs for it :)

Kailai-Wang:
btw my prettier says otherwise:
image

tee-worker/ts-tests/common/utils.ts
0xverin: get it!

PR #1508, opened 2023-03-22
integritee service stop/restart update

### LINKED ISSUE #1486: Support restart tee-worker only without pruning on staging
### Context

Currently when restarting tee-staging, everything is reset (relaychain + parachain + workers).

Sometimes we only want to restart the workers without pruning the state, e.g. when there's a temporary network issue and the worker can't connect to the parachain (we once got DNS error), or when the non-enclave part panics out.

We'd like to provide an additional restart-option to only restart the worker **without pruning the sidechain database**. It helps test the worker resuming too.

Please mind the start parameter difference, if any.

---

niteshb3495:
after decoupling the integritee service with docker containers, the service needs to be stopped/restarted differently

### DIFF
--- a/.github/workflows/dev-sgx-restart.yml
+++ b/.github/workflows/dev-sgx-restart.yml
@@ -8,6 +8,7 @@ on:
         options: 
         - stop
         - restart
+        - restart-worker-only
 jobs:
   action:
     runs-on: tee-staging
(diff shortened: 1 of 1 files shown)


BillyWooo (approved):
LGTM

PR #1512, opened 2023-03-23
pallet_whitelist ready

wangminqi:


pallets/whitelist/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+authors = ['Litentry Dev']
+edition = '2021'
+homepage = 'https://litentry.com'
+name = 'pallet-whitelist'
+repository = 'https://github.com/litentry/litentry-pallets'
```
Kailai-Wang: I think this URL is deprecated - let's just use parachain repo

pallets/whitelist/src/lib.rs
Kailai-Wang, line 1: Do you want to use it as 1) a standalone pallet 2) a pallet that can be connected to other pallets (like `pallet-membership`)

If 1) => we don't need the EnsureOrigin
If 2) => we probably need multiple instance support, as there can be more pallets which use pallet-whitelist

pallets/whitelist/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+		}
+
+		/// Batch adding of new whitelists
+		#[pallet::call_index(1)]
+		#[pallet::weight(100_000)]
+		#[transactional]
+		pub fn batch_add_whitelists(origin: OriginFor<T>, vs: Vec<T::AccountId>) -> DispatchResult {
+			T::WhitelistManagerOrigin::ensure_origin(origin)?;
+			for v in vs {
+				ensure!(!Self::is_whitelist(&v), Error::<T>::WhitelistAlreadyExists);
```
Kailai-Wang: I'm a bit undecided, should we continue or break? Same for the removal.

Kailai-Wang (commented):
Looks good in general except several questions, please check below

pallets/whitelist/src/lib.rs
wangminqi: Just adding a sample implementation for demosntration prupose.

pallets/whitelist/src/lib.rs
wangminqi: break. And I use [transactional] for further making it either all executed or none.

wangminqi:
The pallet is now instantible

Kailai-Wang (approved):
Thanks, I'll take care of the rest:
- remove IMP-mock code
- remove litmus IMP and VCMP

PR #1516, opened 2023-03-24
Refactor IMP and VCMP errors

### LINKED ISSUE #1495: Make TEE errors more explicit
### Context

See the discussion in slack:

Current definition: https://github.com/litentry/litentry-parachain/blob/bdbc1db72da1efae6c085e1c6fff8d69d842c6f3/primitives/core/src/error.rs#L60-L78

It makes sense to consolidate the StfErrors to `AssertionXFailed` error to make it more clear and more explicit for the F/E. In that way we know e.g. an HTTP error happened exactly when building Assertion 4. Something like this:

```
Assertion5Failed { type: StfError, reason: "user shielding key not set" }
Assertion5Failed { type: HTTPRequestError, reason: "server not available" }
```
Or use nested error type.

---

Kailai-Wang:
resolves #1495 

This PR tries to make the TEE errors more F/E friendly by (hopefully) better sorting and enum usage. Additionally:
- `build_assertion` is renamed to `request_vc` in trusted call handling - reason: building assertion should be a sub-task (although crucial) in requesting vc, some errors happen before building assertions.
- resolves most `unwrap()`s in assertion build
- `RequestVCFailed` always has the assertion-type included
- each IMP/VCMP extrinsic has an explicit `xxFailed` error for better routing and clear display
- a few errors are dismissed or merged, see `ErrorDetail`
- refactored `on_success()` in assertion-handler: it shouldn't call `on_failure` internally

I've also found quite some problems, e.g.:
- too many error types in VC building
- inconsistent error type usage (the same type is used here and there, or unrelated)
- duplicate code

We should improve them later


pallets/identity-management-mock/src/lib.rs
```diff
@@ -182,7 +182,7 @@ pub mod pallet {
 		/// identity not exist when removing an identity
 		IdentityNotExist,
 		/// identity should be disallowed
```
BillyWooo: nice. maybe also update the comment

Kailai-Wang: > nice. maybe also update the comment

Thank you, done!

tee-worker/ts-tests/common/utils.ts
```diff
@@ -282,24 +282,25 @@ export async function checkJSON(vc: any, proofJson: any): Promise<boolean> {
...
-export async function checkFailReason(
+export async function checkErrorDetail(
     response: string[] | Event[],
-    expectedReason: string,
+    expectedDetail: string,
     isModule: boolean
 ): Promise<boolean> {
-    let failReason = '';
+    let detail: string = '';
+    // TODO: sometimes `item.data.detail.toHuman()` or `item` is treated as object (why?)
```
0xverin, line 297: JSON.stringify() cannot guarantee to return a string type, it will return an `object` such as `{func:function(){}}`. This might be one of the reasons why you are seeing the error,  can console the `removeErrorIdentities events[i].data.detail.toHuman() `type to confirm,may be it is a function type.

BillyWooo (approved):
LGTM

outofxxx (approved):

PR #1518, opened 2023-03-24
Parallelise the http requests when building assertions

### LINKED ISSUE #1468: Parallelise the http requests when building assertions
### Context

We are currently doing it sequentially. Imagine the user linked 20 identities and requested a VC while we have a very bad network - this could easily be a problem and block the following requests.

A simple solution is to send the identities in a vector all in one go. Should we go for it? Please bear in mind that we need to "measure before optimise" to make sure we don't optimise things prematurely: we should be able to answer the questions: how long does it take to generate a VC with 20+ identities? What about 50 identities? What if under a throttled network?

---

### LINKED ISSUE #1513: Unify the logging format
### Context

The logs in the trusted_call handling are not so well formatted:
- some with indention/spaces `	[BuildAssertion]`, some not
- some with square brackets, some not
- keywords are not consistent, e.g.`[Assertion]` vs `[BuildAssertion]` vs `[AssertionBuild]` vs `[AssertionBuild-A8]`

Medium prio

---

### LINKED ISSUE #1558: `minimum_amount` type mismatch for holder assertions
`minimum_amount` should be human readable

outofxxx:
resolved #1468 
resolved #1513 
resolved #1558 (Because the coupling between this problem and this pr is too high, so this problem is solved in this pr).


outofxxx:
* Send 3 consecutive 0.2 UNIT balance Transfer Bob -> Charlie
send trusted call transfer from 5HVymiZha4iZLiFj4QbQddb9CnGre6eLNsDgZKFq5snF6iZH to 5GgxHgdjt75pDxoZ4UACrwAfD2qKWrPYaK7VaBD6NiqP5XPH: 200000000000, nonce: 0
send trusted call transfer from 5HVymiZha4iZLiFj4QbQddb9CnGre6eLNsDgZKFq5snF6iZH to 5GgxHgdjt75pDxoZ4UACrwAfD2qKWrPYaK7VaBD6NiqP5XPH: 200000000000, nonce: 0
send trusted call transfer from 5HVymiZha4iZLiFj4QbQddb9CnGre6eLNsDgZKFq5snF6iZH to 5GgxHgdjt75pDxoZ4UACrwAfD2qKWrPYaK7VaBD6NiqP5XPH: 200000000000, nonce: 1

* Assert Bob's incognito current nonce...

Assert 0xf073e0349517dcd85f4058d22d8bf585e3027b0d9826a4e2294c407aa55b7605 .nonce failed, expected = 3, actual = 2

[2023-03-24T09:09:12Z ERROR itp_stf_executor::executor] Stf execute failed: InvalidNonce(0, 1)

BillyWooo (approved):
LGTM.

Kailai-Wang:
There are small conflicts with #1516 in terms of logging adjustments. We need to merge one first and resolve the conflict in the other.

tee-worker/litentry/core/assertion-build/src/a4.rs
Kailai-Wang, line 1: `const VC_SUBJECT_TYPE: &str = "LIT Holder";` => is it Holder or Hodler?

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -76,73 +73,87 @@ pub fn build(
...
 				},
 				Identity::Substrate { address, .. } => {
 					let mut address = account_id_to_string(address.as_ref());
 					address.insert_str(0, "0x");
-					debug!("	[AssertionBuild] A4 Substrate address : {}", address);
-
-					addresses.push(address);
+					Ok(address)
 				},
 				Identity::Web2 { address, .. } => match from_utf8(address.as_ref()) {
```
Kailai-Wang: I don't think we need to care about web2 identity

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -62,12 +62,9 @@ pub fn build(
 	let mut client = GraphQLClient::new();
 	let mut found = false;
 	let mut from_date_index = 0_usize;
+	let mut networks: HashMap<VerifiedCredentialsNetwork, HashSet<String>> = HashMap::new();
 
 	for identity in identities.iter() {
```
Kailai-Wang: we have multiple nested loops and twice `match identity`.

I think the whole logic can be dramatically simplified by applying filters to identity iterators.

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -76,73 +73,87 @@ pub fn build(
...
 
-			for (index, from_date) in ASSERTION_FROM_DATE.iter().enumerate() {
-				// if found is true, no need to check it continually
-				if found {
-					from_date_index = index + 1;
-					break
+			if let Ok(address) = address {
+				if let Some(set) = networks.get_mut(&verified_network) {
+					set.insert(address);
 				}
```
Kailai-Wang: The map will be forever empty if I'm not mistaken because the initial `get_mut` will always return `None` and it can never insert any element.

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -76,73 +73,87 @@ pub fn build(
...
+			// Ethereum network
+			(min_balance / (10 ^ 18)) as f64
+		};
+
+		let mut lit_token_addr = String::from("");
+		if verified_network == VerifiedCredentialsNetwork::Ethereum {
+			lit_token_addr = LIT_TOKEN_ADDRESS.to_string();
+		}
+
+		for (index, from_date) in ASSERTION_FROM_DATE.iter().enumerate() {
```
Kailai-Wang, line 170: Why do we need to divide the time into yearly durations? Can TDF only handle data within one year? A4 has only one input parameter `min_amount`.

What do `from_date` and `to_date` mean?

Either the implementation is wrong or the description is wrong, as the subject description says: "Check whether any of the linked accounts hold a minimum amount of LIT **NOW**"

Kailai-Wang (changes requested):
I only checked `a4.rs` and had a few questions. I assume other VCs would have the same.

I need to get myself clear before I can continue the review :(

tee-worker/litentry/core/assertion-build/src/a4.rs
outofxxx: fixed, should be `Hodler`

tee-worker/litentry/core/assertion-build/src/a4.rs
outofxxx: yes, it's a bug, fixed

jingleizhang (approved):
Looks good, thanks!

jingleizhang:
The CI `lit-resume-worker` has failed, but no errors found in the log.
[logs-lit-resume-worker.zip](https://github.com/litentry/litentry-parachain/files/11084289/logs-lit-resume-worker.zip)

tee-worker/litentry/core/assertion-build/src/a10.rs
```diff
@@ -60,42 +60,44 @@ pub fn build(
...
-					}
+				addresses.push(address);
+			}
+		}
+	}
+
+	if !addresses.is_empty() {
+		for (index, from_date) in ASSERTION_FROM_DATE.iter().enumerate() {
+			if found {
+				from_date_index = index + 1;
```
Kailai-Wang: Is this logic correct?

it seems `from_date_index` can only be in the range [2, 7] (i.e. it can never go below 2 as this branch will only be entered when `index` is >= 1)

Apart from that, imagine the user holds since `2023-01-01`, `is_hodler` will only be true when `index == 6`, but the corresponding `from_date_index` is still 0 (as the whole loop ends already).

I don't quite understand why we need to `+1` here

Kailai-Wang (changes requested):
Only checked `a10` and I don't understand the logic...

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -102,98 +99,97 @@ pub fn build(
...
-					Ok(addr) => addresses.push(addr.to_string()),
-					Err(e) => error!(
-						"	[AssertionBuild] A4 parse error Web2 address {:?}, {:?}",
-						address, e
-					),
+	let dates = add_now_to_from_dates(&now);
+	debug!("Assertion A4 dates: {:?}", dates);
+
+	// is_hold default value is true
+	let mut is_hold = networks.is_empty();
```
Kailai-Wang: Imagine the `networks` **is** empty, how should the VC look like? Should it be like:
```
from_date < 2017-01-01
value == false
```
? If so then the logic is incorrect here

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -102,98 +99,97 @@ pub fn build(
...
-				Identity::Web2 { address, .. } => match from_utf8(address.as_ref()) {
-					Ok(addr) => addresses.push(addr.to_string()),
-					Err(e) => error!(
-						"	[AssertionBuild] A4 parse error Web2 address {:?}, {:?}",
-						address, e
-					),
+	let dates = add_now_to_from_dates(&now);
+	debug!("Assertion A4 dates: {:?}", dates);
+
+	// is_hold default value is true
```
Kailai-Wang: Typo? Do you mean the default value should be `false`?

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -102,98 +99,97 @@ pub fn build(
 		identities
 	);
 
+	let evm_min_balance = (min_balance / (10 ^ 18)) as f64;
+	let substrate_min_balance = (min_balance / (10 ^ 12)) as f64;
```
Kailai-Wang: Is it confirmed that F/E will pass in `numbers * (10 ^ decimals)`?

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -102,98 +99,97 @@ pub fn build(
...
+	let dates = add_now_to_from_dates(&now);
+	debug!("Assertion A4 dates: {:?}", dates);
+
+	// is_hold default value is true
+	let mut is_hold = networks.is_empty();
+	for (verified_network, addresses) in networks {
+		// If find a satisfying interval from a network, do not need to query the remaining networks.
+		if hold_from_date != now {
+			break
+		}
```
Kailai-Wang: There is a (not so corner) case: what if the user holds substrate LIT from 2023-01-01 to now, but holds EVM LIT from 2017-01-01 to now? Or vice versa?

We should only break if `hold_from_date == ASSERTION_FROM_DATE.last().unwrap()`, what do you think?

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -102,98 +99,97 @@ pub fn build(
...
-					addresses.push(address);
-				},
-				Identity::Substrate { address, .. } => {
-					let mut address = account_id_to_string(address.as_ref());
-					address.insert_str(0, "0x");
-					debug!("	[AssertionBuild] A4 Substrate address : {}", address);
+	// Including 8 items from NOW -> 2017-01-01
+	// Because it is necessary to determine the continuous holding, the reverse order query is started from NOW.
+	// As long as there are terminals, this interval is taken as the result.
+	// query from NOW -> 2023 -> ... -> 2017
```
Kailai-Wang: I don't get it, why do we have to query `NOW`?

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -102,98 +99,97 @@ pub fn build(
...
+			evm_min_balance
+		};
+
+		for from_date in dates.iter() {
+			let vch = VerifiedCredentialsIsHodlerIn::new(
+				addresses.clone(),
+				from_date.to_string(),
+				verified_network.clone(),
+				token_address.to_string(),
+				q_min_balance,
```
Kailai-Wang: It means for a user we could have 7 * 2 = 14 HTTP requests, even with vectored addresses :(
Can't TDF handle mixed network type?

tee-worker/litentry/core/assertion-build/src/a4.rs
outofxxx: This is the correct expression that should be
```rust
let mut is_hold = true;
is_hold &= !network.is_empty();
```

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -102,98 +102,103 @@ pub fn build(
...
+				addresses.clone(),
+				from_date.to_string(),
+				verified_network.clone(),
+				token_address.to_string(),
+				q_min_balance.to_string(),
+			);
+			match client.check_verified_credentials_is_hodler(vch) {
+				Ok(is_hodler_out) => {
+					for hodler in is_hodler_out.verified_credentials_is_hodler.iter() {
+						is_hold = is_hold || hodler.is_hodler;
```
Kailai-Wang, line 181: I think you need to reset `is_hold` in `for (verified_network, addresses) in networks` - otherwise the second network would always start with `is_hold == true`

Kailai-Wang (changes requested):
It's close - see below.

I only checked A4 though, I hope others are similar.

And please fix CIs :)

tee-worker/litentry/core/assertion-build/src/a4.rs
outofxxx: Yes, should reset is_hold to false every networks loop.

outofxxx:
CI fails here:

```shell
UserShieldingKeySet event data: [
  '0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48',
  '0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d'
]
    1) set user shielding key
    - check user shielding key from sidechain storage after setUserShieldingKey
    - check idgraph from sidechain storage before create
    - create identities
    - check IDGraph before verifyIdentity and after createIdentity
    - verify identities
    - check IDGraph after verifyIdentity
    - verify error identities
    - remove identities
    - check IDGraph after removeIdentity
    - remove prime identity NOT allowed
    - remove error identities
    - set error user shielding key
    - create error identities


  2 passing (2m)
  13 pending
  1 failing

  1) Test Identity
       set user shielding key:

      alice shielding key should be set
      + expected - actual

      -0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48
      +0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d
      
      at Context.<anonymous> (identity.test.ts:94:16)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Kailai-Wang (approved):
Thanks!

PR #1534, opened 2023-03-28
Refactor A8 AssertionNetworks to enum

### LINKED ISSUE #1511: refactoring AssertionNetworks from array to enum
https://github.com/litentry/litentry-parachain/blob/c73cdc19a575039e8f4bd08e5cf73375635da64b/primitives/core/src/assertion.rs#L47

https://github.com/litentry/litentry-parachain/blob/c73cdc19a575039e8f4bd08e5cf73375635da64b/primitives/core/src/assertion.rs#L28-L29

outofxxx:
resolved #1511


primitives/core/src/assertion.rs
```diff
@@ -24,8 +24,18 @@ use sp_runtime::{traits::ConstU32, BoundedVec};
 
 type MaxStringLength = ConstU32<64>;
 pub type ParameterString = BoundedVec<u8, MaxStringLength>;
-pub type Network = BoundedVec<u8, MaxStringLength>;
-pub type AssertionNetworks = BoundedVec<Network, MaxStringLength>;
+pub type AssertionNetworks = BoundedVec<AssertionNetwork, MaxStringLength>;
```
Kailai-Wang: I don't think we need `AssertionNetworks` anymore

tee-worker/litentry/core/data-providers/src/graphql.rs
```diff
@@ -83,6 +83,35 @@ impl From<EvmNetwork> for VerifiedCredentialsNetwork {
 	}
 }
 
+impl From<AssertionNetwork> for VerifiedCredentialsNetwork {
```
Kailai-Wang: If I'm allowed to be a bit picky: VC = Verifiable Credential.

So it should be either `VerifiableCredentialNetwork` or `VCNetwork` or maybe `SupportedNetwork`, and please add a comment in the struct definition why we need this (or what's the difference from `AssertionNetwork`).

What's even better is that we have a trait for each data provider, something like `get_supported_network`, which returns a subset of `AssertionNetwork` => but we can do it later

tee-worker/ts-tests/vc.test.ts
```diff
@@ -26,29 +26,29 @@ describeLitentry('VC test', async (context) => {
...
-    });
+    // step('check user sidechain storage before create', async function () {
+    //     const resp_shieldingKey = await checkUserShieldingKeys(
+    //         context,
+    //         'IdentityManagement',
+    //         'UserShieldingKeys',
+    //         u8aToHex(context.substrateWallet.alice.addressRaw)
+    //     );
+    //     assert.equal(resp_shieldingKey, '0x', 'resp_shieldingKey should be empty before set');
+    // });
```
Kailai-Wang: What's the reason for commenting out these? If it's needed we should have the comment somewhere.

Kailai-Wang (changes requested):

tee-worker/ts-tests/vc.test.ts
outofxxx: Accidentally commented out for local testing, it will be restored

Kailai-Wang:
As mentioned it's OK to leave as it is right now as the refactoring is not desperately needed. But please leave a comment/TODO that we'll refactor out `VerifiedCredentialNetwork` later

Kailai-Wang (approved):

PR #1541, opened 2023-03-28
Add `req_ext_hash` in IMP/VCMP events

### LINKED ISSUE #1272: Associate the errors with the request extrinsics
### Context
Raised by IDH colleagues, we only return the error names/strings when some error happens (e.g. when creating or verifying identities), but there's no way to associate the error to a specific extrinsic request. This makes it difficult for F/E to tell which account/identity the error is regarded, which makes the error message sometimes useless.

How to solve it?

cc @jonalvarezz @kaylawangnow

---

Kailai-Wang:
resolves #1272 

This PR adds `req_ext_hash` to the IMP/VCMP events for better tracking.
For the error events, the `account` will be optional as some errors can happen during the decode where we can't retrieve the extrinsic sender.

TODO: `batch_all` is untested: what should the extrinsic-hash be? (or do we even have an extrinsic hash for individual call in the batch?


pallets/vc-management/src/lib.rs
```diff
@@ -88,29 +88,70 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		// TODO: do we need account as event parameter? This needs to be decided by F/E
-		VCRequested { shard: ShardIdentifier, assertion: Assertion },
+		VCRequested {
```
jingleizhang, line 91: `VCRequested` would be more F/E friendly, if a new parameter `account` is added.

tee-worker/core/parentchain/indirect-calls-executor/src/executor/litentry/create_identity.rs
```diff
@@ -79,17 +79,18 @@ impl CreateIdentity {
 			let enclave_account_id = context.stf_enclave_signer.get_enclave_account()?;
 			let trusted_call = TrustedCall::create_identity_runtime(
 				enclave_account_id,
-				account,
+				account.clone(),
 				identity,
 				metadata,
 				self.block_number,
+				hash_of(extrinsic),
```
jingleizhang, line 86: I don't know the actual value of `extrinsic` and `hash_of(extrinsic)`, so I'm just wondering can F/E predict this hash after it sends the `CreateIdentity` request? 
If the answer is YES, then this is great.

jingleizhang (approved):

tee-worker/core/parentchain/indirect-calls-executor/src/executor/litentry/create_identity.rs
Kailai-Wang: Should be able to do that - once the extrinsic is included in a block, the hash is firm. See the corresponding ts-test adjustments.

pallets/vc-management/src/lib.rs
Kailai-Wang: Yes I can add that :)

PR #1576, opened 2023-04-05
Use an enclave admin as origin

Kailai-Wang:
It's a prerequisite of the incoming "enclave update" PR.

We need it because the `update_scheduled_enclave` can only be parsed if a raw signed origin type calls it - the indirect-call-executor can't parse `sudo` calls.


BillyWooo (approved):
LGTM

PR #1579, opened 2023-04-06
Revisit enclave update

### LINKED ISSUE #1295: Enclave update [3-2] Check mrenclave and scheduled height before produce blocks

Kailai-Wang:
resolves #1295 

This is a big PR which revisits the previous implementation of enclave update, refactors and reimplements some logic if necessary.

Things that are tested:
- sealing of scheduled enclave syncs with parachain
- unsealing of scheduled enclave works after restart
- block production will be pending when MRENCLAVE doesn't match
- shard migration works
- enclave upgrade works after resuming (with some manual operations)

To be tested:
Enclave state migration after upgrade


BillyWooo (approved):
Generally understand the logic. Couldn't get a chance to test by myself. But it looks quite good for me. Let's roll out.

PR #1580, opened 2023-04-06
comment out removing history data; backup log into specific folder;

BillyWooo:
as title
to avoid any possible mistake removing.


Kailai-Wang (approved):
Thanks

PR #1581, opened 2023-04-06
Add more logs

### LINKED ISSUE #1557: More explicit logs
### Context

We've already enabled quite some logs, but still, they might not be enough to dissect the problem we found.
Moreover it would be great if we could:
- remove some logs that don't give out much information (e.g. ocall notification without details)
- add more explicit logs on important timepoints (e.g. block production start), maybe with better formatting/dividers.

---

Kailai-Wang:
resolves #1557


BillyWooo (approved):
Only logs changes. LGTM

PR #1584, opened 2023-04-07
Refactor admin in teerex and VCMP

Kailai-Wang:
This PR:
- use unified `SetAdmin` and its origin setting
- shift the call indexes of extrinsics that should be called by TEE by a big margin
- add genesis_build for admin setting: Litmus => None, rococo => sudo


outofxxx (approved):
lgtm

PR #1585, opened 2023-04-07
Minor GHA adjustments

Kailai-Wang:
A minor PR to adjust GHA:
- remove `sgx restart` CI to avoid misuse
- do not auto-close stale issues (but we still get notified 60 days when an issue becomes stale)


BillyWooo (approved):
LGTM

PR #1587, opened 2023-04-10
parachain launch scripts

### LINKED ISSUE #1564: Script to move VCRegistry snapshot
### Context

precondition: https://github.com/litentry/litentry-parachain/issues/1544

We need a script to move a snapshot of VCRegistries of a given chain to another:
- query the storage of the VCRegistry of a given chain (e.g. staging)
- store it as human-readable format, e.g. JSON, with timestamp
- generate an encoded extrinsic call which adds these entries to a new VCRegistry in another chain, using utility batch (so it should be one call).
- test the workflow.

---

### LINKED ISSUE #1582: Provide hex-encoded calls for initial R1 setup
### Context

See step 7 in the R1 deployment plan. High prio.

---

0xverin:
resolves: #1582
resolves: #1564

- [x] test move vcregistry snapshot(tee-staging->local)
- [x] test add whitelist(add whitelist success and not whitelist can't request IMP&VCMP)

add [setup enclave script](https://github.com/litentry/litentry-parachain/pull/1587/files#diff-e1ca93ff23e63583fc56e02d2e9cf24ecb68eb0ba1e2c072510ff16aa63fdb10);
I think we just need to generate the call encode_hex, putting private key into a public file feels strange. @BillyWooo


scripts/ts-utils/index.ts
```diff
@@ -0,0 +1,16 @@
+const { ApiPromise, WsProvider } = require("@polkadot/api");
+
+//fetchApi is used to fetch data from the chain
+const wsFetchProvider = new WsProvider("wss://tee-staging.litentry.io:443");
+
+//syncApi is used to send transactions to the chain
+const wsSyncProvider = new WsProvider("ws://localhost:9944");
+
```
jingleizhang: Why those two endpoint urls are different?

scripts/ts-utils/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export const initApi = async () => {
+    const fetchApi = await ApiPromise.create({ provider: wsFetchProvider });
+    const syncApi = await ApiPromise.create({ provider: wsSyncProvider });
+
+    return { fetchApi, syncApi };
+}
+
```
jingleizhang: Why this file is called `index` as it just initializes the APIs?

scripts/ts-utils/move-vcregistry-snapshot-script.ts
```diff
@@ -0,0 +1,48 @@
...
+        let tx = syncApi.tx.vcManagement.addVcRegistryItem(
+            data[index].index[0],
+            data[index].vc.subject,
+            data[index].vc.assertion,
+            data[index].vc.hash_
+        );
+        txs.push(tx);
+    }
+
+    const extrinsic = syncApi.tx.sudo.sudo(syncApi.tx.utility.batch(txs));
```
jingleizhang: No pagination? What if the length of data is larger than the maximum count of a batch call.

scripts/ts-utils/setup-enclave-script.ts
```diff
@@ -0,0 +1,38 @@
+
+//run:npx ts-node setup-enclave-script.ts
+
+import { initApi } from "./index";
+const { hexToU8a } = require('@polkadot/util');
+
+//100 token
+const transferAmount = '100000000000000';
+const enclaveAccount = '5EtChN6SSfL7E13fEqKQn4jZztDtZkSiRqfFTv2BKufL5fzZ'
+const mrenclave = 'a552654d1733c4054a3c7e5e86adf26b5d65c072b57b2550fe763821ebac54c6'
```
jingleizhang: Please get these 3 variables from the script's command line args.

jingleizhang (changes requested):

scripts/ts-utils/add-whitelist-script.ts
```diff
@@ -0,0 +1,56 @@
+
+//run:npx ts-node add-whitelist-script.ts
+
+import { initApi } from "./index";
+
+let whitelist = ['5EtChN6SSfL7E13fEqKQn4jZztDtZkSiRqfFTv2BKufL5fzZ'];
```
BillyWooo: This 'whitelist' should be either an input parameter or to read from a file. I prefer the later option.

scripts/ts-utils/index.ts
BillyWooo: I think these two urls should be input parameter for this module.

scripts/ts-utils/setup-enclave-script.ts
BillyWooo: The same I want to comment.

BillyWooo (commented):
The scenario we need to consider here: the utils will be used in an environment where only script itself can be rely on, and also need to consider automation during the deployment and restarting. So we need to avoid any kind of hardcode; to consider the work-flow during automation.

scripts/ts-utils/index.ts
0xverin: `wsFetchProvider`is the endpoint get vcRegistry, `wsSyncProvider` is the endpoint we need to synchronize,encode call need to specify an endpoint,otherwise can't decode,so we should define two endpoint

scripts/ts-utils/index.ts
0xverin: yes

jingleizhang: so please rename the `index.ts` to a more specific name.

scripts/ts-utils/index.ts
jingleizhang: understand. What about to rename `fetch` & `sync` to be more specific names?

scripts/ts-utils/index.ts
0xverin: fixed.

scripts/ts-utils/move-vcregistry-snapshot-script.ts
0xverin: fixed and tested

scripts/ts-utils/setup-enclave-script.ts
0xverin: fixed

0xverin:
> The scenario we need to consider here: the utils will be used in an environment where only script itself can be rely on, and also need to consider automation during the deployment and restarting. So we need to avoid any kind of hardcode; to consider the work-flow during automation.

Fixed,i think you can test it locally with a admin account.

scripts/ts-utils/add-whitelist-script.ts
0xverin: fixed

BillyWooo (approved):
In general, it looks good to me. Do you think it's better to remove 'script' from the files name?

Kailai-Wang:
This PR doesn't link against any issues btw :)

0xverin:
> This PR doesn't link against any issues btw :)

Thanks for the reminder, there was a missing space.

0xverin:
> In general, it looks good to me. Do you think it's better to remove 'script' from the files name?

think so!

scripts/ts-utils/initApis.ts
```diff
@@ -0,0 +1,20 @@
+const { ApiPromise, WsProvider } = require("@polkadot/api");
+const { cryptoWaitReady } = require('@polkadot/util-crypto');
+import colors from "colors";
+//fetchApi is used to fetch data from the chain
+const wsFetchProvider = new WsProvider("wss://tee-staging.litentry.io:443");
```
BillyWooo: I still think `wss://tee-staging.litentry.io:443` should be a parameter.

BillyWooo (changes requested):

jingleizhang (approved):

scripts/ts-utils/initApis.ts
0xverin: Fixed!Thanks for your suggestions. It will help improve my coding and engineering thinking in the future.

BillyWooo (approved):

PR #1605, opened 2023-04-15
Fix empty nonce in AES encryption

### LINKED ISSUE #1603: Empty nonce in AES-encrypted return data
### Context

In the returning payload it seems that the nonce is always 0:

image

It must come from here: https://github.com/litentry/litentry-parachain/blob/e59d92762e8ade82e4f4ae3090dcb1165bb024fa/tee-worker/litentry/primitives/src/lib.rs#L65

Apart from that it might be better to move everything related to aes encryption to a separate mod/file.

---

Kailai-Wang:
resolves #1603

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,9 +57,9 @@ dependencies = [
 
 [[package]]
 name = "aead"
-version = "0.5.1"
+version = "0.5.2"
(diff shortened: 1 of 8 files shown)


BillyWooo (approved):
LGTM

PR #1615, opened 2023-04-18
Ts test case of wrong signature

### LINKED ISSUE #1602: IdentityVerification of wrong signature in ts-test
ts-test need to validate with the wrong signature, and then catch the specified IdentityVerification error.

0xverin:
resolves #1602 

- [x] verify evm identity with wrong signature
- [x] verify twitter identity with evm validation(invalid identity case)

btw,added color to the log output.


0xverin:
There was an issue with the ts-test: when before  start listening to the event before the transaction was confirmed in the block, which could potentially result in listening to the event from the previous transaction. For example, [in this CI resulst](https://github.com/litentry/litentry-parachain/actions/runs/4732369097/jobs/8400446324?pr=1607#step:10:2177), we can see that the singer of the [second event](https://github.com/litentry/litentry-parachain/actions/runs/4732369097/jobs/8400446324?pr=1607#step:10:2230) is still the same as the first event, in fact the second singer is `Charlie`. Based on[ this code snippet](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/litentry/pallets/identity-management/src/lib.rs#L265), it is easy to understand that we need to first determine the shielding key(`Charlie` did not set it before). This issue has now been resolved by modifying the `sendTxsWithUtility` function.

Kailai-Wang (approved):
Thanks

PR #1620, opened 2023-04-19
Return IDGraph when the user shielding key is set

### LINKED ISSUE #1609: Add IDGraph in the UserShieldingKeySet event
### Context

As topic. The requirement comes from F/E and they need to know the IDGraph when e.g. the user resets their shielding key.

---

Kailai-Wang:
resolves #1609


BillyWooo (approved):
LGTM

PR #1624, opened 2023-04-21
ts-test throw specific error instead of timeout

### LINKED ISSUE #1571: ts-test should throw specific error instead of timeout
`listenevent` now throws a timeout error after a long time if it does not catch the correct event. This is inconvenient  to debug.Need to optimize.

0xverin:
as the title;
resolves #1571


Kailai-Wang (commented):
Thanks - I was about to approve it but then I realised there might be problems when testing bulk tx where multiple extrinsics are sent in parallel: a mix of different types of events will be emitted in the same block, e.g. both `CreatedIdentityFailed` and `IdentityCreated`.

This will interfere with the promise resolving as you wait for `CreatedIdentityFailed` but get `IdentityCreated` first - or the other way around - then you run into the reject branch immediately.

Have you run the bulk-vc/identity test?

I believe a better way is either to use index of extrinsics/events in the block or ext-hash. I prefer the former as it's the "official" way to go.

0xverin:
> Thanks - I was about to approve it but then I realised there might be problems when testing bulk tx where multiple extrinsics are sent in parallel: a mix of different types of events will be emitted in the same block, e.g. both `CreatedIdentityFailed` and `IdentityCreated`.
> 
> This will interfere with the promise resolving as you wait for `CreatedIdentityFailed` but get `IdentityCreated` first - or the other way around - then you run into the reject branch immediately.
> 
> Have you run the bulk-vc/identity test?
> 
> I believe a better way is either to use index of extrinsics/events in the block or ext-hash. I prefer the former as it's the "official" way to go.

1. We can resolve mixed events in order by using the '[includes](https://github.com/litentry/litentry-parachain/blob/1571-throw-specific-error/tee-worker/ts-tests/common/transactions.ts#L113)' method and [sorting by signers](https://github.com/litentry/litentry-parachain/blob/1571-throw-specific-error/tee-worker/ts-tests/common/transactions.ts#L134),so we can parse events in order. I have tested this approach and confirmed it's effectiveness. Thank you for reminding me. I forgot to test `bulk-vc/identity` here. I have fixed some errors, but the reason for the failure was not the issue we were concerned about; it was due to another problem.

2. I think we need to add `bulk-vc/identity` to CI test(To save time,just use 2 accounts to run testflow.)

3. I have previously tried to filter events using txhash, but it didn't seem to match successfully. I don't understand the reason and couldn't find any relevant documentation. Based on my previous experience with ethers, filtering events using txhash is a basic operation, which is why I resorted to using signer filtering when I was unable to filter events. If you have any relevant documentation or links, please send them to me. Thank you.

Kailai-Wang:
>I have previously tried to filter events using txhash, but it didn't seem to match successfully. I don't understand the reason and couldn't find any relevant documentation. Based on my previous experience with ethers, filtering events using txhash is a basic operation, which is why I resorted to using signer filtering when I was unable to filter events. If you have any relevant documentation or links, please send them to me. Thank you.

I think IDHub already uses it, you could ask @jonalvarezz @grumpygreenguy 

Will index work? I thought in a block the event-index can be associated with extrinsic-index, see e.g. subscan: 
Maybe @YoshiyukiSakura knows more.

image

Kailai-Wang:
>I think we need to add bulk-vc/identity to CI test(To save time,just use 2 accounts to run testflow.)

Yes please - I suggest adding it in this PR already.

YoshiyukiSakura:
plz get in touch on Slack so we can discuss 🙌

0xverin:
Please take a close look at [this piece of code](https://github.com/litentry/litentry-parachain/pull/1624/files#diff-00583d14df51e911a6b575e45e173a1e068e18e9178008139722f61238cfe40eR106). I filter events based on `section`, `method`, and `extrinsic index`. Actually, I'm not very clear about whether this is a offical way. I referred to `idhub F/E` and [this repo](https://github.com/zeeve-cargo/axtend/blob/11d7e1844a21a14a5c4cce569d245761684d3de4/tests/util/substrate-rpc.ts#L90), and they seem to be the same.

tee-worker/ts-tests/bulk_identity.test.ts
```diff
@@ -36,36 +38,36 @@ describeLitentry('multiple accounts test', 10, async (context) => {
...
+                .batch(txs)
+                .signAndSend(context.substrateWallet.alice, async (result: SubmittableResult) => {
+                    if (result.status.isInBlock) {
+                        console.log(`Transaction included at blockHash ${result.status.asInBlock}`);
+                        resolve(result.status);
+                    } else if (result.status.isInvalid) {
+                        console.log(`Transaction is ${result.status}`);
+                    }
+                });
+        });
```
arieljbirnbaum: Do we want to:
- `resolve` when `result.status.isFinalized`
- `reject` when `result.status.isInvalid`

? :thinking: 

Compare `signAndSend` at `ts-tests/tests/utils.ts`

0xverin: Fixed.

0xverin: Here's a special case where [`disable_vc` and `revoke_vc`](https://github.com/litentry/litentry-parachain/blob/ad3575322a31aba111588a2596519511538b218f/pallets/vc-management/src/lib.rs#L235) happen instantaneously and are modified directly in the parachain without going through the TEE. Therefore, if we wait for `result.status.isFinalized` before listening, it's possible that the block containing the changes has already passed. Thus, we need to start listening at `result.status.asInBlock`. @zhouhuitian

0xverin:
[This failure](https://github.com/litentry/litentry-parachain/actions/runs/4880700297/jobs/8708752102?pr=1624) arose from the following context: I used `Promise.all` to send multiple transactions from different accounts concurrently, but `Promise.all `could not ensure the transactions were sent in order, and therefore, the events returned were not in order either. In the CI failure logs, the event of Bob's transaction event was returned before Alice's transaction event. [In the previous version](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/common/transactions.ts#LL126C14-L126C14), I added sorting by signer address, which guaranteed the events to be returned in order, but I feel this approach is too complicated and not elegant. Do you have any ideas? @grumpygreenguy

arieljbirnbaum:
@Verin1005 we're not resolving #1630 here; please update the description

0xverin:
> @Verin1005 we're not resolving #1630 here; please update the description

Updated!

0xverin:
@grumpygreenguy I am working on fixing this pr.

tee-worker/docker/lit-bulk-identity-test.yml
```diff
@@ -0,0 +1,25 @@
+services:
+  lit-bulk-identity-test:
```
arieljbirnbaum, line 2: Is this supposed to enable a new check on CI? I don't see it being executed in this PR :thinking:

tee-worker/ts-tests/common/transactions.ts
```diff
@@ -179,45 +73,124 @@ export async function sendTxsWithUtility(
...
+                        //Then, it sorts the events based on this index so that the resulting event array is sorted according to the order of the signers array.
+                        const signerIndexA = signerToIndexMap[a.event.data.find(signerMatches)!.toHex()];
+                        const signerIndexB = signerToIndexMap[b.event.data.find(signerMatches)!.toHex()];
+                        return signerIndexA - signerIndexB;
+                    });
+
+                //There is no good compatibility method here.Only successful and failed events can be filtered normally, but it cannot filter error + successful events, which may need further optimization
+                const eventsToUse = filtered_events_with_signer.length > 0 ? filtered_events_with_signer : events_in_extrinsic;
+
+                events = [...events, ...eventsToUse];
```
arieljbirnbaum: Now I'm even more confused; how did it work before without this? :sweat_smile:

0xverin: This is the [key](https://github.com/litentry/litentry-parachain/pull/1624/files#diff-00583d14df51e911a6b575e45e173a1e068e18e9178008139722f61238cfe40eR161). :P

tee-worker/ts-tests/bulk_vc.test.ts
```diff
@@ -49,22 +50,23 @@ describeLitentry('multiple accounts test', 2, async (context) => {
...
-                .signAndSend(context.substrateWallet.alice, async (result: SubmittableResult) => {
-                    if (result.status.isInBlock) {
-                        console.log(`Transaction included at blockHash ${result.status.asInBlock}`);
+                .signAndSend(context.substrateWallet.alice, (result: SubmittableResult) => {
+                    console.log(`Current status is ${result.status.isFinalized}`);
+                    if (result.status.isFinalized) {
                         resolve(result.status);
                     } else if (result.status.isInvalid) {
                         console.log(`Transaction is ${result.status}`);
+                        reject(result.status);
```
arieljbirnbaum:
```suggestion
                    console.log(`Current status is ${result.status}`);
                    if (result.status.isFinalized) {
                        resolve(result.status);
                    } else if (result.status.isInvalid) {
                        reject(result.status);
```
?

0xverin: Fixed

tee-worker/docker/lit-bulk-identity-test.yml
0xverin: Fixed!

tee-worker/ts-tests/common/transactions.ts
```diff
@@ -158,7 +158,6 @@ export async function listenEvent(api: ApiPromise, section: string, methods: str
                     }
                     return (
                         phase.isApplyExtrinsic &&
-                        phase.asApplyExtrinsic.eq(index) &&
```
arieljbirnbaum: Don't we need this to make sure that each event gets mapped to the correct extrinsic? :thinking:

tee-worker/ts-tests/bulk_identity.test.ts
```diff
@@ -7,66 +7,65 @@ import {
...
             return web3Signer.substrateWallet;
         });
         ethereumSigners = context.web3Signers.map((web3Signer) => {
             return web3Signer.ethereumWallet;
         });
     });
     step('send test token to each account', async () => {
-        const txs: SubmittableExtrinsic<ApiTypes>[] = [];
-        for (let i = 0; i < substraetSigners.length; i++) {
+        const txs: any = [];
```
arieljbirnbaum, line 38: Must we? :sadpanda:

tee-worker/ts-tests/common/transactions.ts
```diff
@@ -1,24 +1,22 @@
...
-import { u8aToHex } from '@polkadot/util';
 import { expect } from 'chai';
 import colors from 'colors';
+import { HexString } from '@polkadot/util/types';
+import { Codec } from '@polkadot/types/types';
+import { u8aToHex } from '@polkadot/util';
 //transactions utils
 export async function sendTxUntilInBlock(api: ApiPromise, tx: SubmittableExtrinsic<ApiTypes>, signer: KeyringPair) {
-    return new Promise<{ block: string; txHash: string }>(async (resolve, reject) => {
+    return new Promise<SubmittableResult>(async (resolve, reject) => {
```
arieljbirnbaum, line 14: Nice :+1:

tee-worker/ts-tests/common/transactions.ts
0xverin: We are using an indexed event for signer addresses sorting here. If we include an  `phase.asApplyExtrinsic.eq(index)`, the events will be returned one by one, and as a result, the sorting of the signer addresses will become invalid(but we need sort by signer address).

tee-worker/ts-tests/bulk_identity.test.ts
0xverin: I modified this type [here](https://github.com/litentry/litentry-parachain/pull/1703/files#diff-e9b9bfd1cf7ea674b2a3a4c883e38a401b6a06c08f90e00d69d60f6abbdd4b08R41).

tee-worker/ts-tests/common/transactions.ts
0xverin: Additionally, I think using `section+method+signer.address` can filter the events we need. Although it may not be official, it's the approach we can currently use.

arieljbirnbaum (approved):

PR #1656, opened 2023-05-02
import first, then check MRENCLAVE

BillyWooo:
as title, 
minor change during testing.


Kailai-Wang:
I've pondered over it a bit - let's keep it unchanged for the moment.

Mainly because [this comment](https://github.com/integritee-network/worker/issues/1231#issuecomment-1487103735): _there absolutely mustn't be a state change when a worker fails to produce a block_. Based on it we shouldn't pop the import queue prior to checking MRENCLAVE. It will cause problems in a multi-worker setup.

Again, the `startBlock` isn't meant to be used for such scenario - we were using it in a hacky way to recover from a bad state.

BillyWooo:
This is just a trick try-out, not a solution. So let's close it.

PR #1662, opened 2023-05-04
Revert to old version of listenEvent

arieljbirnbaum:
Attempt to address the issue in [this comment](https://github.com/litentry/litentry-parachain/pull/1624#issuecomment-1534374083).

@Verin1005 I tried to match events to extrinsics [the canonical way](https://polkadot.js.org/docs/api/cookbook/blocks#how-do-i-map-extrinsics-to-their-events); the main issue I kept running into was being unable to determine the index within the block for individual extrinsics bundled in a single `batchAll` call inside `sendTxsWithUtility`.

The closest I could find in terms of documentation was [this question about `batch` on StackExchange](https://substrate.stackexchange.com/questions/174/correlate-events-with-extrinsics-in-a-utility-batch-call-in-polkadot), but I can't quite tell if `batchAll` emits the same set of events :P

To unblock the fix for #1571, I'm just restoring the previous version of `listenEvent` (modulo a tiny refactoring).

### DIFF
--- a/tee-worker/ts-tests/common/utils.ts
+++ b/tee-worker/ts-tests/common/utils.ts
@@ -777,7 +777,7 @@ export async function assertInitialIDGraphCreated(api: ApiPromise, signer: Keyri
     // check identity in idgraph
     const expected_identity = api.createType(
         'LitentryIdentity',
-        await buildIdentityHelper(u8aToHex(signer.addressRaw), 'TestNet', 'Substrate')
+        await buildIdentityHelper(u8aToHex(signer.addressRaw), 'LitentryRococo', 'Substrate')
     ) as LitentryIdentity;
 
     assert.equal(JSON.stringify(event.idGraph[0][0]), JSON.stringify(expected_identity));
--- a/tee-worker/ts-tests/common/utils.ts
+++ b/tee-worker/ts-tests/common/utils.ts
@@ -777,10 +777,10 @@ export async function assertInitialIDGraphCreated(api: ApiPromise, signer: Keyri
     // check identity in idgraph
     const expected_identity = api.createType(
         'LitentryIdentity',
-        await buildIdentityHelper(u8aToHex(signer.addressRaw), 'LitentryRococo', 'Substrate')
+        await buildIdentityHelper(u8aToHex(signer.addressRaw), 'TestNet', 'Substrate')
     ) as LitentryIdentity;
 
-    assert.isTrue(isEqual(event.idGraph[0][0], expected_identity));
+    assert.equal(JSON.stringify(event.idGraph[0][0]), JSON.stringify(expected_identity));
     // check identityContext in idgraph
     assert.equal(event.idGraph[0][1].linking_request_block, 0);
     assert.equal(event.idGraph[0][1].verification_request_block, 0);
--- a/tee-worker/ts-tests/identity.test.ts
+++ b/tee-worker/ts-tests/identity.test.ts
@@ -119,7 +119,7 @@ describeLitentry('Test Identity', 0, (context) => {
         // the main address should be already inside the IDGraph
         const main_identity = await buildIdentityHelper(
             u8aToHex(context.substrateWallet.alice.addressRaw),
-            'TestNet',
+            'LitentryRococo',
             'Substrate'
         );
         const identity_hex = context.api.createType('LitentryIdentity', main_identity).toHex();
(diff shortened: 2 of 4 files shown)


Kailai-Wang:
> Btw do we need #1624 first, or is this PR supposed to supersede that one?

This fix is supposed to go into #1624 (once it's working, of course :) ); I opened it as a separate branch so I could try things out in parallel.

tee-worker/ts-tests/common/utils.ts
```diff
@@ -777,10 +777,10 @@ export async function assertInitialIDGraphCreated(api: ApiPromise, signer: Keyri
     // check identity in idgraph
     const expected_identity = api.createType(
         'LitentryIdentity',
-        await buildIdentityHelper(u8aToHex(signer.addressRaw), 'LitentryRococo', 'Substrate')
+        await buildIdentityHelper(u8aToHex(signer.addressRaw), 'TestNet', 'Substrate')
```
arieljbirnbaum:
```suggestion
        await buildIdentityHelper(u8aToHex(signer.addressRaw), 'LitentryRococo', 'Substrate')
```

tee-worker/ts-tests/identity.test.ts
```diff
@@ -101,21 +101,25 @@ describeLitentry('Test Identity', 0, (context) => {
...
+        await sleep(6000);
         assert.equal(resp_shieldingKey, aesKey, 'resp_shieldingKey should be equal aesKey after set');
     });
 
     step('check idgraph from sidechain storage before create', async function () {
         // the main address should be already inside the IDGraph
         const main_identity = await buildIdentityHelper(
             u8aToHex(context.substrateWallet.alice.addressRaw),
-            'LitentryRococo',
+            'TestNet',
```
arieljbirnbaum:
```suggestion
            'LitentryRococo',
```

tee-worker/ts-tests/identity.test.ts
```diff
@@ -570,7 +574,7 @@ describeLitentry('Test Identity', 0, (context) => {
         // remove prime identity
         const substratePrimeIdentity = await buildIdentityHelper(
             u8aToHex(context.substrateWallet.alice.addressRaw),
-            'LitentryRococo',
+            'TestNet',
```
arieljbirnbaum:
```suggestion
            'LitentryRococo',
```

Kailai-Wang:
I'll leave it to @Verin1005 to review as he is the one who is the most familiar.

`BatchAll` events should be nothing more than a sequential execution of all calls inside a batch. I select a random such call on polkadot: https://polkadot.statescan.io/#/extrinsics/15636868-4?tab=events&page=1

arieljbirnbaum:
> I'll leave it to @Verin1005 to review as he is the one who is the most familiar.
> 
> `BatchAll` events should be nothing more than a sequential execution of all calls inside a batch. I select a random such call on polkadot: https://polkadot.statescan.io/#/extrinsics/15636868-4?tab=events&page=1

@Kailai-Wang that's a helpful example, thanks!

0xverin:
> Attempt to address the issue in [this comment](https://github.com/litentry/litentry-parachain/pull/1624#issuecomment-1534374083).
> 
> @Verin1005 I tried to match events to extrinsics [the canonical way](https://polkadot.js.org/docs/api/cookbook/blocks#how-do-i-map-extrinsics-to-their-events); the main issue I kept running into was being unable to determine the index within the block for individual extrinsics bundled in a single `batchAll` call inside `sendTxsWithUtility`.
> 
> The closest I could find in terms of documentation was [this question about `batch` on StackExchange](https://substrate.stackexchange.com/questions/174/correlate-events-with-extrinsics-in-a-utility-batch-call-in-polkadot), but I can't quite tell if `batchAll` emits the same set of events :P
> 
> To unblock the fix for #1571, I'm just restoring the previous version of `listenEvent` (modulo a tiny refactoring).

1. The `utility.batchAll` can be sorted in order. [This issue](https://github.com/polkadot-js/api/issues/4683) should be able to answer the question, so I don't think we need to worry about its external index association. Therefore, we don't need to worry about whether it returns in order or not. If I understand this issue correctly, it is actually already returning in order, as long as the filtering is done correctly.

1. What I'm actually concerned about is event filtering when we have concurrent transactions from multiple accounts. In the ts-test, we use [promise.all](https://github.com/litentry/litentry-parachain/blob/1571-try-fix-event-matching/tee-worker/ts-tests/common/transactions.ts#L34).For example, if we have 10 accounts simultaneously sending transactions, `promise.all` does not guarantee sequential execution or sequential return events,so we encounter error when parsing event data. Although most cases are in order, according to this [CI error](https://github.com/litentry/litentry-parachain/actions/runs/4880700297/jobs/8708752102?pr=1624), we may encounter special cases where returns `BOB` event first. That's why I added sorting based on `signer.address` in the event. This is one solution, but since I came up with it myself, I don't think it's very elegant. I wanted to find an official solution for event filtering in the case of multiple account transactions, but I couldn't find one.

PR #1665, opened 2023-05-06
ts-test use generated types

### LINKED ISSUE #1644: Ts-test uses automatically generated types.
Now that the ts-test is able to generate types automatically, we need to update any custom types defined in the ts-test to use the automatically generated ones. refer to #1345

0xverin:
resolves #1644 
Most of the generated types still need to be redefined manually, for example, we need to manually convert `bytes` to `string`. The only type that can be used directly from the generated types is the network type.

### DIFF
--- a/tee-worker/ts-tests/bulk_identity.test.ts
+++ b/tee-worker/ts-tests/bulk_identity.test.ts
@@ -15,6 +15,7 @@ import { handleIdentityEvents } from './common/utils';
 import { assert } from 'chai';
 import { listenEvent, multiAccountTxSender } from './common/transactions';
 import { u8aToHex } from '@polkadot/util';
+import { ApiTypes, SubmittableExtrinsic } from '@polkadot/api/types';
 
 //Explain how to use this test, which has two important parameters:
 //1.The "number" parameter in describeLitentry represents the number of accounts generated, including Substrate wallets and Ethereum wallets.If you want to use a large number of accounts for testing, you can modify this parameter.
@@ -35,7 +36,7 @@ describeLitentry('multiple accounts test', 10, async (context) => {
         });
     });
     step('send test token to each account', async () => {
-        const txs: any = [];
+        const txs: SubmittableExtrinsic<ApiTypes>[] = [];
         for (let i = 0; i < substraetSigners.length; i++) {
             //1 token
             const tx = context.api.tx.balances.transfer(substraetSigners[i].address, '1000000000000');
--- a/tee-worker/ts-tests/common/call.ts
+++ b/tee-worker/ts-tests/common/call.ts
@@ -10,9 +10,10 @@ export async function sendRequest(
     api: ApiPromise
 ): Promise<WorkerRpcReturnValue> {
     const resp = await wsClient.sendRequest(request, { requestId: 1, timeout: 6000 });
-    const resp_json = api.createType('WorkerRpcReturnValue', resp.result).toJSON() as unknown as WorkerRpcReturnValue;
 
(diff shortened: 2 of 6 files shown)


tee-worker/ts-tests/bulk_identity.test.ts
```diff
@@ -35,7 +36,7 @@ describeLitentry('multiple accounts test', 10, async (context) => {
         });
     });
     step('send test token to each account', async () => {
-        const txs: any = [];
```
arieljbirnbaum, line 38: <3

tee-worker/ts-tests/common/call.ts
```diff
@@ -1,25 +1,24 @@
...
+import { RequestBody, WorkerRpcReturnValue } from '../common/type-definitions';
 //rpc call
 export async function sendRequest(
     wsClient: WebSocketAsPromised,
-    request: any,
+    request: RequestBody,
     api: ApiPromise
 ): Promise<WorkerRpcReturnValue> {
     const resp = await wsClient.sendRequest(request, { requestId: 1, timeout: 6000 });
-    const resp_json = api.createType('WorkerRpcReturnValue', resp.result).toJSON() as WorkerRpcReturnValue;
```
arieljbirnbaum, line 14: Wondering, shouldn't we validate the payload here to make sure it conforms to the type? :thinking:

tee-worker/ts-tests/common/utils.ts
```diff
@@ -416,7 +419,7 @@ export async function checkUserShieldingKeys(
     };
     let resp = await sendRequest(context.tee, request, context.api);
 
-    return resp.value;
```
arieljbirnbaum: Same here; I think it'd be safer to throw if `typeof resp.value !== 'string'` :thinking:

arieljbirnbaum (approved):

Kailai-Wang:
Shall we resolve the questions before merging the PR, or can this be done afterwards?

arieljbirnbaum:
> Shall we resolve the questions before merging the PR, or can this be done afterwards?

I don't think it's a blocker, but if it's not too hard (and I don't think it should be; if in doubt we could use [Zod](https://zod.dev/)), it might make sense to add it in right away so we don't forget later :)

tee-worker/ts-tests/common/utils.ts
Verin1005: I modified the usage of WorkerRpcReturnValue, it seems that we still need to manually define it here, as using the generated one is not good enough.

tee-worker/ts-tests/common/call.ts
0xverin: I modified the usage of `WorkerRpcReturnValue`, it seems that we still need to manually define it here, as using the generated has other types that we don't need.

tee-worker/ts-tests/common/utils.ts
Verin1005: Is it better to have this comment in the modified code? Otherwise, it might be confusing to read this part directly.Just a small friendly reminder

tee-worker/ts-tests/common/call.ts
0xverin: btw;Is it better to have this comment in the modified code? Otherwise, it might be confusing to read this part directly.Just a small friendly reminder.:)

arieljbirnbaum: This comment was prompted by the `as WorkerRpcReturnValue`, so I put it on that line :D

PR #1671, opened 2023-05-09
Add maximum length for IDGraph

### LINKED ISSUE #1648: Have a maximum length for IDGraph
### Context

Currently we don't have an upper bound for IDGraph vector, but we do have a limit when pushing the it into `vec_identity` when building assertions.

It makes less sense if identities in IDGraph can't participate in the assertion building. IMO it's better to cap it on IDGraph instead of `vec_identity`

https://github.com/litentry/litentry-parachain/blob/93737c4164463966e99e937d78df50dd9d5837d8/tee-worker/app-libs/stf/src/trusted_call_litentry.rs#L217-L228

---

kziemianek:
resolves #1648 

adds a limit of created identities per account


kziemianek:
In the next commit I'll remove limit from the `request_vc` function.

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -73,6 +73,9 @@ pub mod pallet {
 		/// maximum delay in block numbers between creating an identity and verifying an identity
 		#[pallet::constant]
 		type MaxVerificationDelay: Get<ParentchainBlockNumber>;
+		/// maximum number of identities an account can have, if you change this value to lower some accounts may exceed this limit
+		#[pallet::constant]
+		type AccountIdentitiesLimit: Get<u32>;
```
kziemianek: Right now account can have in total (limit + 1) identities because prime identity is not counted. I think we should count it too, then user will be able to create (limit -1) identities because prime is created automaticaly while shielding key is set.

kziemianek: Then we will be able to simplify id_graph_stats logic and use counters there.

0xverin: One question, if we limit the quantity, then the number of identities we retrieve from idgraph will be limited to this quantity, and will no longer be the latest 20 identities like before, is that correct?Then there is another important issue to consider. The value of the limit is adjustable. For example, if 20 identities have already been set and the limit is then changed to 10, what happens? Will I no longer be able to use the identities I set before? Maybe this is a important issue that needs careful consideration.

kziemianek: The scope of changes in this PR is mainly focused on preventing account from creating more idenitites than given limit.
In my opinion getter call used to retrieve latest 20 identities shouldn't be affected by these changes. How are they requested ?
In case of lowering the limit, you will be still able to use your 20 identities. At the worst scenario some accounts may have more identities than allowed, maybe we should consider different naming if it's confusing.

tee-worker/app-libs/sgx-runtime/src/lib.rs
```diff
@@ -274,6 +274,7 @@ impl pallet_imt::Config for Runtime {
 	type ManageOrigin = EnsureRoot<AccountId>;
 	type MaxMetadataLength = ConstU32<128>;
 	type MaxVerificationDelay = ConstU32<{ 30 * MINUTES }>;
+	type AccountIdentitiesLimit = ConstU32<64>;
```
Kailai-Wang: About the naming: how about `MaxIDGraphLength` or `MaxIdentityLength`? Just to be consistent with other limits `Max...`

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -317,6 +326,28 @@ pub mod pallet {
 	}
 
 	impl<T: Config> Pallet<T> {
+		fn insert_identity_with_limit(
+			owner: &T::AccountId,
+			identity: &Identity,
+			context: IdentityContext<T>,
+		) -> Result<(), ()> {
```
Kailai-Wang: Maybe it's simpler to have the same error as the extrinsic so we don't have to map_err, like `DispatchResult`

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -317,6 +326,28 @@ pub mod pallet {
...
+			let count = AccountIdentityContextCount::<T>::get(owner);
+			if count >= T::AccountIdentitiesLimit::get() {
+				return Err(())
+			}
+			// return Err in case of overflow
+			let new_count = count.checked_add(1).ok_or(())?;
+			AccountIdentityContextCount::<T>::insert(&owner, new_count);
+			IDGraphs::<T>::insert(&owner, &identity, context);
```
Kailai-Wang: `owner` is already a reference, we don't need to take it again :)

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -317,6 +326,28 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		fn remove_identity_with_limit(owner: &T::AccountId, identity: &Identity) {
+			let count = AccountIdentityContextCount::<T>::get(owner);
```
Kailai-Wang: Maybe it's still safer to check the existence of `owner` in `AccountIdentityContextCount`, or use `try_mutate`. The reason is _theoretically_ `AccountIdentityContextCount` and `IDGraph` don't need to be always in sync - imagine we forget to modify these two maps simultaneously in later code, such check could help us to discover the problem earlier

tee-worker/litentry/pallets/identity-management/src/tests.rs
```diff
@@ -71,6 +71,33 @@ fn create_identity_works() {
 	});
 }
 
+#[test]
+fn cannot_create_more_identites_for_account_than_limit() {
+	new_test_ext(true).execute_with(|| {
+		for i in 1..65 {
```
Kailai-Wang: it might be better to use the const type than the hardcoded `65`

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -317,6 +326,28 @@ pub mod pallet {
 	}
 
 	impl<T: Config> Pallet<T> {
+		fn insert_identity_with_limit(
```
Kailai-Wang, line 330: we should use this fn too when inserting the prime identity => so that the `AccountIdentityContextCount` gets incremented. Ideally we could have a test case for it :)

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -317,6 +326,28 @@ pub mod pallet {
...
+			identity: &Identity,
+			context: IdentityContext<T>,
+		) -> Result<(), ()> {
+			let count = AccountIdentityContextCount::<T>::get(owner);
+			if count >= T::AccountIdentitiesLimit::get() {
+				return Err(())
+			}
+			// return Err in case of overflow
+			let new_count = count.checked_add(1).ok_or(())?;
+			AccountIdentityContextCount::<T>::insert(&owner, new_count);
```
Kailai-Wang: `try_mutate` might have a better readability for this logic

Kailai-Wang (commented):
Thanks! The logic is basically fine but I still see a few small problems

Kailai-Wang (commented):
Thanks! The logic is basically fine but I still see a few small problems

tee-worker/litentry/pallets/identity-management/src/lib.rs
kziemianek: You mean something like this (meanwhile i've renamed few things) ?
```
		fn remove_identity_with_limit(owner: &T::AccountId, identity: &Identity) {
			let graph_len_result = IDGraphLens::<T>::try_get(owner);
			if let Ok(graph_len) = graph_len_result {
				if graph_len == 0 {
					warn!(
						"Detected IDGraphLens inconsistency, found len 0 while removing identity"
					);
					IDGraphLens::<T>::remove(owner);
				} else {
					let new_graph_len = graph_len - 1;
					if new_graph_len == 0 {
						IDGraphLens::<T>::remove(owner);
					} else {
						IDGraphLens::<T>::insert(owner, new_graph_len);
					}
				}
			} else {
				warn!("Detected IDGraphLens inconsistency, missing IdGraphLen while removing identity");
			}
			IDGraphs::<T>::remove(&owner, &identity);
		}
```

tee-worker/litentry/pallets/identity-management/src/lib.rs
kziemianek: I'll add relevant storage assertions in tests

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -317,15 +327,51 @@ pub mod pallet {
...
 	impl<T: Config> Pallet<T> {
-		pub fn get_id_graph(who: &T::AccountId) -> Vec<(Identity, IdentityContext<T>)> {
+		fn insert_identity_with_limit(
+			owner: &T::AccountId,
+			identity: &Identity,
+			context: IdentityContext<T>,
+		) -> Result<(), DispatchError> {
+			IDGraphLens::<T>::try_mutate(owner, |len| {
+				let new_len = len.checked_add(1).ok_or(Error::<T>::IDGraphLenLimitReached)?;
+				if new_len > T::MaxIDGraphLength::get() {
```
BillyWooo, line 337: Why need to check again? I think previous line already checked.

kziemianek: Not really. The previous line checks only for overflow, while the second checks against the defined limit.

Kailai-Wang:
You got CI passed in the first submission, that was awesome 👍 

Could you please:
- resolve confilicts
- resolve the review conversation that you believe is done

Then I'll have a re-review :)

kziemianek:
> You got CI passed in the first submission, that was awesome +1
> 
> Could you please:
> 
> * resolve confilicts
> * resolve the review conversation that you believe is done
> 
> Then I'll have a re-review :)

Thank you Kai.
Yes, I'll do this and I'll re-request all reviewers once everything that is planned is done.

kziemianek:
This is only an improvement of the code to check the limit as soon as possible.
Some accounts may have more identities than the limit after lowering the limit - I suggest leaving it as it is, for now it shouldn't be a problem.

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -317,15 +327,51 @@ pub mod pallet {
...
+					if new_graph_len == 0 {
+						IDGraphLens::<T>::remove(owner);
+					} else {
+						IDGraphLens::<T>::insert(owner, new_graph_len);
+					}
+				}
+			} else {
+				warn!("Detected IDGraphLens inconsistency, missing IdGraphLen while removing identity");
+			}
+			IDGraphs::<T>::remove(owner, identity);
```
Kailai-Wang: Maybe this can be further simplified using `try_mutate_exists`

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -317,15 +327,51 @@ pub mod pallet {
...
+				let new_len = len.checked_add(1).ok_or(Error::<T>::IDGraphLenLimitReached)?;
+				if new_len > T::MaxIDGraphLength::get() {
+					return Err(Error::<T>::IDGraphLenLimitReached.into())
+				}
+				*len = new_len;
+				Result::<(), DispatchError>::Ok(())
+			})?;
+			IDGraphs::<T>::insert(owner, identity, context);
+			Ok(())
+		}
```
Kailai-Wang, line 345: `IDGraphs::<T>::insert` could be put inside the `try_mutate` block to save a `?` and return value :)

Kailai-Wang (approved):
Looks good, thanks!

I only found two minor issues - but we can leave them as it is too.

Kailai-Wang (approved):
Looks good, thanks!

I only found two minor issues - but we can leave them as it is too.

tee-worker/litentry/pallets/identity-management/src/lib.rs
kziemianek: Maybe, but for clarity I prefer to put in `try_mutate` only the logic related to mutating that `Map`, not any other.

Kailai-Wang: Yes, I misread it sorry

tee-worker/litentry/pallets/identity-management/src/lib.rs
kziemianek: done 🙃

Kailai-Wang (approved):

tee-worker/ts-tests/common/utils.ts
```diff
@@ -523,6 +529,9 @@ export async function handleIdentityEvents(
             case 'Failed':
                 results.push(events[index].data.detail.toHuman());
                 break;
+            case 'CreateIdentityFailed':
```
0xverin, line 530: Maybe can remove `CreateIdentityFailed`, as `Failed` has a standardized parsing format.

kziemianek: Ok, but I still need to add `case 'CreateIdentityFailed':`, otherwise it's not matched and no events are returned from `handleIdentityEvents`. Next, if I call `checkErrorDetail` with an empty list of events it gives a false positive result, I don't know if there was an error with the given message or not.

PR #1674, opened 2023-05-10
fix: scripts to run tee-worker ts tests on a SGX machine

wli-pro:


tee-worker/README.md
```diff
@@ -42,6 +42,22 @@ make
 ### Integration Tests
 See [docker/README.md](docker/README.md)
 
+### Quickly launch parachain, tee worker and run ts-test against tee-worker
+
+```
+source "$HOME/.cargo/env"
```
kziemianek: you only need to run it once after rustup installation so I think it is not necessary to include it here.

tee-worker/README.md
```diff
@@ -42,6 +42,22 @@ make
...
+source /opt/intel/sgxsdk/environment
```
kziemianek: I suggest adding comment that location of this file might differ across environments

tee-worker/README.md
```diff
@@ -42,6 +42,22 @@ make
...
+# run clean if only SGX_MODE change
+make clean
+make SGX_MODE=SW
+./local-setup/launch.py local-setup/github-action-config-one-worker.json
+# run script below if container started
```
kziemianek: i'm not sure about the rest part because I stopped once I checked the logs

tee-worker/README.md
wli-pro: added it as comment in case new beginner doesn't have it in bash profile.

tee-worker/README.md
wli-pro: added the comment now.

tee-worker/README.md
```diff
@@ -42,6 +42,24 @@ make
...
+source /opt/intel/sgxsdk/environment
+# run clean if only SGX_MODE change
+make clean
+make SGX_MODE=SW
+./local-setup/launch.py local-setup/github-action-config-one-worker.json
+# run the ts test script below if container started
+cd  ./tee-worker/ts-tests
+yarn
+yarn run test-identity:local
+# other ts test
```
kziemianek: I don't know what you mean by this comment 🙃

kziemianek (approved):

Kailai-Wang:
Thank you! Could you please shift this section to the very beginning of this README - before the original Integritee doc (so that we know it's from us)

This README needs to be better structured as well - but there's another issue for it

Issue #1673, opened 2023-05-09
Use augmented and generated types in tee-worker/ts-tests


Kailai-Wang:
### Context

If I understand it correctly we are not using the auto-generated types in any case:
- neither the types generated from the metadata,
- nor the types generated from the definition

We are using the "raw" definitions that we add manually, see `tee-worker/ts-tests/common/utils.ts`:

```
    // teeTypes is imported from './type-definitions';
    const { types } = teeTypes;
    const api = await ApiPromise.create({
        provider,
        types,
    });
```

This can be very error-prone and needs to be rectified.

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

PR #1675, opened 2023-05-10
Fix the sidechain metadata and type generation

### LINKED ISSUE #1672: Wrong metadata in tee-worker ts-tests
### Context

It seems the current [litentry-metadata.json](https://github.com/litentry/litentry-parachain/blob/93715df9c4da0290b6082a1e5552c6ffab53524c/tee-worker/ts-tests/litentry-metadata.json) was generated from an integritee-node. Additionally the generation command hooks with the 9944 port, which is wrong.

---

Kailai-Wang:
resolves #1672 

It should give you the real sidechain metadata. The augmented types are re-generated too using `"@polkadot/typegen": "^9.14.2"`.

TODO:
- upgrade the polkadot/* package dependencies and re-generate the types (we might need some adjustments)
- #1673


tee-worker/core/rpc-client/src/mock.rs
```diff
@@ -95,6 +95,10 @@ impl DirectApi for DirectClientMock {
 		RuntimeMetadataPrefixed::decode(&mut metadata.as_slice()).map_err(|e| e.into())
 	}
 
+	fn get_state_metadata_raw(&self) -> Result<String> {
+		unimplemented!()
```
kziemianek, line 99: Have you considered returning `self.metadata` or maybe we don't need this mock ?

kziemianek (approved):
I generated those types locally and recieved the same output.

tee-worker/core/rpc-client/src/mock.rs
Kailai-Wang: I'm not aware it's used anywhere - so it's just to make it compile

PR #1676, opened 2023-05-10
Direct invocation demo code

### LINKED ISSUE #1658: Provide direct invocation e2e sample code
### Context

Including:

- [x] F/E code to construct such a request
- [x] handling logic inside the worker
- [ ] response emission inside the worker - is it better to do it synchronously or via parachain event?
- [x] F/E code to handle the response

---

Kailai-Wang:
This PR has addressed most of #1658 

It shows how to construct a direct trusted call and send it, currently supported trusted call:
- `set_user_shielding_key_direct`
- `create_identity_direct`

It also demonstrates the public query to e.g. retrieve TEE's shielding key.

It includes a basic RPC response handling as well.
The subscription to parachain header and the subsequent handling are not written as we have the code already I believe.

Please bear with my crude ts code.

See `ts-tests/examples/direct-invocation/README.md` for more information.

cc @jonalvarezz


arieljbirnbaum (approved):

BillyWooo (approved):

tee-worker/ts-tests/examples/direct-invocation/index.ts
```diff
@@ -0,0 +1,96 @@
...
+// in order to handle self-signed certificates we need to turn off the validation
+// TODO add self signed certificate
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+
+const WebSocketAsPromised = require('websocket-as-promised');
+const WebSocket = require('ws');
+const keyring = new Keyring({ type: 'sr25519' });
+
+const PARACHAIN_WS_ENDPINT = 'ws://localhost:9944';
+const WORKER_TRUSTED_WS_ENDPOINT = 'wss://localhost:2000';
```
jonalvarezz, line 24: If I understood correctly from Nitesh some time ago, this endpoint is not publicly available for security reasons. Is it still the case?

tee-worker/ts-tests/examples/direct-invocation/util.ts
```diff
@@ -0,0 +1,194 @@
...
+) => {
+    const [variant, argType] = trustedCall;
+    const call = parachain_api.createType('TrustedCall', {
+        [variant]: parachain_api.createType(argType, params),
+    });
+    const payload = Uint8Array.from([
+        ...call.toU8a(),
+        ...nonce.toU8a(),
+        ...hexToU8a(mrenclave),
+        ...hexToU8a(mrenclave), // should be shard, but it's the same as MRENCLAVE in our case
```
jonalvarezz, line 76: "in our case" here means our Parachain regardless of the network, right? How likely it is to change in the future?

jonalvarezz (commented):
Thanks for the examples, If that's all what it takes, it doesn't look complex 🙌

tee-worker/ts-tests/examples/direct-invocation/index.ts
Kailai-Wang: See https://github.com/litentry/litentry-parachain/issues/1010 
We don't see security issues - only server load (similar to any other rpc service provider)

tee-worker/ts-tests/examples/direct-invocation/util.ts
Kailai-Wang: Yes it's network unrelated.
> How likely it is to change in the future?

Highly improbable, even if it's going to change, that's not in near future

tee-worker/ts-tests/examples/direct-invocation/util.ts
```diff
@@ -0,0 +1,194 @@
...
+// given an encoded trusted operation, construct a request bytes that are sent in RPC request parameters
+export const createRequest = async (
+    wsp: WebSocketAsPromised,
+    parachain_api: ApiPromise,
+    mrenclave: string,
+    teeShieldingKey: KeyObject,
+    top: Uint8Array
+) => {
+    let cyphertext = compactAddLength(bufferToU8a(encryptWithTeeShieldingKey(teeShieldingKey, top)));
+    return parachain_api.createType('Request', { shard: hexToU8a(mrenclave), cyphertext }).toU8a();
```
jonalvarezz, line 193: @Kailai-Wang, `cyphertext` -> `ciphertext` – Is it a typo?

Kailai-Wang: No - In rust the struct field is `cyphertext` and it's from upstream - so we didn't change it.

And according to Bing, cipher and cypher should be alternative forms:
image

jonalvarezz: IDHub uses `ciphertext` 😭😭😭

PR #1677, opened 2023-05-11
Add shellcheck and fix warnings

### LINKED ISSUE #1670: Fix shell scripts warnings
### Context
Similier to `rust clippy` command, there's a static analysis tool for shell scripts too.  [shellcheck ](https://github.com/koalaman/shellcheck). It's simple to use  `shellcheck xxx.sh`. More further, we can add it to `Makefile`, for example
```Makefile
check-scripts:
    # Fail if any of these files have warnings
    shellcheck scripts/*.sh
```

shell scripts contains in 
1. scripts
2. tee-work/scripts

Here's some warnings and suggestions from scripts:

```shell

Did you mean: 
$PARACHAIN_BIN --alice --collator --force-authoring --tmp --chain "$CHAIN"-dev \


In launch-local-binary.sh line 105:
  --bootnodes /ip4/127.0.0.1/tcp/30336/p2p/$RELAY_ALICE_IDENTITY &> "para.alice.log" &
                                           ^-------------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
  --bootnodes /ip4/127.0.0.1/tcp/30336/p2p/"$RELAY_ALICE_IDENTITY" &> "para.alice.log" &


In launch-local-binary.sh line 109:
cd "$ROOTDIR/ts-tests"
^--------------------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

Did you mean: 
cd "$ROOTDIR/ts-tests" || exit


In launch-local-binary.sh line 118:
cd "$ROOTDIR/ts-tests"
^--------------------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

Did you mean: 
cd "$ROOTDIR/ts-tests" || exit
```

Kailai-Wang:
resolves #1670 

It should be straightforward. The severity level is set to `warning` to avoid spam.

### DIFF
--- a/scripts/generate-upstream-patch.sh
+++ b/scripts/generate-upstream-patch.sh
@@ -2,7 +2,6 @@
 
 set -eo pipefail
 
-
 cleanup() {
 	rm -rf "$1"
 	echo "cleaned up $1"
@@ -10,10 +9,12 @@ cleanup() {
 
 print_help() {
 	echo "Usage:"
-	echo "	$0 [-p <tag|branch|commit-hash>] [-w <tag|branch|commit-hash>]"
-	echo "	without any parameter, the script will generate upstream patch for both pallets and tee-worker"
-	echo "		-p specify the tag|branch|commit-hash for upstream pallets"
-	echo "		-w specify the tag|branch|commit-hash for upstream worker"
+	echo "$0 [-h] [-p <tag|branch|commit-hash>] [-w <tag|branch|commit-hash>]"
+	echo ""
+	echo "without any parameter, the script will generate upstream patch for both pallets and tee-worker"
+	echo "	-h print this message"
+	echo "	-p specify the tag|branch|commit-hash for upstream pallets"
+	echo "	-w specify the tag|branch|commit-hash for upstream worker"
 }
 
 check_upstream() {
@@ -52,7 +53,8 @@ generate_upstream_patch() {
 	echo "fetch upstream_$TARGET"
 	git fetch -q "upstream_$TARGET"
 
-	local tmp_dir=$(mktemp -d)
+	local tmp_dir
+	tmp_dir=$(mktemp -d)
 	cd "$tmp_dir"
 	echo "cloning $UPSTREAM_URL to $tmp_dir"
 	git clone -q $UPSTREAM_URL repo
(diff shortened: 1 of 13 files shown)


kziemianek:
Can we run these checks as a part of `scripts/pre-commit.sh` ?

Kailai-Wang:
> Can we run these checks as a part of `scripts/pre-commit.sh` ?

I could add it - it's not enforced by CI tho, as shell script is just tooling script, it either works or fails. Normally we are fine if it does the job.

Edit: done.

scripts/generate-upstream-patch.sh
```diff
@@ -52,7 +53,8 @@ generate_upstream_patch() {
 	echo "fetch upstream_$TARGET"
 	git fetch -q "upstream_$TARGET"
 
-	local tmp_dir=$(mktemp -d)
+	local tmp_dir
+	tmp_dir=$(mktemp -d)
```
arieljbirnbaum, line 57: Really? That's an odd convention imho but if it makes the bot happy then okay :D

scripts/geth/run_geth.sh
```diff
@@ -57,7 +57,7 @@ start="${GETH_BIN} --datadir ${DATADIR} \
     --miner.etherbase 0xff93B45308FD417dF303D6515aB04D9e89a750Ca
 "
 if [ ${DOCKER} = "local" ]; then
-    eval ${start} 2>&1 &>'geth.log' &
+    eval ${start} &>'geth.log' &
```
arieljbirnbaum, line 60: Just to make sure; if we're running locally then we want the output stream to go into `geth.log` and the error stream to go into the terminal?

scripts/generate-upstream-patch.sh
Kailai-Wang: Yea: https://www.shellcheck.net/wiki/SC2155

scripts/geth/run_geth.sh
BillyWooo: Similar question: why remove it?

Kailai-Wang: doesn't `&>` redirect both stdout and stderr to `geth.log`? It supersedes the previous redirection. So `2>&1` is redundant here

BillyWooo: Yes. You are right. Merged into one step is enough.

BillyWooo (approved):

PR #1678, opened 2023-05-11
Using the timestamp in parentchain block as VC's issuance timestamp

### LINKED ISSUE #1282: Using the timestamp in parentchain block as VC's issuance timestamp
Currently the block number is used as VC's issuance timestamp. This needs to be changed to the timestamp in parentchain blocks.

outofxxx:
resolved #1282 

Three fields in VC content may affect the F/E:
1. issuanceBlockNumber   ->  issuanceTimestamp
2. expirationBlockNumber -> expirationTimestamp
4. createdBlockNumber     -> createdTimestamp


Kailai-Wang:
May I know the background of this PR? It seems we don't need anything from the parachain, if we need timestamps, we can simply get it locally.

outofxxx:
> 

As described in the issue, the `blocknumber` in the VC needs to be replaced with `timestamp`. The timestamp here can be from the parentchain or sidechain. This pr implements the timestamp extrinsic parsed from the parentchain block.

tee-worker/core/parentchain/indirect-calls-executor/src/lib.rs
```diff
@@ -300,7 +300,7 @@ where
 			// Found VerifyIdentity extrinsic
 			let verify_identity = VerifyIdentity { block_number: parentchain_block_number };
 			// Found RequestVC extrinsic
-			let request_vc = RequestVC { block_number: parentchain_block_number };
+			let request_vc = RequestVC;
 			// Found BatchAll extrinsic
 			let batch_all = BatchAll { block_number: parentchain_block_number };
```
BillyWooo, line 305: do we still block number here too?

tee-worker/litentry/core/credentials/src/lib.rs
```diff
@@ -229,44 +229,33 @@ pub struct Credential {
...
+		debug!("generate credential from template, who: {:?}", account_id_to_string(&who),);
 
 		let mut vc: Self =
 			serde_json::from_str(s).map_err(|err| Error::ParseError(format!("{}", err)))?;
 		vc.issuer.mrenclave = shard.encode().to_base58();
 		vc.issuer.name = LITENTRY_ISSUER_NAME.to_string();
 		vc.credential_subject.id = account_id_to_string(who);
-		vc.issuance_block_number = bn;
-		vc.expiration_block_number = None;
+		vc.issuance_timestamp = now_as_millis();
```
BillyWooo, line 246: Underneath of `now_as_millis()`, there is SystemTime::now(). So will this part of code be running within Enclave? Is it still secure?

BillyWooo (commented):
In general, I agreed with this approach to get local timestamp, instead of block number. Just worry about the security concern within one of my comments.

BillyWooo (commented):
In general, I agreed with this approach to get local timestamp, instead of block number. Just worry about the security concern within one of my comments.

tee-worker/core/parentchain/indirect-calls-executor/src/lib.rs
outofxxx: No longer needed in `RequestVC`

tee-worker/litentry/core/credentials/src/lib.rs
outofxxx: I think it is ok, sidechain also uses this method to generate timestamp

BillyWooo (approved):

BillyWooo:
Please resolve the conflicts, then merge it. Thanks.

Issue #1684, opened 2023-05-11
Identity ts-test fails to detect `wrongWeb2Handle` error


Kailai-Wang:
### Context

A bug discovered when fixing #1680:

I intentionally failed the twitter identity verification, which can be seen in the screenshot below. But the identity ts-test didn't manage to detect it and happily continued.

<img width="1718" alt="Screenshot 2023-05-12 at 00 18 39" src="https://github.com/litentry/litentry-parachain/assets/7630809/77cf27ef-1ff6-4616-8dcc-124fcd816b42">

I believe the problem is 
```
const verified_event_datas = await handleIdentityEvents(context, aesKey, alice_resp_events, 'IdentityVerified');
```
only collects the `IdentityVerified` events.

And of course the assertion later would succeed :D
```
assertIdentityVerified(context.substrateWallet.alice, verified_event_datas);
```

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

wli-pro:
I am checking this one.

0xverin:
The console printing is just for display purposes and does not have any other significance. It also does not indicate that the "ts-test" is continuing execution. Therefore, it will still eventually `timeout`. I suggest fixing #1624 first and then merging. cc @grumpygreenguy  This way, this issue will be resolved as well.

#1624 testing result:
<img width="1538" alt="image" src="https://github.com/litentry/litentry-parachain/assets/104152026/0809c04d-0c4a-46de-a054-b103cc89e188">

PR #1685, opened 2023-05-11
Twitter handle verification should be case insensitive

### LINKED ISSUE #1680: Twitter username should be case insensitive
### Context

As topic, we should take that into consideration when verifying the Twitter identity.

---

Kailai-Wang:
resolves #1680 

There is still a bug in ts-test: #1684 
It should have failed with the old implementation but it didn't.


BillyWooo:
We need to keep an eye on this topic: https://beebom.com/discord-username-removes-four-digits/

BillyWooo (approved):

PR #1695, opened 2023-05-15
fix: resolve port conflicts with ENV vars 

### LINKED ISSUE #1661: Resolve port conflicts to allow multiple dev instances per machine
### Context

The port numbers (and other env constants) are hardcoded; that makes it impossible for multiple instances to run on the same machine. This is a problem if developers need to share a machine.

### Task

- Find all the hardcoded conflicting constants
- If possible, allocate them automatically to avoid conflicts
- If the above doesn't work, override them via a local config file

### Acceptance criteria

- At least two developers can run and test the parachain on the same server _simultaneously_ without conflicts.

felixfaisal:
Notion Page used for discussion: 
https://www.notion.so/web3builders/Port-Conflicts-Proposal-628f38941369483b8f004722a99d83bb?pvs=4

- All hardcoded ports have been replaced with Environment Variables 
- `.env.example` has been provided to set the `.env` ports 

WIP 
- [x] Working on port offseting

Please use this command for testing 
```
./local-setup/launch.py local-setup/development-worker.json dev 100
```
`development-worker.json` : config file for development setup with single worker and env vars 
`dev`: developer mode 
`100`: port offset from the original ports defined in `.env.dev` 

Resolves #1661


docker/rococo-parachain-launch-config.yml
```diff
@@ -13,13 +13,13 @@ relaychain:
     - --execution=wasm
   nodes:
     - name: alice
-      wsPort: 9946
-      rpcPort: 9936
-      port: 30336
+      wsPort: ${AliceWSPort:-9946}
+      rpcPort: ${AliceRPCPort:-9936}
+      port: ${AlicePort:-30336}
```
arieljbirnbaum, line 18: Adding the defaults here sounds a bit dangerous imo; someone can forget to set up their local env files and the defaults here will fall back to the Forbidden Ports instead of rightfully complaining :D

felixfaisal: This for the CI basically, if we don't add it here then we need to maintain separate config files for development and CI

arieljbirnbaum: > then we need to maintain separate config files for development and CI

This sounds like a Good Thing :tm: :)

ts-tests/tests/utils.ts
```diff
@@ -19,6 +19,7 @@ export function loadConfig() {
     require('dotenv').config();
     switch (process.env.NODE_ENV) {
         case 'development':
+            return require('../config.local.json')
```
BillyWooo, line 22: where is the file `config.local.json`? or how it should look like?

tee-worker/local-setup/launch.py
```diff
@@ -44,12 +50,54 @@ def run_worker(config, i: int):
...
+    ## Load environment file
+    load_dotenv('.env')
+    ## Check Ports and Automatically Reallocate
+    check_all_ports()
+
     print('Starting litentry-parachain in background')
 
     with open(config_path) as config_file:
         config = json.load(config_file)
+        config['workers'][0]['flags'] = [flag.replace('$CollatorWSPort', os.environ.get('CollatorWSPort', '')) for flag in config['workers'][0]['flags']]
```
BillyWooo: Looks like here only checks `worker0`. The config_file could have multiple worker configs.

tee-worker/local-setup/launch.py
```diff
@@ -44,12 +50,54 @@ def run_worker(config, i: int):
...
+    new_port = int(port) + int(OFFSET)
+    while not is_port_open(str(new_port)):
+        new_port = int(port) + int(OFFSET)
+
+    # Set the new port value in the environment variable
+    os.environ[env_name] = str(new_port)
+    print("Port for {} changed to: {}".format(env_name, os.environ.get(env_name)))
+
+# Function to iterate over all ports and automatically reallocate
+def check_all_ports():
```
BillyWooo: maybe think twice about the function name. It not only `check`, but also `reallocate`.

BillyWooo (commented):
Check reviewed the code here. `launch.py` in general looks good for me. I have left some comments. Will try to get a chance to try it on a dev machine.

kziemianek:
I tried to run it locally but it throws error:
```
Traceback (most recent call last):
  File "/home/kziemianek/projects/litentry/litentry-parachain/tee-worker/./local-setup/launch.py", line 168, in <module>
    if main(process_list, args.config, args.parachain) == 0:
  File "/home/kziemianek/projects/litentry/litentry-parachain/tee-worker/./local-setup/launch.py", line 90, in main
    check_all_ports()
  File "/home/kziemianek/projects/litentry/litentry-parachain/tee-worker/./local-setup/launch.py", line 77, in check_all_ports
    if is_port_open(os.environ.get(x)):
  File "/home/kziemianek/projects/litentry/litentry-parachain/tee-worker/./local-setup/launch.py", line 57, in is_port_open
    sock.bind(('127.0.0.1', int(port)))
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```
If there is any setup I'm missing please include instructions, I think I'm missing at least `.env` file.

felixfaisal:
> 

Yes, You are missing the `env` file therefore the above error

tee-worker/local-setup/.env.example
Kailai-Wang, line 1: It could be a flavour issue, but personally, the env file seems a bit superfluous - I could have changed the `one-worker.json` and `docker/rococo-parachain-launch-config.yml` directly if I want specific ports.

How about keeping `one-worker.json` unchanged and just adding the offset in py script? Additionally, maybe not everyone is used to the env file (if we don't want to run ts-tests).

Anyway, if we go for it, we probably need to add the copy-up into the script like `cp tee-worker/local-setup/.env.example tee-worker/local-setup/.env`, so that everyone could start the network with one line of command - instead of asking: where can I find the env file / how should I define the values.

felixfaisal: So I made changes, Now the script will also generate config files for TS Tests based on the same offset port, I also created a separate json called `development-worker.json` in `local-setup` and added an extra mode `dev` in `launch.py`, Also created `.env.dev`, Also the offset argument in CLI 
So if someone wants to start the system, they can just do this 
```
./local-setup/launch.py local-setup/development-worker.json dev 100
```
Here `dev` is the mode, and `100` is the offset value for the ports, if someone wants to explicitly use ports, they can edit `.env.dev` 

Let me know what you think?

ts-tests/tests/utils.ts
felixfaisal: This will be generated by `launch.py` it will look same as `config.example.json`

tee-worker/local-setup/launch.py
felixfaisal: So I moved this to only happen when in `dev` mode wherein we will only have a single worker, Let me know what you think

tee-worker/local-setup/launch.py
felixfaisal: I changed the function named to be `check_all_ports_and_reallocate()`

kziemianek:
Works for me, please update the documentation, for example:
https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/README.md?plain=1#L3-L19

Kailai-Wang:
It works basically, thanks! 🤝 

I've found a few small problems though:
-  the command-line arg`local` and `dev` seems duplicate - they are both connecting to a locally launched parachain/substrate-chain. Can we only have one of them?
- a follow-up problem: the stuff is not adequately cleaned up using `dev` - I mean the docker containers are not shut down - mostly because the `GracefulKiller` is only applied to `local`
- so far the solution works for parachain launching with docker. What about I changed some parachain code and want to launch it with native binary? The `launch-local-binary.sh` might needs to be adjusted.

arieljbirnbaum:
Another scenario to check is whether it'll work with `integritee-node` as parentchain: https://www.notion.so/web3builders/How-to-run-parachain-tee-worker-4a3332d470cc4aa29333b33bfa95fbfd?pvs=4#192275f2a9a34f0fa2d09d74ca42e419

(needed for faster local testing)

felixfaisal:
All the tested scenarios: 
- Launch local docker setup ✅ 
- Launch local binary setup ✅ 
- Launch local docker setup with offseting ports ✅ 
- Launch local binary setup with offseting ports ✅ 
- Use remote to connect to other node (Integritee) ✅

Kailai-Wang (approved):
Thanks!

I didn't try out every case myself, but it looks good according to the reported tests 🙌

PR #1701, opened 2023-05-16
Use local crates when compiling tee-worker

### LINKED ISSUE #1691: Use local crates when compiling tee-worker
### Context

I accidentally found that we've been using [upstream pallet crates](https://github.com/integritee-network/pallets) in the worker compilation, e.g.:

```
- teerex-primitives
- sgx-verify
- common-primitives
- ...
```

We should use the local crates, which are forked from upstream and can contain our changes.
There can be more places than listed above, please double-check it

---

outofxxx:
resolved: #1691 

After code inspection, all the crates used in the worker are local crates from pallets, except this crate(sgx-verify) is related to the dcap feature.

### DIFF
--- a/tee-worker/Cargo.lock
+++ b/tee-worker/Cargo.lock
@@ -752,14 +752,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "common-primitives"
-version = "0.1.0"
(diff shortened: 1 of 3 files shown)


Kailai-Wang (approved):
Looks good, thanks!

PR #1702, opened 2023-05-16
Implement direct invocations for all requests

### LINKED ISSUE #1681: Follow-up of DI: implement direct calls for all requests
### Context

In https://github.com/litentry/litentry-parachain/pull/1676, only `setUserShieldingKey` and `createIdentity` have a "direct" version. We should extend it to cover other requests as well.

---

Kailai-Wang:
resolves #1681 

As topic.

Some major changes that are worth noting:
- renaming: there're no `preflight` or `runtime` suffixes, no `direct` suffixes either. Reasons: we don't differentiate between direct and indirect calls on the processing level, they'll be equally handled with the same logic. For requests that require a two-step processing, e.g. identity verification - a callback trusted call is defined, e.g. `create_identity` and `create_identity_callback`. Those callbacks should never be triggered directly: we also have a signature guard against it.
- each trusted call always has an extra `AccountId` for the signature verification - it's also a preparation for the "delegation" functionality later - it means we _could_ allow others to send requests on behalf of the user, if authorised.
- refactoring: the logic is simplified, especially about the error handling. Each request has a `*_internal` helper fn to capture the errors and handle them on a upper level.


primitives/core/src/error.rs
```diff
@@ -30,13 +30,15 @@ pub type ErrorString = BoundedVec<u8, MaxStringLength>;
...
 	ImportError,
+	// the direct or indirect request comes from an unauthorised sender
+	UnauthorisedSender,
 	// generic error when executing STF, the `ErrorString` should indicate the actual reason
 	StfError(ErrorString),
-	// error when sending stf request to the receiver
+	// error when sending stf request to the receiver fails
 	SendStfRequestFailed,
-	// error when challenge code not found
+	// error when challenge code can not found
```
arieljbirnbaum: If we're going there :)

```suggestion
	// error when challenge code can not be found
```

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -471,124 +462,125 @@ where
...
+					Err(e) => {
+						debug!("pushing error event ... error: {}", e);
+						add_call_from_imp_error(
+							calls,
+							node_metadata_repo,
+							Some(account),
+							e.to_imp_error(),
+							hash,
+						);
+					},
```
arieljbirnbaum, line 502: Looks pretty similar across cases; can we factor it out? :thinking:

tee-worker/cli/src/trusted_base_cli/commands/litentry/set_user_shielding_key.rs
```diff
@@ -29,32 +29,31 @@ use log::*;
...
 		let mut key = UserShieldingKeyType::default();
 
 		hex::decode_to_slice(&self.key_hex, &mut key).expect("decoding shielding_key failed");
 
-		let top: TrustedOperation = TrustedCall::set_user_shielding_key_preflight(
-			root.public().into(),
-			who,
+		let top: TrustedOperation = TrustedCall::set_user_shielding_key(
+			who.public().into(),
+			who.public().into(),
```
arieljbirnbaum, line 52: Is this change as intended? :thinking:

Kailai-Wang: It is. For now it won' accept the root's signature.

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: That's already one-liner (of a function call). If we want to make it even shorter, then macro is probably the only way. But normally macro would reduce the readability, so IMO it's a style issue.

primitives/core/src/error.rs
Kailai-Wang: Thanks, fixed!

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
```diff
@@ -19,186 +19,111 @@ extern crate sgx_tstd as std;
...
-		let sender = StfRequestSender::new();
-		sender
-			.send_stf_request(request)
-			.map_err(|_| StfError::SetUserShieldingKeyFailed(ErrorDetail::SendStfRequestFailed))
-	}
-
-	pub fn set_user_shielding_key_runtime<NodeMetadataRepository>(
-		node_metadata_repo: Arc<NodeMetadataRepository>,
-		calls: &mut Vec<OpaqueCall>,
+	pub fn set_user_shielding_key_internal(
```
kziemianek, line 40: Maybe it's good time to refactor into struct-like variants :

```
pub enum TrustedCall {
...
	set_user_shielding_key(SetUserShieldingKeyCall),
...
}
...

#[derive(PartialEq, Encode, Decode, Clone, Debug, Eq)]
pub struct SetUserShieldingKeyCall {
	pub signer: AccountId,
	pub who: AccountId,
	pub key_type: UserShieldingKeyType,
	pub hash: H256
}
...

impl SetUserShieldingKeyCall {

	pub fn invoke(&self, parent_ss58_prefix: u16) -> StfResult<UserShieldingKeyType> {
		ensure!(
			is_authorised_signer(&self.signer, &self.who),
			StfError::SetUserShieldingKeyFailed(ErrorDetail::UnauthorisedSender)
		);
		IMTCall::set_user_shielding_key { who: self.who.clone(), key: self.key_type, parent_ss58_prefix }
			.dispatch_bypass_filter(RuntimeOrigin::root())
			.map_or_else(|e| Err(StfError::SetUserShieldingKeyFailed(e.error.into())), |_| Ok(self.key_type))
	}

}

...

			TrustedCall::set_user_shielding_key(call) => {
				debug!("set_user_shielding_key, who: {}", account_id_to_string(&data.who));
				let parent_ss58_prefix =
					node_metadata_repo.get_from_metadata(|m| m.system_ss58_prefix())??;
				let account = SgxParentchainTypeConverter::convert(data.who.clone());
				let c = node_metadata_repo
					.get_from_metadata(|m| m.user_shielding_key_set_call_indexes())??;

				match call.invoke() {
					Ok(key) => {
						debug!("pushing user_shielding_key_set event ...");
						let id_graph =
							IMT::get_id_graph_with_max_len(&who, RETURNED_IDGRAPH_MAX_LEN);
						calls.push(OpaqueCall::from_tuple(&(
							c,
							account,
							aes_encrypt_default(&key, &id_graph.encode()),
							hash,
						)));
					},
					Err(e) => {
						debug!("pushing error event ... error: {}", e);
						add_call_from_imp_error(
							calls,
							node_metadata_repo,
							Some(account),
							e.to_imp_error(),
							hash,
						);
					},
				}
				Ok(())
			},


```

This would be more verbose but should improve the readability of the code. Also, some variables have improper names like `key: UserShieldingKeyType`.

Personally, I don't have an opinion which version is better. I'm leaving it for your consideration.

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
```diff
@@ -19,186 +19,111 @@ extern crate sgx_tstd as std;
...
+		shard: &ShardIdentifier,
 	) -> StfResult<()> {
-		ensure_enclave_signer_account(&enclave_account)?;
+		ensure!(
+			is_authorised_signer(&signer, &who),
+			StfError::VerifyIdentityFailed(ErrorDetail::UnauthorisedSender)
+		);
 		let code = IdentityManagement::challenge_codes(&who, &identity)
 			.ok_or(StfError::VerifyIdentityFailed(ErrorDetail::ChallengeCodeNotFound))?;
+		let _ = IdentityManagement::user_shielding_keys(&who)
```
kziemianek: If you don't need a value then it's better to check for key existence, you will reduce storage reads.

Consider adding following method to SGX `IdentityManagement` pallet:

```
		pub fn user_shielding_key_exists(who: &T::AccountId) -> bool {
			UserShieldingKeys::<T>::contains_key(who)
		}
```

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
```diff
@@ -274,45 +179,62 @@ impl TrustedCallSigned {
...
-		identity: Identity,
-		metadata: Option<MetadataOf<Runtime>>,
 		bn: ParentchainBlockNumber,
-		parent_ss58_prefix: u16,
-	) -> StfResult<ChallengeCode> {
-		IMTCall::create_identity {
+	) -> StfResult<UserShieldingKeyType> {
+		// important! The signer has to be enclave_signer_account, as this TrustedCall can only be constructed internally
+		ensure_enclave_signer_account(&signer)
+			.map_err(|_| StfError::VerifyIdentityFailed(ErrorDetail::UnauthorisedSender))?;
```
kziemianek: Consider naming consistency: you are checking `signer` and throwing `UnauthorisedSender`.

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
```diff
@@ -274,45 +179,62 @@ impl TrustedCallSigned {
...
+			verification_request_block: bn,
 		}
 		.dispatch_bypass_filter(RuntimeOrigin::root())
-		.map_err(|e| StfError::CreateIdentityFailed(e.into()))?;
+		.map_err(|e| StfError::VerifyIdentityFailed(e.into()))?;
 
-		// generate challenge code
-		let code = generate_challenge_code();
+		// remove challenge code
+		IMTCall::remove_challenge_code { who, identity }
```
kziemianek, line 206: Shouldn't removing challenge code be a part of `IMTCall::verify_identity`?
Are there situations in which we want to verify identity and do not remove the challenge code?

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -662,34 +651,51 @@ where
...
-					))),
-					Ok(e) => warn!("error getting vc_issued call indexes: {:?}", e),
-					Err(e) => warn!("error getting vc_issued call indexes: {:?}", e),
+				debug!(
+					"request_vc_callback, who: {}, assertion: {:?}",
+					account_id_to_string(&who),
+					assertion
+				);
+				let account = SgxParentchainTypeConverter::convert(who.clone());
+				let c = node_metadata_repo.get_from_metadata(|m| m.vc_issued_call_indexes())??;
```
kziemianek: Consider finding more descriptive name for this `c` variable.

kziemianek (commented):
Nice work Kai! I dropped few thoughts. In general it looks good to me 🙃

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
outofxxx: These codes come from upstream, and I think it's best to provide a PR to upstream for modification if possible, otherwise the code for each merge of upstream may become more cumbersome.

Kailai-Wang: Thanks for the suggestion, yes we probably aren't going to change it as we want fewer deviations from upstream codebase when possible.

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: Sure 👍 I changed it to `call_index`

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
Kailai-Wang: Not now, but might be in the future. That's why we have a dedicated sidechain extrinsic for it (meaning the user might be able to clear it explicitly).

Besides that, there's a `set_challenge_code` fn, it would be strange if there's no counterpart of it.

Hence we will probably leave it unchanged.

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
Kailai-Wang: I changed it to `UnauthorisedSigner`

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
Kailai-Wang: Good catch, fixed.

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
kziemianek: We can have `remove_challenge_code` as well and same logic inside `verify_identity` if these functions should always be called together in case of veryfing identity

kziemianek (approved):

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
Kailai-Wang: That's indeed better for the current situation, thanks.

One slight concern is what if `remove_challenge_code` somehow fails, should the identity still considered verified? (My preference is yes). In that case it might be better to separate those two, and we shouldn't use `?` on `remove_challenge_code` ofc

kziemianek: @Kailai-Wang sorry for late reply.
From the user's perspective, your preference sounds like the correct way of handling such errors. We could monitor dangling challenge codes and think later about what to do with them.

PR #1703, opened 2023-05-16
Use augmented and generated types

### LINKED ISSUE #1667: ts-test needs to delay the querying on the sidechain storage.
https://github.com/litentry/litentry-parachain/actions/runs/4901317838/jobs/8752534190?pr=1665

The CI error is caused by querying the state immediately after processing the event, while the storage has not been updated yet. Have previously checked this with @zhouhuitian . Therefore, when querying sidechain storage in ts-test, need to wait for a moment before querying.

### LINKED ISSUE #1673: Use augmented and generated types in tee-worker/ts-tests

### LINKED ISSUE #1679: Use latest `polkadot/*` packages in ts-test
### Context

Follow-up task of https://github.com/litentry/litentry-parachain/pull/1675

both parachain ts-tests/ and tee-worker/ts-tests
The types need to be re-generated.

---

0xverin:
resolves #1673 
resolves #1667
resolves #1679 (@polkadot/api @polkadot/types @polkadot/kering)

`@polkadot/typegen` still cannot be upgraded,[this issue comment](https://github.com/polkadot-js/api/issues/5540#issuecomment-1469858228) has been mentioned.

1. I used the latest typegen to generate types for `litentry-parachain-metadata`. Because when I use `litentry-sidechain-metadata`, I found that the types are incorrect. 
2. All the generated types have been successfully imported, and in the code, we can see what type it is, as shown in the screenshot below:
image

It is no longer like this:
image


Kailai-Wang:
There's some misunderstanding, I think.

We mainly need the primitives types for sidechain, e.g. `LitentryIdentity`, `LitentryValidationData` etc. We hope these types are generated instead of being manually defined. That's why we need the sidechain metadata.

The parachain metadata only adds types/extrinsics for the parachain, which is nice-to-have, but not the main goal of the issue.
I'm unsure if we can merge both metadata into one, please check it. 

Both parachain and sidechain have a pallet `IdentityManagement`, if that's a problem, we should check if metadata supports alias. If not, we need to change the pallet name in the worst case.

0xverin:
> 

After checking, it is not possible to directly generate `LitentryIdentity` and `LitentryValidationData` using any kind of metadata. 

- The argument-types used for these [two types](https://github.com/litentry/litentry-parachain/blob/1673-use-generated-types/tee-worker/ts-tests/interfaces/augment-types.ts#L8) are also being manually defined. 
- I have examined the [code of idhub](https://github.com/litentry/identity-hub/blob/develop/libs/chain-types/src/lib/interfaces/augment-types.ts#L8), and the result is the same.So, if the `sidechain` doesn't have the types we need, I believe there may be no need for a merge. 
- I couldn't find any information on merging metadata in either the official documentation or the GitHub repository. I believe it may not be possible to merge metadata.

arieljbirnbaum:
> * I have examined the [code of idhub](https://github.com/litentry/identity-hub/blob/develop/libs/chain-types/src/lib/interfaces/augment-types.ts#L8), and the result is the same.

You mean that the generated types in the idhub code are the same we have? :thinking:

Verin1005:
> 

As of now, it appears to be the case.

0xverin:
> 

As of now, it appears to be the case. 

> > * I have examined the [code of idhub](https://github.com/litentry/identity-hub/blob/develop/libs/chain-types/src/lib/interfaces/augment-types.ts#L8), and the result is the same.
> 
> You mean that the generated types in the idhub code are the same we have? 🤔

Not entirely,they made changes to the content of `identity/definitions`, so `augment-types.ts` will have some differences. What I actually meant was that the generation rules are the same. However, that is not enough for us. I am currently thinking about how to include the types generated from `sidechain-meta` as well.

arieljbirnbaum (approved):

0xverin:
Updated:

1. There was a mistake earlier:when querying the storage in the `sidechain`, the data passed to createType should come from the `sidechain's` construction, not the `parachain's`. The reason it appears to work is that the data structure of `LitentryIdentity` is correct and matches the expected format.

2. Although we have generated the `LitentryPrimitivesIdentity` type from `sidechain`, the data composed of this structure is still the same as before. I don't think I have made a mistake(if I have,plz correct me,thanks!). Here is an [example](https://github.com/shivam-devops/phala-blockchain/blob/b5b17cf1d0dcd9efc5af0b5ac0302f037e32b401/scripts/js/src/console.js#LL518C33-L518C33) of how Phala uses [XcmVersionedXcm type](https://github.com/litentry/litentry-parachain/blob/4355e7bef629594ca9ee69d968f04318a5ad12b1/tee-worker/ts-tests/parachain-interfaces/types-lookup.ts#L4429). Additionally, if the returned structure is not like the one constructed by `buildIdentityHelper`,, it is unlikely to succeed.So,I didn't change `LitentryIdentity` type. 

3. This pr, with the addition of `tsconfig.json`, type checking has been enabled, which means that when I pull other PRs, I encounter type check errors locally. It was my mistake, and in fact, the type checking should have been added to ts-test from the beginning instead of modifying the types now. Type checking is something that should be done from the start.

tee-worker/ts-tests/bulk_vc.test.ts
```diff
@@ -72,7 +68,7 @@ describeLitentry('multiple accounts test', 10, async (context) => {
     step('test requestVc with multiple accounts', async () => {
         let txs: TransactionSubmit[] = [];
         for (let i = 0; i < substrateSigners.length; i++) {
-            const tx = context.api.tx.vcManagement.requestVc(context.mrEnclave, assertion_type);
+            const tx = context.api.tx.vcManagement.requestVc(context.mrEnclave, assertion_type!);
```
arieljbirnbaum: Can `assertion_type` be undefined here? :thinking:

tee-worker/ts-tests/common/utils/identity-helper.ts
```diff
@@ -47,7 +46,7 @@ export async function buildIdentityHelper(
 export async function buildIdentityTxs(
     context: IntegrationTestContext,
     signers: KeyringPair[] | KeyringPair,
-    identities: LitentryIdentity[],
+    identities: any[],
```
arieljbirnbaum: Can we use a less promiscuous type here?

tee-worker/ts-tests/common/utils/identity-helper.ts
```diff
@@ -61,8 +60,9 @@ export async function buildIdentityTxs(
         const identity = identities[k];
         let tx: SubmittableExtrinsic<ApiTypes>;
         let nonce: number;
-        const encod_identity = api.createType('LitentryIdentity', identity).toU8a();
-        const ciphertext_identity = encryptWithTeeShieldingKey(teeShieldingKey, encod_identity).toString('hex');
+        const encod_identity = context.sidechainRegistry.createType('LitentryPrimitivesIdentity', identity);
```
arieljbirnbaum:
```suggestion
        const encoded_identity = context.sidechainRegistry.createType('LitentryPrimitivesIdentity', identity);
```

Also, what's our convention for variable identifiers? `snake_case` or `camelCase`?

tee-worker/ts-tests/identity.test.ts
```diff
@@ -116,10 +118,10 @@ describeLitentry('Test Identity', 0, (context) => {
         // the main address should be already inside the IDGraph
         const main_identity = await buildIdentityHelper(
             u8aToHex(context.substrateWallet.alice.addressRaw),
-            'LitentryRococo',
+            process.env.NODE_ENV === 'local' ? 'TestNet' : 'LitentryRococo',
```
arieljbirnbaum: Should we rather centralize this somewhere rather than having these checks and magic constants scattered all over the place? :thinking:

tee-worker/ts-tests/package.json
```diff
@@ -5,9 +5,12 @@
...
-		"gen-type:meta": "ts-node --skip-project node_modules/.bin/polkadot-types-from-chain --package . --output ./interfaces --endpoint ./litentry-sidechain-metadata.json",
+		"gen-type:parachain": "yarn gen-type-parachain:defs && yarn gen-type-parachain:meta && yarn format",
+		"gen-type-parachain:defs": "ts-node --skip-project node_modules/.bin/polkadot-types-from-defs  --package . --input ./parachain-interfaces --endpoint ./litentry-parachain-metadata.json",
+		"gen-type-parachain:meta": "ts-node --skip-project node_modules/.bin/polkadot-types-from-chain --package . --output ./parachain-interfaces --endpoint ./litentry-parachain-metadata.json",
+		"gen-type:sidechain": "yarn gen-type-sidechain:defs && yarn gen-type-sidechain:meta && yarn format",
+		"gen-type-sidechain:defs": "ts-node --skip-project node_modules/.bin/polkadot-types-from-defs  --package . --input ./sidechain-interfaces --endpoint ./litentry-sidechain-metadata.json",
+		"gen-type-sidechain:meta": "ts-node --skip-project node_modules/.bin/polkadot-types-from-chain --package . --output ./sidechain-interfaces --endpoint ./litentry-sidechain-metadata.json",
```
arieljbirnbaum, line 13: Do we have these documented somewhere? If so then that probably needs updated as well :thinking:

arieljbirnbaum (commented):
LGTM; left minor comments/questions

tee-worker/ts-tests/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+	"compilerOptions": {
+		// this is specific with augmented overrides
+		"paths": {
+			"@polkadot/api/augment": ["./parachain-interfaces/augment-api.ts"],
+			"@polkadot/types/augment": ["./parachain-interfaces/augment-types.ts"],
+			"@polkadot/types/lookup": ["./sidechain-interfaces/types-lookup.ts"]
```
Kailai-Wang, line 7: These three paths are a mixture of parachain and sidechain interfaces, is this intended?

tee-worker/ts-tests/common/utils/identity-helper.ts
Kailai-Wang: I would agree - either `encoded_identity` or `enc_identity` if you want a shorter name, but not `encod` :P

tee-worker/ts-tests/common/type-definitions.ts
```diff
@@ -1,31 +1,43 @@
...
+export type WorkerRpcReturnString = {
+    vec: string;
+};
 
+export type BatchCall = Vec<Call> | (string | Uint8Array | IMethod<AnyTuple, any> | Call)[];
+export type WorkerRpcReturnValue = {
+    value: `0x${string}`;
+    do_watch: boolean;
+    status: DirectRequestStatus;
+};
```
Kailai-Wang: Why do we still need `WorkerRpcReturnString` and `WorkerRpcReturnValue`? Aren't they deprecated in favor of the definitions in `interfaces/identity/`?

Kailai-Wang (commented):
I'd expect things like `LitentryIdentity` and  `LitentryValidationData` would be deleted from `type-definitions.ts` and types in metadata are used instead. Isn't it the case?

tee-worker/ts-tests/common/type-definitions.ts
0xverin: fixed!

tee-worker/ts-tests/common/utils/identity-helper.ts
0xverin: fixed!

tee-worker/ts-tests/common/utils/identity-helper.ts
0xverin: > Also, what's our convention for variable identifiers? snake_case or camelCase?

I prefer snake_case as the uniform format for variables.

0xverin:
Updated:

1. Removed custom types related to identity and replaced them with generated types.
2. Modified parts of the code that involved assertions to use interfaces from the autogenerated types.

tee-worker/ts-tests/batch.test.ts
```diff
@@ -88,8 +98,13 @@ describeLitentry('Test Batch Utility', 0, (context) => {
...
-        for (let i = 0; i < event_datas.length; i++) {
+        const event_datas = (await handleIdentityEvents(
+            context,
+            aesKey,
+            resp_events,
+            'IdentityCreated'
+        )) as IdentityGenericEvent[];
+        for (let i = 0; i < [event_datas].length; i++) {
             assertIdentityCreated(context.substrateWallet.alice, event_datas[i]);
         }
```
arieljbirnbaum: Isn't this the same as

```suggestion
        assertIdentityCreated(context.substrateWallet.alice, event_datas);
```

?

tee-worker/ts-tests/batch.test.ts
```diff
@@ -88,8 +98,13 @@ describeLitentry('Test Batch Utility', 0, (context) => {
...
             ['IdentityCreated']
         );
-        const event_datas = await handleIdentityEvents(context, aesKey, resp_events, 'IdentityCreated');
-        for (let i = 0; i < event_datas.length; i++) {
+        const event_datas = (await handleIdentityEvents(
+            context,
+            aesKey,
+            resp_events,
+            'IdentityCreated'
+        )) as IdentityGenericEvent[];
```
arieljbirnbaum: Why do we need the `as` here? :thinking:

tee-worker/ts-tests/identity.test.ts
0xverin: fixed!

tee-worker/ts-tests/bulk_vc.test.ts
0xverin: Because this [type definition](https://github.com/litentry/litentry-parachain/blob/1673-use-generated-types/tee-worker/ts-tests/common/type-definitions.ts#L100) uses the "?" operator.

tee-worker/ts-tests/batch.test.ts
0xverin: Because I modified the return type of [handleIdentityEvents](https://github.com/litentry/litentry-parachain/blob/1673-use-generated-types/tee-worker/ts-tests/common/utils/identity-helper.ts#L118).

arieljbirnbaum: Then shouldn't we instead handle the `string` case in the caller and/or change the type of `assertIdentityCreated` accordingly? (bonus points for a more informative error type for the failure case as opposed to a naked `string`)

tee-worker/ts-tests/bulk_vc.test.ts
arieljbirnbaum: By its use in `VCRequested` I'm assuming that the `Assertion` type is supposed to encode _one_ assertion with its params, correct? Then shouldn't it rather be sth like

```
type Assertion = {
  A1: string;
} | {
  A2: [string]
} | ...
```
?

0xverin: I agree with your suggestion. Union types are indeed much better than optional types.Fixed it.

tee-worker/ts-tests/tsconfig.json
0xverin: They are different. The first two refer to the detection of API types (for example, when using api.tx, we need the parachain type). The last one refers to the configuration when we need to import a specific type (for example, importing LitentryPrimitivesIdentity).

tee-worker/ts-tests/batch.test.ts
0xverin: It might be necessary to rewrite assertIdentityCreated to handle the data internally, as it seems more appropriate.

tee-worker/ts-tests/batch.test.ts
0xverin: What do you think about adding a `handleIdentityErrorEvent` specifically for handling error messages? I'm considering whether it would be too cumbersome to have `assertIdentityCreated` handle all the data.I can also separate the two return types and eliminate the need for as assertions.:p

arieljbirnbaum: > What do you think about adding a handleIdentityErrorEvent specifically for handling error messages?

Could work; not sure how you'd want to integrate it with the batch case (e.g. if you have a batch where some transactions are supposed to succeed and some are supposed to fail, how would it look using `handleIdentityErrorEvent`?)

tee-worker/ts-tests/tsconfig.json
Kailai-Wang: What if the types that we need are in the parachain types-lookup?

0xverin: I have contemplated merging them together: `"@polkadot/types/lookup":[parachain..., sidechain...]`, but it is not supported. There are related [issues](https://github.com/polkadot-js/api/issues/4969) mentioning this, so I think ultimately Rust needs to merge `identity types` into `parachain`.

tee-worker/ts-tests/batch.test.ts
0xverin: Currently, what I'm thinking is to rewrite the `assertIdentityCreated/Verified/Removed`methods so that the obtained events are directly passed to them for processing, instead of handling them collectively in `handleIdentityEvents`. The advantages of doing this are:

1. Improved code readability and maintainability, as it separates the responsibilities.
2. We can use the `LitentryPrimitivesIdentity` interface when asserting, making it easier to perform checks.

arieljbirnbaum: That sounds very reasonable :+1:

0xverin:
Major updates:

1. Refactored all the [assertions](https://github.com/litentry/litentry-parachain/pull/1703/files#diff-5952befd50a0aae3ef5dedd7f0947610a2c5f27e7b55a3b5c05efa7ed392e32e).
2. I have made an effort to use the `LitentryPrimitivesIdentity interface` for data comparison between events and idGraph. I believe this approach is reasonable([usage](https://github.com/litentry/litentry-parachain/pull/1703/files#diff-5952befd50a0aae3ef5dedd7f0947610a2c5f27e7b55a3b5c05efa7ed392e32eR70)).

Once the CI passes, it is ready for review.

tee-worker/ts-tests/identity.test.ts
```diff
@@ -12,24 +12,24 @@ import {
...
 import { Event } from '@polkadot/types/interfaces';
 import { ethers } from 'ethers';
-const substrateExtensionIdentity: LitentryPrimitivesIdentity = {
+const substrateExtensionIdentity = {
     Substrate: {
         address: '0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48', //Bob
         network: 'Litentry',
     },
-} as any;
+} as unknown as LitentryPrimitivesIdentity;
```
arieljbirnbaum: I get why we need the `as` when interfacing with the (mis-)augmented API, but why the assertion here?

If I get it correctly this is the _input_ we want to pass to `createType('LitentryPrimitivesIdentity', <_>)`, but from what we saw earlier _that_ type is actually different from how `LitentryPrimitivesIdentity` is declared, no? For instance, this input doesn't have the `type` or `isSubstrate` props.

This might still be a job for the `ExtractEnumVariant` we discussed earlier :thinking:

0xverin: >This might still be a job for the ExtractEnumVariant we discussed earlier

Your understanding is correct. If we want perfection, we do need to do it that way(ExtractEnumVariant). However, you will find that in our ts-test, we mostly use interfaces for `LitentryPrimitivesIdentity`. I'm thinking that when assembling the parameters, we can define the `LitentryPrimitivesIdentity` type ,but need  pass it to `createType` with ` Substrate: {
        address: '0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48', 
        network: 'Litentry',
    }.` Do you think this is significant? Would it be enough for IDHub to know the  `LitentryPrimitivesIdentity` which interfaces it has?

If they want to define `LitentryPrimitivesIdentity`, would it be enough for them to assemble their own `ExtractEnumVariant`?

0xverin:
Now there is a final decision. Do we want to assemble the `ExtractEnumVariant` type to fully reconstruct the `LitentryPrimitivesIdentity` type? Personally, I believe it may be more meaningful for IDHub to understand how to use the interfaces of `LitentryPrimitivesIdentity`. They may need more interfaces to perform certain state checks (I inferred this from looking at their code, as they also use as unknown). They can refer to our usage of assert to utilize the `LitentryPrimitivesIdentity` interface. @Kailai-Wang @grumpygreenguy :P

Since this PR has been open for a while, I would like to make the final confirmation in the last stage. We can create a separate PR for the "looks-up" issue because that [seems to be a TypeScript problem](https://github.com/polkadot-js/api/issues/5664).

tee-worker/ts-tests/identity.test.ts
arieljbirnbaum: > If we want perfection, we do need to do it that way(ExtractEnumVariant). 

I don't think we want _perfection_, but we should at least strive to have accurate definitions for the input and output datatypes in the API (even if we cannot get them to be _inferred_ correctly in each usage). 

> However, you will find that in our ts-test, we mostly use interfaces for LitentryPrimitivesIdentity.

I don't follow what you mean by this :thinking: 

> but need pass it to `createType` with  `Substrate: { address: '0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48',  network: 'Litentry', }`.

My point is that AFAICT this value (`{ Substrate: { ... } }`) that we're passing to `createType` is _not_ a `LitentryPrimitivesIdentity`, but rather some other type that at the moment doesn't have a name. We can derive that type _from_ `LitentryPrimitivesIdentity` with something like `ExtractEnumVariant`, but they're not the same type and I think it only makes sense to make that distinction explicit.

Kailai-Wang:
I still see many conflicting files for this PR.
Can we resolve these ASAP and get it merged (if it's ready)? This PR has been there for 3 weeks :(

tee-worker/ts-tests/identity.test.ts
0xverin: We have decided to address this matter in another PR/issue.

PR #1709, opened 2023-05-18
chore: #1668 split test common utils

### LINKED ISSUE #1668: Dissect `tee-worker/ts-tests/common/utils.ts`
### Context

`tee-worker/ts-tests/common/utils.ts` is only getting more bloated when we add more helpers. Maybe it makes sense to split it into smaller files depending on the applied areas, e.g.:
- crypto/enc/dec
- interaction with parachain
- interaction with sidechain
- ...

---

wli-pro:
fixed #1668  

All existing tests passed:
image


tee-worker/ts-tests/common/utils/test.ts
0xverin, line 1: I think the file name `integration_setup` more clearer. :)

tee-worker/ts-tests/common/utils/context.ts
```diff
@@ -0,0 +1,154 @@
...
+        teeShieldingKey,
+        mrEnclave,
+        ethersWallet,
+        substrateWallet,
+        metaData,
+        web3Signers,
+    };
+}
+
+export async function handleVcEvents(
```
0xverin: We have `identity.ts` now. Would it be better to move it into `vc.ts`? Additionally, wouldn't it be clearer to name them `identity_helper` and `vc_helper`?

tee-worker/ts-tests/common/utils/assertion.ts
```diff
@@ -0,0 +1,206 @@
...
+        isModule ? (detail = JSON.stringify(item.data.detail.toHuman())) : (detail = JSON.stringify(item));
+        assert.isTrue(
+            detail.includes(expectedDetail),
+            `check error detail failed, expected detail is ${expectedDetail}, but got ${detail}`
+        );
+    });
+    return true;
+}
+
+export async function buildValidations(
```
0xverin: I feel that it would be better to move this to `identity.ts`. This is the functionality for constructing identity validations.

0xverin (commented):
Thanks!I think it's mostly fine, but perhaps a slightly clearer file naming would be better.

tee-worker/ts-tests/common/utils/identity.ts
```diff
@@ -0,0 +1,224 @@
...
+    LitentryValidationData,
+    SubstrateNetwork,
+    Web2Network,
+} from '../type-definitions';
+import { decryptWithAES, encryptWithTeeShieldingKey } from './crypto';
+import { ApiTypes, SubmittableExtrinsic } from '@polkadot/api/types';
+
+const base58 = require('micro-base58');
+
+export async function createTrustedCallSigned(
```
kziemianek: this is a dead code, can we get rid of it ?
also i doubt it's specific to identity.

0xverin: Yeap,we can get rid of it, the use case update is [here](https://github.com/litentry/litentry-parachain/blob/1668-split-test-utils/tee-worker/ts-tests/examples/direct-invocation/util.ts#L58)

0xverin (approved):
Thanks! LGTM!

PR #1717, opened 2023-05-21
Use `WorkerRpcReturnValue` types from interface

### LINKED ISSUE #1714: Use consistent types for `WorkerRpcReturnValue` and `WorkerRpcReturnString`
### Context

We have some inconsistent type definitions for these two types, which can cause some problems in decoding as they have incompatible types.

https://github.com/litentry/litentry-parachain/blob/d3f6570b256bd5fed0a637bea2d511c4e433bb5c/tee-worker/ts-tests/common/type-definitions.ts#L23-L31

versus

https://github.com/litentry/litentry-parachain/blob/d3f6570b256bd5fed0a637bea2d511c4e433bb5c/tee-worker/ts-tests/interfaces/identity/definitions.ts#L3-L10

We should stick to the second definition.

---

Kailai-Wang:
resolves #1714 

This PR uses the `WorkerRpcReturnValue` type from interface generation and removes `WorkerRpcReturnString` value to make sure the RPC responses are consistently handled.

related: https://github.com/litentry/litentry-parachain/issues/1673


tee-worker/ts-tests/common/call.ts
```diff
@@ -1,29 +1,44 @@
...
+        console.log('Rpc trusted operation execution failed, hash: ', res.value.toHex());
     }
-    return resp_json;
+
+    return res;
+}
+
+// decode the returned bytes as string
+// please note we shouldn't use toU8a(), which encodes the Bytes instead of converting
+export function decodeRpcBytesAsString(value: Bytes): string {
```
0xverin, line 33: Wouldn't it be better to move this to`common/helpers.ts`?This file is responsible for exporting the convert functionality. :)

0xverin (commented):
I believe overall it should be fine. Thank you.

tee-worker/ts-tests/common/call.ts
Kailai-Wang: > Wouldn't it be better to move this to`common/helpers.ts`?This file is responsible for exporting the convert functionality. :)

I thought it was a follow-up of `sendRequest`, so I put it in the same place

0xverin (approved):

PR #1728, opened 2023-05-24
fix: #1684 use block number instead of timeout limit in test

### LINKED ISSUE #1684: Identity ts-test fails to detect `wrongWeb2Handle` error

wli-pro:
…il test assertion when there is a VerifyIdentityFailed event from chain

fixed #1684 

The error shown here
Screenshot 2023-05-12 at 00 18 39

 will fail the test now.

### DIFF
--- a/tee-worker/ts-tests/common/utils/context.ts
+++ b/tee-worker/ts-tests/common/utils/context.ts
@@ -12,7 +12,8 @@ import { IntegrationTestContext, teeTypes, EnclaveResult, Web3Wallets } from '..
 const crypto = require('crypto');
 
 // maximum milliseconds that we wait in listening events before we timeout
-const listenTimeoutInMilliSeconds = 3 * 60 * 1000;
+// set to 5 minutes as identity-hub to ensure all tests have enough time to get events
+const listenTimeoutInMilliSeconds = 5 * 60 * 1000;
 
 export async function getListenTimeoutInBlocks(api: ApiPromise) {
     const slotDuration = await api.call.auraApi.slotDuration();
--- a/tee-worker/ts-tests/common/utils/context.ts
+++ b/tee-worker/ts-tests/common/utils/context.ts
@@ -11,14 +11,8 @@ import { IntegrationTestContext, teeTypes, EnclaveResult, Web3Wallets } from '..
 
 const crypto = require('crypto');
 
-// maximum milliseconds that we wait in listening events before we timeout
-// set to 5 minutes as identity-hub to ensure all tests have enough time to get events
-const listenTimeoutInMilliSeconds = 5 * 60 * 1000;
-
-export async function getListenTimeoutInBlocks(api: ApiPromise) {
-    const slotDuration = await api.call.auraApi.slotDuration();
-    return listenTimeoutInMilliSeconds / parseInt(slotDuration.toString());
-}
+// maximum block number that we wait in listening events before we timeout
+export const defaultListenTimeoutInBlockNumber = 15;
 
 export async function initWorkerConnection(endpoint: string): Promise<WebSocketAsPromised> {
     const wsp = new WebSocketAsPromised(endpoint, <Options>(<unknown>{
--- a/tee-worker/ts-tests/identity.test.ts
+++ b/tee-worker/ts-tests/identity.test.ts
@@ -353,7 +353,7 @@ describeLitentry('Test Identity', 0, (context) => {
             context.substrateWallet.alice,
             alice_txs,
             'identityManagement',
-            ['IdentityVerified', 'VerifyIdentityFailed'] // allow possible failed events to be collected
+            ['IdentityVerified']
         );
         let bob_txs = await buildIdentityTxs(
             context,
@@ -368,7 +368,7 @@ describeLitentry('Test Identity', 0, (context) => {
             context.substrateWallet.bob,
             bob_txs,
             'identityManagement',
-            ['IdentityVerified', 'VerifyIdentityFailed'] // allow possible failed events to be collected
+            ['IdentityVerified']
         );
         const verified_event_datas = await handleIdentityEvents(context, aesKey, alice_resp_events, 'IdentityVerified');
         const [substrate_extension_identity_verified] = await handleIdentityEvents(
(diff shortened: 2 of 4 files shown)


Kailai-Wang:
Have you tested it? Like intentionally fail the identity test 👀

tee-worker/ts-tests/identity.test.ts
```diff
@@ -353,7 +353,7 @@ describeLitentry('Test Identity', 0, (context) => {
             context.substrateWallet.alice,
             alice_txs,
             'identityManagement',
-            ['IdentityVerified']
+            ['IdentityVerified', 'VerifyIdentityFailed'] // allow possible failed events to be collected
```
kziemianek: Shouldn't we collect other possible events in other cases, for example `RemoveIdentityFailed` in case of removing identity ?

Another way is to wait for example 10 blocks and strictly check what events were emitted and do not pass if there is something more than expected.

tee-worker/ts-tests/identity.test.ts
```diff
@@ -368,7 +368,7 @@ describeLitentry('Test Identity', 0, (context) => {
             context.substrateWallet.bob,
             bob_txs,
             'identityManagement',
-            ['IdentityVerified']
+            ['IdentityVerified', 'VerifyIdentityFailed'] // allow possible failed events to be collected
         );
         const verified_event_datas = await handleIdentityEvents(context, aesKey, alice_resp_events, 'IdentityVerified');
```
kziemianek: I checked it and it seems that in case of `VerifyIdentityFailed` it will fail at this line.

kziemianek (commented):
In case of `VerifyIdentityFailed` shouldn't `sendTxsWithUtility` timeout ?

```
await sendTxsWithUtility(
            context,
            context.substrateWallet.bob,
            bob_txs,
            'identityManagement',
            ['IdentityVerified']
``` 
@Verin1005 ?

tee-worker/ts-tests/identity.test.ts
0xverin: In fact,we don't want to accommodate unexpected events in the expected process; We will throw unexpected events. In fact, it will eventually timeout because we expect to receive three `IdentityVerified` events, but in reality, we only receive two. So, [events.length === txsLength ](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/common/transactions.ts#L163)cannot be true. I suggest closing this pr,merge [this pr](https://github.com/litentry/litentry-parachain/pull/1624) first.The issue with @Kailai-Wang has been optimized in the pr,we will throw an unexpected event.

tee-worker/ts-tests/identity.test.ts
0xverin: > 



> In case of `VerifyIdentityFailed` shouldn't `sendTxsWithUtility` timeout ?
> 
> ```
> await sendTxsWithUtility(
>             context,
>             context.substrateWallet.bob,
>             bob_txs,
>             'identityManagement',
>             ['IdentityVerified']
> ```
> 
> @Verin1005 ?

I found that the description of #1684 may be incorrect. If you continue testing, you will eventually encounter a `timeout`. The console only[ prints the current event](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/common/transactions.ts#L106) and its content without any further meaning, and the "ts-test" will not succeed [here](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/common/transactions.ts#L163). Therefore, it will eventually timeout. This [pr](https://github.com/litentry/litentry-parachain/pull/1624) addresses the optimization related to the timeout issue.

tee-worker/ts-tests/identity.test.ts
wli-pro: I confirmed the current implementation in dev branch is fine, as in it will only try to get the IdentityVerified event, before timeout (which will fail test).
In this case, the current dev branch can already work with the VerifyIdentityFailed event. Evidence as below:

image

wli-pro:
reverted the extra event check, and verified it's working: https://github.com/litentry/litentry-parachain/pull/1728#discussion_r1204972330
In this PR, it's now just some minor typing fix in this PR.

tee-worker/ts-tests/common/utils/context.ts
```diff
@@ -12,7 +12,8 @@ import { IntegrationTestContext, teeTypes, EnclaveResult, Web3Wallets } from '..
 const crypto = require('crypto');
 
 // maximum milliseconds that we wait in listening events before we timeout
-const listenTimeoutInMilliSeconds = 3 * 60 * 1000;
+// set to 5 minutes as identity-hub to ensure all tests have enough time to get events
```
0xverin: Do we need to increase the timeout here? I feel like it may not be necessary to extend the time. Normally, we expect to get the result within 15 blocks (`bulk_identity/vc` may vary depending on the situation). So, what I mean is, I'm not sure of the significance of increasing the time here. If it is necessary, please let me know. Thank you. :)

kziemianek: I agree with @Verin1005 

I would ask if we can go a little bit further and have exact block number expectations for invocation to be finished.
It would help us detect situations when changes causes invocations to take longer than expected.

For example we have CreateIdentity call, and we expect this call to be finished within 3 blocks.

wli-pro:
Testing results after using blocknumber instead of timeout time:

test-batch:local:
image

test-identity:local:
image

test-bulk-identity:local
image

tee-worker/ts-tests/common/utils/context.ts
wli-pro: Used 15 block number for now, removed the timeout completely.

0xverin (approved):

0xverin:
btw, plz modify this pr's title, change it to a suitable title. Thanks! :)

kziemianek (approved):

PR #1729, opened 2023-05-24
Add public getter of account nonce

### LINKED ISSUE #1710: Follow-up of DI: public getter of account nonce
### Context

To construct and sign a trusted call, we'd need to include the account nonce to avoid replay/double-spend attack.

We don't have such direct getter yet (or do we?) - if that's the case, we should implement it.

---

outofxxx:
resolved #1710


BillyWooo (approved):
LGTM

Kailai-Wang (changes requested):
Thank you!

I think the implementation part is OK, but maybe it's not worth a dedicated test, as you probably need to use it anyway in the ts-test.

So can you please:
- remove the test and CLI 
- remove the nonce from trusted getter (also from ts-test definitions)
- (optional) modify the demo-code of DI to use this getter to retrieve nonce instead of hardcoding

Kailai-Wang:
Thank you, the logic looks good.

I'll approve it first, the changes in ts-test can be done in a follow-up PR

Kailai-Wang (approved):

PR #1732, opened 2023-05-24
Implement trusted getter

### LINKED ISSUE #1706: Follow-up of DI: implement synchronous TrustedGetter
### Context

For DI we can not only write the state, but also read it.

The codebase is generally there, we only need to add our getters for:
- user shielding key
- IDGraph
- maybe challenge code?

Ideally with ts code examples.

---

Kailai-Wang:
resolves #1706 

Includes the demo code and updated metadata.


tee-worker/app-libs/stf/src/getter.rs
```diff
@@ -188,8 +195,15 @@ impl ExecuteGetter for TrustedGetterSigned {
 			// litentry
 			TrustedGetter::user_shielding_key(who) =>
 				IdentityManagement::user_shielding_keys(&who).map(|key| key.encode()),
-			TrustedGetter::id_graph_stats(_who) =>
-				IdentityManagement::id_graph_stats().map(|item| item.encode()),
+			TrustedGetter::id_graph(who) => Some(IdentityManagement::get_id_graph(&who).encode()),
+			TrustedGetter::challenge_code(who, identity) =>
+				IdentityManagement::challenge_codes(&who, &identity).map(|code| code.encode()),
+			// TODO: we need to re-think it
```
kziemianek, line 201: You mean previous implementation returned whole id graph stats, not only slice of that `AccountId` ?

tee-worker/ts-tests/examples/direct-invocation/util.ts
```diff
@@ -93,6 +93,26 @@ export const createSignedTrustedCall = (
     });
 };
 
+export const createSignedTrustedGetter = (
+    parachain_api: ApiPromise,
+    trustedGetter: [string, string],
+    account: KeyringPair,
+    params: Array<any>
```
kziemianek: `params: Array<any> | any` ?

then you woudn't need to check `params.length` and extract first element in case of `len == 1`.

kziemianek (approved):

tee-worker/ts-tests/examples/direct-invocation/util.ts
arieljbirnbaum: `params: Array<any> | any` would be the same as `params: any` but more noisy :P

kziemianek: next step is to get rid of `any` :)

Kailai-Wang: Thanks :) updated

tee-worker/app-libs/stf/src/getter.rs
Kailai-Wang: Yes - with this enabled, anyone can query the IDGraph stats.

PR #1737, opened 2023-05-26
atomic IMT pallet operations

### LINKED ISSUE #1705: Double-check the atomicity/state mutation of trusted call execution
### Context

Follow-up from https://github.com/litentry/litentry-parachain/pull/1702#discussion_r1196023618:
there're cases where multiple sidechain extrinsics are called within a single trusted call execution, e.g.: https://github.com/litentry/litentry-parachain/blob/62d4f2a3471292c1ccabd110e3a3e3de11115f1e/tee-worker/app-libs/stf/src/trusted_call_litentry.rs#L197-L210
 
What if `remove_challenge_code` somehow fails? Should the identity be considered verified? I believe yes. But then we'll have a `VerifiyIdentityFailed` event, which is misleading to F/E.

We need to avoid that. For example, ignore the execution result of `remove_challenge_code` as the state of Identity is already mutated, and we should tell F/E `IdentityVerified` in any case.

---

kziemianek:
Merge multiple IMT pallet operations into one

fixes: #1705


kziemianek:
I checked it locally and multiple calls to pallet in single trusted call might result in inconsistent pallet state because storage transaction is open per pallet call. It's better to avoid such multiple calls whenever possible.

Kailai-Wang (approved):
Looks good, thanks!

PR #1746, opened 2023-06-02
add compact

### LINKED ISSUE #1433: Try to use compact codec when possible
### Context

Especially for the extrinsics that are already large and called frequently. It helps reduce the ecall/ocall buffer size when passing encoded extrinsics around, hopefully.

See https://substrate.stackexchange.com/questions/679/when-to-use-compact-on-extrinsic-arguments/686#686

Note we have to bump transaction version if it's changed.

---

songzhouran:


tee-worker/core/parentchain/indirect-calls-executor/src/lib.rs
```diff
@@ -238,7 +238,7 @@ where
 		&self,
 		block_hash: H256,
 		extrinsics: Vec<H256>,
-		block_number: <<ParentchainBlock as ParentchainBlockTrait>::Header as Header>::Number,
+		block_number: u32,
```
felixfaisal: I think this trait implements `TryFrom<u32>` have you tried doing something like 
```rust
let block: u32 = block_number.try_into(); 
```

pallets/teerex/src/lib.rs
```diff
@@ -344,7 +344,7 @@ pub mod pallet {
 		pub fn shield_funds(
 			origin: OriginFor<T>,
 			incognito_account_encrypted: Vec<u8>,
-			amount: BalanceOf<T>,
+			#[pallet::compact] amount: BalanceOf<T>,
```
Kailai-Wang: I think we can leave out `shield_funds` and `unshield_funds` as these two are not called by us :)

tee-worker/core-primitives/types/src/lib.rs
```diff
@@ -47,7 +47,7 @@ pub type CallIndex = [u8; 2];
 
 // pallet teerex
 pub type ConfirmCallFn = (CallIndex, ShardIdentifier, H256, Vec<u8>);
-pub type ShieldFundsFn = (CallIndex, Vec<u8>, Balance, ShardIdentifier);
+pub type ShieldFundsFn = (CallIndex, Vec<u8>, codec::Compact<u128>, ShardIdentifier);
```
Kailai-Wang: We should have used `codec::Compact<Balance>` if that's possible - but since we are not going to change shielding_funds - let's just revert the changes

tee-worker/core/parentchain/indirect-calls-executor/src/lib.rs
```diff
@@ -248,7 +248,10 @@ where
 		})??;
 
 		let root: H256 = merkle_root::<Keccak256, _>(extrinsics);
-		Ok(OpaqueCall::from_tuple(&(call, block_hash, block_number, root)))
+		let parentchain_block_number: ParentchainBlockNumber =
+			block_number.try_into().map_err(|_| Error::ConvertParentchainBlockNumber)?;
+		let block: u32 = parentchain_block_number;
```
Kailai-Wang: Does `codec::Compact(block_number)` work? As `<<ParentchainBlock as ParentchainBlockTrait>::Header as Header>::Number` should implement `AtLeast32BitUnsigned`. 

If it doesn't work - `codec::Compact(parentchain_block_number)` should be enough, we don't need the extra `block` variable here

Kailai-Wang (commented):
Thank you - I've left some comments

Basically I think it makes sense to change only those extrinsics that are called frequently :)

Btw can you please link the related issue using the auto-close github keyword?

Kailai-Wang:
superseded by #1758

Issue #1628, opened 2023-04-22
Have an automated release process for the worker and enclave


Kailai-Wang:
### Context

We are still missing the release process for the TEE-related stuff:
- scripts to automate the release-build and packaging
- version and release cycle management
- differentiate between dev and prod version? (where to build it for HW mode?)

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
@felixfaisal you might be the best person for this - are you interested in it? :)

felixfaisal:
Yup, I can take this up

BillyWooo:
Just some comment during deployment:
- deploy binary package (no dependency with repo)
- save config file (key/tokens/etc.) to a specific location on server

PR #1750, opened 2023-06-06
feat: automate release process for worker

### LINKED ISSUE #1686: Automate the enclave update process
### Context

Imagine we want to apply bug fixes and feature implementations to the staging worker, so far it's been done manually, see this [notion page](https://www.notion.so/web3builders/How-to-upgrade-enclave-one-worker-edfaf5871b4441579f9471074032ed1e).

Ideally we'd have a process to automate it, either via GHA or at least a local script.

---

felixfaisal:
This PR creates a script for automating upgrading the worker. 

Resolves #1686
 
Documentation: [Notion](https://www.notion.so/web3builders/Release-Automation-Documentation-e1243577bf9e4d92806875d5c6043d86?pvs=4)


scripts/ts-utils/move-vcregistry-snapshot.js
0xverin, line 1: We don't need to submit compiled files here. Why is it being added now? I think we can simply run "npx ts-node xxx.ts" directly.

scripts/ts-utils/setup_enclave.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Also need to set the port
+
+cd ../scripts/ts-utils/
+yarn install
+yarn tsc
+yarn node setup-enclave.js  $ENCLAVE_ACCOUNT $NEW_MRENCLAVE $SCHEDULE_UPDATE_BLOCK
```
0xverin: `node setup-enclave.js` => `npx ts-node setup-enclave.ts` It should work if you modify the code like this.

0xverin (commented):
When I initially worked on this, I also followed the process of compiling before running. However, considering that it was inconvenient to compile every time I made changes to the TypeScript code, I switched to using the "npx ts-node" approach. So you can make the changes accordingly and delete the compiled files.

scripts/ts-utils/setup-enclave.ts
```diff
@@ -34,11 +35,39 @@ async function transfer(api: any, Alice: any) {
             });
     });
 }
-async function updateScheduledEnclave(api: any, signAccount: any) {
+async function setTeerexAdmin(api: any, Alice: any) {
     return new Promise(async (resolve, reject) => {
         await api.tx.sudo
-            .sudo(api.tx.teerex.updateScheduledEnclave(0, hexToU8a(`0x${mrenclave}`)))
-            .signAndSend(signAccount, ({ status, events, dispatchError }) => {
+            .sudo(api.tx.teerex.setAdmin("esqZdrqhgH8zy1wqYh1aLKoRyoRWLFbX9M62eKfaTAoK67pJ5"))
```
0xverin, line 38: One small suggestion is to avoid using hardcoded values in script-type code. Instead, consider either reading from a configuration file or accepting input from the command line.:p

scripts/ts-utils/move-vcregistry-snapshot.js
felixfaisal: Not very familiar with TS, Will remove and use this command 😅

scripts/ts-utils/setup_enclave.sh
felixfaisal: Done ✅

0xverin: Thanks! 3

0xverin:
A small suggestion regarding thoroughness. Since these changes(ts-utils) won't be included in the CI for testing, if possible, it would be helpful to provide logs or screenshots demonstrating successful local testing. Of course, I trust that you have already tested it successfully locally. :)

felixfaisal:
No worries 
Screenshot 2023-06-12 at 12 49 43 PM
This is the log for the most recent upgrade testing i did with the changes I commited, Still working on removing the hardcoded account address

0xverin:
Thanks! I believe there are no issues with the `TypeScript` part. The remaining code can be reviewed by others.

Makefile
```diff
@@ -32,7 +32,7 @@ build-all:
 
 .PHONY: build-node ## Build release node with `tee-dev` feature
 build-node:
-	cargo build --locked -p $(call pkgid, $(NODE_BIN)) --release --features=tee-dev
+	~/.cargo/bin/cargo build --locked -p $(call pkgid, $(NODE_BIN)) --release --features=tee-dev --package=litentry-collator
```
Kailai-Wang: I don't think this change is necessary:
`-p $(call pkgid, $(NODE_BIN))` should point to `litentry-collator` already

scripts/ts-utils/endpoint.json
```diff
@@ -1,4 +1,4 @@
 {
-  "fetchEndpoint": "wss://tee-staging.litentry.io:443",
-  "defaultEndpoint": "wss://tee-staging.litentry.io/"
+  "fetchEndpoint": "ws://localhost:11944",
+  "defaultEndpoint": "ws://localhost:11944"
```
Kailai-Wang: Why this change?

scripts/ts-utils/setup-enclave.ts
```diff
@@ -10,10 +10,11 @@ import colors from "colors";
 const transferAmount = "100000000000000";
 const enclaveAccount = process.argv[2];
 const mrenclave = process.argv[3];
-const accountPassword = process.argv[4];
+const block = process.argv[4];
+// const accountPassword = process.argv[4];
 
 //put account in private.json
-const private_account_pair = require("./private.json");
+// const private_account_pair = require("./private.json");
```
Kailai-Wang: Please remove commented lines if we don't need them :)

scripts/ts-utils/setup-enclave.ts
Kailai-Wang: Yes - or at least having a comment about where this address comes from :)

scripts/ts-utils/setup-enclave.ts
```diff
@@ -66,20 +95,21 @@ async function updateScheduledEnclave(api: any, signAccount: any) {
 
 async function main() {
 
+    // NOTE: If we are using development environment, we can use Alice instead of moving around with keys
```
Kailai-Wang, line 94: Do you expect this enclave update script to only work for dev, or is it suitable for prod/staging too?

scripts/ts-utils/setup-enclave.ts
```diff
@@ -66,20 +95,21 @@ async function updateScheduledEnclave(api: any, signAccount: any) {
...
+    // signAccount.toJson(accountPassword);
 
     const { defaultAPI } = await initApi();
     const Alice = keyring.addFromUri("//Alice", { name: "Alice default" });
 
     await transfer(defaultAPI, Alice);
-
-    await updateScheduledEnclave(defaultAPI, signAccount);
+    await setTeerexAdmin(defaultAPI, Alice);
+    await updateScheduledEnclave(defaultAPI, Alice, block);
```
Kailai-Wang, line 104: Hmm - there can be a timing issue here.

The `await` would return as soon as the tx was included in the block, but it doesn't mean the change has taken effect. The sidechain would need to sync parachain blocks to actually execute it.

tee-worker/local-setup/launch-parachain.py
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env python3
+"""
+Launch handily a local dev setup consisting of one integritee-node and some workers.
+
+Example usage: `./local-setup/launch.py /local-setup/simple-config.py`
```
Kailai-Wang: Is the comment up-to-date?
I see no difference between `launch-parachain.py` and `launch-worker.py` :(

tee-worker/local-setup/launch.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+#!/bin/bash
```
Kailai-Wang: duplicate shebang

tee-worker/local-setup/launch.sh
Kailai-Wang, line 1: Any comment why we need this sh script over the original py?

tee-worker/local-setup/release.py
Kailai-Wang, line 1: I'm still a bit confused why we need both `release.py` and `release.sh`, the same applies to `upgrade.py|sh`. Can we have one unified version, or are they actually doing two different things and both are required?

tee-worker/scripts/litentry/migrate_worker.sh
```diff
@@ -0,0 +1,36 @@
...
+# Copy integritee-service binary and enclave_signed.so to ./tmp/w0
+if [ "$PRODUCTION" = "1" ]; then
+  cp ./bin/integritee-service ./tmp/w0
+  cp ./bin/enclave.signed.so  ./tmp/w0
+  cd ./tmp/w0 || exit
+else
+  cp ./bin/integritee-service ./tmp/w0
+  cp ./bin/enclave.signed.so ./tmp/w0
+  cd ./tmp/w0 || exit
+fi
```
Kailai-Wang: What's the diff here? 👀

Kailai-Wang (commented):
Thanks! Apart from the comments I have some more thoughts:
- the PR doesn't seem to solve #1628 completely (which is fine, we just need to mark #1686 only)
- basically, we have two needs regarding automated deployment:
   - reset everything including both parachain and worker, which means stop the current process, clean up, pull latest code, compile and launch everything
   - enclave update on staging or internal, where we only update enclave to a new mrenclave without resetting parachain or worker's old data
Does the PR address both, or one of them?
- I'm a bit confused about both py and sh versions, maybe one is enough?
- It's nice to manage the service via things like systemd

Kailai-Wang:
[Some additional things](https://www.notion.so/web3builders/How-to-upgrade-enclave-one-worker-edfaf5871b4441579f9471074032ed1e?pvs=4#3a66c1ae219b46f389838e178f96e7f4) to be taken into account when updating the enclave

scripts/ts-utils/endpoint.json
felixfaisal: Initially, I was testing with non-`9944` ports, Updated ✅

scripts/ts-utils/setup-enclave.ts
felixfaisal: So here there are two things, It is considered to work for any environment that is using the `dev` chain because here the `sudo` is set to be `Alice`. This is not suitable for Mainnet wherein I guess we shouldn't be calling `sudo` calls via script because then we would have to store the private key on the server running this script. If we consider staging to be a network using `dev` chain then this should be alright.

tee-worker/local-setup/release.py
felixfaisal: I initially wrote `release.py` and `upgrade.py` and Yang mentioned to use bash instead for deploying to avoid having an extra layer of python dependencies since the python script itself is calling the bash commands internally. So I created `*.sh` versions of them and tested them, Removing them now as they are no longer needed ✅

tee-worker/local-setup/launch-parachain.py
felixfaisal: We don't need this, I was just testing things 😅 , Removed ✅

tee-worker/local-setup/launch.sh
felixfaisal: Was using `launch.sh` and `upgrade.sh` for testing the whole scenario and upgrading, Before I created `release.sh`. We don't need this now, Removing ✅

tee-worker/scripts/litentry/migrate_worker.sh
felixfaisal: Removed the IF Block ✅

tee-worker/local-setup/launch.sh
felixfaisal: Removed, No longer needed ✅

scripts/ts-utils/setup-enclave.ts
felixfaisal: Added a comment about this address ✅

scripts/ts-utils/setup-enclave.ts
felixfaisal: Removed ✅

scripts/ts-utils/setup-enclave.ts
felixfaisal: Discussed on PM

scripts/ts-utils/setup-enclave.ts
Kailai-Wang: The staging's root isn't Alice though 😂 (otherwise anyone could do anything on tee-staging parachain)

I'm not sure if you can find an easy way to import/sign the tx with external accounts, it's fine if not - the staging's parachain is unlikely to be reset very often. 

Resetting the internal is more frequent where Alice is the root

scripts/ts-utils/setup-enclave.ts
```diff
@@ -66,20 +92,21 @@ async function updateScheduledEnclave(api: any, signAccount: any) {
 
 async function main() {
 
+    // NOTE: If we are using development environment, we can use Alice instead of moving around with keys
     const keyring = new Keyring({ type: "sr25519" });
-    let signAccount = keyring.addFromJson(private_account_pair);
+    // let signAccount = keyring.addFromJson(private_account_pair);
```
Kailai-Wang: Let's remove commented lines :) same below

tee-worker/scripts/litentry/clean_worker.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+pids=$(pgrep -f "integritee-service")
```
Kailai-Wang: Would `killall integritee-service` work?

tee-worker/scripts/litentry/get_sync_block.sh
Kailai-Wang, line 1: I find this script is unused/commented out

tee-worker/scripts/litentry/migrate_worker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Copy integritee-service binary and enclave_signed.so to ./tmp/w0
```
Kailai-Wang: I think in general it's always good practice to `cd` to the desired working dir first - you never know who is calling the script in which location.

Same applies to other `sh` scripts, you can refer to other existing scripts

tee-worker/scripts/litentry/release/release.sh
```diff
@@ -0,0 +1,336 @@
...
+  echo "Starting Working Service"
+  systemctl --user start worker.service
+  # TODO: Check for block production via Logs
+
+  echo "Parachain and Worker Service have started succesfully, You can check logs at /log/worker0.log"
+  print_divider
+
+}
+
+succesful_parse_of_update_schedule_enclave(){
```
Kailai-Wang: typo: `successful`

tee-worker/scripts/litentry/release/release.sh
```diff
@@ -0,0 +1,336 @@
...
+ [Install]
+ WantedBy=default.target
+ "
+   local service_filename="${service_name}.service"
+   echo "$service_template" > "$service_filename"
+   echo "Service file \"${service_filename}\" generated successfully."
+}
+
+generate_worker_service_file() {
+  local command="/bin/bash -c  'cd ${WORKER_DIR}/tmp/w0 && ./integritee-service --clean-reset -P 2000 -w 2001 -r 3443 -h 4545 --running-mode mock --enable-mock-server --parentchain-start-block 0 run --skip-ra --dev'"
```
Kailai-Wang: The `running-mode` is different on staging and internal

Kailai-Wang: same applies to `--skip-ra --dev` which only exists for development env (internal).

Generally I still favor to put the config into a json file, like the current setup :(

tee-worker/scripts/litentry/release/release.sh
```diff
@@ -0,0 +1,336 @@
...
+  local command="/bin/bash -c 'cd ${WORKER_DIR}/tmp/w0 && ./integritee-service -P 2000 -w 2001 -r 3443 -h 4545 --running-mode mock --enable-mock-server --parentchain-start-block ${LATEST_FINALIZED_BLOCK} run --skip-ra --dev'"
+  local service_name='worker'
+  local description='Worker Service for Litentry Side chain'
+  local working_directory='/opt/worker/'
+
+  generate_service_file_for_worker "${service_name}" "${description}" "${command}" "${working_directory}"
+}
+
+generate_parachain_service_file() {
+  local command="/bin/bash -c 'source /home/faisal/.nvm/nvm.sh && cd ${PARACHAIN_SOURCE} && scripts/launch-local-binary.sh rococo && sleep infinity'"
```
Kailai-Wang: this command is guaranteed not working 😂

tee-worker/scripts/litentry/release/release.sh
```diff
@@ -0,0 +1,336 @@
...
+  if [[ $? -eq 0 ]]; then
+      mrenclave_value=$(echo "$output" | awk '{print $2}')
+      echo "MRENCLAVE value: $mrenclave_value"
+      export NEW_MRENCLAVE="$mrenclave_value"
+  else
+      echo "Failed to extract MRENCLAVE value."
+  fi
+}
+
+latest_sync_block(){
```
Kailai-Wang: Is it sidechain block? If so it's better to explicitly say it in the function name -  and this function seems to do more than getting the sidechain block. More importantly it sets the scheduled update block I believe

tee-worker/scripts/litentry/release/release.sh
```diff
@@ -0,0 +1,336 @@
...
+  else
+    ./local-setup/launch.sh
+  fi
+fi
+
+# TODO: Need to also perform clean reset of the service
+if [ "$CLEAN_RESET" = "true" ]; then
+  # Assuming the worker has already stopped
+  echo "Preparing to Upgrade Worker"
+fi
```
Kailai-Wang: This is a bit confusing - `clean_reset` normally happens in **resetting** the worker, not upgrading

tee-worker/scripts/litentry/release/release.sh
```diff
@@ -0,0 +1,336 @@
...
+if [ "$UPGRADE_WORKER" = "1" ]; then
+  # Assuming the worker has already stopped
+  echo "Preparing to Upgrade Worker"
+  upgrade_worker
+  exit 0
+fi
+
+# Example usage
+# generate_service_file "my_service" "My Service Description" "/path/to/my_script.sh" "/path/to/working_directory"
+if [ "$PRODUCTION" = "1" ]; then
```
Kailai-Wang: Where is `PRODUCTION` set?

Basically I'm not a fan of controlling the script logic via envs, which is not apparent enough for the users. I prefer to do it via command line args, so IMO we should either have one script, with the second param specifying the action to take (e.g. `upgrade` or `reset`) followed by the respective args, or we have two individual scripts for upgrade and reset.

Kailai-Wang (changes requested):
I looked into it in more detail, while I like some of the ideas, I must admit it could have been better if we had broken this task into smaller ones.

I feel that it could be reworked/improved in a cleaner way, some of major issues that I found:
- no README, I notice there's a notion doc, but it's a bit too much if I need to check notion each time when I run this script. It would be great if we have some "brief manual", self-explanatory is even better (like `--help` or usage printing)
- the whole workflow is not clear enough. As I mentioned in one of the comments, IMO we should either have one script, with the second param specifying the action to take (e.g. upgrade or reset) followed by corresponding args, or we have two individual scripts for upgrade and reset, respectively.
- two many small `sh` scripts make it hard to comprehend and maintain. Some could have been part of the main script as functions
- script code logic could be improved: like better naming, removal of duplicate code, more explicit control of different branches etc

scripts/ts-utils/setup-enclave.ts
Kailai-Wang, line 1: Maybe this one should be moved into tee-worker/ dir? Because it's enclave related

Kailai-Wang:
So in general, we'd like to reach the goal that even first-time users know by browsing the script:
- what can the script do
- how to do it 
- how to control distinct features
- what are the differences between features (e.g. different args, execution logic etc)

I haven't got a clear answer to these for the current impl :(

scripts/ts-utils/setup-enclave.ts
felixfaisal: Removed ✅

tee-worker/scripts/litentry/clean_worker.sh
felixfaisal: It actually did not work, but this is no longer needed, Will be removed ✅

tee-worker/scripts/litentry/release/release.sh
felixfaisal: This script is no longer used, parsing JSON instead in `deploy.sh`

tee-worker/scripts/litentry/release/release.sh
felixfaisal: This script is no longer used, Using different commands based on `launch-local-binary.sh` and generate service files for collator, Relay chain node Alice and Relay chain node Bob

tee-worker/scripts/litentry/release/release.sh
felixfaisal: Updated, Using a JSON config ✅

tee-worker/scripts/litentry/release/release.sh
felixfaisal: Updated in new script `deploy.sh`

tee-worker/scripts/litentry/release/release.sh
felixfaisal: No longer needed, Using a new script `deploy.sh`

tee-worker/scripts/litentry/release/release.sh
felixfaisal: Using CLI in new script `deploy.sh`

tee-worker/scripts/litentry/migrate_worker.sh
felixfaisal: All small scripts moved into main script `deploy.sh`

tee-worker/scripts/litentry/setup_enclave.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Also need to set the port
+
+cd $ROOTDIR/scripts/ts-utils/
+yarn install
+npx ts-node setup-enclave.ts  $ENCLAVE_ACCOUNT $NEW_MRENCLAVE $SCHEDULE_UPDATE_BLOCK
```
Kailai-Wang: With these envs it seems that this script isn't supposed to be called alone, is it? If so we should put those in the script where these envs are set, to make sure they always have the correct value IMO

scripts/ts-utils/setup-enclave.ts
felixfaisal: Moved ✅

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+export ROOTDIR=$(git rev-parse --show-toplevel)
+
+
+function print_divider() {
+  echo "------------------------------------------------------------"
+}
+
+# Function to display the script's usage
+function display_help() {
+  echo "Usage: ./release.sh launch --build"
```
Kailai-Wang: This needs to be adjusted

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+  echo "  - It is mandatory to provide a JSON config file for the worker."
+  echo "  - jq is required to be installed on the system "
+  echo ""
+  echo "For more information or assistance, please contact Faisal."
+}
+
+generate_service_file() {
+  if [ "$#" -ne 5 ]; then
+    echo "Usage: generate_service_file <service_name> <description> <command> <working_directory> <log_file_path>"
+    return 1
```
Kailai-Wang: `return` is probably not enough, we need `exit` here

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+      file1="${files[0]}"
+      file2="${files[1]}"
+      if [[ $file1 -ot $file2 ]]; then
+          echo "Deleting the older shard folder: $file1"
+          rm -r "$file1"
+      else
+          echo "Deleting the older shard folder: $file2"
+          rm -r "$file2"
+      fi
+  fi
```
Kailai-Wang: While it should work too, I still prefer deleting the exact old enclave-shard folder when possible

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+  # JSON-RPC request payload
+  request='{"jsonrpc":"2.0","id":1,"method":"chain_getHeader","params":[]}'
+
+  # Make the JSON-RPC request and retrieve the latest finalized block
+  response=$(curl -s -H "Content-Type: application/json" -d "$request" http://$PARACHAIN_HOST:$PARACHAIN_PORT)
+  hex_number=$(echo "$response" | grep -oP '(?<="number":")[^"]+')
+  dec_number=$(printf "%d" "$hex_number")
+
+
+  # Store the latest finalized block numbe  r in an environment variable
```
Kailai-Wang: extra spaces

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+
+function build_parachain(){
+  cd $ROOTDIR
+  make build-node
+}
+
+function build_worker(){
+  cd $ROOTDIR/tee-worker/
+  source /opt/intel/sgxsdk/environment
+  SGX_COMMERCIAL_KEY=/home/faisal/litentry-parachain/tee-worker/enclave-runtime/Enclave_private.pem SGX_PRODUCTION=1 make
```
Kailai-Wang: `SGX_COMMERCIAL_KEY=/home/faisal/...`

This path will not work

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+    mrenclave_value=$(echo "$output" | awk '{print $2}')
+    echo "MRENCLAVE value: $mrenclave_value"
+    export OLD_MRENCLAVE="$mrenclave_value"
+  else
+    echo "Failed to extract MRENCLAVE value."
+    exit 1
+  fi
+fi
+
+# Move log files  to log-backup
```
Kailai-Wang: extra space

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+export CONFIG=$(cat $config)
+
+if [ "$discard" = true ]; then
+  echo "Cleaning the existing state for Parachain and Worker."
+  stop_running_services
+  rm -rf /tmp/parachain_dev/
+  rm -rf $ROOTDIR/tee-worker/tmp
+fi
+
+# Get Old MRENCLAVE
```
Kailai-Wang: `old` I believe :)

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+    exit 1
+  fi
+fi
+
+# Move log files  to log-backup
+if [ -d "log" ]; then
+  new_folder_name=$(date +"log-backup/log-%Y%m%d-%H%M%S")
+  cp -r "log" "$new_folder_name"
+  echo "Backup log into $new_folder_name"
+fi
```
Kailai-Wang: Can you backup everything or just the whole folder `tmp/w${i}` before doing any change?
cc @BillyWooo please confirm if that's enough

The backup should be carried out on both restart (even though it's a reset) and upgrade

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,640 @@
...
+
+  # Store the latest finalized block numbe  r in an environment variable
+  export LATEST_FINALIZED_BLOCK=${dec_number}
+
+  echo "Latest finalized block number: $LATEST_FINALIZED_BLOCK"
+}
+
+function build_parachain(){
+  cd $ROOTDIR
+  make build-node
```
Kailai-Wang: `build-node` is with `tee-dev` feature active - if on production you probably need to compile it without this feature flag.

Additionally, I believe copy from the docker image `litentry/litentry-parachain:tee-prod` would be easier, maybe you want to add an extra flag to control if you want to build the parachain binary from source or copy from docker image

Kailai-Wang (commented):
Thanks, it looks much better now.

Have you tested both functionalities, at least locally?

And you probably missed this comment: https://github.com/litentry/litentry-parachain/pull/1750#issuecomment-1610891055

tee-worker/scripts/litentry/setup_enclave.sh
felixfaisal: Yeah it's not supposed be called alone, Let me move them

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I actually couldn't find the relation between `MRENCLAVE` -> `Shard Folder Name` , that is why I decided to delete the older folder and I think the same was mentioned in the instructions for upgrading, Is there a way to find this relationship ?

Kailai-Wang: MRENCLAVE is the hex presentation, shard is the same as MRENCLAVE, but the shard folder is using the base58 form I believe.

If you find it hard to do conversion from command line we can modify `make mrenclave` to output both formats

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: So building is an optional flag with `--build`, If we want to use separate build process, we can pre-build. I'll look into getting the binary

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Done

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Removed

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Put in correct directory

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Done

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Done

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Done

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I think I should be able to do in bash, I'll update this code ✅

felixfaisal: Done ✅

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Done, 
So added a flag `--copy-from-docker` so user can enter their own docker image to pull from else it will default to `litentry/litentry-parachain:tee-prod`

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I took the logic to backup the logs from the python script for launching the worker, I used the same to back up the tmp folder 

```
if [ -d "$ROOTDIR/tee-worker/tmp/w$i" ]; then
          new_folder_name=$(date +"$ROOTDIR/tee-worker/tmp/w$i-%Y%m%d-%H%M%S")
          mkdir -p new_folder_name
          cp -r $ROOTDIR/tee-worker/tmp/w$i $new_folder_name
          echo "Backing up, previous worker binary $new_folder_name"
      fi
```
This will take place during restart and upgrade-worker
CC: @BillyWooo

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Included `Production` build that builds without the `tee-dev` feature

BillyWooo (approved):
LGTM. I think it's good to go.

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -0,0 +1,721 @@
...
+  if [ "$COPY_FROM_DOCKER" = true ]; then
+    docker pull litentry/litentry-parachain:tee-prod
+    img_id=$(docker create litentry/litentry-parachain:tee-prod)
+    docker cp $img_id:/usr/local/bin/litentry-collator $ROOTDIR/
+    docker rm -v $img_id
+  else
+    if [ "$PRODUCTION" = 1 ]; then
+      cd $ROOTDIR
+      # It builds without the `tee-dev` feature
+      make "build-runtime-$CHAIN"
```
Kailai-Wang, line 553: I think you still need to build the node, just without the `tee-dev` feature

Kailai-Wang (approved):
Thank you - I approve it

I still believe it's better to control `PRODUCTION`or not from the command line, because it's hidden in the code and people can easily forget it while reading the help message

But maybe let's use the script for a while and do it in another PR if we really see the needs

Kailai-Wang:
And let's not forget to delete the branch after being merged :)

Issue #1752, opened 2023-06-06
Accept wrapped message when verifying the signature of trusted calls 


Kailai-Wang:
### Context

A similar problem to https://github.com/litentry/litentry-parachain/issues/1137
F/E couldn't get the signature verification of DI requests working as the browser extension js forcibly add the wrapped `<Bytes>` to the raw payload (so the user always signs the wrapped version in real scenarios).

We need to adjust the rust code to accept the wrapped version too, ideally with ts-test.

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Issue #1725, opened 2023-05-22
DI port on the instance not reachable in browser


Kailai-Wang:
### Context

Might be a mismatch of TLS signature scheme:
https://github.com/integritee-network/worker/issues/1316

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
Found this too:
https://security.stackexchange.com/questions/269725/what-is-the-current-april-2023-browser-support-for-ed25519-certificate-signatu

Kailai-Wang:
And

https://issuetracker.google.com/issues/231334600

Kailai-Wang:
Tried ecdsa - it looked better now.

With port forwarding (to localhost) and the experimental flag `chrome://flags/#allow-insecure-localhost` I'm able to connect to `wss//localhost:2000` in browser:

![image](https://github.com/litentry/litentry-parachain/assets/7630809/5c830a7b-5431-41cf-aed6-82e3e83e502b)

The only question now is if IDHub can accept self-signed-cert by default.

hanwencheng:
Is it a blocker?

Kailai-Wang:
> Is it a blocker?

Not a blocker - IDHub is using proxy and port forwarding for that
But it's still nice to get it completely solved (which is not so trivial)

Kailai-Wang:
Something to read: https://gramine.readthedocs.io/en/stable/attestation.html#mid-level-ra-tls-interface

PR #1761, opened 2023-06-08
Remove the usage of challenge code

### LINKED ISSUE #1739: Replace challenge code with sidechain account nonce
### Context

Why did we need the challenge code?
Because we wanted to add random salt to identity verification so that we could differentiate between **this time** and **next time**. A fabricated example:
- I link my Twitter account
- I unlink it
- I transfer my Twitter account to someone else, and I don't actually own the old Twitter account anymore
 
Without a challenge code, I can still verify my old Twitter account by providing the tweet ID for the old verification message.

The drawback of the challenge code is that it forcibly splits the verification process into two steps - you can only post a verification message once you get the challenge code, which worsens the user experience.

So we need something deterministic, constantly changing and ideally known by both the client and TEE-worker.

I'm thinking if we can use the sidechain nonce of the litentry address (or the primary address) to replace challenge code, in this way we can do the linking and verification with one request. The nonce is different each time, yet it's predictable so the TEE-worker knows what the expected verification message would be, and the client can construct such message without communicating with tee-worker beforehand.

The nonce retrieval is public, but it doesn't matter. Even if it's somehow eavesdropped the attacker still can't prove his ownership of both identities.

Again the challenge code was not there for the verification itself, but as something to prevent replay forgery of a verification.

Additionally we want to add another source of randomness by including the user shielding key to calculate the message digest, see the discussion [here](https://litentry.slack.com/archives/C03KR96989G/p1685227072451249). So the raw verification message is calculated 
```
blake2_256(<nonce> + shieldingKey.encrypt(<primary account> + <identity-to-be-linked>).ciphertext)
```

, where `<>` means SCALE-encoded.

---

Kailai-Wang:
resolves #1739 

This is a big PR trying to:
- remove IMP mock - it's not used anymore
- replace the challenge code with sidechain nonce and encrypted identities
- deprecate `create_identity` and `verify_identity` by merging them into one extrinsic: `link_identity`
- remove `metadata` when linking identities - it's never been used so far
- simplify `IdentityContext`. Now it has two fields only: `linkBlock` and `status`
- simplify mock-server getter
- adjust ts-test
- remove some unnecessary test cases

TODO:
- re-benchmark the IMP pallet
- avoid using hardcoded sidechain-nonce in ts-test
- add example code for DI
- other TODOs in ts-test

I'd like to delay the merging of this PR until the other two issues are solved, so that we can have a release with those fixes for IDHub without adding this new `link_identity` feature:
-  https://github.com/litentry/litentry-parachain/issues/1752
~~https://github.com/litentry/litentry-parachain/issues/1725~~ it's not a blocker right now (but needs to be fixed asap)


Kailai-Wang:
I still need to resolve the conflicts, but basically it can be reviewed

edit: conflicts resolved.

primitives/core/src/key.rs
```diff
@@ -30,11 +30,13 @@ pub type ShardIdentifier = sp_core::H256;
 pub type MrenclaveType = [u8; 32];
 
 // we use 256-bit AES-GCM as user shielding key
+// TODO: use constants from `ring` crate, e.g. ring::aead::NONCE_LEN
```
kziemianek, line 33: can this be done within this PR ?

runtime/litmus/src/weights/mod.rs
```diff
@@ -28,7 +28,7 @@ pub mod pallet_democracy;
 pub mod pallet_drop3;
 pub mod pallet_extrinsic_filter;
 pub mod pallet_identity;
-pub mod pallet_identity_management;
+// pub mod pallet_identity_management;
```
kziemianek, line 31: can this be removed?

runtime/rococo/src/weights/mod.rs
```diff
@@ -27,7 +27,7 @@ pub mod pallet_democracy;
 pub mod pallet_drop3;
 pub mod pallet_extrinsic_filter;
 pub mod pallet_identity;
-pub mod pallet_identity_management;
+// pub mod pallet_identity_management;
```
kziemianek, line 30: maybe this can be removed aswell ?

tee-worker/app-libs/stf/src/getter.rs
```diff
@@ -93,7 +91,6 @@ impl TrustedGetter {
 			// litentry
 			TrustedGetter::user_shielding_key(account) => account,
 			TrustedGetter::id_graph(account) => account,
-			TrustedGetter::challenge_code(account, _) => account,
```
kziemianek, line 96: Just wanted to confirm that we can ignore all ongoing create identity processes.
Users who started process before new version deployment wouldn't be able to end it.

tee-worker/core/parentchain/indirect-calls-executor/src/executor.rs
```diff
@@ -120,13 +120,8 @@ impl<
...
 		let block_hash = block.hash();
 
-		// let block_number_convert: u32 = block_number.try_into();
-		let mut block_num: u32 = 0;
-		if let Ok(number) = block_number.try_into() {
-			block_num = number;
-		}
-
 		debug!("Scanning block {:?} for relevant xt", block_number);
+		let block_number_u32: u32 = block_number.try_into().unwrap_or_default();
```
kziemianek, line 124: What do you think about removing this variable? It's used only in one place and the expression never fails.

tee-worker/litentry/core/identity-verification/src/lib.rs
```diff
@@ -49,28 +46,11 @@ mod web3;
 mod error;
 use error::{Error, Result};
 
-pub fn verify(r: &IdentityVerificationRequest) -> Result<()> {
+pub fn verify(r: &IdentityLinkRequest) -> Result<()> {
```
kziemianek: I think we should keep the old naming here, we are in the context of verification, not identity linking.

tee-worker/litentry/core/identity-verification/src/web3/mod.rs
```diff
@@ -112,25 +121,27 @@ fn verify_substrate_signature_internal(
...
 	validation_data: &Web3CommonValidationData,
 ) -> Result<()> {
-	let msg = get_expected_raw_message(who, identity, code);
+	let msg = get_expected_raw_message(who, identity, sidechain_nonce, key, nonce);
 	let digest = compute_evm_msg_digest(&msg);
 	let evm_address = if let Identity::Evm { address, .. } = identity {
 		address
 	} else {
-		return Err(Error::VerifyIdentityFailed(ErrorDetail::InvalidIdentity))
+		return Err(Error::LinkIdentityFailed(ErrorDetail::InvalidIdentity))
```
kziemianek, line 134: Same here, I think it's unnecessarily coupled to identity linking usecase.

tee-worker/litentry/core/mock-server/src/twitter_official.rs
```diff
@@ -15,45 +15,51 @@
...
-	F: Fn(&AccountId, &Identity) -> ChallengeCode + Send + Sync + 'static,
+	F: Fn(&Sr25519Pair) -> UserShieldingKeyType + Send + Sync + 'static,
 {
 	warp::get()
 		.and(warp::path!("2" / "tweets" / u32))
 		.and(warp::query::<HashMap<String, String>>())
-		.map(move |_tweet_id, p: HashMap<String, String>| {
-			log::info!("query_tweet");
+		.map(move |tweet_id: u32, p: HashMap<String, String>| {
+			println!("query_tweet, tweet_id: {}", tweet_id);
```
kziemianek, line 36: can we use log instead ?

tee-worker/litentry/core/stf-task/sender/src/lib.rs
```diff
@@ -67,13 +68,15 @@ use sp_std::prelude::Vec;
 /// https://www.notion.so/web3builders/Sidechain-block-importer-and-block-production-28292233b4c74f4ab8110a0014f8d9df
 
 #[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
-pub struct IdentityVerificationRequest {
+pub struct IdentityLinkRequest {
```
kziemianek: From my understanding we are still requesting verification, and if its ok then linking is performed.

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -194,150 +147,62 @@ pub mod pallet {
...
-		pub fn remove_challenge_code(
-			origin: OriginFor<T>,
-			who: T::AccountId,
-			identity: Identity,
-		) -> DispatchResult {
-			T::ManageOrigin::ensure_origin(origin)?;
 			ensure!(
-				ChallengeCodes::<T>::contains_key(&who, &identity),
-				Error::<T>::ChallengeCodeNotExist
+				!IDGraphs::<T>::contains_key(&who, &identity),
```
kziemianek, line 159: Ensuring the identity cannot be linked to another account is out of the scope of this PR ?

runtime/litmus/src/weights/mod.rs
Kailai-Wang: It will to be activated very soon - in the next PR I'll re-benchmark the pallet and enable this.

Wanted to add a comment but forgot it :(

tee-worker/core/parentchain/indirect-calls-executor/src/executor.rs
Kailai-Wang: I see your point - yes it makes sense.

I'm probably not going to change it in this PR tho as we are going to change it anyway in an upcoming PR where we want to remove the unnecessary types in the indirect-call-executor (talked with @felixfaisal)

tee-worker/litentry/core/mock-server/src/twitter_official.rs
Kailai-Wang: `log` won't print anything.
I haven't looked into it - maybe because it's globally overridden by the sgx version

I know we have a mixture of `println` and `log`, but that's mainly from upstream. They seem to use `println` in non-enclave and `log` in enclave.

tee-worker/litentry/core/identity-verification/src/web3/mod.rs
Kailai-Wang: I'll be against this though. The error will be bubbled to F/E in the end.

There's only one `link_identity` ext now, so you either expect `IdentityLinked` or `LinkIdentityFailed`. Having additional error events might be confusing.
What do you think?

tee-worker/litentry/core/identity-verification/src/lib.rs
Kailai-Wang: Sounds good. I'll revert it - it's indeed a verification request.

tee-worker/litentry/core/stf-task/sender/src/lib.rs
Kailai-Wang: Reverted!

tee-worker/app-libs/stf/src/getter.rs
Kailai-Wang: Yes - it's still test version in the end - we can wait a bit if there're unhandled requests, or we simply upgrade it and ignore all pending requests, if any.

primitives/core/src/key.rs
Kailai-Wang: There're multiple places - I'll create an issue for e.g. zhouran so that they can get familiar with the codebase

runtime/rococo/src/weights/mod.rs
Kailai-Wang: Same as above.

tee-worker/litentry/pallets/identity-management/src/lib.rs
Kailai-Wang: That's right, it belongs to EVM-sign-in feature or a separate issue if need.

tee-worker/litentry/core/identity-verification/src/web3/mod.rs
kziemianek: I though the error could be mapped, but yes on the other hand it would make the code more complex.
I'm still in favor of getting rid of all `link` mentions here if it's possible but we can keep it as it is as well if you are not convinced enough 👍🏼

kziemianek (approved):

PR #1762, opened 2023-06-09
fix: change the method of determining the Substrate network.

0xverin:
[The previous Substrate network type determination](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/common/helpers.ts#L59) was incorrect, as it was not related to whether the network was locally launched. In fact, it is associated with the SS58 registry. This PR aims to fix this issue.

### DIFF
--- a/tee-worker/ts-tests/common/helpers.ts
+++ b/tee-worker/ts-tests/common/helpers.ts
@@ -56,7 +56,7 @@ export function identity(data: HexString | Uint8Array): Uint8Array {
     return u8aToU8a(data);
 }
 
-// https://github.com/paritytech/ss58-registry/blob/main/c.json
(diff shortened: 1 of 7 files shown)


tee-worker/ts-tests/common/helpers.ts
```diff
@@ -56,4 +56,13 @@ export function identity(data: HexString | Uint8Array): Uint8Array {
     return u8aToU8a(data);
 }
 
-export const env_network = process.env.NODE_ENV === 'local' ? 'TestNet' : 'LitentryRococo';
+// https://github.com/paritytech/ss58-registry/blob/main/c.json
```
Kailai-Wang: I think it's better to point it to our rust codebase, as not every ss58prefix is listed in the registry JSON, e.g. `LitentryRococo`, `TestNet`.

Kailai-Wang (approved):
Looks good, thanks!

Some comments are adjusted in #1761 btw

kziemianek (approved):

PR #1790, opened 2023-06-19
Twitter follower assertion

### LINKED ISSUE #1789: Twitter follower

kziemianek:
Can be used to generate VC proving that account is following specified twitter acount at generation time.

closes: #1789

### DIFF
--- a/primitives/core/src/assertion.rs
+++ b/primitives/core/src/assertion.rs
@@ -52,6 +52,7 @@ pub enum Assertion {
 	A10(ParameterString),                                   // (minimum_amount)
 	A11(ParameterString),                                   // (minimum_amount)
 	A13(u32),                                               // (Karma_amount) - TODO: unsupported
+	TwitterFollower(ParameterString)						// (target_twitter_screen_name)
 }
 
 pub const ASSERTION_FROM_DATE: [&str; 7] = [
--- a/primitives/core/src/assertion.rs
+++ b/primitives/core/src/assertion.rs
@@ -51,8 +51,9 @@ pub enum Assertion {
 	A9,
 	A10(ParameterString),                                   // (minimum_amount)
 	A11(ParameterString),                                   // (minimum_amount)
+	A12(ParameterString),									// (target_twitter_screen_name)
 	A13(u32),                                               // (Karma_amount) - TODO: unsupported
-	TwitterFollower(ParameterString)						// (target_twitter_screen_name)
+
 }
 
 pub const ASSERTION_FROM_DATE: [&str; 7] = [
--- a/tee-worker/litentry/core/assertion-build/src/twitter_follower.rs
+++ b/tee-worker/litentry/core/assertion-build/src/twitter_follower.rs
@@ -1,0 +1,109 @@
+// Copyright 2020-2023 Litentry Technologies GmbH.
+// This file is part of Litentry.
+//
+// Litentry is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Litentry is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+#[cfg(all(feature = "std", feature = "sgx"))]
+compile_error!("feature \"std\" and feature \"sgx\" cannot be enabled at the same time");
+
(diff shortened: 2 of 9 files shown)


kziemianek:
I haven't tested it locally yet.

tee-worker/litentry/core/assertion-build/src/twitter_follower.rs
```diff
@@ -0,0 +1,109 @@
...
+};
+use log::*;
+use std::{format, vec::Vec};
+
+const VC_SUBJECT_DESCRIPTION: &str = "The user has followed {:?}";
+const VC_SUBJECT_TYPE: &str = "A follower of the {:?}";
+const VC_SUBJECT_TAG: [&str; 1] = ["Twitter"];
+
+pub fn build(
+	twitter_screen_name: ParameterString,
```
Kailai-Wang: What is twitter_screen_name? Is it some phrase in twitter API?

primitives/core/src/assertion.rs
```diff
@@ -52,6 +52,7 @@ pub enum Assertion {
 	A10(ParameterString),                                   // (minimum_amount)
 	A11(ParameterString),                                   // (minimum_amount)
 	A13(u32),                                               // (Karma_amount) - TODO: unsupported
+	TwitterFollower(ParameterString)						// (target_twitter_screen_name)
```
Kailai-Wang: We've had bad experiences in giving the assertion a descriptive name because we could have so many similar assertions that differ only slightly. That's why we came up with those `A*` and leave the explanation in the comments.

I'm a bit inclined to the originally suggested `Decoded2023_1` name - although it doesn't have a clear meaning - it gives the benefits that we immediately recognise it's used for decoded. Under the hood it uses the same implementation as some A* assertions, which is fine

Kailai-Wang (commented):
The logic looks good - I only have comments for some details

tee-worker/litentry/core/assertion-build/src/twitter_follower.rs
kziemianek: It's a query parameter.

https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/follow-search-get-users/api-reference/get-friendships-show

primitives/core/src/assertion.rs
kziemianek: I don't treat these assertions as strongly coupled to Decoded, that's why I suggested different name, but I'm not certain about that. Definetly it makes sense to follow the convention so I can name it `a12` as well. Do you think this assertion will stay with us for longer ?

BillyWooo: I think it will. Even not so long, it won't hurt to follow the convention.

tee-worker/litentry/core/assertion-build/src/twitter_follower.rs
BillyWooo: Why don't just copy a5.rs and remove the last `retweet` part?

BillyWooo (commented):
In general, I think it's fine to clone a5.rs and adapt a bit. That's enough. How do you think?

tee-worker/litentry/core/assertion-build/src/twitter_follower.rs
kziemianek: I imagine user providing twitter account username he is following, not tweet id of that account. That's why I introduced this slightly change.

kziemianek:
> In general, I think it's fine to clone a5.rs and adapt a bit. That's enough. How do you think?

@BillyWooo this is the adapted copy of `a5`.

tee-worker/litentry/core/assertion-build/src/twitter_follower.rs
Kailai-Wang: All good - yes I can imagine the F/E will give the target username/handle as input parameter

primitives/core/src/assertion.rs
Kailai-Wang: `A12` would be the safest name.

As I previously mentioned I'm only a bit concerned about assertions that are quite similar yet still differ a bit. E.g. for the decoded LIT/DOT Holder assertions, they are similar to A4/A7 except that we only count from 2023-01-01 to now. In this sense, giving them a dedicated `A*` name seems overkill. That's why I suggested `Decoded2023_*` series: maybe we can even drop them after the event.

kziemianek: If we plan to stay longer with existing assertions we can:

- add another one similar to A4/A7
- add another one but more parametrized, so it could cover also A4/A7 and mark A4/A7 as deprecated and remove it in the future.
- change A4/A7 so they can cover the new use case as well (potentially breaking change).

If the second/third option sounds valuable let's schedule changes.

For now let's stay with the new assertion named A*, even for token holding, what do you think?

Kailai-Wang: Sure let's go for `A*` for now

Kailai-Wang (approved):

PR #1796, opened 2023-06-21
Add IDGraph getter and twitter identity linking in DI-sample

### LINKED ISSUE #1763: Follow-up of #1761: add linkIdentity in DI sample-code
### Context

As topic, one of the follow-up TODOs of https://github.com/litentry/litentry-parachain/pull/1761

We want to add the `linkIdentity` request sending/receiving to DI sample code.

---

Kailai-Wang:
### Context

resolves #1763 

It fixes the twitter link error and checks the idgraph after the linking.

TODO: somehow `createType('Option<Vec<(....)>>')` doesn't work, I need to split the decoding into two steps.

### DIFF
--- a/tee-worker/app-libs/stf/src/getter.rs
+++ b/tee-worker/app-libs/stf/src/getter.rs
@@ -183,7 +183,9 @@ impl ExecuteGetter for TrustedGetterSigned {
 			// litentry
 			TrustedGetter::user_shielding_key(who) =>
 				IdentityManagement::user_shielding_keys(&who).map(|key| key.encode()),
-			TrustedGetter::id_graph(who) => Some(IdentityManagement::get_id_graph(&who).encode()),
+			// use get_id_graph_with_max_len for deterministic result
+			TrustedGetter::id_graph(who) =>
+				Some(IdentityManagement::get_id_graph_with_max_len(&who, usize::MAX).encode()),
 			// TODO: we need to re-think it
 			//       currently, _who is ignored meaning it's actually not a "trusted" getter.
 			//       In fact, in the production no one should have access to the concrete identities
--- a/tee-worker/app-libs/stf/src/getter.rs
+++ b/tee-worker/app-libs/stf/src/getter.rs
@@ -183,9 +183,8 @@ impl ExecuteGetter for TrustedGetterSigned {
 			// litentry
 			TrustedGetter::user_shielding_key(who) =>
 				IdentityManagement::user_shielding_keys(&who).map(|key| key.encode()),
-			// use get_id_graph_with_max_len for deterministic result
 			TrustedGetter::id_graph(who) =>
-				Some(IdentityManagement::get_id_graph_with_max_len(&who, usize::MAX).encode()),
+				Some(IdentityManagement::get_id_graph(&who, usize::MAX).encode()),
 			// TODO: we need to re-think it
 			//       currently, _who is ignored meaning it's actually not a "trusted" getter.
 			//       In fact, in the production no one should have access to the concrete identities
--- a/tee-worker/litentry/pallets/identity-management/src/lib.rs
+++ b/tee-worker/litentry/pallets/identity-management/src/lib.rs
@@ -243,13 +243,9 @@ pub mod pallet {
 			IDGraphs::<T>::remove(owner, identity);
 		}
 
-		pub fn get_id_graph(who: &T::AccountId) -> IDGraph<T> {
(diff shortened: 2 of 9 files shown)


tee-worker/app-libs/stf/src/getter.rs
```diff
@@ -183,7 +183,9 @@ impl ExecuteGetter for TrustedGetterSigned {
 			// litentry
 			TrustedGetter::user_shielding_key(who) =>
 				IdentityManagement::user_shielding_keys(&who).map(|key| key.encode()),
-			TrustedGetter::id_graph(who) => Some(IdentityManagement::get_id_graph(&who).encode()),
+			// use get_id_graph_with_max_len for deterministic result
```
kziemianek: What kind of problems was previous implementation causing?

kziemianek:
I think it should be also labeled as breaking change if the trusted getter is not returning the whole idgraph anymore.

tee-worker/app-libs/stf/src/getter.rs
Kailai-Wang: You will get non-deterministic/out-of-order results because it's random access to a map.
Having a (predefined) sorted result makes the verification easier IMO

Kailai-Wang:
> I think it should be also labeled as breaking change if the trusted getter is not returning the whole idgraph anymore.

It is :) It's returning `IdentityManagement::get_id_graph_with_max_len(&who, usize::MAX)`

tee-worker/app-libs/stf/src/getter.rs
kziemianek: Do you mean assertions in di-examples and potentially ts-tests ?

Kailai-Wang: no only the trustedGetters for IDGraph.

Ts-tests use the value in AesOutput where it's already sorted.

kziemianek: I think we should sort `IdGraph` in `get_id_graph` anyway, what do you think ?

Kailai-Wang: or maybe only keep a single version of `get_id_graph` which is with MAX_LENGTH? You use `usize::MAX` if you want a full IDGraph

kziemianek: yes, that's even better.

kziemianek: I think every caller of this getter will face the same ordering problems as we now.

kziemianek:
> > I think it should be also labeled as breaking change if the trusted getter is not returning the whole idgraph anymore.
> 
> It is :) It's returning `IdentityManagement::get_id_graph_with_max_len(&who, usize::MAX)`

Yes, you are right. I somehow didn't noticed it's `usize::MAX`.

tee-worker/app-libs/stf/src/getter.rs
Kailai-Wang: I updated to have only one `get_id_graph` now

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -243,13 +243,9 @@ pub mod pallet {
...
-		pub fn get_id_graph(who: &T::AccountId) -> IDGraph<T> {
-			IDGraphs::<T>::iter_prefix(who).collect::<_>()
-		}
-
 		// get the most recent `max_len` elements in IDGraph
-		pub fn get_id_graph_with_max_len(who: &T::AccountId, max_len: usize) -> IDGraph<T> {
-			let mut id_graph = Self::get_id_graph(who);
+		pub fn get_id_graph(who: &T::AccountId, max_len: usize) -> IDGraph<T> {
+			let mut id_graph = IDGraphs::<T>::iter_prefix(who).collect::<IDGraph<T>>();
 			id_graph.sort_by(|a, b| Ord::cmp(&b.1.link_block, &a.1.link_block));
```
kziemianek, line 249: I think this can still lead to non deterministic ordering but with much lower probability (several identities created within same block).

kziemianek (approved):
Looks fine to me 🙃

tee-worker/litentry/pallets/identity-management/src/lib.rs
Kailai-Wang: Strictly speaking that's true - but that should only happen very rarely without batching

PR #1801, opened 2023-06-21
token holder assertion

### LINKED ISSUE #1792: DOT/LIT Token Hodler assertion
- Assert that account have been consistently holding at least 10 LIT tokens from date 2023 Jan 1st 00:00:00 UTC till now on blockchain: Litentry.
- Assert that account have been consistently holding at least 10 DOT tokens from date 2023 Jan 1st 00:00:00 UTC till now on blockchain: Polkadot.

kziemianek:
Adds assertion for proving of constantly holding selcted token in provided years.
Please do not merge until I confirm it works e2e. I'm opening PR because we are short on time for Decoded.

closes: #1792


primitives/core/src/assertion.rs
```diff
@@ -23,8 +23,24 @@ use sp_runtime::{traits::ConstU32, BoundedVec};
...
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, TypeInfo, MaxEncodedLen, Hash)]
+pub enum NativeToken {
+	LIT,
+	DOT,
+}
+
+impl From<NativeToken> for SupportedNetwork {
+	fn from(value: NativeToken) -> Self {
+		match value {
+			NativeToken::LIT => SupportedNetwork::Litentry,
```
Kailai-Wang, line 38: The LIT network actually includes ETH, Litmus and Litentry

primitives/core/src/assertion.rs
```diff
@@ -53,7 +69,7 @@ pub enum Assertion {
 	A11(ParameterString),                                   // (minimum_amount)
 	A12(ParameterString),									// (target_twitter_screen_name)
 	A13(u32),                                               // (Karma_amount) - TODO: unsupported
-
+	A14(ParameterString, NativeToken, ParameterYears)	    // (min_balance, native_token, [years -> eg [2021, 2022]])
```
Kailai-Wang, line 80: I'm actually inclined to make it simple, because they are specifically made for the decoded event. Maybe not (yet) a good time to generalise it.

For the decoded event we know the starting date is 2023-01-01 (from feature request), so we can save one param.

The "native_token" can be parameterised, but then you need to include the token symbol in the final VC. Alternatively it can be omitted, and we simply have two VCs representing LIT and DOT respectively. 

For the building part we could simply copy a4 and a7, with the starting date fixated at 2023-01-01 and that's it :P
What do you think?

Kailai-Wang: A even more lazy way is not to have new assertions at all - we simply use the old a4 and a7. The VC verifier will decide if it meets the requirement 😂

primitives/core/src/assertion.rs
kziemianek: For Decoded - it's explicitly specified that for LIT we prove holding on Litentry, DOT on Polkadot.
Are you sure we should check more networks? that's why I named them `NativeToken`

primitives/core/src/assertion.rs
kziemianek: If you feel that current a4 and a7 are sufficient then let's use them.
The situation I want to avoid is adding and then removing assertions, or keeping assertions that are not going to be used anymore.

Kailai-Wang: Let me double-check it with Kayla
I'm personally fine with adding and removing assertions, as long as they are trivial and don't cost too much effort

Kailai-Wang: I have talked with @kaylawangnow, let's reuse A4/A7 for simplicity.

It would be however great if we can still test it - like find an account that started to hold before 2023-01-01 and check the generated VC.

kziemianek: I think the token could be included this way:

```
"assertions": [
      {
        "and": [
          {
            "src": "$minimum_amount",
            "op": "==",
            "dst": "1"
          },
          {
            "src": "$token",
            "op": "==",
            "dst": "LIT"
          },
          {
            "src": "$from_date",
            "op": "<",
            "dst": "2022-01-01"
          },
          {
            "src": "$to_date",
            "op": ">=",
            "dst": "2022-12-31"
          }
        ]
      }
    ],
   
```

Kailai-Wang: You mean the merge of the HOLDER VCs. Yes it should be possible, most likely we'll do it after the decoded event.

cc @BillyWooo @zhouhuitian

Kailai-Wang: I created the issue: https://github.com/litentry/litentry-parachain/issues/1813

kziemianek: > You mean the merge of the HOLDER VCs. Yes it should be possible, most likely we'll do it after the decoded event.
> 
> cc @BillyWooo @zhouhuitian

Not really, I just posted possible solution to: _The "native_token" can be parameterised, but then you need to include the token symbol in the final VC_

primitives/core/src/assertion.rs
kziemianek: If we want to query more networks then this can be changed.

tee-worker/litentry/core/assertion-build/src/a14.rs
```diff
@@ -0,0 +1,210 @@
...
+			ErrorDetail::ParseError,
+		)
+	})?;
+
+	let mut client = GraphQLClient::new();
+	let supported_network: SupportedNetwork = token.clone().into();
+	let addresses: Vec<String> = collect_addresses_from_identities(identities, supported_network);
+
+	if addresses.is_empty() {
+		return Err(Error::RequestVCFailed(
```
kziemianek, line 79: Other token holder assertions generates empty VC (without assertions) when there are not related identities linked. 

What's the purpose of such empty VC? I prefer to fail, what you think?

tee-worker/litentry/core/assertion-build/src/a14.rs
```diff
@@ -0,0 +1,210 @@
...
+	for from_date in assertion_dates.iter() {
+		let vch = VerifiedCredentialsIsHodlerIn::new(
+			addresses.clone(),
+			from_date.to_string(),
+			SupportedNetwork::Litentry,
+			"".to_string(),
+			q_min_balance.to_string(),
+		);
+		let result = client.verified_credentials_is_hodler(vch).map_err(|e| {
+			Error::RequestVCFailed(
```
kziemianek, line 105: This is another distinction.
If we cannot do all queries then we fail, instead of issuing incomplete VC, what you think ?

tee-worker/litentry/core/assertion-build/src/a14.rs
```diff
@@ -0,0 +1,210 @@
...
+
+		assert_eq!(
+			vec!["2020-01-01", "2021-01-01", "2022-01-01"],
+			prepare_assertion_dates(&years).unwrap()
+		)
+	}
+
+	#[test]
+	pub fn build_works() {
+		unimplemented!()
```
kziemianek, line 208: Let's have unit tests covering this assertion, eventually let's write unit tests for others as well.

tee-worker/litentry/core/assertion-build/src/a14.rs
```diff
@@ -0,0 +1,210 @@
...
+
+	let assertion_dates = prepare_assertion_dates(&years).map_err(|_| {
+		Error::RequestVCFailed(
+			Assertion::A14(min_balance.clone(), token.clone(), years.clone()),
+			ErrorDetail::ParseError,
+		)
+	})?;
+	let mut results: Vec<(String, bool)> = Vec::with_capacity(assertion_dates.len());
+	for from_date in assertion_dates.iter() {
+		let vch = VerifiedCredentialsIsHodlerIn::new(
```
kziemianek, line 97: I assume this request will answer the question wheter the given address was holding a token since `from_date` to end of that year. Can anybody confirm ?

primitives/core/src/assertion.rs
Kailai-Wang: Ah yes.
I mean additionally they should be merged, given that they follow similar logic using similar code

tee-worker/litentry/core/assertion-build/src/a14.rs
Kailai-Wang: No it's since `from_date` to `now` IIRC

Kailai-Wang:
For the decoded event A4/A7 are sufficient - maybe it makes sense to put off working on `A14` but focus on the correctness of A4/A7 🤔 🙌

Kailai-Wang:
Closing it as we decided to reuse A4/A7

PR #1809, opened 2023-06-22
fix: peer update mechanism 

### LINKED ISSUE #1546: Improve the peer updating mechanism
### Context

Before broadcasting the sidechain blocks, the worker searches and updates the peers. I found the current implementation suboptimal:
- it should use a set of URLs instead of a vector
- it should only search for peers which have the same MRENCLAVE as the current worker => or maybe only MRENCLAVEs that are on the whitelist

---

felixfaisal:
Resolves #1546 

- [x] Use `HashSet<Url>` instead of `Vector<Url>` 
- [x] Search only for whitelist mrenclave peers -> Already done as the enclave registry has enclaves with whitelisted mrenclave peers

### DIFF
--- a/tee-worker/service/src/main.rs
+++ b/tee-worker/service/src/main.rs
@@ -97,6 +97,7 @@ use sp_core::{
 };
 use sp_keyring::AccountKeyring;
 use std::{
+	collections::HashSet,
 	env,
 	fs::File,
 	io::Read,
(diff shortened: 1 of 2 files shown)


kziemianek (approved):

Kailai-Wang (approved):
Thanks!

Issue #1782, opened 2023-06-18
Benchmarking litentry-parachain throws an error


Kailai-Wang:
### Context

See https://github.com/litentry/litentry-parachain/actions/runs/5298733242/jobs/9591887467

```
2023-06-17 16:34:56 Join candidates failed in genesis with error DispatchErrorWithPostInfo { post_info: PostDispatchInfo { actual_weight: None, pays_fee: Pays::Yes }, error: Module(ModuleError { index: 45, error: [7, 0, 0, 0], message: Some("CandidateBondBelowMin") }) }   
```

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
Fixed by #1812

PR #1812, opened 2023-06-22
debug: fix for benchmark prod env

wangminqi:
This PR is for #1782.
The success benchmark weight test after fixing is here.
https://github.com/litentry/litentry-parachain/actions/runs/5342286471

The original issue is caused by genesis balance of staking collator account is less than 5000 unit LIT (the minimal staking amount requirement in Litentry prod environment)


node/res/genesis_info/litentry.json
Kailai-Wang, line 1: Is it possible to solve it without changing genesis? 😂 
Although the chain was launched and genesis can't be changed anymore, we still prefer a way without changing it

wangminqi: I see. Then in that way, we can make a special benchmark compile for runtime parameters regarding minimal staking when running benchmark.

However, I feel changing genesis is fine. Since we already changed it when introducing parachain staking genesis. (which the pallet does not exist in the first place)

wangminqi:
Benchmark is actually working. 
https://github.com/litentry/litentry-parachain/actions/runs/5386991687
The weight generating workflow here.

Kailai-Wang (approved):
Sounds good, thank you!
Can you please update the litentry weights file in this PR too?

wangminqi:
> Sounds good, thank you! Can you please update the litentry weights file in this PR too?

Sure. The weight file update PR is now together.

PR #1814, opened 2023-06-22
Request participation VC from delegatee account

### LINKED ISSUE #1797: Request participation vc on the user's behalf
### Context

Preparation for decoded.
We'd like to request VC proving the user's participation in decoded event on the users' behalf.

---

Kailai-Wang:
### Context

resolves #1797 

Instead of having a generic delegated vc-request mechanism, this PR tries to keep it simple without changing the extrinsic interface.

`A13` represents the event's participation and can only be requested by a pre-added delegatee account.
Here's an example of generated `A13`:

```
{"@context":["https://www.w3.org/2018/credentials/v1","https://w3id.org/security/suites/ed25519-2020/v1"],"id":"0xebf5981c8de515b594a1cfe0127b26490a324316c3b36291392e19506ef7d2e8","type":["VerifiableCredential"],"credentialSubject":{"id":"90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22","description":"The user has participated in polkadot decoded 2023","type":"Basic participation proof","tag":["Polkadot decoded 2023"],"assertions":[{"and":[{"src":"$event_name","op":"==","dst":"Polkadot Decoded 2023"},{"src":"$has_participated","op":"==","dst":"true"}]}],"values":[true],"endpoint":"http://localhost:9933"},"issuer":{"id":"af25d49917abc778105261a8ce44bbc902f30280e1a0f4b439b113b3162a9c8d","name":"Litentry TEE Worker","mrenclave":"7d6qKEt4UuEJvs9w2wctct1RKDrMSaYseyyfDwkCyJyx"},"issuanceTimestamp":1687438578340,"proof":{"createdTimestamp":1687438578342,"type":"Ed25519Signature2020","proofPurpose":"assertionMethod","proofValue":"b045a52bb23d7ca78c1bbd9d226cf3a541e68c6b04d059727e9f68df39be84c00f13ce1127d5b4db0b60976616f1c9fc22d9e09d013f8fbc27f00ced341df001","verificationMethod":"af25d49917abc778105261a8ce44bbc902f30280e1a0f4b439b113b3162a9c8d"}}
```


pallets/vc-management/src/lib.rs
```diff
@@ -218,19 +239,45 @@ pub mod pallet {
...
+		pub fn add_delegatee(origin: OriginFor<T>, account: T::AccountId) -> DispatchResult {
+			let _ = T::DelegateeAdminOrigin::ensure_origin(origin)?;
+			// we don't care if `account` already exists
+			Delegatee::<T>::insert(account.clone(), ());
+			Self::deposit_event(Event::DelegateeAdded { account });
+			Ok(())
+		}
+
+		/// remove an account from the delegatees
+		#[pallet::call_index(1)]
```
kziemianek, line 254: Why you push these calls at the beginning ?

pallets/vc-management/src/tests.rs
```diff
@@ -44,6 +45,42 @@ fn request_vc_works() {
 	});
 }
 
+#[test]
+fn request_vc_with_authorised_delegatee_works() {
```
kziemianek: I would name it `request_vc_13_with_authorised_delegatee_works` to indicate that we should have such assertion for every vc that ensures delegatee is a caller.

pallets/vc-management/src/lib.rs
```diff
@@ -218,19 +239,45 @@ pub mod pallet {
...
+		#[pallet::weight(<T as Config>::WeightInfo::remove_delegatee())]
+		pub fn remove_delegatee(origin: OriginFor<T>, account: T::AccountId) -> DispatchResult {
+			let _ = T::DelegateeAdminOrigin::ensure_origin(origin)?;
+			ensure!(Delegatee::<T>::contains_key(&account), Error::<T>::DelegateeNotExist);
+			Delegatee::<T>::remove(account.clone());
+			Self::deposit_event(Event::DelegateeRemoved { account });
+			Ok(())
+		}
+
+		#[pallet::call_index(2)]
```
arieljbirnbaum, line 264: Q: does this change in `call_index` affect the encoding? More to the point, is this a breaking change?

kziemianek (approved):

pallets/vc-management/src/lib.rs
kziemianek: call indices change is a breaking change

pallets/vc-management/src/lib.rs
```diff
@@ -82,6 +84,15 @@ pub mod pallet {
 	#[pallet::getter(fn admin)]
 	pub type Admin<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
 
+	// delegatees who can request (and receive) VCs on users' behalf,
+	// some VCs can only be requested by delegatee accounts (e.g. A13)
+	// delegatees and admins are different:
+	// - admins are meant to manage the pallet state manually, e.g. schema, vcRegistry
+	// - delegatees can request VCs for users, similar to `proxied account`
```
arieljbirnbaum, line 91: Is the delegation restricted to a specific assertion (e.g. Alice can allow Bob to request an `A13` for Alice but that doesn't allow him to request an `A4`)?

If not then it seems to open the door to some big privacy issues (as in, instead of asking Alice for a specific `A6`, Evil Inc. can nudge Alice into giving them a blanket delegation so they can milk _all_ her info) :warning: :thinking:

pallets/vc-management/src/lib.rs
```diff
@@ -175,6 +192,10 @@ pub mod pallet {
 
 	#[pallet::error]
 	pub enum Error<T> {
+		/// a delegatee doesn't exist
+		DelegateeNotExist,
```
arieljbirnbaum, line 196: :nail_care: the grammar in the error name feels a bit broken tbh, but it's already broken in the other error names so maybe it's better to keep it consistent? :sweat_smile:

pallets/vc-management/src/lib.rs
```diff
@@ -175,6 +192,10 @@ pub mod pallet {
...
+		/// a `request_vc` request from unauthorised user
+		UnauthorisedUser,
```
arieljbirnbaum: Q: are we using British or American spelling? :3

pallets/vc-management/src/lib.rs
```diff
@@ -218,19 +239,45 @@ pub mod pallet {
...
+			Self::deposit_event(Event::DelegateeAdded { account });
+			Ok(())
+		}
+
+		/// remove an account from the delegatees
+		#[pallet::call_index(1)]
+		#[pallet::weight(<T as Config>::WeightInfo::remove_delegatee())]
+		pub fn remove_delegatee(origin: OriginFor<T>, account: T::AccountId) -> DispatchResult {
+			let _ = T::DelegateeAdminOrigin::ensure_origin(origin)?;
+			ensure!(Delegatee::<T>::contains_key(&account), Error::<T>::DelegateeNotExist);
```
arieljbirnbaum, line 258: I'm a bit confused, `add_delegatee` is idempotent but `remove_delegatee` is not; either is fine I suppose but I'm not sure I see a reason for the difference :thinking:

kziemianek:
@Kailai-Wang please generate meta and ts-types

pallets/vc-management/src/lib.rs
Kailai-Wang: It's a breaking change in the sense that you can't use the cached index/metadata to construct the offline extrinsic. Let me bump the runtime version and update the parachain metadata to be more explicit.

pallets/vc-management/src/tests.rs
Kailai-Wang: Good point, thanks

pallets/vc-management/src/lib.rs
Kailai-Wang: I wanted them to be consistent with IMP :(

pallets/vc-management/src/lib.rs
Kailai-Wang: > Is the delegation restricted to a specific assertion (e.g. Alice can allow Bob to request an A13 for Alice but that doesn't allow him to request an A4)?

That's true - delegation only works for specific assertion. As written in the PR description, it's not a generic delegation system. We want it basically for the decoded event where we'd like to request some VC (in this case `A13`) for users on site

pallets/vc-management/src/lib.rs
Kailai-Wang: Yes it's intentionally written in this way to be consistent with other pallets 😂 

but maybe it's better to look through similar errors in substrate and adjust them all in our code

Kailai-Wang: I'll leave it to another small issue

pallets/vc-management/src/lib.rs
Kailai-Wang: Thanks, let's stick to American spelling - I'll check it (and other places)

pallets/vc-management/src/lib.rs
Kailai-Wang: Maybe I don't get your point here but I was thinking adding delegatee can be "enforced" (so we don't care if the account is delegatee or not) but you can't remove it from delegatee if it's not in the list.

Maybe we could have saved one DB write if the account to be added is already a delegatee when `add_delegatee`, but we'll need another error. So personally I think either way is fine

Kailai-Wang:
> @Kailai-Wang please generate meta and ts-types

I'll regenerate those, but I'm not touching the `Assertion` type before we sort out the uniqueness of the definition (that's why I asked in slack)

kaylawangnow:
Thanks for the awesome work, I have some modifications about the metadata, pls update:

description":"The user has a Polkadot Decoded 2023 Litentry Booth Special Badge",
"type":"Decoded 2023 Basic Special Badge"

{"src":"$has_claimed_badge","op":"==","dst":"true"}

pallets/vc-management/src/lib.rs
arieljbirnbaum: What I meant is, if you try adding the same delegatee twice then the worker doesn't complain, but if you try removing the same delegatee twice then it throws an error. Either way of handling it can be OK; I was just wondering why they're different from each other here :)

pallets/vc-management/src/lib.rs
arieljbirnbaum: OK so _only_ `A13` can be requested by a delegatee? e.g. if Daisy is a delegatee she can request an `A13` for Alice but not an `A4` or `A6`? 

(also, is Daisy a delegatee _for Alice_ or _globally_? :thinking:)

Kailai-Wang: That's true.

Daisy is a delegatee globally - she can request A13 for any user. This is the feature request we received.

PR #1825, opened 2023-06-27
add achainable_rest_key

### LINKED ISSUE #1824: Add a new config entry for achainable REST API
### Context

The old graphql API key is **not** compatible with the REST API key, they can't be used interchangeably.

However we are mixing them before #1654 is complete, yet we have only one config entry in the setting. It means A4/A7 and A14 wouldn't work simultaneously.

To work around this we'd like to introduce a new entry like `achainable_rest_key` to the global data provider settings. After that we shall do an enclave upgrade.

---

outofxxx:
fix #1824


Kailai-Wang (approved):
Thank you for the prompted PR!

PR #1829, opened 2023-06-27
add account parameter to set_scheduled_mrenclave

### LINKED ISSUE #1827: Add `account` parameter to `set_scheduled_mrenclave` trusted cli command
Right now this command is always executed in the context of Alice.
To make this command more flexible we need to add  a third parameter `account` in ss58check format.

kziemianek:
User can specify any acount (sudo).

fixes: #1827

### DIFF
--- a/tee-worker/cli/src/trusted_base_cli/commands/litentry/set_scheduled_mrenclave.rs
+++ b/tee-worker/cli/src/trusted_base_cli/commands/litentry/set_scheduled_mrenclave.rs
@@ -32,23 +32,26 @@ use teerex_primitives::MrEnclave;
 
 #[derive(Parser)]
 pub struct SetScheduledMrenclaveCommand {
+	/// AccountId in ss58check format
+	account: String,
+	/// Sidechain block number from which mrenclave is in effect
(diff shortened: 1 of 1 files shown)


tee-worker/cli/src/trusted_base_cli/commands/litentry/set_scheduled_mrenclave.rs
```diff
@@ -32,23 +32,26 @@ use teerex_primitives::MrEnclave;
...
 	bn: SidechainBlockNumber,
-	//hex encoded
+	/// Hex encoded mrenclave
 	mrenclave: String,
 }
 
 impl SetScheduledMrenclaveCommand {
 	pub(crate) fn run(&self, cli: &Cli, trusted_cli: &TrustedCli) {
-		let root = get_pair_from_str(trusted_cli, "//Alice");
+		let who = get_pair_from_str(trusted_cli, &self.account);
```
Kailai-Wang, line 45: This will not work, unfortunately. Because you can't get the pair from public address except for well known keys like `//Alice`

Kailai-Wang (changes requested):

PR #1840, opened 2023-06-30
move set scheduled mrenclave to direct rpc call

### LINKED ISSUE #1830: Replace `set_scheduled_mrenclave` trusted call with worker direct api call
In some circumstances `set_scheduled_mrenclave` trusted call cannot be executed, one of the examples is when `tee-worker` detects `mrenclave` mismatch and is not able to produce blocks anymore, thus cannot execute any trusted calls:

```
[2023-06-26T10:42:30Z WARN  aura] Skipping sidechain block 550 due to mismatch MRENCLAVE, current: Ok("cd104e995d478c063c0aa3cb1d02daa56c3008d0f336eeabcb771f2111f09823"), expect: Ok("456f56313143575432397376474d4259556a77597663764a6a61336779653561")
```

Because this call is altering only node's local state, we can move logic from trusted call to worker's direct api call.

Tasks:

- [ ] Remove `set_scheduled_mrenclave` trusted call
- [ ] Add `mrenclave__update_scheduled` rpc method available only in dev mode

kziemianek:
Replace trusted cli command with worker direct rpc call

close: #1830


tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -199,6 +202,36 @@ where
...
+		let mrenclave_update_scheduled_name: &str = "mrenclave_updateScheduled";
+		io.add_sync_method(mrenclave_update_scheduled_name, move |params: Params| {
+			match params.parse::<(SidechainBlockNumber, String)>() {
+				Ok((bn, mrenclave)) =>
+					return match hex::decode(&mrenclave) {
+						Ok(mrenclave) => {
+							let mut enclave_to_set: MrEnclave = [0u8; 32];
+							enclave_to_set.copy_from_slice(&mrenclave);
+							return match GLOBAL_SCHEDULED_ENCLAVE.update(bn, enclave_to_set) {
+								Ok(()) => Ok(json!("{}")),
```
Kailai-Wang: Would you like to set the return value as 
```
RpcReturnValue {
				do_watch: false,
				value: new_mrenclave,
				status: DirectRequestStatus::Ok,
			}
```
Just to be consistent with other rpc call?

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -199,6 +202,36 @@ where
 		}
 	});
 
+	if cfg!(not(feature = "production")) {
+		// mrenclave_updateScheduled
+		// params: sidechainBlockNumber, hex encoded mrenclave
+		let mrenclave_update_scheduled_name: &str = "mrenclave_updateScheduled";
```
Kailai-Wang: regarding the name, would `state_updateScheduledEnclave` sound better? To be resonant with the parachain extrinsic

Kailai-Wang (approved):
Looks good, thanks. I only have two comments.

Btw is it tested? Ideally with the situation where no blocks can be produced due to MRENCLAVE mistmatch

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
kziemianek: I thought `state` refers to shared state ( the one replicated across the nodes) so I decided to not use it.
I can change it to `state_*`, this call is present only in `dev` mode anyway.

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
kziemianek: I'm not sure this call should return value and why only `new_mrenclave` instead of tuple or even whole map.
But yes, in general I think I need to change the response.

kziemianek:
> Looks good, thanks. I only have two comments.
> 
> Btw is it tested? Ideally with the situation where no blocks can be produced due to MRENCLAVE mistmatch

I  tested it with this command

`echo '{"id":1,"jsonrpc":"2.0","method":"mrenclave_updateScheduled","params":[21, "6173647361736466726577617364667265777161736466726577716173646672"]}' | ./websocat.x86_64-unknown-linux-musl -n1 -k -B 99999999 wss://localhost:2000`

I still need to test the case described by you.

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
Kailai-Wang: `RpcReturnValue` only has a `value` field for hex string, so we need encoded tuple/map if we want, I was just lazy and use mrenclave as the value, similar to substrate event :(

kziemianek:
@Kailai-Wang I tested it locally and it works as expected.
Once we hit the point where sidechain cannot produce block because of mrenclave mismatch, the direct call can be executed and update mrenclave which will eventually resume block production.

Kailai-Wang:
Thanks! Feel free to merge it once CI passes

PR #1835, opened 2023-06-27
Separate parachain-api and sidechain-api into workspaces

arieljbirnbaum:
### Context

Attempt to resolve the augmentation issues from having both the parachain and sidechain API in the same project; following the lead from [this comment](https://github.com/polkadot-js/api/issues/5664#issuecomment-1573162122) from the polkadot.js dev.

The gist of it is to separate each API into its own package, so that essentially each has its own "copy" of the polkadot-js modules to augment.

Added the `C1-noteworthy` label since it can be relevant if client applications have the same problem.

### Implementation Notes

- Currently this is on a separate folder from the main `ts-tests` to avoid interfering with CI. Also, for the moment this only has the direct invocation example script, to iron out integration issues with ts-node before adding mocha into the mix.
- Currently the build outputs for the API workspaces are committed; perhaps we should build them on the fly instead.

### Status

- :heavy_check_mark: `parachain-api` and `sidechain-api` packages compile cleanly under `tsc`.
- :x: `ts-node` fails to run the example script

#### Reproduction steps

In the `tee-worker/ts-tests-with-workspaces` directory:

```
yarn install
yarn workspace parachain-api build
yarn workspace sidechain-api build
yarn workspace tests di-examples
```

PR #1846, opened 2023-07-05
Fix `resuming_worker` test

### LINKED ISSUE #1805: lit-resume-worker fails occassionally
### Context

See e.g. https://github.com/litentry/litentry-parachain/actions/runs/5332795556/jobs/9662533244

It's a timeout, looks like some disorder or race condition in the initialisation setup.

---

arieljbirnbaum:
Fixes #1805 and adds some changes to make the process handling slightly less naive.
Also implements #1124.

The root cause for the sporadic "timeout" bug was that the way the test parsed the worker output to extract the shard number; the output stream was read in chunks (as opposed to line-by-line), so if the line containing the shard number didn't line up with the chunk boundary, the shard number would remain blank. Furthermore, the script didn't check for this, so it just kept hitting the RPC endpoint for `latestBlock` with a blank shard and getting a null response in return.

This PR fixes the stream handling logic by wrapping each stream in a `readline` interface.
It also refactors the logic for handling subprocesses, to facilitate debugging and further changes. This allowed adding the missing testcase detailed in #1124.


tee-worker/ts-tests/resuming_worker.test.ts
```diff
@@ -1,151 +1,290 @@
...
+            discord_litentry_url: 'http://localhost:19527',
+            discord_auth_token: '',
+            achainable_url: 'http://localhost:19527',
+            achainable_auth_key: '',
         },
-    };
+        null,
+        4
+    );
+    fs.writeFileSync(`${workingDir}/worker-config-mock.json`, data);
```
felixfaisal: I don't see this file mentioned in the worker command-line arguments in the function `generateWorkerCommandArguments` , Are you sure we need this file? I might be missing something here.

arieljbirnbaum: I don't know; it was there before so I didn't want to mess with it :D

My guess would be that the worker picks it up when `--enable-mock-server` is passed, but I can't find an explicit mention about it in the code.

arieljbirnbaum: (although I did forget to switch to `path.join` for this one, so thanks for pointing it out :grin:)

felixfaisal: I don't think it picks it up, Because I use `--enable-mock-server` and it doesn't give me an error

felixfaisal: Well, It's a trivial operation, I just wanted to check if I have missed this in my restart script 🤣

arieljbirnbaum: Well, let's see what happens if I yeet it from the test then :D

arieljbirnbaum: It worked :raised_hands: 
Good catch @felixfaisal; thanks!

Kailai-Wang:
Thank you @grumpygreenguy!

Would you mind adding a bit of PR description to explain what the root cause was and how the PR would solve it, please? It would help with the review

arieljbirnbaum:
@Kailai-Wang 

> Would you mind adding a bit of PR description to explain what the root cause was and how the PR would solve it, please? It would help with the review

Thanks for pointing that out! I updated the PR description; please let me know if that clears it up :)

Kailai-Wang (approved):
I didn't check every line, but the overall logic looks good.

If it does the job I'm happy to merge it :)

arieljbirnbaum:
(To prevent a conflict with #1835, this also removes the usage of `__dirname` in favor of `os.tmpdir()`)

0xverin:
>(To prevent a conflict with https://github.com/litentry/litentry-parachain/pull/1835, this also removes the usage of __dirname in favor of os.tmpdir())

I solved the problem locally, well, I'm using your. Thanks!

PR #1848, opened 2023-07-05
disable state_getStorage in prod

### LINKED ISSUE #1781: Disable sidechain `state_getStorage` in prod
### Context

Follow-up of DI: after we expose the DI port we should disallow the query of the storage for prod sidechain - for obvious reasons :)

Currently it's handled by the `state_getStorage` RPC call.

Please note it should be a compilation-time path deviation.

---

kziemianek:
`state_getStorage` direct rpc call will be not available if `production` feature is enabled.

closes: #1781


tee-worker/cli/demo_shielding_unshielding.sh
```diff
@@ -118,6 +138,7 @@ function assert_account_state()
 }
 
 echo "* Query on-chain enclave registry:"
+echo ${CLIENT}
```
kziemianek: I'll remove this line.

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -304,6 +258,56 @@ where
...
+						},
+						Err(e) => {
+							let error_msg = format!("load shard failure due to: {:?}", e);
+							return Ok(json!(compute_hex_encoded_return_error(error_msg.as_str())))
+						},
+					}
+				},
+				Err(_err) => Ok(json!(compute_hex_encoded_return_error("parse error"))),
+			}
+		});
```
BillyWooo, line 309: Is it only switch the code position?

kziemianek: Yes, it's moved inside condition `if cfg!(not(feature = "production"))` so this rpc method won't be present with `production` feature on.

Kailai-Wang (approved):
Looks good thanks!

PR #1850, opened 2023-07-05
proper request_vc trusted call definition

kziemianek:
Parentchain block number was removed in #1678


BillyWooo (approved):
LGTM

Kailai-Wang:
Do we need to regenerate the metadata/types for this change?

Kailai-Wang:
Nvm I'll regenerate it in my PR #1854

Issue #1851, opened 2023-07-06
Trusted CLI nonce command fails


kziemianek:
Apparently, after the latest changes to the nonce retrieval mechanism the nonce trusted command stopped working.

```
kziemianek@kziemianek:~/projects/litentry/litentry-parachain/tee-worker/target/release$ ./integritee-cli trusted --mrenclave 56kLvcE2kkxmNQ3Jii1SiPseXzmSUBWrZheomiUm1epG nonce //Alice
thread 'main' panicked at 'get nonce error!: MspcReceiver(RecvError)', cli/src/trusted_base_cli/commands/nonce.rs:39:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The worker dies with following message:
```
thread '<unnamed>' panicked at 'storage cannot be called outside of an Externalities-provided environment.', /home/kziemianek/projects/litentry/litentry-parachain/tee-worker/core-primitives/substrate-sgx/sp-io/src/lib.rs:131:9
```

songzhouran:
<img width="563" alt="image" src="https://github.com/litentry/litentry-parachain/assets/3433393/bf69dcc7-618a-4b15-9ecd-b1ef8724af45">

songzhouran:
It works on my local and there is no panic.

Kailai-Wang:
Hmm I got panic too :( 

<img width="1303" alt="image" src="https://github.com/litentry/litentry-parachain/assets/7630809/46e8a6aa-b11b-40df-8bf8-42dd11144331">

PR #1852, opened 2023-07-06
add trusted cli nonce command test

kziemianek:
Adds check to see if trusted cli command `nonce` works.

relates: #1851

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -523,6 +523,7 @@ jobs:
           - test_name: lit-batch-test
           - test_name: lit-bulk-identity-test
           - test_name: lit-bulk-vc-test
+          - test_name: lit-nonce
 
     steps:
       - uses: actions/checkout@v3
--- a/tee-worker/cli/lit_nonce.sh
+++ b/tee-worker/cli/lit_nonce.sh
@@ -43,7 +43,6 @@ echo "Using trusted-worker uri $WORKER1URL:$WORKER1PORT"
 echo ""
 
 ACC=//Bob
-KEY="22fc82db5b606998ad45099b7978b5b4f9dd4ea6017e57370ac56141caaabd12"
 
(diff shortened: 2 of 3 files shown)


Kailai-Wang:
I feel like having a dedicated nonce script is a bit overkill, the `nonce == 0` test case doesn't bring a lot either as that's the default value.

Can we reuse some existing test cases to do it?
E.g. `demo_shielding_unshielding.sh` where we have nonce retrieval too, but through `getStorage`. Maybe we can make adjustments there?

Kailai-Wang:
Actually it is not necessarily related to #1851, that issue seems more like a bug in implementation :(

kziemianek:
> I feel like having a dedicated nonce script is a bit overkill, the `nonce == 0` test case doesn't bring a lot either as that's the default value.

The idea behind this test is to verify whether the command is working as expected with real node, not to check nonce default value 🙃 . We can change assertion but this one seemed most straightforward to me.

kziemianek:
This should be implicitly verified by https://github.com/litentry/litentry-parachain/pull/1848/files#diff-a79aa284ab66e6a189272e6ecd6c428f9f179792e6835691992128dfdcb19e69R104 too, so if you feel like this script is an overkill, we can close this. Another check like this will also extend CI check by next ~10 min which we probably want to avoid.

Kailai-Wang:
> This should be implicitly verified by https://github.com/litentry/litentry-parachain/pull/1848/files#diff-a79aa284ab66e6a189272e6ecd6c428f9f179792e6835691992128dfdcb19e69R104 too, so if you feel like this script is an overkill, we can close this. Another check like this will also extend CI check by next ~10 min which we probably want to avoid.

Yes I'll prefer that (including it when testing other functionalities)
Partially because the setup effort to even test one such simple RPC is not trivial 😂 

Btw the `polling` in that script might be still needed, in our experience, the live query can easily fail as it's too early for the state to change.

Issue #1813, opened 2023-06-22
Merging of HOLDER VCs


Kailai-Wang:
### Context

We seem to have some duplicate code in the HOLDER VCs: `a4/a7/a10/a11`, they should have been merged with parameterised token symbol. 

After merging, such token symbol should be included in the VC too.
Please note this will be a breaking change for IDHub.

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

PR #1854, opened 2023-07-06
Move network information from `Identity` to `IdentityContext`

### LINKED ISSUE #1841: Move identity network type to IdentityContext
### Context

Many blockchains "share" the same key pair construction mechanism, which means owning one private key would grant you access to the same public key or address on different networks.

Examples:
- EVM based: ETH, Polygon, BSC -> they have the same public key and address
- Substrate based: relaychains + parachains -> they have the same public key, but different addresses due to different ss58 prefixes
 
We [had discussions](https://litentry.slack.com/archives/C03KR96989G/p1680718905532839?thread_ts=1680708645.416489&cid=C03KR96989G) about whether we should treat the same public key on different networks as distinct identities, the decision was _yes_ at that time. Reasons include:
- it gives the user finer control over their identities, e.g. they can link their polkadot address to the IDGraph and intentionally not link their kusama address
- without explicit network types, calculating specific VC would be unrealistic, e.g. the number of governance activities over all linked substrate identities

During the implementation of EVM-signin feature, we realised that we might be able to simplify this process - by moving the network type information away from `Identity` itself to `IdentityContext`.

**Before:**

```
pub enum Identity {
	Substrate { network: SubstrateNetwork, address: Address32 },
	Evm { network: EvmNetwork, address: Address20 },
	Web2 { network: Web2Network, address: IdentityString },
}

pub struct IdentityContext<T: Config> {
	// the sidechain block number at which the identity is linked
	pub link_block: BlockNumberOf<T>,
	// the identity status
	pub status: IdentityStatus,
}
```

**After:**
```
pub enum Identity {
        // web2
        Twitter(IdentityString),
        Discord(IdentityString),
        Github(IdentityString),
        // web3
	Substrate(Address32),
	Evm(Address20),
	
}

pub struct IdentityContext<T: Config> {
	// the sidechain block number at which the identity is linked
	pub link_block: BlockNumberOf<T>,
        // a list of networks on which this identity is used
        pub networks: Vec<Network>, 
	// the identity status
	pub status: IdentityStatus,
}
```

**Rationale:**
- **Foremost**: the TrustedCall/DI is authorised using signature, and **signature is ignorant of network types**. It's impossible to specify the identity network that can access/manage the IDGraph. So you can't say: _we only allow litentry identities to manage the IDGraph, not polkadot identities_. It doesn't make sense because the signature doesn't contain the network information, they are just calculated from key pairs
- it makes the linking process easier for the user. They don't have to repeatedly link identities with different networks despite sharing the same public key (thus the same validation data). They only link it once.
- it reduces the IDGraph size.
- it _might_ lower the chance of cheating, given that one identity can only exist in one IDGraph and you can't unlink it.

### Proposal

Meanwhile, we want to maintain flexibility, and the user should still have complete control over things like identities on which network should take part in the assertion building.

While we don't distinguish network types when adding identities to IDGraph, sometime, somewhere, we need to ask the user:  for which networks do you want your identities in IDGraph to be used? It can be 3 such moments:
1. when the user links an identity
2. when the user builds a VC 
3. somewhere in between

We propose a hybrid solution of 1 and 3.

Example:
- when the user links a substrate identity (EVM-based is similar), we give the user the initial possibility to specify which networks the identity should be used for. In UI it's something like a multi-select box, where we have some default networks that are ticked already (e.g. Litentry, polkadot, kusama, phala...) but the user is free to check or uncheck the network. Just so you know, it's a finite set of networks, mostly dependent on the data provider, so it won't be 100 parachain networks.
- we provide an extrinsic or DI called `set_identity_network` so that user can update the network list at any time
- the change will take effect in the next VC generation. So for example, when the user only sets `kusama` as active network, only activities on `kusama` will be considered when the user requests the VC "_number of governance activities over all linked substrate identities_", no polkadot, no litentry or any other substrate networks.

---

Kailai-Wang:
### Context

Resolves #1841 
It's a **breaking change** for the client.

This PR tries to:
- relocate the network information from the `Identity` struct to `IdentityContext` struct, it uses `Vec<Web3Network>` to highlight that it's only related to web3 identities
- restructure the `Identity` and `Web3Network` to be more flattened, see see `network.rs` - **please let me know the feedback** (tagging @jonalvarezz on this too)
- replace `SupportedNetwork` and `IndexingNetwork` with `Web3Network` in assertion building
- specify the supported networks for each `Assertion`: `get_supported_web3networks`
- simplify the assertion building logic
- unify the logic of `A4/A7/A10/A11`, but probably needs further refactoring, see #1813 
- remove the deprecated `mock-tee-primitives` package
- fix a validation data signing bug in ts-test

TODO:
- add an interface to allow users to update the web3networks, will follow this up once this PR is merged (because I want to make sure the fundamentals are aligned before moving on)
- maybe more testing regarding the networks in ts-test


pallets/identity-management/src/lib.rs
```diff
@@ -202,7 +212,7 @@ pub mod pallet {
 			user: T::AccountId,
 			encrypted_identity: Vec<u8>,
 			encrypted_validation_data: Vec<u8>,
-			// I don't think it has to be encrypted - no problem to adjust it if required
+			encrypted_web3networks: Vec<u8>,
```
BillyWooo, line 223: why it's not used?

tee-worker/cli/src/base_cli/commands/litentry/link_identity.rs
```diff
@@ -66,6 +66,7 @@ impl LinkIdentityCommand {
 		let tee_shielding_key = get_shielding_key(cli).unwrap();
 		let encrypted_identity = tee_shielding_key.encrypt(&identity.unwrap().encode()).unwrap();
 
+		// TODO: the params are incorrect - and need to be reworked too
```
BillyWooo, line 69: which params are incorrect?

pallets/identity-management/src/lib.rs
Kailai-Wang: It's used in indirect calls

tee-worker/cli/src/base_cli/commands/litentry/link_identity.rs
Kailai-Wang: The extrinsic parameters - they are a bit outdated

primitives/core/src/network.rs
```diff
@@ -0,0 +1,103 @@
...
+	Eq,
+	PartialOrd,
+	Ord,
+	Hash,
+	TypeInfo,
+	MaxEncodedLen,
+	EnumIter,
+)]
+pub enum Web3Network {
+	// substrate
```
kziemianek, line 50: I'm not sure if this is good structure. For example what if we want to add new substrate network, would it be added at the end of 'substrate' or after 'evm' values ? I think one way might be breaking change as this is enumerated type.

Have you considered something like this?
```
pub enum Web3Network {
	Substrate(SubstrateNetwork),
	Evm(EvmNetwork)
}

pub enum SubstrateNetwork {
	Polkadot,
	Kusama,
	Litentry,
	Litmus,
	LitentryRococo,
	Khala,
	SubstrateTestnet, // when launched it with standalone (integritee-)node
}

pub enum EvmNetwork {
	Ethereum,
	Polygon,
	BSC,
}
```

This would also reduce the need for `is_substrate` and `is_evm` methods.

primitives/core/src/network.rs
```diff
@@ -0,0 +1,103 @@
...
+
+	// evm
+	Ethereum,
+	Polygon,
+	BSC,
+}
+
+impl Web3Network {
+	pub fn is_substrate(&self) -> bool {
+		matches!(
```
kziemianek, line 67: Have you considered using the normal match as it's exhaustive and requires you handling a new variant. Adding following test could also prevent from such mistakes:

```
	#[test]
	fn is_evm_network() {
		Web3Network::iter().for_each(|network| {

			assert_eq!(network.is_evm(), match network {
				Web3Network::Polkadot => false,
				Web3Network::Kusama => false,
				Web3Network::Litentry => false,
				Web3Network::Litmus => false,
				Web3Network::LitentryRococo => false,
				Web3Network::Khala => false,
				Web3Network::SubstrateTestnet => false,
				Web3Network::Ethereum => true,
				Web3Network::Polygon => true,
				Web3Network::BSC => true
			})
		})
	}
```

tee-worker/litentry/primitives/src/identity.rs
```diff
@@ -65,135 +58,59 @@ impl From<[u8; 32]> for Address32 {
...
+/// We only include the network categories (substrate/evm) without concrete types
+/// see https://github.com/litentry/litentry-parachain/issues/1841
 #[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub enum Identity {
-	Substrate { network: SubstrateNetwork, address: Address32 },
-	Evm { network: EvmNetwork, address: Address20 },
-	Web2 { network: Web2Network, address: IdentityString },
+	// web2
+	Twitter(IdentityString),
```
kziemianek, line 71: Potentialy same problem as with `Web3Network`

kziemianek (commented):
We are very often leaving todos in code, do we have any strategy for resolving them 🙃  ?

primitives/core/src/network.rs
BillyWooo: It was something like that before the change. One more structure layer in between. Either way has its stand. I think here Kai wants to decrease the layers.

Kailai-Wang: Hi @kziemianek, yes I intentionally did so after pondering it 🥲 see the comment just above this struct:
```
/// supported web3 networks
/// use a flattened style to avoid overly nested structure like:
/// {
///   Substrate(SubstrateNetwork),
///   Evm(EvmNetwork),
/// }
```

The nested style would complicate the construction of such type (also from F/E side). Compare this: 
```
Web3Network::Litentry
```
vs
```
Web3Network::Substrate(SubstrateNetwork::Litentry)
```
Both "Substrate" are kind of redundant, but you need to have some names for the enum types, which is a bit awkward. You can have extra `From` or `TryFrom` implementation but that's not as straightforward either.

The value retrieval and comparison/branching typically need more effort (`if let`/`match`) in code which lowers the readability. 

If some of the enum types (e.g. `Ethereum`) need further subdivision, that's even more layers. `is_substrate` and `is_evm` would take a one-time maintenance effort so I'm fine with it.

Adding new values is a fair point, but either way it would need a runtime upgrade to take effect => it will have a new spec_version number => it signals the client it's a breaking change => the client needs to refresh/renew the metadata eventually. That said I'll keep it categorised when adding new values.

So in general I consider the flattened structure a bit more advantageous but I'm open to suggestions.

primitives/core/src/network.rs
Kailai-Wang: That's a good point - I'll add test cases to prevent errors due to being silently ignored

Kailai-Wang:
> We are very often leaving todos in code, do we have any strategy for resolving them 🙃 ?

In general we should follow up with issues if it's not so trival.

I normally leave `TODO`s since:
- it's better/clearer for a separate PR
- it's not a trivial effort
- I accidentally find something that shouldn't belong to this PR, to not forget about it
- I know it can be improved, but have no idea right now

Maybe I've left too many `TODO`s - my apology.

tee-worker/litentry/primitives/src/identity.rs
Kailai-Wang: Yes same consideration as above

primitives/core/src/network.rs
kziemianek: _Adding new values is a fair point, but either way it would need a runtime upgrade to take effect => it will have a new spec_version number => it signals the client it's a breaking change => the client needs to refresh/renew the metadata eventually. That said I'll keep it categorised when adding new values._

This might not be true for clients that are using DI only.

This scaries me a bit more, because if you didn't notice the change, you might be sending old values and everything will work fine, except the results will be not the ones you expected.

kziemianek: Keeping this structure but not forcing new variants to be added "in between" should avoid such problems.

Kailai-Wang: That's akin to the substrate extrinsic index - where adding new extrinsic is backwards compatible, but if you change the order that's a breaking change and you have to bump the `spec_version`/`transaction_version`.

I'd argue even if you split into `evm` and `substrate` groups you can still fall into this problem once you remove/change the existing fields. The same applies to other enums too (remember the famous `trusted_call`).

I'm not sure how IDHub maintains the generated types, but maybe we can minimise the risk by having a proper version and metadata update mechanism, like utilise the runtime version in sidechain too - once the client sees a different runtime version, it should realise something is broken and they should re-fetch the metadata and re-generate the types.

tee-worker/litentry/pallets/identity-management/src/identity_context.rs
```diff
@@ -32,12 +32,14 @@ pub enum IdentityStatus {
...
+	pub web3networks: BoundedWeb3Network,
 	// the identity status
 	pub status: IdentityStatus,
 }
 
 impl<T: Config> IdentityContext<T> {
-	pub fn new(link_block: BlockNumberOf<T>) -> Self {
-		Self { link_block, status: IdentityStatus::Active }
+	pub fn new(link_block: BlockNumberOf<T>, web3networks: BoundedWeb3Network) -> Self {
+		Self { link_block, web3networks, status: IdentityStatus::Active }
```
kziemianek: What do you think about removing duplicates from `web3networks` here ?

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -129,17 +131,23 @@ pub mod pallet {
...
 		) -> DispatchResult {
 			T::ManageOrigin::ensure_origin(origin)?;
-			// we don't care about the current key
-			UserShieldingKeys::<T>::insert(&who, key);
 
-			let prime_id = Self::build_prime_identity(&who, parent_ss58_prefix)?;
+			let prime_id = Self::build_prime_identity(&who)?;
 			if IDGraphs::<T>::get(&who, &prime_id).is_none() {
-				let context = <IdentityContext<T>>::new(<frame_system::Pallet<T>>::block_number());
+				// TODO: shall we activate all available networks for the prime id?
```
kziemianek, line 141: The more I think about it the more I'm convinced that we should let user decide when setting shielding key.

Apart from that, does it make sense to set "evm" networks for `Identity::Substrate` ?

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -151,18 +159,19 @@ pub mod pallet {
...
 				!IDGraphs::<T>::contains_key(&who, &identity),
 				Error::<T>::IdentityAlreadyLinked
 			);
-			let prime_id = Self::build_prime_identity(&who, parent_ss58_prefix)?;
+			let prime_id = Self::build_prime_identity(&who)?;
 			ensure!(identity != prime_id, Error::<T>::LinkPrimeIdentityDisallowed);
 
-			let context = <IdentityContext<T>>::new(<frame_system::Pallet<T>>::block_number());
+			let context =
+				<IdentityContext<T>>::new(<frame_system::Pallet<T>>::block_number(), web3networks);
```
kziemianek, line 181: I think we should make sure that linked identity and networks belong to same type of chain ("substrate"/"evm").

Having such mix doesn't make sense and should be prevented from happening.

If applied, it would be good to have a test for it.

tee-worker/litentry/primitives/src/identity.rs
```diff
@@ -65,135 +58,59 @@ impl From<[u8; 32]> for Address32 {
...
-	}
-
 	pub fn is_web2(&self) -> bool {
-		matches!(self, Identity::Web2 { .. })
+		matches!(self, Self::Twitter(..) | Self::Discord(..) | Self::Github(..))
 	}
 
 	pub fn is_web3(&self) -> bool {
-		matches!(self, Identity::Evm { .. } | Identity::Substrate { .. })
+		matches!(self, Self::Substrate(..) | Self::Evm(..))
```
kziemianek, line 86: Might be worth to have same tests as for `Web3Network` 🙃?

tee-worker/litentry/core/stf-task/sender/src/lib.rs
```diff
@@ -87,7 +88,7 @@ pub struct AssertionBuildRequest {
 	pub shard: ShardIdentifier,
 	pub who: AccountId,
 	pub assertion: Assertion,
-	pub vec_identity: Vec<Identity>,
+	pub vec_identity: Vec<(Identity, Vec<Web3Network>)>,
```
kziemianek: What do you think about promoting it to the new type, for example 'IdentityWithNetworks` ?
This tuple is widely used across assertions, which makes code less readable.

tee-worker/litentry/core/assertion-build/src/a1.rs
```diff
@@ -21,37 +21,27 @@ compile_error!("feature \"std\" and feature \"sgx\" cannot be enabled at the sam
...
 const VC_A1_SUBJECT_TAG: [&str; 1] = ["Litentry Network"];
 
-pub fn build(
-	identities: Vec<Identity>,
-	shard: &ShardIdentifier,
-	who: &AccountId,
-) -> Result<Credential> {
-	debug!("Assertion A1 build, who: {:?}", account_id_to_string(&who));
+pub fn build(req: &AssertionBuildRequest) -> Result<Credential> {
+	debug!("Assertion A1 build, who: {:?}", account_id_to_string(&req.who));
```
kziemianek, line 31: We don't want to have such sensitive information to appear in logs or is it jus irrevelant ?

tee-worker/litentry/core/assertion-build/src/a7.rs
```diff
@@ -21,64 +21,48 @@ compile_error!("feature \"std\" and feature \"sgx\" cannot be enabled at the sam
...
+pub fn build(req: &AssertionBuildRequest, min_balance: ParameterString) -> Result<Credential> {
+	debug!("Assertion A7 build, who: {:?}", account_id_to_string(&req.who),);
 
 	let q_min_balance = vec_to_string(min_balance.to_vec()).map_err(|_| {
 		Error::RequestVCFailed(Assertion::A7(min_balance.clone()), ErrorDetail::ParseError)
 	})?;
 
 	let mut client = AchainableClient::new();
-	let mut addresses = vec![];
+	let identities = transpose_identity(&req.vec_identity);
```
kziemianek, line 42: I think we should filter out all "identities" having `Web3Network` other than Polkadot as it doesn't make sense to ask for them in context of this assertion or it may lead to false positive results.

tee-worker/litentry/core/assertion-build/src/a8.rs
```diff
@@ -137,39 +81,14 @@ pub fn build(
 		},
 		Err(e) => {
 			error!("Generate unsigned credential failed {:?}", e);
-			Err(Error::RequestVCFailed(Assertion::A8(index_networks), e.into_error_detail()))
+			// In fact, it should never fail
+			let bounded_web3networks =
+				req.assertion.get_supported_web3networks().try_into().unwrap();
```
kziemianek, line 86: One day it may panic 😂

kziemianek (commented):
I gave it a second more careful look.

tee-worker/litentry/core/assertion-build/src/a7.rs
Kailai-Wang: Maybe this is what you meant? https://github.com/litentry/litentry-parachain/blob/08a7b9201035753dc8d344b69bc09e80ac3c25be/tee-worker/app-libs/stf/src/trusted_call_litentry.rs#L141-L148

tee-worker/litentry/core/assertion-build/src/a8.rs
Kailai-Wang: Noooo - then rust is broken.

The best thing I can do is use default value if `try_into` fails 😂 
If we fall into this branch, it means something is wrong when generating VC and we need to report it back to parachain, so we can't have any other error types than `RequestVCFailed`

tee-worker/litentry/pallets/identity-management/src/identity_context.rs
Kailai-Wang: Do you mean the client could insert duplicate types in this vector?

tee-worker/litentry/primitives/src/identity.rs
Kailai-Wang: Good call

tee-worker/litentry/pallets/identity-management/src/lib.rs
BillyWooo: I would suggest not to do so. If we let the user to decide when to set shielding key, it's out of control. The user may not set any shielding key. 
What do you mean by **set "evm" networks for `Identity::Substrate`** ?

Kailai-Wang: I agree it needs to be discussed. To me it's more like a product/user experience issue, so I look for opinions of @kaylawangnow @jonalvarezz @mikuczi

One opposing view would be: when the user sets the shielding key, it's confusing that they need to select the identity type - at least less natural than linking an identity. If they want to change the network information of the prime account, they can use the `set_identity_web3network` at any time

tee-worker/litentry/pallets/identity-management/src/identity_context.rs
kziemianek: Yes, but whatever the reason is, we want to prevent from having same network enabled multiple times.

tee-worker/litentry/core/assertion-build/src/a7.rs
kziemianek: Yes, but the purpose of this code is to reduce request size, not to prevent assertions from making wrong requests.
I think it could be worth to add redundant check. Better safe than sorry.

tee-worker/litentry/pallets/identity-management/src/lib.rs
kziemianek: > I would suggest not to do so. If we let the user to decide when to set shielding key, it's out of control. The user may not set any shielding key.

@BillyWooo What I mean is that the setting user shielding key use cause should invole picking networks for prime identity by user explicitly.

> What do you mean by set "evm" networks for Identity::Substrate ?

@BillyWooo This code will result in substrate identity being linked to user's IdGraph with all web3 networks activated, even Ethereum, Polygon and BSC. Will it be fine?

```
let web3networks = get_all_web3networks()
    .try_into()
    .map_err(|_| Error::<T>::Web3NetworkLenLimitReached)?;
let context = <IdentityContext<T>>::new(
    <frame_system::Pallet<T>>::block_number(),
    web3networks,
);
Self::insert_identity_with_limit(&who, &prime_id, context)?;
```

Kailai-Wang: Currently people link identities via IDHub, where it comes down to two distinct categories: substrate or evm. So as long as F/E send the right types it's all good.

But yes from code logic we should have a sanity check here. A follow-up question is how to handle it? Do you simply throw out an error, or silently leave out the unmatched network types? It may be another user experience problem @kaylawangnow @jonalvarezz

tee-worker/litentry/core/assertion-build/src/a7.rs
Kailai-Wang: I don't get it - the `assertion_networks` already contains the desired network, doesn't it? E.g. for `A7` it would be `Web3Network::Polkadot`

tee-worker/litentry/core/stf-task/sender/src/lib.rs
Kailai-Wang: Will do :)

tee-worker/litentry/core/assertion-build/src/a7.rs
kziemianek: Yes, the code in current state should work fine.

tee-worker/litentry/core/assertion-build/src/a6.rs
```diff
@@ -40,39 +34,29 @@ const VC_A6_SUBJECT_TAG: [&str; 1] = ["Twitter"];
...
-	shard: &ShardIdentifier,
-	who: &AccountId,
-) -> Result<Credential> {
-	debug!(
-		"Assertion A6 build, who: {:?}, identities: {:?}",
-		account_id_to_string(&who),
-		identities
-	);
+pub fn build(req: &AssertionBuildRequest) -> Result<Credential> {
+	debug!("Assertion A6 build, who: {:?}", account_id_to_string(&req.who),);
```
kziemianek, line 38: We don't want to have such sensitive information to appear in logs or is it jus irrevelant ?

tee-worker/litentry/core/assertion-build/src/a1.rs
kziemianek: This should point to other assertion 🙃 
https://github.com/litentry/litentry-parachain/pull/1854/files/08a7b9201035753dc8d344b69bc09e80ac3c25be#r1255746447

Kailai-Wang: Irrelevant I think - you will see who has requested a VC (via extrinsics) - who has received a VC (via `VCIssued` event) - so it should be fine here

tee-worker/litentry/pallets/identity-management/src/lib.rs
Kailai-Wang: I currently throw out an error if the types don't match, because 99% of users (exception: IDHub team) will link it using IDHub so I don't expect it to go wrong. If it however does, it must be a bug

primitives/core/src/network.rs
jonalvarezz: I also agree that we should look for structures that keep breaking changes to a minimum or are at least backward compatible in the future without translating the problem to clients.

put another way, as a client, if you give to pick between ugly and nested structures that require mininum to non maintenance and nice structures that require me to watch every single thing you change, I'd very much prefer the former.

Now, I leave it to you folks to find such a structure and keep a good balance between both trade-offs 👍

kziemianek (approved):
Looks good to me.
It's great that week ago the idea came up and we already have an implementation. Nice pace 🚀

tee-worker/litentry/pallets/identity-management/src/lib.rs
jonalvarezz: > The more I think about it the more I'm convinced that we should let user decide when setting shielding key

This is a very fair point. My reasoning from a user perspective is that since:
1. The user can update their linked networks at anytime, including disabling previously activated networks, and
2. This happens only for the main (prime) account

It is alright. 

I'd suggest to however leave this to clients, i.e., don't do any automatic network activation here but accept a `web3network` vec with the list of networks to activate as part of setting the shielding key, if any. 

This enable clients to do it under the hood or to display the option to the user. IDHub, as requested by Product, will do it automatically at the beginning but it leaves the door open to change it at anytime.

Additionally, another reason to not perform any automatic activation here is that we are using setUserShieldingKey as a recovery method, e.g., you forgot your password. It wouldn't be nice to get networks activating as a side effect.

Kasper: If the IDHub goes with this automatic network linking when setting shielding key, we will make sure to let the user know what will happen so they know it upfront.

jonalvarezz: > But yes from code logic we should have a sanity check here. A follow-up question is how to handle it? Do you simply throw out an error, or silently leave out the unmatched network types? 

Throwing an error if any of networks doesn't match sounds totally sane. A partial success would be more complex to properly handle. Make it fail out loud 

> I currently throw out an error if the types don't match, because 99% of users (exception: IDHub team) will link it using IDHub so I don't expect it to go wrong. If it however does, it must be a bug

Do we have the data to deny or confirm that IDHub is the only client out there using our Parachain? It is public, anyone could be using it and toying around with it. as far as I know, we are not monitoring it extensively either 😄

Futhermore, I don't think it is healthy for us, as a public parachain, to think only in terms of IDHub. We should always use the word "clients" and tackle issues holistically as service provider. This is, to not minimize the impact of issues because we can translate and control them within the IDHub.

Kailai-Wang: > Do we have the data to deny or confirm that IDHub is the only client out there using our Parachain? 

No - it's completely open.
But if not using IDHub, probably only tech guys can construct the parameters correctly by looking up the code, if they are that far, they should know the handling of mismatching networks in our code.

Kailai-Wang:
Please don't merge it now - we are still having discussions about the struct/enum :(

BillyWooo (approved):
LGTM. Excellent. I think it's good to go.

Kailai-Wang:
We had some discussions about whether we should go back to the old structure and change the network to a list of vectors, like this:
```
pub enum Identity {
	Substrate(Address32, BoundedVec<SubstrateNetwork, ..>),
	Evm(Address20, BoundedVec<EvmNetwork, ...>),
	Web2(IdentityString, Web2Network),
}
```

but it will be problematic if we need to pass in those identities as parameters in e.g. `remove_identity` because we need to correctly construct those network types.

PR #1858, opened 2023-07-07
fix: downgrade srtool image compatible with polkadotv0.9.39

### LINKED ISSUE #1819: Building wasm with srtool fails
### Context

See https://github.com/litentry/litentry-parachain/actions/runs/5373385699/jobs/9747713246

Is it because of two low polkadot version, combined with the fact that upstream has been using stable rustc?

See https://github.com/paritytech/substrate/pull/13580

Or is it because srtool is using a too new rustc version? (`1.70`) Maybe we can solve it by downgrading the srtool image version.

It needs a bit of research.
The goal is we can successfully build wasm with srtools and the built wasm can be used for runtime ugprade.

---

felixfaisal:
Resolves #1819 

The issue as correctly identified by @Kailai-Wang was due to the low version of polkadot being used. We can in the future refer to the release notes of Polkadot which mention against which version of `srtools` and `rustc` version it was tested against to build the `wasm`. 

I've tested the fix on my fork [here](https://github.com/felixfaisal/litentry-parachain/actions/runs/5485044759)

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -39,6 +39,7 @@ jobs:
         with:
           chain: ${{ github.event.inputs.chain }}-parachain
           runtime_dir: runtime/${{ github.event.inputs.chain }}
+          tag: "1.66.0"
 
       - name: Summary
         run: |


Kailai-Wang (commented):
Thanks! There might be multiple places where srtool is used (IIRC)

felixfaisal:
Hey @Kailai-Wang 
Not sure with which `polkadot` version we will be releasing the litentry parachain runtime, if with `9.39` then I can downgrade everywhere it is being used atleast in GitHub Actions.

Kailai-Wang:
> Hey @Kailai-Wang Not sure with which `polkadot` version we will be releasing the litentry parachain runtime, if with `9.39` then I can downgrade everywhere it is being used atleast in GitHub Actions.

Next one would be `0.9.42` I believe

felixfaisal:
So `polkadotv0.9.42` is tested against  `rustc 1.69.0` as mentioned in the [release notes](https://github.com/paritytech/polkadot/releases/tag/v0.9.42). So should I wait until PRs for updating versions are merged and then update the `srtools` image or focus only on `0.9.39` for now and downgrade it. Let me know what you think.

Kailai-Wang:
Yes let's wait a bit till 0.9.42 is done and re-test the compatibility with variant srtool version :)

BillyWooo (approved):
I would suggest to merge it first. After 0.9.42 is done, update the tools again.

felixfaisal:
So based on the above: 
The `srtools` in the actions is being used in `build_wasm` and `create_release_draft`, I will not update `create_release_draft` as it makes no sense if we are planning on the next release to use `0.9.42`. I think `build_wasm` is more important as of now, I'm assuming it will be used to perform runtime upgrades for the current parachain. Therefore I will merge this PR. Thanks!

PR #1862, opened 2023-07-09
enclave-runtime clippy rules

### LINKED ISSUE #1227: No panic in enclave
### Context
As in this comment: https://github.com/integritee-network/worker/pull/1134#discussion_r1083861780

Let's double-check the code and try to avoid panic as possible

---

kziemianek:
Activates `clippy::unwrap_used` lint. 
Unwrap ignored in tests and init code. 

fixes: #1227


Kailai-Wang (commented):
Thanks for the PR!
Do you consider allowing panic for `service`, which is non-enclave part? It would actually help because that's the desired behavior (we'd like it to warn and quit if e.g. something isn't initialised successfully

tee-worker/service/src/main.rs
```diff
@@ -162,17 +163,22 @@ fn main() {
...
 	// build the entire dependency tree
 	let tokio_handle = Arc::new(GlobalTokioHandle {});
+	#[allow(clippy::unwrap_used)]
 	let sidechain_blockstorage = Arc::new(
 		SidechainStorageLock::<SignedSidechainBlock>::new(PathBuf::from(&SIDECHAIN_STORAGE_PATH))
 			.unwrap(),
 	);
 	let node_api_factory =
 		Arc::new(NodeApiFactory::new(config.node_url(), AccountKeyring::Alice.pair()));
+	#[allow(clippy::unwrap_used)]
```
BillyWooo: Do you have any good idea to remove this unwrap? Actually it frequently causes trouble.

tee-worker/service/src/main.rs
```diff
@@ -896,6 +939,7 @@ fn check_we_are_primary_validateer(
 	Ok(enclave_count_of_previous_block == 0)
 }
 
+#[allow(clippy::unwrap_used)]
```
felixfaisal: Can we have better panic messages for these unwraps, This is usually caused when the `config.json` is not according to the format we expect, So a panic message mentioning exactly that could be more helpful.

felixfaisal (changes requested):

tee-worker/service/src/enclave/api.rs
```diff
@@ -52,10 +52,9 @@ pub fn enclave_init(config: &Config) -> EnclaveResult<Enclave> {
 	if use_token {
 		match File::open(&token_file) {
 			Err(_) => {
-				info!(
-					"[-] Token file {} not found! Will create one.",
-					token_file.as_path().to_str().unwrap()
-				);
+				#[allow(clippy::unwrap_used)]
+				let file_str = token_file.as_path().to_str().unwrap();
```
arieljbirnbaum: Maybe sth like `token_file.as_path().to_string_lossy()` would avoid the need to `unwrap` altogether?

tee-worker/service/src/main.rs
arieljbirnbaum: Rust noob here, but if the `enclave_init` (or similar critical initialization) fails then shouldn't the error be bubbled up with `?` :thinking:?

kziemianek:
> Thanks for the PR! Do you consider allowing panic for `service`, which is non-enclave part? It would actually help because that's the desired behavior (we'd like it to warn and quit if e.g. something isn't initialised successfully

I moved rule to `enclave-runtime`.

BillyWooo (approved):
It should be good enough to go.

PR #1871, opened 2023-07-11
Use plain `Vec<>` in sidechain runtime

### LINKED ISSUE #1867: Use plain `Vec<>` in sidechain
### Context

In sidechain runtime it's not mandatory to use bound types or derive `MaxEncodedLen` trait, mainly because we won't have problems verifying the proof size (see [here](https://substrate.stackexchange.com/questions/7926/prepare-pallets-for-weights-v2)) for sidechain blocks - they are validated locally and don't need to be sent elsewhere.

Using plain `Vec<>` would increase readability and (hopefully) avoid unnecessary conversion back and forth.

---

Kailai-Wang:
### Context

resolves #1867 - there's no `BoundedVec` in sidechain pallet storage now

This PR renames `vec_identity` to `identities` too


kziemianek (approved):
The code looks fine. 
I think reducing the usage of `BoundedWeb3Network` across the code is a step in a good direction.
Even if we don't need bounded types in storage because we are not limited by the proof size I would consider keeping it for two reasons:
- it keeps storage records size limited
- it's going to be used across all pallets in substrate and should be treated as good practice

Apart from that, it would be good to have clear decision if/when we should use bounded types in sidechain's runtime pallets. Does it depend or do we always use/not use them ?

Kailai-Wang:
> Apart from that, it would be good to have clear decision if/when we should use bounded types in sidechain's runtime pallets. Does it depend or do we always use/not use them ?

This is a good point. From my perspective, we can avoid using it wherever we can. An exception is that we want to have an upper bound for the collections for whatever reason, then we can either do it via `Bounded...` or set a runtime const

PR #1877, opened 2023-07-12
Rename `G_DATA_PROVIDERS` to `GLOBAL_DATA_PROVIDER`

Kailai-Wang:
### Context

A trivial PR for renaming:
- to be consistent with other global variables
- `G_` is a bit confusing
- `DataProvidersStatic` => `DataProviderConfig`

This PR also updates the metadata and shortens the VC assertion list in ts-test, see comments:
```
// TODO: keep the list short, the manual types will be solved in #1878
//       more extensive testing of assertion will be done in #1873
```


kziemianek (commented):
What do you think about naming it `GLOBAL_CONFIG_PROVIDER` ?

kziemianek:
Data providers are:

- Achainable
- Discord
- Twitter

`DataProvidersStatic` struct holds only configuration for mentioned above providers.

Kailai-Wang:
Then `GLOBAL_DATA_PROVIDER_CONFIG`? "data provider" is a term widely used in discussions, so I'd like to keep both words

Then `DataProvidersStatic` should be changed to `DataProviderStaticConfig` too?

kziemianek:
> 

Sounds good. Maybe we can remove this `Static` word too so `DataProviderStaticConfig` -> `DataProviderConfig`. For me both versions are fine, I just prefer the latter.

Kailai-Wang:
Sure - I added a bit more renaming :)

BillyWooo (approved):
LGTM. 
Also thanks @kziemianek for the suggestions.

kziemianek (approved):

PR #1888, opened 2023-07-13
1880 Add a direct RPC to return MRENCLAVE

### LINKED ISSUE #1880: Add a direct RPC to return MRENCLAVE
### Context

As topic.
IIRC the only way to get MRENCLAVE for the moment is to parse the teerex registry on parachain. We had a few problems with that, e.g. #1820 

MRENCLAVE is public information. Similar to the worker's shielding key, we should be able to send the request directly to the worker and get it back. Under DI it further verifies that the target worker is the one you want to speak to.

---

songzhouran:
### Context

resolves #1880
add rpc to get enclave

### DIFF
--- a/tee-worker/core/rpc-client/src/direct_client.rs
+++ b/tee-worker/core/rpc-client/src/direct_client.rs
@@ -36,7 +36,7 @@ use std::{
 	thread::JoinHandle,
 };
 use substrate_api_client::RuntimeMetadataPrefixed;
-use teerex_primitives::Request;
+use teerex_primitives::{MrEnclave, Request};
 
 pub use crate::error::{Error, Result};
 
@@ -58,6 +58,7 @@ pub trait DirectApi {
 	fn get_state_metadata_raw(&self) -> Result<String>;
 	fn get_next_nonce(&self, shard: &ShardIdentifier, account: &AccountId) -> Result<u32>;
 
+	fn get_mrenclave(&self) -> Result<MrEnclave>;
 	fn send(&self, request: &str) -> Result<()>;
 	/// Close any open websocket connection.
 	fn close(&self) -> Result<()>;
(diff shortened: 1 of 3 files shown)


tee-worker/core/rpc-client/src/direct_client.rs
```diff
@@ -215,6 +216,19 @@ impl DirectApi for DirectClient {
...
+	fn get_mrenclave(&self) -> Result<MrEnclave> {
+		let jsonrpc_call: String =
+			RpcRequest::compose_jsonrpc_call("get_mrenclave".to_string(), Default::default())?;
+
+		// Send json rpc call to ws server.
+		let response_str = self.get(&jsonrpc_call)?;
+
+		let mrenclave: MrEnclave = decode_from_rpc_response::<MrEnclave>(&response_str)?;
+
+		info!("[+] Got enclave: {:?}", mrenclave);
```
kziemianek, line 228: I think this log might be too noisy.

kziemianek (approved):

tee-worker/core/rpc-client/src/direct_client.rs
songzhouran: Thank you for your comments! Can I use mrenclave.to_base58() or hex::encode(mrenclave)?

kziemianek: I thought about getting rid of entire log entry but I see we do similar logging in other methods so it's fine.
I'll resolve this conversation 🙃 .

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -218,6 +218,25 @@ where
 		Ok(json!(json_value))
 	});
 
+	// get_mrenclave
+	let get_mrenclave_name: &str = "get_mrenclave";
```
kziemianek, line 222: Hi @songzhouran I missed this small thing.
What do you think about renaming it to `state_getEnclave` 
Just to be consistent with https://github.com/litentry/litentry-parachain/blob/e27b733b84675d22438a9c4d5af0504900a540bc/tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs#L243

PR #1890, opened 2023-07-14
Switch to achainable api

### LINKED ISSUE #1654: Switch to new achainable REST API
### Context

The graphQL APIs are deprecated.

---

outofxxx:
fix: #1654 

Adding the `noteworthy` label for clients as the parameters for `A4/A7/A10/A11` are practically ignored for now.


tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -320,27 +142,26 @@ impl AchainablePost for AchainableClient {
...
-					Err(Error::AchainableError("Invalid response".to_string()))
-				},
+			Ok(res) => Ok(res),
 			Err(e) => Err(Error::RequestError(format!("{:?}", e))),
 		}
 	}
 }
 
-pub trait AchainableResultParser {
+pub trait AchainableSystemLabelParser {
```
BillyWooo: why it's named `SystemLabel`? The return struct is also SystemLabel?

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -320,27 +142,26 @@ impl AchainablePost for AchainableClient {
...
 	type Item = bool;
-	fn parse(value: serde_json::Value) -> Result<Self::Item, Error> {
-		if let Some(b) = value.as_bool() {
-			Ok(b)
+	fn parse(response: serde_json::Value) -> Result<Self::Item, Error> {
+		if let Some(value) = response.get("result") {
+			if let Some(b) = value.as_bool() {
+				Ok(b)
+			} else {
+				Err(Error::AchainableError("Invalid response".to_string()))
```
BillyWooo: Can we give more detail error message? At least to differentiate from the follow one.

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -358,6 +179,232 @@ fn check_achainable_label(
...
+	) -> Result<bool, Error> {
+		if index >= PATH_LENS {
+			return Err(Error::AchainableError("Wrong index".to_string()))
+		}
+
+		let path = if *network == Web3Network::Ethereum {
+			A4_ERC20_LIT_ETHEREUM_PATHS[index]
+		} else if *network == Web3Network::Litentry {
+			A4_LIT_LITENTRY_PATHS[index]
+		} else {
```
BillyWooo, line 244: Could it be `network` outside of these three predefined type lists?

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -358,6 +179,232 @@ fn check_achainable_label(
...
+	fn is_holder(&mut self, holder_type: &str, address: &str, index: usize) -> Result<bool, Error> {
+		if index >= PATH_LENS {
+			return Err(Error::AchainableError("Wrong index".to_string()))
+		}
+
+		let path = if holder_type == "A7" {
+			A7_DOT_PATHS[index]
+		} else if holder_type == "A10" {
+			A10_WBTC_PATHS[index]
+		} else {
```
BillyWooo: Same logic as above. If holder_type goes out of the predefined three lists, what will happen? From the logic here: all the rest goes into A11_ETH_PATHS. Is that true?

BillyWooo (commented):
As this is a temp solution for API replacement, so we don't spend too much on the quality of structure/code/maintenance/etc. Please have a thorough discuss about the architecture of new implementation with SystemLabel, then start refactor/implementation.

tee-worker/litentry/core/data-providers/src/achainable.rs
outofxxx: Added conditional judgment and filtered inappropriate situations in advance

BillyWooo (approved):
LGTM. Thanks

tee-worker/litentry/core/assertion-build/src/a10.rs
```diff
@@ -43,60 +44,33 @@ pub fn build(req: &AssertionBuildRequest, min_balance: ParameterString) -> Resul
 
 	let mut client = AchainableClient::new();
 	let identities = transpose_identity(&req.identities);
+	let addresses = identities
+		.into_iter()
+		.filter(|(network, _)| *network == Web3Network::Ethereum)
```
Kailai-Wang: I don't think you need the `filter` here => it should have the desired network types already, see `get_supported_web3networks` and its usage.

Same applies to other assertion building.

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -358,6 +179,243 @@ fn check_achainable_label(
...
+		if index >= PATH_LENS {
+			return Err(Error::AchainableError("Wrong index".to_string()))
+		}
+
+		if *network != Web3Network::Ethereum
+			|| *network != Web3Network::Litentry
+			|| *network != Web3Network::Litmus
+		{
+			return Err(Error::AchainableError("Unsupported network".to_string()))
+		}
```
Kailai-Wang: Maybe we should continue without breaking

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -358,6 +179,243 @@ fn check_achainable_label(
...
+
+		let params = ReqParams::new(path);
+		let body = ParamsAccount::new(address).into();
+		let resp = self.post(params, &body)?;
+
+		Self::parse(resp)
+	}
+}
+
+// consistently holding at least 5 DOT tokens
```
Kailai-Wang, line 256: Shouldn't it be 10 DOTs?

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -358,6 +179,243 @@ fn check_achainable_label(
...
+	"/v1/run/label/7f28c5cb-64c4-4880-9242-3cde638a57d4",
+	"/v1/run/label/0afc7c00-a1be-47aa-9903-2d99d2970091",
+	"/v1/run/label/078b2f54-4515-4513-9c67-33c30081b758",
+];
+
+pub trait AchainableHoldingAssertion {
+	fn is_holder(&mut self, holder_type: &str, address: &str, index: usize) -> Result<bool, Error>;
+}
+impl AchainableHoldingAssertion for AchainableClient {
+	fn is_holder(&mut self, holder_type: &str, address: &str, index: usize) -> Result<bool, Error> {
```
Kailai-Wang: Why not use `assertion` type?

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -358,6 +179,243 @@ fn check_achainable_label(
...
+			Web3Network::Ethereum => path = "/v1/run/label/8e19fb04-57fc-4537-ac93-d6fa7cff5bbe",
+			Web3Network::Khala => path = "/v1/run/label/f6a5cbe7-605a-4f9f-8763-67f90f943fb4",
+			_ => {
+				error!("Unsupported network: {:?}", network);
+				is_unsupported_network = true;
+			},
+		}
+		if is_unsupported_network {
+			return Err(Error::AchainableError("Unsupported network".to_string()))
+		}
```
Kailai-Wang: Same here - we can just ignore it

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -358,6 +179,243 @@ fn check_achainable_label(
...
+
+// TODO:
+// This is a compromise. We need to judge the range of the sum of transactions of all linked accounts, but the achanable api
+// currently only judges the range of a single account, so the current approach is to parse the returned data through
+// an assertion such as under 1 to calculate the sum, and then perform interval judgment.
+pub trait AchainableTotalTransactionsParser {
+	fn parse_total_transactions(response: serde_json::Value) -> Result<u64, Error> {
+		if let Some(value) = response.get("label") {
+			if let Some(value) = value.get("display") {
+				if let Some(displays) = value.as_array() {
```
Kailai-Wang: Can we use `add_then` here?

Kailai-Wang (changes requested):
Thanks for the PR! In general it looks fine but I do have some small comments.

Additionally, one major concern is: does it mean that the parameter `minimum_amount` in the assertion building is neglected due to hardcoding? If so please add proper comment in the code and create a follow-up issue for it.

tee-worker/litentry/core/data-providers/src/achainable.rs
outofxxx: According to [the description here](https://idhub.litentry.io/scores/web3-user-score-demo), so it should be 5 DOTs.

tee-worker/litentry/core/assertion-build/src/a10.rs
outofxxx: Here is to filter the address set of the specified network, because different assertions support different networks.

Kailai-Wang: Yes I mean:
> it should have the desired network types already, see get_supported_web3networks and its usage.

Did you check this:

https://github.com/litentry/litentry-parachain/blob/61f77ac1f57e0197b214b17231b5dcf076c3910b/primitives/core/src/assertion.rs#L59-L75

and this:

https://github.com/litentry/litentry-parachain/blob/61f77ac1f57e0197b214b17231b5dcf076c3910b/tee-worker/app-libs/stf/src/trusted_call_litentry.rs#L148

... so it's redundant unless we are talking about different things 🤔

outofxxx: Oh, yes, i see that, will fix it.

tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -106,57 +109,48 @@ pub fn build(req: &AssertionBuildRequest, min_balance: ParameterString) -> Resul
 	//    value: true
 	// ]
 	for (network, addresses) in identities {
+		if network != Web3Network::Ethereum
```
kziemianek: This should be already filtered out by https://github.com/litentry/litentry-parachain/blob/6da97441101348d8b1bfc486087514fd2c57088c/tee-worker/app-libs/stf/src/trusted_call_litentry.rs#L138-L151 as @Kailai-Wang mentioned. 

Is there any reason you kept it?

kziemianek (approved):

Kailai-Wang (approved):

PR #1891, opened 2023-07-14
Add missing `cli` mount in docker-compose files

Kailai-Wang:
### Context

As topic, without this the `cli/` changes won't take effect without rebuilding tee docker image.


BillyWooo (approved):
Thanks. Yes, it saves lots of time. Should drop that from upstream.

PR #1892, opened 2023-07-14
feat: Release enclave binary

### LINKED ISSUE #1628: Have an automated release process for the worker and enclave

felixfaisal:
I think the **********PR********** might be better served if I can provide a full context and the documentation will generally be helpful for future iterations: 

- It updates the `yml` file for `create-release-draft.yml` and includes building enclave and packaging it as an artifact
- It updates the `generate-release-notes.sh` to include the `release notes` for the enclave such as the `rustc_version` , `mrenclave` etc.

So currently, I'm including only the following information for the `Enclave` Release Notes: 
```
rustc: $RUSTC_VERSION
mrenclave: $MRENCLAVE
upstream_commit: $UPSTREAM_COMMIT
```
Let me know if you'd like me to include anything else in the workflow 

Resolves #1628


scripts/generate-release-notes.sh
```diff
@@ -143,6 +143,23 @@ sha1sum of genesis wasm   : $GENESIS_WASM_HASH
 EOF
 fi
 
+# release notes for enclave binary 
+if [ "$2" != "runtime" ] && [ "$2" != "client" ] && [ "$2" != "both"]; then
```
BillyWooo: Why can't have an option of `"$2" == "enclave"` ?

scripts/generate-release-notes.sh
```diff
@@ -143,6 +143,23 @@ sha1sum of genesis wasm   : $GENESIS_WASM_HASH
 EOF
 fi
 
+# release notes for enclave binary 
+if [ "$2" != "runtime" ] && [ "$2" != "client" ] && [ "$2" != "both"]; then 
+   echo "Generating Release Notes for Enclave"
+   local UPSTREAM_COMMIT=$(cat tee-worker/upstream_commit)
```
BillyWooo: why need `UPSTREAM_COMMIT` here? Or in another words: Parachain repo commit hash is less important?

BillyWooo (commented):
The flow looks fine. Have you got a chance to test on `prod-tee-1` ?

scripts/generate-release-notes.sh
felixfaisal: I kind of tried following the existing pattern in the script, let me elaborate: 
```bash
if [ "$2" != "client" ]; then
  echo "## Runtime" >> "$1"
  ```
  A snippet, wherein `$2` != `client` we build the `Runtime`, So i used the same pattern, wherein if not `client`, `runtime` and `both`, we build the `enclave`. This is not very intuitive imo but would require good refactoring if we were to update existing pattern. 
  CC: @Kailai-Wang

File: scripts/generate-release-notes.sh (hunk 2 above)
Comment (felixfaisal) [reply scripts/generate-release-notes.sh]: I wasn't sure, what all needs to be included in the `release notes` but I thought `UPSTREAM_COMMIT` could be good to track the progress of our worker against the upstream when we release a version from our side.

File: .github/workflows/create-release-draft.yml [hunk 3]
```diff
@@ -13,6 +13,11 @@ on:
         description: runtime
         required: true
         default: true
+      release_enclave:
```
Kailai-Wang, line 16: Does it always contain `enclave.signed.so` and the worker binary?

.github/workflows/create-release-draft.yml
```diff
@@ -148,6 +157,42 @@ jobs:
             ${{ env.GENESIS_RELEASE }}-genesis-state
             ${{ env.GENESIS_RELEASE }}-genesis-wasm
 
+  ## Build the enclave and package config files
+  generate_package_and_artifact:
```
Kailai-Wang: maybe something like `build-enclave` would be more appropriate for the job name

.github/workflows/create-release-draft.yml
```diff
@@ -277,6 +322,8 @@ jobs:
 
       - name: Generate release notes
         run: |
+          export RUSTC_VERSION=${{ needs.generate_package_and_artifact.outputs.rustc_version }}
+          export MRENCLAVE_OUTPUT=${{ needs.generate_package_and_artifact.outputs.mrenclave }}
```
Kailai-Wang: I don't see them in the `needs` list?

Kailai-Wang (commented):
Thank you, looks good for the most part.

The most important question: have you tried it? :) Do you have something like a generated release draft?

scripts/generate-release-notes.sh
BillyWooo: I don't think `UPSTREAM_COMMIT` should be included in the release note. It's mainly used for internal usage to track against the upstream worker. I would suggest the current repo git commit hash is more important and valuable. Take the short hash value please.

scripts/generate-release-notes.sh
BillyWooo: What I mean is: why not replace the whole line with `if [ "$2" == "enclave" ]; then` ? Otherwise what if another new parameter coming in, what will happen here?

.github/workflows/create-release-draft.yml
BillyWooo: Don't get what you mean here. The package will be created by `./tee-worker/scripts/litentry/release/generate_release_pkg.sh`. So `enclave.signed.so` and worker binary are created together.

Kailai-Wang: Is there any case we only want to release one of them?

Similar to substrate client and runtime

scripts/generate-release-notes.sh
felixfaisal: Removed ✅

scripts/generate-release-notes.sh
felixfaisal:
```
if [ "$2" = "enclave" ] || [ "$2" = "all" ]; then 
``` 
Updated to use this

.github/workflows/create-release-draft.yml
felixfaisal: Yes, if you only select `enclave`, it will release only the `enclave` 
```
          elif [ "${{ github.event.inputs.release_enclave }}" = "true" ]; then
            release_type="enclave"
          elif [ "${{ github.event.inputs.release_enclave }}" = "true" ] && [ "${{ github.event.inputs.release_client }}" = "true" ] && [ "${{ github.event.inputs.release_runtime }}" = "true" ]; then
            release_type="all"
```

.github/workflows/create-release-draft.yml
felixfaisal: Done ✅

.github/workflows/create-release-draft.yml
felixfaisal: Included it now 😅

.github/workflows/create-release-draft.yml
Kailai-Wang: It will only release `enclave` but the enclave still contains both binary (integritee-worker) and enclave (.so), right? I mean is there any case we only want to release one of them. Like only the worker part but the enclave doesn't change.

That's why I said "similar to substrate client and runtime".

Not saying we should differentiate them, I'm just raising the question

Kailai-Wang:
> Do you have something like a generated release draft?

ping 🔼

.github/workflows/create-release-draft.yml
felixfaisal: I see, As of now, it doesn't do that. Both `integritee-worker` and `enclave` are packaged together.

felixfaisal:
So testing this GHA has been difficult to do independently, because it requires building and releasing, and mocking it was tough, However I've tested some parts independently, 
1. Building the enclave and set it artifacts and outputs 
- https://github.com/litentry/litentry-parachain/actions/runs/5551356714/jobs/10137391942 
2. Generating Release Scripts: 
- https://github.com/felixfaisal/litentry-parachain/actions/runs/5543051626/jobs/10118562372 (Not a good mock-up) 

I could not find a better way to test these changes, Let me know if I can test it elsewhere 
CC: @Kailai-Wang @BillyWooo

Kailai-Wang:
You simply run https://github.com/litentry/litentry-parachain/actions/workflows/create-release-draft.yml on your feature branch

.github/workflows/create-release-draft.yml
BillyWooo: @felixfaisal I don't think you answered @Kailai-Wang 's question. He wanted to know: would we have situation that only 'worker' or 'enclave.signed.so' (one of them) needed to be released?
TBH, it could happen. But I would say very rare. 
And we can build them separately (it's easy to do), we just don't do it. Normally we build all together.

Kailai-Wang:
btw it's a bit strange that building enclave only took 6min: https://github.com/litentry/litentry-parachain/actions/runs/5610322811

Maybe it reuses the existing code.
We'd better checkout the repo to a different folder (like `mktemp -d `) and do the compilation ...

Kailai-Wang:
Whenever it's ready for review, please attach a link to the **release draft** for reference, thanks :)

.github/workflows/create-release-draft.yml
```diff
@@ -148,43 +158,43 @@ jobs:
             ${{ env.GENESIS_RELEASE }}-genesis-state
             ${{ env.GENESIS_RELEASE }}-genesis-wasm
 
-  ## test again the built docker image ##
```
BillyWooo, line 151: Why we simply remove these testing steps?

felixfaisal: Hey @BillyWooo I removed them because it was failing and not related to the changes I've made, So that I can only test the changes I've made. I will include them back.

felixfaisal:
So I had a few succesfull tests completed 
- https://github.com/litentry/litentry-parachain/actions/runs/5660062718/job/15338016011 (both) 
- https://github.com/litentry/litentry-parachain/actions/runs/5655699432/job/15323894884 (all) 
- https://github.com/litentry/litentry-parachain/actions/runs/5654230410 (enclave) 

I will add the steps I removed for testing

BillyWooo (approved):
thanks

Kailai-Wang (changes requested):
A few questions:

- why was the `release-xxx.tar.gz` file still included in the [release](https://github.com/litentry/litentry-parachain/releases/tag/untagged-89b09d8665e2a4b87955) with `client` and `runtime` only?
- the `rustc` version is wrong - you need the one from the `tee-worker/` folder
- can you align the indention with the existing dependencies? (like polkadot dependency)
- we should include the hash of the `binary` and `enclave.signed.so` too
- did you differentiate between a `dev` build and `prod` build?

felixfaisal:
> A few questions:
> 
> * why was the `release-xxx.tar.gz` file still included in the [release](https://github.com/litentry/litentry-parachain/releases/tag/untagged-89b09d8665e2a4b87955) with `client` and `runtime` only?
> * the `rustc` version is wrong - you need the one from the `tee-worker/` folder
> * can you align the indention with the existing dependencies? (like polkadot dependency)
> * we should include the hash of the `binary` and `enclave.signed.so` too
> * did you differentiate between a `dev` build and `prod` build?

 why was the release-xxx.tar.gz file still included in the [release](https://github.com/litentry/litentry-parachain/releases/tag/untagged-89b09d8665e2a4b87955) with client and runtime only?
It was included because of this step that I had left behind, which I removed in this commit - https://github.com/litentry/litentry-parachain/pull/1892/commits/4f1a539b6f7bd3a859bf5d66d8366b98b952af90

I've incorporated all the other changes, except for 
> * did you differentiate between a `dev` build and `prod` build?

For now the binaries are built on production mode, because `generate_relase_pkg.sh` builds it on Production, Should I create an input to allow us to chose whether production or dev? 

This is the recent run: 
https://github.com/litentry/litentry-parachain/releases/tag/untagged-f0f321cd562d7eebdbfb

Kailai-Wang:
Thanks!

Can you rerun it with the "previous release" set? I want to check the changelog and collected PRs.

felixfaisal:
Here's the draft with the previous release having changelog  - https://github.com/litentry/litentry-parachain/releases/tag/untagged-8229ac39f97c3c61c335

Kailai-Wang:
Looks good!
Better would be that we list those PRs with C0 and C1 first to highlight them even more, which are followed by the PRs without any tags, ideally with toggle list.

But that should be out of this PR's scope.

Kailai-Wang (approved):
I approve it - one minor thing is we need shasum for both binary and enclave. In this PR there's only one, which is not enough, and please mind the upper/lower case letters :) (to be consistent with other metadata)

PR #1893, opened 2023-07-14
set proper name for mrenclave rpc

### LINKED ISSUE #1880: Add a direct RPC to return MRENCLAVE

songzhouran:
### Context

- set proper name for mrenclave rpc
- the issue #1880 is already resolved.


kziemianek:
@songzhouran please change title and description:

- it only renames things
- the issue is already resolved.

BillyWooo (approved):

Issue #1715, opened 2023-05-19
Allow EVM-based sign-in and request


Kailai-Wang:
### Context

Thanks to DI, we can accept queries from EVM-based identities now. It means users don't even need to own a parachain account to create the IDGraph, as they don't _have to_ send extrinsic anymore.

However, our current data structure and interface must be adjusted to support this:

#### IDGraph data structure change

Current IDGraph is a double map with keys `(litentry-account-address, identity)`, it's going to be changed:
https://www.notion.so/web3builders/EVM-Sign-in-IDGraph-change-4e33973861c84c8c8858ad97d4fe74ab

#### TrustedCall interface and signature verification change

Typically `TrustedCall` uses the first arg (of type `AccountId`) as the account which signs the request, with the signature defined as `sp_runtime::MultiSignature`. We probably need to extend it to [IdentityMultiSignature](https://github.com/litentry/litentry-parachain/blob/d8903604b191a2f6afb295bf6e353289162b0810/tee-worker/litentry/primitives/src/validation_data.rs#L31), and the account should be the more generic [Identity](https://github.com/litentry/litentry-parachain/blob/d8903604b191a2f6afb295bf6e353289162b0810/tee-worker/litentry/primitives/src/identity.rs#L124)

Obviously the signature verification needs to be adjusted too

#### nonce mechanism change

When creating DI requests, a nonce is needed to avoid replay attack or duplicate execution. The nonce is retrieved from `System::account_nonce` for now, but it's only for substrate accounts.

For EVM identities we probably need to have our own storage in pallet, or take use of `pallet-evm` from [frontier](https://github.com/paritytech/frontier).

#### Summary

There's more than expected to be done here.

I could have created multiple sub-tasks for it, but those changes are closely connected. So, ideally:
- the implementation can be done by one person to keep consistent
- the changes should be submitted as one single PR, as any individual change alone would break the current codebase

We can have multiple people working on the same feature branch tho (e.g. for ts-tests)

Meanwhile, I see the upstream has quite some evm-related implementations already, maybe we can refer to them to avoid duplicate effort.
E.g. https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/app-libs/stf/src/evm_helpers.rs

That's some research debt from https://github.com/litentry/litentry-parachain/issues/996

### Scope:
- [x] ethereum signature verification
- [x] evm id graphs
- ~~[ ] identity archiving~~
- [x] identity can be linked to 1 idgraph only

---
[...]

PR #1895, opened 2023-07-16
identity can be linked only once

kziemianek:
Adds a rule that identity can be linked only once.

related: #1715


Kailai-Wang (approved):
This change looks fine, thanks!

Now that we add the `LinkedIdentities` - I'm thinking this might be changed again later when we add the `set_prime_identity` feature. Even the `IDGraph` storage - I might need to break it into two storages with some unique identifiers as indexes

PR #1911, opened 2023-07-20
1899 ts-tests: retrieve nonce from sidechain

### LINKED ISSUE #1899: ts-tests: retrieve nonce from sidechain
### Context

This comment from @Kailai-Wang in `identity.test.ts`:

> being lazy - the nonce here is hardcoded
> it's better to retrieve the starting nonce from the sidechain and increment
> it for each such request, similar to the construction of substrate tx
> However, beware that we should query the nonce of the enclave-signer-account
> not alice or bob, as it's the indirect calls are signed by the enclave signer

### Task

- [ ] Replace hardcoded nonce with enclave-signer nonce from sidechain

songzhouran:
### Context

resolves #1899


tee-worker/ts-tests/integration-tests/identity.test.ts
```diff
@@ -24,6 +24,33 @@ import type { LitentryValidationData, Web3Network } from 'parachain-api';
 import type { TransactionSubmit } from './common/type-definitions';
 import type { HexString } from '@polkadot/util/types';
 import { ethers } from 'ethers';
+import { sendRequest } from './common/call';
+import * as base58 from 'micro-base58';
+
+async function getWorkerAddress(base58mrEnclave: string, context: any) : Promise<string> {
```
arieljbirnbaum:
```suggestion
async function getWorkerAddress(base58mrEnclave: string, context: IntegrationTestContext) : Promise<string> {
```

arieljbirnbaum: (same for other instances of `context`)

tee-worker/ts-tests/integration-tests/identity.test.ts
```diff
@@ -24,6 +24,33 @@ import type { LitentryValidationData, Web3Network } from 'parachain-api';
...
+    if(len > 0) {
+        for( let i = len - 1; i > 0; i--){
+            nonce *= 16;
+            nonce += u8aValue[i];
+        }
+    }
+    console.log("res value",u8aValue);
+    console.log("nonce is:", nonce);
+    return nonce;
+}
```
arieljbirnbaum, line 55: I'm confused -- does this give a different value than `res.value.toHuman()`?

arieljbirnbaum (changes requested):

tee-worker/ts-tests/integration-tests/identity.test.ts
songzhouran: u8a is array like this [ 16, 15, 0, 0, 0 ].  toHuman() is like this: 0x0f000000. Which would you perfer to print?

arieljbirnbaum: What I mean is, why do we need to decode the nonce by hand instead of using the baked-in functionality?

tee-worker/ts-tests/integration-tests/identity.test.ts
```diff
@@ -38,16 +38,12 @@ async function getWorkerAddress(base58mrEnclave: string, context: IntegrationTes
...
-    if(len > 0) {
-        for( let i = len - 1; i > 0; i--){
-            nonce *= 16;
-            nonce += u8aValue[i];
-        }
-    }
-    console.log("res value in array", u8aValue, "in Hex:", res.value.toHuman());
+    if(resHex){
+        nonce = parseInt('0x' + resHex.slice(2)?.match(/../g)?.reverse().join(''));
+    }
```
arieljbirnbaum: Does it work if you say
```
const nonce = context.api.createType('Index', res.value.toHex()).toNumber()
```
?

songzhouran: This will cause endian problem. We must reverse the string byte by byte(every two character). For example, res.value.toHex() = 0x0f000000, we will firstly convert it to 0x0000000f.

0xverin: Decoding it here is not the correct approach. We should use `createType` to decode it. If it's about obtaining the sidechainNonce, [there](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/integration-tests/examples/direct-invocation/util.ts#L246) has been an implementation for that. I'm not sure what the difference is between you and him.

arieljbirnbaum: Like @Verin1005 said; if there's an endianness issue in the encoding, that should already be accounted for in the metadata (and thus handled by `createType`); if that doesn't work then that's something we need to dig into.

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -103,6 +104,67 @@ where
 		Ok(json!(json_value.to_hex()))
 	});
 
+	// author_getEnclaveAccountId
+	let state_account_id = state.clone();
+	let author_get_enclave_account_id: &str = "author_getEnclaveAccountId";
```
Kailai-Wang: You can actually get it via parachain storage.

If you really need it, it can be derived from the shielding key pair I think:
https://github.com/litentry/litentry-parachain/blob/55c2e1642408e8475164dd665a9cbd219c2a5bc0/tee-worker/core-primitives/stf-state-handler/src/state_initializer.rs#L61

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -103,6 +106,32 @@ where
 		Ok(json!(json_value.to_hex()))
 	});
 
+	// author_getEnclaveId
+	let rsa_pubkey_name: &str = "author_getEnclaveId";
```
Kailai-Wang: I think `EnclaveId` is really confusing - how about `author_getEnclaveSignerAccount`

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -103,6 +106,32 @@ where
...
+	io.add_sync_method(rsa_pubkey_name, move |_: Params| {
+		let shielding_key_repository = match GLOBAL_SHIELDING_KEY_REPOSITORY_COMPONENT.get() {
+			Ok(s) => s,
+			Err(e) => {
+				let error_msg: String = format!("{:?}", e);
+				debug!("{:?}", e);
+				return Ok(json!(compute_hex_encoded_return_error(error_msg.as_str())))
+			},
+		};
+		let vc_pubkey = shielding_key_repository
```
Kailai-Wang: `vc_pubkey` is confusing too, although they are the sam key, I suggest `enclave_signer_account` or `enclave_account`

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -103,6 +106,32 @@ where
...
+			.retrieve_key()
+			.and_then(|keypair| keypair.derive_ed25519().map(|keypair| keypair.public().to_hex()))
+			.ok();
```
Kailai-Wang: `ok()` leaves an `Option` behind, that's probably why you need `slice(1)`  below.

~~We should unwrap it's value - either a direct `unwrap()` or handle the error if you want~~

tee-worker/ts-tests/integration-tests/identity.test.ts
```diff
@@ -214,14 +247,15 @@ describeLitentry('Test Identity', 0, (context) => {
             },
         };
         const bobSubject = await buildIdentityFromKeypair(new PolkadotSigner(context.substrateWallet.bob), context);
-
+        const nonce9 = await getNonce(base58mrEnclave, workerAddress, context);
```
Kailai-Wang: It's strange that nonce`9` is used here and later again `nonce` is named. Can we use `const nonce` here?

tee-worker/ts-tests/integration-tests/identity.test.ts
```diff
@@ -22,9 +22,34 @@ import { assert } from 'chai';
...
+import type { IntegrationTestContext, TransactionSubmit } from './common/type-definitions';
 import type { HexString } from '@polkadot/util/types';
 import { ethers } from 'ethers';
+import { sendRequest } from './common/call';
+import * as base58 from 'micro-base58';
+import { decodeRpcBytesAsString } from './common/call';
+
+async function getWorkerAddress(context: IntegrationTestContext): Promise<string> {
+    const requestAcc = { jsonrpc: '2.0', method: 'author_getEnclaveId', params: [], id: 1 };
+    const resAcc = await sendRequest(context.tee, requestAcc, context.api);
```
Kailai-Wang: Would you like to be consistent with other impl?
Like 
```
function getEnclaveSignerAccount ....
const request = ...
const res = ...
const enclaveSignerAccount = ...
```

@grumpygreenguy might have better ideas

Kailai-Wang (changes requested):
I assume it works but I have some comments regarding the code style :(

I understand it's a flavor problem but in general we need to be more cautious and consistent

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
arieljbirnbaum: I thought we wanted to avoid `unwrap` (and panics in general)?

Kailai-Wang: I corrected myself during a call with Zhouran earlier but forgot to edit the comment, sorry!

Indeed we should avoid panics in enclave, so errors should be handled here.

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
arieljbirnbaum: Looks like it's actually the public key tho, so maybe `enclave_signer_public_key`? Also, since that's the value that the method returns, I'd argue that that's what it should be named.

tee-worker/ts-tests/integration-tests/identity.test.ts
arieljbirnbaum: My 2 picoLIT: it should do what it says and say what it does :D And what it does is just calling the other method so I think the name should reflect just that. Assuming it's `enclave_signer_public_key`, then I'd expect something like

```
async function getEnclaveSignerPublicKey(...) {
    const request = { ... }
    const response = await sendRequest(...)
    if (!response.status.isOk) {
        throw new Error(...)
    return response.value
}
```

(on that note: as a reader, without any additional context, I have no idea what the `Acc` in `requestAcc` is supposed to stand for; there's no other request/response pair in the function body tho, so it can just be left as `request` and `response`).

Kailai-Wang (approved):
Looks fine from my perspective.

@grumpygreenguy @Verin1005 can you please review it too? :) 

@songzhouran can you please merge `dev` to this branch to detect any incompatibilities early?

songzhouran:
> Looks fine from my perspective.
> 
> @grumpygreenguy @Verin1005 can you please review it too? :)
> 
> @songzhouran can you please merge `dev` to this branch to detect any incompatibilities early?

Merging dev done.

arieljbirnbaum (approved):

PR #1920, opened 2023-07-21
Make some DI responses synchronous

### LINKED ISSUE #1883: Making some DI responses synchronous
### Context

To further reduce the waiting time, we'd like to make some DI responses synchronous, that is, do not post the responses as parachain event, but return to the caller as rpc results.

For the following calls:
- [x] set_user_shielding_key
- [x] activate_identity
- [x] deactivate_identity
- [x] set_identity_networks
- [x] link_identity

For VC issuance it needs to change parachain storage so we'll keep it unchanged. 
It's medium prio now but still, it would be great if we can finish it soon.

---

Kailai-Wang:
### Context

~~Partially implemented #1883, it's divided into multiple PRs as otherwise the PR would be too big~~

solves #1883 

This PR:
- laid down skeleton to make response synchronous
- apply parentchain effects for failed STFs too
- store the RPC responses to top pool
- embed `top_hash` into `TrustedOperationStatus`
- devise a way to deliver synchronous responses for streamed trustedCalls (e.g. `link_identity`)
- refactor some of the trusted calls - they should return `Err` so that the synchronous DI response shows `Invalid`
- responses to these calls are synchronous now:
   - set_user_shielding_key
   - activate_identity
   - deactivate_identity
   - set_identity_networks
   - link_identity

**The parachain callbacks are preserved to make CI backwards compatible**. If you send one of the DI requests above, you should see both the rpc response **AND** the parachain event. 

Now the `WorkerRpcReturnValue` would have the following structure, example of `set_user_shielding_key`:
```
setUserShieldingKey call returned {
  value: '0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48fcc5ab7556950296f4f3042fd955b1a281f7683ad07cf54819ab4d4921c636f8c61a250d01bc98d1b3dd8a4e49bfaa1f17fe9d0a086e950a40c58a6a20f3152f00d0964c0fc62a4b7640706f048bda3725e56e2882631916e924d9d8133e2d628a8317593fe5a245dcf7f969e0',
  do_watch: false,
  status: {
    TrustedOperationStatus: [
      [Object],
      '0x0c2612dc5372b7ba83cbc33b1f2023cb287a7c6546ed79b53921b04dc109ca7e'
    ]
  }
}
```

`value` would contain the encoded `*Response` struct, please refer to the test code in DI-sample.

TODO (see comment in `trusted_call.rs`)
- [x] #1928 
- [x] #1929


Kailai-Wang:
Hold on - maybe I can finish the streamed trustedCall part

tee-worker/litentry/core/stf-task/receiver/src/lib.rs
```diff
@@ -136,6 +137,10 @@ where
 			return Ok(())
 		}
 
+		// swap the hash in the rpc connection registry to make sure furthre RPC responses go to
+		// the right channel
+		self.author_api.swap_rpc_connection_hash(*old_top_hash, top.hash());
```
kziemianek, line 142: It's a little bit hacky I feel. I totally agree this is very smart and fast way of achieving desired outcome.

In general how I undestand it is that the user is submitting one extrinsic (TOP) and receives response from another one.

We have got a request identifier, I wonder if we can assign tops to it and track them.
It would require much more work but would be clearer.

Another idea is to add steps to TOP.
Most of TOPs would be single stepped but `link_identity` could be divided into `pre_verification` and `post_verification` steps.

kziemianek, line 142: Have you tested it in multiworker setup? 
I can imagine scenario in which first TOP is processed by worker 1 but second by worker 2. Will rpc client receive correct response ?

According to this comment TOPs might not be synchronized, so maybe this is not a real problem now but may occour in future.
https://github.com/litentry/litentry-parachain/blob/7e096092915036bcff513dbb334528d0a75395f6/tee-worker/sidechain/consensus/aura/src/block_importer.rs#L145 - I couldn't find a code doing sync job.

Kailai-Wang: > Have you tested it in multiworker setup?

Nope, but I expect it to work.

Because the top pool is **not** synced between workers, so if you submit your request to worker 1 only worker 1 will be able to process it and reply to you.

Worker 2 will get the processing result as sidechain blocks.

I don't see it changing in the near future

Kailai-Wang: Yes I have TODOs in the code that maybe it's better to rely on `req_ext_hash` as the connection registry key - but it would require you to pass this hash all the way down to deep in top pool and change the listener/watch key as well.

I'm unsure if it would cause any problem because they are indeed two different tops (with different signers, different nonces), so they **should differ** from tx pool's perspective.

However, from the requester, we want it to behave as if it's a continuous, integral single request. By swapping the hash we basically say, hey I know it's technically two different trusted calls, but I want you two to share the same channel to pass back responses - so we keep their distinction until the last minute when we reply to the client - actually not bad I'd say 🤔 

What can be improved here is we'd better identify the other trusted calls by `req_ext_hash`, not by the value `0x01`, but since we have full control over the processing of trusted calls (= we know which calls will generate `0x01` and we identity both old and new hashes when swapping) it should be fine

kziemianek: It sounds like once we get more workers the time to process this operation will increase linerary and the problem will come back sooner or later.

kziemianek: I mean if we have Y workers online then:

- first TOP will be processed in block X
- second TOP will be processed in block X+Y

Kailai-Wang: Yes that's a legit assumption. It is out of the scope of this PR but we'll need to address it sooner or later when we start multi-worker setup. We need to have an issue to track this.

II won't have such problem because the tx pool is synced from the parachain.

tee-worker/core/rpc-client/src/direct_client.rs
```diff
@@ -218,7 +218,7 @@ impl DirectApi for DirectClient {
 
 	fn get_state_mrenclave(&self) -> Result<MrEnclave> {
 		let jsonrpc_call: String =
-			RpcRequest::compose_jsonrpc_call("state_getEnclave".to_string(), Default::default())?;
+			RpcRequest::compose_jsonrpc_call("state_getMrenclave".to_string(), Default::default())?;
```
kziemianek, line 221: The rpc method is still `	let state_get_enclave_name: &str = "state_getEnclave";`

kziemianek (commented):
I noticed several warnings in tee-worker logs while running `test-identity:local` ts-tests aganist this implementation

```
[2023-07-24T21:07:06Z INFO  itc_direct_rpc_server::rpc_responder] set response value
[2023-07-24T21:07:06Z WARN  itp_top_pool::watcher] failed to set response value to rpc client: InvalidConnectionHash
```
It looks like it doesn't make sense to invoke `set_rpc_response_value` for TOPs that originates from indirect calls.

Kailai-Wang:
> I noticed several warnings in tee-worker logs while running `test-identity:local` ts-tests aganist this implementation
> 
> ```
> [2023-07-24T21:07:06Z INFO  itc_direct_rpc_server::rpc_responder] set response value
> [2023-07-24T21:07:06Z WARN  itp_top_pool::watcher] failed to set response value to rpc client: InvalidConnectionHash
> ```
> 
> It looks like it doesn't make sense to invoke `set_rpc_response_value` for TOPs that originates from indirect calls.

I believe you'll see such warnings even for DI requests due to potential early hash swap. The swap is called in another thread, and there's no guarantee that the previous top is already imported in the sidechain block at that time. In fact, for most of the time it's **not yet** included in the block.

Therefore later when the previous top does get included in the block, the top pool watcher tries to send out the response and finds out the channel doesn't exist any more, hence the warning. It's expected behavior and should be fine in such case.

tee-worker/core/rpc-client/src/direct_client.rs
Kailai-Wang: Good catch, thanks!

tee-worker/core-primitives/stf-executor/src/lib.rs
```diff
@@ -57,26 +57,39 @@ pub mod mocks;
...
+/// Litentry:
+/// we have made a few changes:
+/// - additionally for the success case, we add the encoded rpc response that will be
+///   passed back to the requester when the top status is `InSidechainBlock`
+/// - for failed top, we apply the parachain effects too
 #[derive(Clone, Debug, PartialEq)]
 pub enum ExecutionStatus {
-	Success(H256, Vec<OpaqueCall>),
-	Failure,
+	Success(H256, Vec<OpaqueCall>, Vec<u8>),
```
kziemianek, line 68: Can we make this `Option<Vec<u8>>` because execution status of `TrustedOperation::indirect_call` shouldn't contain rpc response ?

tee-worker/core-primitives/stf-executor/src/lib.rs
```diff
@@ -151,6 +175,20 @@ where
...
+	// For now only successfully executed tops have a hash (see `get_executed_operation_hashes` above),
+	// but that doesn't matter, because the failed top rpc response will stay unchanged
+	pub fn get_rpc_responses_values(&self) -> Vec<(H256, Vec<u8>)> {
+		self.executed_operations
+			.iter()
+			.filter_map(|ec| {
+				ec.status
+					.get_executed_operation_hash()
+					.map(|h| (h, ec.status.get_rpc_response_value()))
+			})
```
kziemianek, line 189: You will need to add this 

```
			// we are interested only in executed_operations with response
			.filter_map(|(h, r)| r.map(|r| (h, r)))
```

tee-worker/core-primitives/stf-executor/src/executor.rs
```diff
@@ -124,16 +126,20 @@ where
...
 			self.node_metadata_repo.clone(),
 		) {
-			error!("Stf execute failed: {:?}", e);
-			return Ok(ExecutedOperation::failed(top_or_hash))
-		}
+			Err(e) => {
+				error!("Stf execute failed: {:?}", e);
+				return Ok(ExecutedOperation::failed(top_or_hash, extrinsic_call_backs))
+			},
+			Ok(res) => res,
```
kziemianek, line 141: Here we should only care about res coming from indirect calls

```
			Ok(res) => match trusted_operation {
				TrustedOperation::indirect_call(_) => Some(res),
				_ => None
			},
```

kziemianek (commented):
By making these changes and using rpc return values only for indirect calls I managed to get rid of all `[WARN  itp_top_pool::watcher] failed to set response value to rpc client: InvalidConnectionHash` log entries.

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -218,9 +218,9 @@ where
 		Ok(json!(json_value))
 	});
 
-	// state_getEnclave
-	let state_get_enclave_name: &str = "state_getEnclave";
-	io.add_sync_method(state_get_enclave_name, |_: Params| {
+	// state_getMrenclave
+	let state_get_mrenclave_name: &str = "state_getMrenclave";
+	io.add_sync_method(state_get_mrenclave_name, |_: Params| {
```
kziemianek, line 223: I think we should rename `state_updateScheduledEnclave` -> `state_updateScheduledMrenclave` also.

Kailai-Wang: Updating "enclave" makes sense too as it basically says: I'd like to update the enclave schedule, only enclave with _this_ mrenclave can produce blocks start from _this_ block number, so we schedule the enclave behaviour as a whole, which includes both mrenclave and sidechain_block_number as criteria.

I'm fine with either way though if you insist on renaming

tee-worker/core-primitives/stf-executor/src/lib.rs
Kailai-Wang: This is a valid concern, I'll leave it to another PR. In fact, I created a related issue earlier: https://github.com/litentry/litentry-parachain/issues/1922

It's about distinguishing II and DI in a broader sense

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
kziemianek: Ok, i see the difference 🙃

tee-worker/litentry/core/stf-task/receiver/src/lib.rs
BillyWooo: @kziemianek  The `pre_xxx` and `post_xxx`, we had it before for indirect invocation. Do you mean to switch back?

tee-worker/core-primitives/stf-executor/src/executor.rs
Kailai-Wang: same above - how about another PR dedicated for this problem?

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -254,25 +252,32 @@ where
...
 	// Update:
 	// see discussion in https://github.com/integritee-network/worker/issues/1232
 	// my current thoughts are:
-	// - we should return Err() if the STF execution fails
+	// - we should return Err() if the STF execution fails, the parentchain effect will get applied regardless
 	// - the failed top should be removed from the pool
-	// - however, the failed top hash needs to be included in the sidechain block
+	// - however, the failed top hash needs to be included in the sidechain block (still TODO)
+	//
+	// Almost every (Litentry) trusted call has a `H256` as parameter, this is used as the requent identifier.
```
BillyWooo, line 262: 'requent identitier' -> 'request identifier' ?

tee-worker/core-primitives/stf-executor/src/lib.rs
kziemianek: ok 👍🏼

kziemianek (approved):
Looks good 👍🏼

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: Yes thanks 👍 - I'll fix it in some PR next

Issue #1861, opened 2023-07-08
Reivew VC fields


Kailai-Wang:
### Context

VC undergoes variant transmission channels: top pool buffer, internal mpsc, ocall, extrinsic.

Reducing the VC size would help improve the performance/latency.

However, before doing that, we need to review the structure to see if we have proper fields at all.

### Tasks:
- compare the current VC structure with w3c standards or even VCs generated from other projects:
   - are we missing any fields?
   - do we wrongly populate any fields?
   - are all additional fields necessary?
 - based on that, check if we can shorten the VC size by e.g. removing redundant fields, or populating it with a shorter representation.
 - (optional) using any "standard" VC parser tool to parse our VC, does it work?

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

kziemianek:
* **@context**
  * ```["https://www.w3.org/2018/credentials/v1", "https://w3id.org/security/suites/ed25519-2020/v1"]```
  * let's keep it as it is
* **id**
  * generated randomly, example: `0x90d9a1227dfd524693f927db7fc8aa6b39231d1ca7efd96c7f53b70604536190`
  * It should meet following requirements:  </br> The id property MUST express an identifier that others are expected to use when expressing statements about a specific thing identified by that identifier. </br> The id property MUST NOT have more than one value. </br> The value of the id property MUST be a URI.
* **type**
  * `["VerifiableCredential"]`
  * We can use more specific types but that would increase size of generated VCs, example: `["VerifiableCredential", "UniversityDegreeCredential"]`
* **credentialSubject**
  * ```{"id":"5c5d33eb2187f2a769e0fdf34a1c4c3a24b83223f6e46e9aad89a9f39b2baf10","description":"The user has obtained an ID-Hubber role in a Litentry Discord channel","type":"Discord ID-Hubber Role Verification","tag":["Discord"],"assertions":[{"and":[{"src":"$verified_discord_account","op":">","dst":"0"},{"src":"$has_joined","op":"==","dst":"true"},{"src":"$discord_guild_id","op":"==","dst":"807161594245152800"}]}],"values":[false],"endpoint":"wss://tee-staging.litentry.io"}```
  * We can get rid of enpoint because there is issuer field. If we know who issues VC, then we should be able to find actual endpoint. In our Rust model we also have data_source but it's not used so it's not present in generated VCs
  * We can make descriptions less verbose.
[...]

PR #1941, opened 2023-07-25
remove endpoint from credential subject

kziemianek:
Removes endpoint from credential subject as this information is unnecessary.

relates #1861


BillyWooo (approved):
LGTM

kziemianek:
Let's keep this field because it might be needed in the future.

PR #1942, opened 2023-07-25
remove vc issuer fields

kziemianek:
Remove redundant VC's issuer fields.

Issuer field already contains AccountId of enclave which generated VC.
Mrenclave can be retrieved from parachain teerex pallet (storage or events).
I'm not sure about `name` as it serves as general information about what system generated this VC.

relates #1861


kziemianek:
Maybe we can put `litentry-worker:` in `id` so it will be obvious that it's issued by litentry worker identified by given `AccountId`, example: `litentry-worker:13d25960c4bcd91a691ddf9609d741ffe1653aab7146890af56e61bf4518c4b7`

BillyWooo (approved):
LGTM

kziemianek:
The decision is to keep it as it is.

PR #1952, opened 2023-07-26
Rename new_default to new

outofxxx:
### Context

This name `new_default` doesn't look suitable, rename it to `new`


Kailai-Wang (approved):
I'm fine with it

PR #1954, opened 2023-07-27
fix: remove SubstrateTestNet

### LINKED ISSUE #1918: Remove `SubstrateTestnet` from `Web3Network`
I think we don't need this network anymore and it can be safely removed from [here](https://github.com/litentry/litentry-parachain/blob/f63d9d6cb64b76c185448178c3a0d0cf907c32bb/primitives/core/src/network.rs#L57).

This would be a breaking change.

This type was used to determine network type for prime identity. Currently we do not rely on ss58 address format to determine network type.

felixfaisal:
Resolves #1918

### DIFF
--- a/primitives/core/src/network.rs
+++ b/primitives/core/src/network.rs
@@ -31,6 +31,8 @@ pub type BoundedWeb3Network = BoundedVec<Web3Network, ConstU32<MAX_WEB3NETWORK_L
 ///   Evm(EvmNetwork),
 /// }
 /// TODO: theoretically this should the the union of the supported networks of all data providers
+/// TODO: Using enum is not a great idea when using combined with scale codec, The ordering might
+/// cause inconsistencies
 #[derive(
 	Encode,
 	Decode,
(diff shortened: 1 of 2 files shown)


kziemianek (changes requested):
I think you need to remove occurence of this variant from other files too.

For example it is used in di examples 
https://github.com/litentry/litentry-parachain/blob/055b19d358089b0ed2f9571a23badbd82c638f6d/tee-worker/ts-tests/integration-tests/examples/direct-invocation/example.ts#L366

I think you should get chain metadata and rebuild ts types too.

primitives/core/src/network.rs
```diff
@@ -48,18 +50,16 @@ pub type BoundedWeb3Network = BoundedVec<Web3Network, ConstU32<MAX_WEB3NETWORK_L
...
 	// substrate
-	Polkadot,
-	Kusama,
-	Litentry,
-	Litmus,
-	LitentryRococo,
-	Khala,
-	SubstrateTestnet, // when launched it with standalone (integritee-)node
-
+	Polkadot = 0,
```
Kailai-Wang, line 53: Does the integer assignment bring anything? It's encoded by SCALE anyway

0xverin:
>For example it is used in di examples

see [code](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/integration-tests/examples/direct-invocation/example.ts#L366)

Kailai-Wang (commented):
And yes there should be more places where this network is used

primitives/core/src/network.rs
arieljbirnbaum: I take it the intent here is to preserve the SCALE encoding by keeping the integer value unchanged; not sure if that's something we want to mess with / rely on.

kziemianek: I think yes, and we should try to apply `#[codec(index = 15)]` to all enums in api contract definitions.

felixfaisal:
I will be closing this PR for now, because i don't want to poke around with `TS` code without understanding it atleast partially if not fully. As this is a low priority issues, it should be fine for now.

PR #1957, opened 2023-07-28
DI error responses

### LINKED ISSUE #1928: Display error details in synchronous DI responses
### Context

Follow-up of #1883.

Once #1920 gets merged, we should further adjust the error handling to include it in the DI responses, preferably in the `value` field: hex string of encoded error type. See also the comment in `trusted_call.rs`.

It should be at least tested via DI-sample. 

---

kziemianek:
* Feeds `WorkerRpcReturnValue` `value` with tee-worker error in case of STF errors.
* Adds standarized Response struct

fixes: #1928


tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -168,6 +168,35 @@ impl TrustedCall {
 		}
 	}
 
+	pub fn req_hash(&self) -> Option<&H256> {
```
kziemianek: Can be renamed to `call_id`. I think it's worth to abstract it a little bit and make `stf` less req/rpc aware.

tee-worker/core-primitives/stf-executor/src/executor.rs
```diff
@@ -136,14 +139,17 @@ where
 		) {
 			Err(e) => {
 				error!("Stf execute failed: {:?}", e);
-				return Ok(ExecutedOperation::failed(top_or_hash, extrinsic_call_backs))
+				let rpc_response_value: Vec<u8> = trusted_operation.req_hash().map(|h| {
```
kziemianek, line 142: @Kailai-Wang I think we can do the same to successful TOPs. TOPs can return result value that implements scalec's `Encode` and in case `req_has` is present encode it here and put it in `ExecutedOperation`. What do you think ?

This could lead us to removal of `https://github.com/litentry/litentry-parachain/blob/055b19d358089b0ed2f9571a23badbd82c638f6d/tee-worker/app-libs/stf/src/trusted_call.rs#L280`

tee-worker/core-primitives/stf-executor/src/error.rs
```diff
@@ -86,3 +88,9 @@ impl From<itp_node_api::metadata::Error> for Error {
 		Self::NodeMetadata(e)
 	}
 }
+
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
```
arieljbirnbaum: Doesn't it need to derive `TypeInfo`?

tee-worker/core-primitives/stf-executor/src/lib.rs
```diff
@@ -53,27 +53,28 @@ pub mod executor_tests;
...
-/// - additionally for the success case, we add the encoded rpc response that will be
-///   passed back to the requester when the top status is `InSidechainBlock`
+/// - we add the encoded rpc response that will be passed back to the requester
 /// - for failed top, we apply the parachain effects too
 #[derive(Clone, Debug, PartialEq)]
 pub enum ExecutionStatus {
-	Success(H256, Vec<OpaqueCall>, Vec<u8>),
-	Failure(Vec<OpaqueCall>),
+	Success(H256, Vec<OpaqueCall>, RpcResponseValue),
+	Failure(H256, Vec<OpaqueCall>, RpcResponseValue),
```
arieljbirnbaum, line 70: Two questions:

1. This is now isomorphic to `(H256, Vec<OpaqueCall>, RpcResponseValue, boolean)`; could/should we factor that out? (can be a named enum instead of a naked boolean ofc)
2. Is the success/failure value not included already (albeit encoded) in the `RpcResponseValue`?

tee-worker/ts-tests/integration-tests/identity-direct-invocation.test.ts
```diff
@@ -120,6 +121,15 @@ describe('Test Identity (direct invocation)', function () {
         */
         assert.isTrue(res.do_watch.isFalse);
         assert.isTrue(res.status.asTrustedOperationStatus[0].isInvalid);
+        assertWorkerError(
+            context,
+            requestIdentifier,
+            (v) => {
+                assert.isTrue(v.isLinkIdentityFailed);
```
arieljbirnbaum: :nail_care: for easier debugging, either go with

```
assert.equal(v.type, 'LinkIdentityFailed')
```

or print out the actual type in the message:

```
assert.isTrue(v.isLinkIdentityFailed, `expected LinkIdentityFailed, received ${v.type} instead`)
```

tee-worker/core-primitives/stf-executor/src/error.rs
kziemianek: Nothing is going to consume generated metadata so we don't need it.

tee-worker/core-primitives/stf-executor/src/lib.rs
kziemianek: 1. I prefer to keep it as it is because i like this top level distinction, it feels natural that you have success and failure execution status. Additionaly there is high possibility that in the future it would become nonisomorphic again.
2. even if it is, its hidden behind encoded bytes so there is no easy and cheap way to get access to it.

tee-worker/app-libs/stf/src/lib.rs
```diff
@@ -203,4 +204,13 @@ impl TrustedOperation {
 			_ => None,
 		}
 	}
+
+	pub fn req_hash(&self) -> Option<&H256> {
+		match self {
+			TrustedOperation::direct_call(c) => c.call.req_hash(),
+			TrustedOperation::indirect_call(c) => c.call.req_hash(),
+			//todo: getters should also contain req_hash ?
```
Kailai-Wang, line 214: Not now - but maybe later

tee-worker/core-primitives/top-pool-author/src/author.rs
```diff
@@ -344,8 +344,8 @@ where
 		self.process_top(ext, shard, TopSubmissionMode::SubmitWatch)
 	}
 
-	fn set_rpc_response_value(&self, rpc_responses_value: Vec<(TxHash<TopPool>, Vec<u8>)>) {
-		self.top_pool.set_rpc_response_value(rpc_responses_value)
+	fn set_rpc_response_values(&self, rpc_responses_values: Vec<(TxHash<TopPool>, Vec<u8>)>) {
+		self.top_pool.set_rpc_response_values(rpc_responses_values)
```
Kailai-Wang: I've been struggling if I should use singular or plural, but in the end I chose singular because a `Vec` is a singular 😂  it can have many elements but I refer `Vec` as the `value` (and it should be `response` not `responses` - that's my typo).

What do you think?

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: Yea I like the name `call_id` or simply `id` or `identifier`.
In my knowledge, `id` could mean `index` too which might lead to confusion as it is _normally_ implemented as an integer.

tee-worker/core-primitives/stf-executor/src/lib.rs
```diff
@@ -84,10 +85,17 @@ impl ExecutionStatus {
 		}
 	}
 
-	pub fn get_rpc_response_value(&self) -> Vec<u8> {
+	pub fn get_operation_hash(&self) -> Option<H256> {
```
Kailai-Wang: Seems we'll never get `None`? How about a plain `H256`?

tee-worker/core-primitives/stf-executor/src/error.rs
```diff
@@ -86,3 +88,9 @@ impl From<itp_node_api::metadata::Error> for Error {
...
+pub(crate) struct ErrorResponse<T: Encode> {
```
Kailai-Wang: do you think it's better to go to `trusted_call_rpc_response.rs`?

Kailai-Wang: And shall we always position `req_ext_hash` as the last field to be consistent with other `*Response` structs?

tee-worker/core-primitives/stf-executor/src/executor.rs
Kailai-Wang: Hmm I'm not sure if I get you. You can do this for `ErrorResponse` because the failed TOP has the error of the same type. The success TOP would have different Response types even though they are all encodable. How would you write the `Ok(res)` here? if you mean `res` is already encoded, I don't see much difference from the current impl..

Kailai-Wang (commented):
Looks good in general

tee-worker/core-primitives/stf-executor/src/executor.rs
kziemianek: They can be wrapped into enum (just like `StfError`) and handled in the same way.

tee-worker/core-primitives/top-pool-author/src/author.rs
kziemianek: If this is intentional then it's fine, i'll revert 🙃

tee-worker/core-primitives/stf-executor/src/error.rs
kziemianek: If we make `Response` generic then it will be enforced
```
pub(crate) struct Response<T: Encode> {
	pub req_ext_hash: H256,
	pub value: T,
}
```

tee-worker/core-primitives/stf-executor/src/executor.rs
kziemianek: @Kailai-Wang please check this [commit](https://github.com/litentry/litentry-parachain/commit/84de0ee5af337a9039c299828e719dd3e0eef66e)

tee-worker/app-libs/stf/src/trusted_call.rs
kziemianek: I'll use `identifier`

tee-worker/core-primitives/top-pool-author/src/author.rs
Kailai-Wang: Thanks! Can you do me a favour and remove the `s` in the middle? `rpc_responses_value ` => `rpc_response_value `

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -787,9 +804,7 @@ where
 				IMTCall::set_identity_networks { who, identity, web3networks }
 					.dispatch_bypass_filter(ita_sgx_runtime::RuntimeOrigin::root())
 					.map_err(|e| Self::Error::Dispatch(format!(" error: {:?}", e.error)))?;
-				let res = SetIdentityNetworksResponse { req_ext_hash: hash };
```
arieljbirnbaum, line 790: I think we need to keep this here for compatibility until clients (i.e. IDHub) are no longer relying on it to detect the response, no?

tee-worker/bin/integritee-cli
arieljbirnbaum, line 1: Is this supposed to be committed? o_O

tee-worker/ts-tests/integration-tests/substrate_ii_identity.test.ts
```diff
@@ -126,6 +127,18 @@ describe('Test Identity (direct invocation)', function () {
...
+                assert.isTrue(
+                    v.asLinkIdentityFailed.isUserShieldingKeyNotFound,
+                    `expected UserShieldingKeyNotFound, received ${v.asLinkIdentityFailed.type} instead`
+                );
+            },
+            res
+        );
 
         const events = await eventsPromise;
         await assertFailedEvent(context, events, 'LinkIdentityFailed', 'UserShieldingKeyNotFound');
```
arieljbirnbaum: Is this check passing? If so, please disregard my previous comment about backwards compatibility, I must have misunderstood the code :)

tee-worker/bin/integritee-cli
kziemianek: No, these binaries were renamed in the meanwhile so that's why they appeared in commit. Thanks!

tee-worker/app-libs/stf/src/trusted_call.rs
kziemianek: Marked pull request as breaking.
The clients will need to adapt, sadly there is no backward compatibility guaranteed :(

tee-worker/ts-tests/integration-tests/substrate_ii_identity.test.ts
kziemianek: Tests pass 🙃

tee-worker/bin/integritee-cli
Kailai-Wang, line 1: We don't need to commit this

tee-worker/bin/integritee-service
Kailai-Wang, line 1: Same as above

tee-worker/ts-tests/parachain-api/prepare-build/interfaces/identity/definitions.ts
```diff
@@ -206,37 +206,73 @@ export default {
...
+                DeactivateIdentityFailed: "(ErrorDetail)",
+                ActivateIdentityFailed: "(ErrorDetail)",
+                RequestVCFailed: "(Assertion, ErrorDetail)",
+                SetScheduledMrEnclaveFailed: "Null",
+                SetIdentityNetworksFailed: "(ErrorDetail)",
+                InvalidAccount: "Null",
+                UnclassifiedError: "Null",
+            },
+        },
+        Response: {
```
Kailai-Wang: I think the name `Response` is a bit too generic, do you have a better one?

Kailai-Wang (commented):
Since your `Response` always contains `req_ext_hash` as the first field, do you need to change the logic in `RpcResponder::update_status_event`?

Actually if you don't change that I don't expect the two-step trustedCalls to work 🤔

tee-worker/app-libs/stf/src/trusted_call.rs
arieljbirnbaum: Well if the tests are passing that seems to indicate that the Old Way (ie listening for the response on the parachain events) still works, no? :3

tee-worker/ts-tests/parachain-api/prepare-build/interfaces/identity/definitions.ts
kziemianek: @Kailai-Wang what do you think about `TrustedOperationResult` ?

kziemianek:
> Since your `Response` always contains `req_ext_hash` as the first field, do you need to change the logic in `RpcResponder::update_status_event`?
> 
> Actually if you don't change that I don't expect the two-step trustedCalls to work 🤔

Should work, at least tests pass. 

There is `TrustedCallResult::Streamed` which is represended by encoded `true` value so it should work as before.

```
			// true means that there are more results to come, see rpc_responder
			Self::Streamed => true.encode(),
```

How is this related to `req_ext_hash` ?

Kailai-Wang:
The problem is the `rpc_response_value`, which contains the req_ext_hash in the new impl. And this code snippet:

```
		// Litentry:
		// a trick to use `result.value` as the flag to forcily watch this connection.
		// If it's `true.encode()` it means we have streamed trustedCalls and there's more to come
		// TODO: how about relying on req_ext_hash? See comment in `trusted_call.rs`
		let do_watch = continue_watching(&status_update)
			|| (result.value == true.encode()
				&& matches!(
					result.status,
					DirectRequestStatus::TrustedOperationStatus(
						TrustedOperationStatus::InSidechainBlock(_),
						_
					)
				));
```

It seems `result.value == true.encode()` can never be true here as `value` is actually encoded `Response`

tee-worker/ts-tests/parachain-api/prepare-build/interfaces/identity/definitions.ts
Kailai-Wang: Maybe `TrustedOperationResponse`?  Another dumb question is why we use `Vec<u8>` for `value`, not `TrustedCallResult`? It seems in code that's the only source for encoding

kziemianek: I think because for some cases the `value` can hold encoded string

```
	pub fn from_error_message(error_msg: &str) -> Self {
		RpcReturnValue {
			value: error_msg.encode(),
			do_watch: false,
			status: DirectRequestStatus::Error,
		}
	}
```

kziemianek: In some cases it can hold encoded string error 

```
	pub fn from_error_message(error_msg: &str) -> Self {
		RpcReturnValue {
			value: error_msg.encode(),
			do_watch: false,
			status: DirectRequestStatus::Error,
		}
	}
```

Kailai-Wang: I see, thanks. So I need to decode it to different types based on the `status` field in the `RpcReturnValue`?

kziemianek:
> Since your `Response` always contains `req_ext_hash` as the first field, do you need to change the logic in `RpcResponder::update_status_event`?
> 
> Actually if you don't change that I don't expect the two-step trustedCalls to work 🤔

That's true. 
I though we have test case covering this.

tee-worker/app-libs/stf/src/trusted_call.rs
kziemianek: In terms of response channel - yes it works as before. 
Currently client can:

- in case of II listen to parachain events
- in case of DI wait for response on WS or listen to parachain events

kziemianek: By saying there is no backward compatibility i meant that RPC has slightly changed.

tee-worker/ts-tests/parachain-api/prepare-build/interfaces/identity/definitions.ts
kziemianek: Yes, not very convenient.

Kailai-Wang:
> > Since your `Response` always contains `req_ext_hash` as the first field, do you need to change the logic in `RpcResponder::update_status_event`?
> > Actually if you don't change that I don't expect the two-step trustedCalls to work 🤔
> 
> That's true. I though we have test case covering this.

The original web3 verification sample covered this.
But that's not the case anymore after #1989 is merged.

The two-step trustedcalls are:
- web2 verification
- vc request
So we need to have one of them to cover the case :(

kziemianek:
> > > Since your `Response` always contains `req_ext_hash` as the first field, do you need to change the logic in `RpcResponder::update_status_event`?
> > > Actually if you don't change that I don't expect the two-step trustedCalls to work 🤔
> > 
> > 
> > That's true. I though we have test case covering this.
> 
> The original web3 verification sample covered this. But that's not the case anymore after #1989 is merged.
> 
> The two-step trustedcalls are:
> 
> * web2 verification
> * vc request
>   So we need to have one of them to cover the case :(

True, i'm working on adding it.

kziemianek:
@Kailai-Wang there is already test verifying linking of web2 identity:
https://github.com/litentry/litentry-parachain/blob/385c54864e3e45f18621466e75c22806f3dad0f1/tee-worker/ts-tests/integration-tests/di_substrate_identity.test.ts#L48

tee-worker/core-primitives/stf-executor/src/executor.rs
```diff
@@ -313,3 +330,9 @@ fn into_map(
 ) -> BTreeMap<Vec<u8>, Option<Vec<u8>>> {
 	storage_entries.into_iter().map(|e| e.into_tuple()).collect()
 }
+
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub(crate) struct Response {
```
Kailai-Wang: Hmm would you like to rename this to `TrustedOperationResponse` or `TrustedOperationResult`?

tee-worker/core-primitives/stf-executor/src/executor.rs
```diff
@@ -136,19 +139,33 @@ where
...
-
-		if let StatePostProcessing::Prune = post_processing {
-			state.prune_state_diff();
-		}
+			Ok(result) => {
+				let rpc_response_value: Vec<u8> = trusted_operation.req_hash().map(|h| {
+					let encoded_result = result.get_encoded_result();
+					if encoded_result == true.encode() {
+						encoded_result
+					} else {
```
Kailai-Wang, line 156: Shall we add a comment regarding this special handling?

Kailai-Wang (approved):
Thanks!

PR #1978, opened 2023-08-02
change vc proof signature message

kziemianek:
Change a message to sign from VC's json to mrenclave value in order to create vc's proof signature

relates: #1861

### DIFF
--- a/tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
+++ b/tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
@@ -100,10 +100,7 @@ where
 		credential.credential_subject.set_endpoint(credential_endpoint);
 
 		credential.issuer.id = account_id_to_string(&enclave_account);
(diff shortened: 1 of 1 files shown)


outofxxx:
Changing to mrenclave does not reduce the size of the VC, and if issue. mrenclave is used, then if the signatures of all VC generated by the same worker are the same

kziemianek:
> Changing to mrenclave does not reduce the size of the VC, and if issue. mrenclave is used, then if the signatures of all VC generated by the same worker are the same

Hi, what are your concerns ?
VCs can still be only generated by registered enclaves and submitted to parachain vc registry only by them. There was an discussion [here](https://www.notion.so/web3builders/Review-VC-fields-d7132757fcdc4554b53b1f4ab93ab507)

Kailai-Wang (approved):
Looks good to me :)

kziemianek:
Merging it. @zhouhuitian if you have any concerns or doubts we can discuss suggested changes again.

PR #1980, opened 2023-08-03
fix: allow root to run the deploy script 

### LINKED ISSUE #1963: Allow root user to run the deploy scripts
### Context
The services in our internal server have been going down often. After investigation, we've identified the following: It looks like, the systemctl service with “--user” parameter is stopping all the services associated to that user (user developer in our case). User manager is limited until any session is active for a user (even after user’s logout). Once the session is closed, all the services started by the user will be stopped.  

As we are considering the server as a prod server, I strongly recommend starting a service with root user and not with system user.

### Task
Make a few tweaks to the scripts to run it as a root user, specifically the parts related 
- Storing the service files 
- Starting the service files 
- Target directories for the binaries 

### Acceptance criteria
We should be able to run the script on a local envrionment before we test it on `TEE-Internal`

felixfaisal:
Resolves #1963


tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -189,14 +213,24 @@ function restart_parachain() {
   local service_name="relay-alice"
   local description="Alice Node for Relay Chain"
   local working_directory="$TMPDIR"
-  local log_file=/tmp/parachain_dev/relay.alice.log
+  if [ "$ROOT" = true ]; then 
+    local log_file=/opt/parachain_dev/relay.alice.log
+  else 
+    local log_file=/tmp/parachain_dev/relay.alice.log
+  fi 
   local command="$POLKADOT_BIN --chain $ROCOCO_CHAINSPEC --alice --tmp --port ${AlicePort:-30336} --ws-port ${AliceWSPort:-9946} --rpc-port ${AliceRPCPort:-9936}"
```
Kailai-Wang: I think we have a hidden problem here:

we shouldn't use `--tmp`. We should use `--base-path` to persist the storage: imagine we want to restart the parachain (or the server simply reboots)

This applies to relaychains + parachain

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -205,29 +239,49 @@ function restart_parachain() {
...
   local service_name="para-alice"
   local description="Parachain Collator for Litenry Parachain"
   local working_directory="$TMPDIR"
-  local log_file=/tmp/parachain_dev/para.alice.log
+  if [ "$ROOT" = true ]; then 
+      local log_file=/opt/parachain_dev/para.alice.log
+  else 
+      local log_file=/tmp/parachain_dev/para.alice.log
+  fi 
   local command=
```
Kailai-Wang: is this one redundant?

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -189,14 +213,24 @@ function restart_parachain() {
...
+    local log_file=/opt/parachain_dev/relay.alice.log
+  else 
+    local log_file=/tmp/parachain_dev/relay.alice.log
+  fi 
   local command="$POLKADOT_BIN --chain $ROCOCO_CHAINSPEC --alice --tmp --port ${AlicePort:-30336} --ws-port ${AliceWSPort:-9946} --rpc-port ${AliceRPCPort:-9936}"
 
   generate_service_file "${service_name}" "${description}" "${command}" "${working_directory}" "${log_file}"
-  cp ./${service_name}.service ~/.config/systemd/user/
+  if [ "$ROOT" = true ]; then
+    cp ./${service_name}.service /etc/systemd/system/
```
Kailai-Wang: Did you test it? 

I believe `/etc/systemd/system/` needs the root, I don't see you using `sudo` here

Kailai-Wang (changes requested):
Actually, I'm a bit inclined to only keeping the root systemd variant.

It's a deployment script, I can't think of a use case where a specific user needs to deploy this (and gets terminated when the user logs out) - if there's such requirement, then using `./local-setup/launch.py` might be a better solution.

What do you think?

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I see, I was using the same commands as `launch-local-binary.sh`, I will update and test this, but I have another questions, if it persists the data, do you have an idea where it stores it, and does it also mean if we restart with the persisted data we don't have to `regsiter_parachain` again?

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: This is redundant, will update it

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I assumed when we call the script with `root` user, the caller will have permissions to do this, Let me know what you think.

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: > and does it also mean if we restart with the persisted data we don't have to regsiter_parachain again?

yes.

the stored data is exactly under `--base-path`
We need 3 different folders for 3 dbs: 2 relaychains + 1 parachain

**We should test** it (e.g. restarting the parachain service) after it's deployed to make sure the data is actually persisted.

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: Do you mean `sudo ./deploy.sh` - that's fine too

But as I wrote:
> It's a deployment script, I can't think of a use case where a specific user needs to deploy this (and gets terminated when the user logs out) - if there's such requirement, then using ./local-setup/launch.py might be a better solution.

So it means we should always execute it as root (thus we can keep the `root = true` branch only) What do you think?

felixfaisal: I'm also inclined towards having a single branch which is called by `root` ,  That would mean if anyone wants to deploy on servers, they should always have `root` permission. I like this, but my worry is that then we should give `root` permission to anyone who wants to deploy, if that's fine, then I guess we could have single branch.

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I like this, This makes the system to be able to get back on server reboots and the data persists, so we don't have to call any scripts. It will take me some time to test this, but this should be a good improvement.

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: Yea that should be solved at the account permission level - I'm fine for now

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: Thanks! We have to be time conscious too - ideally we can get it done today

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: No longer branching

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Single branch for root use only now

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Included `--base-path` , also tested restarting all together, It worked fine ✅  , the chain data will be purge only if we restart using the `--discard` flag, and  we reset the worker without losing the data on chain

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -55,14 +55,13 @@ generate_service_file() {
 Description=${description}
 
 [Service]
+ExecStartPre=/bin/mv ${log_file} ${log_file}-backup
```
Kailai-Wang: Can it error out? Shall we make sure `log_file` exists before we `mv`?

Kailai-Wang: Like

```
[ -d "$log_file" ] && /bin/mv "${log_file}" "${log_file}-backup" 
```

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -72,18 +71,21 @@ Description=${description}
 After=network.target
 
 [Service]
+ExecStartPre=/bin/mv ${log_file} ${log_file}-backup
```
Kailai-Wang: same as above

Kailai-Wang (approved):
Thanks. I think we might need some cleanups / refactoring afterwards.

E.g. maybe it will look better if have clear service template files, like
- relaychain.service.template
- parachain.service.template
- worker.service.template

But let's do it later. I'm happy to approve it as long as it works.

**Please also mind the name difference if you want to use this script on staging.**

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Done ✅

Kailai-Wang:
It's already approved from my side :) please mind the name changes on staging and test it out once it's not used for stress tests

PR #1981, opened 2023-08-03
vc did ids

kziemianek:
Use DID format for credentialSubject and issuer ids

relates #1861


Kailai-Wang (approved):
Looks good to me - so now both the credential.subject.id and issuer.id are in did form

PR #1984, opened 2023-08-03
Use consistent types in ts-test #1917

### LINKED ISSUE #1917: Use consistent types in ts-test
### Context

A (trivial) example is we should maybe prefer `HexString` over `0x${string}` - and they need to be consistent.

It applies to other types that we should use from a single source of truth.

---

songzhouran:
### Context
related #1917


arieljbirnbaum:
Good change; thanks @songzhouran :raised_hands: 
Please do not close issue #1917 though; I'm sure there are plenty more inconsistencies like this one lurking around :P

arieljbirnbaum (approved):

Issue #1916, opened 2023-07-21
Abort VC building in case of data provider errors


Kailai-Wang:
### Context

It seems that we continue building VCs even when a data provider error is present. This would lead to wrong results in VC.

We should abort the process instead.

Can we have a test case for it?

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

BillyWooo:
related issue: https://github.com/litentry/litentry-parachain/issues/1870
The refactor may affect the logic/structure. So take this into consideration.

Issue #1826, opened 2023-06-27
Create new assertions based on current new Achainable API (tag4: DeFi)


outofxxx:
### Context
Our data provider TDF's product [Achainable](https://achainable-stage.vercel.app/?achainableLabels=true&userLabels=true&searchText=&chain=All+chains&tag=All+tags&view=list) has developed new API to query data. There are quite many labels provided by Achainable. They are grouped by tags: Account, Balance, Dotsama, DeFi, etc.

Task
Covert existed labels into assertions. Tag: DeFi

### Acceptance criteria
What's the definition of done, if applicable?

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

hanwencheng:
What is the follow up for the new assertions?

Kailai-Wang:
Is this done? @zhouhuitian

outofxxx:
There will be a [new issue](https://github.com/litentry/litentry-parachain/issues/1944) to track the specific new assertions to be made. According to @BillyWooo 's suggestion, these Tag-series issues (including https://github.com/litentry/litentry-parachain/issues/1718 / https://github.com/litentry/litentry-parachain/issues/1719 / https://github.com/litentry/litentry-parachain/issues/1720) will be temporarily retained and will be discussed later.

BillyWooo:
As discussed [here](https://github.com/litentry/litentry-parachain/pull/1986#issuecomment-1667263786), we can close this issue for the moment. If further new assertion requirement comes, create new issue.

Issue #1718, opened 2023-05-22
Create new assertions based on current new Achainable API (tag1: Account)


BillyWooo:
### Context
Our data provider TDF's product [Achainable](https://achainable-stage.vercel.app/?achainableLabels=true&userLabels=true&searchText=&chain=All+chains&tag=All+tags&view=list) has developed new API to query data. There are quite many labels provided by Achainable. They are grouped by tags: Account, Balance,  Dotsama, Bridge, etc.

### Task
Covert existed labels into assertions. Start from tag: Account.

### Acceptance criteria
What's the definition of done, if applicable?

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

outofxxx:
Pay attention to the naming issue when replacing the new API, refer to the discussion here

https://github.com/litentry/litentry-parachain/pull/1721

Kailai-Wang:
Have we done everything? Can we close this issue (and its siblings)?

kziemianek:
@BillyWooo / @zhouhuitian what's the status of this issue ?

BillyWooo:
As discussed [here](https://github.com/litentry/litentry-parachain/pull/1986#issuecomment-1667263786), we can close this issue for the moment. If further new assertion requirement comes, create new issue.

PR #1986, opened 2023-08-04
Add new achainaable assertions

### LINKED ISSUE #1944: create new Ethereum related assertions
### Context
After refactor of Achainable API with SystemLabel, we should be able to create some more new assertions. 
Ethereum:
- class of xxxx (year)
- ETH Holder (only check balance)
- Contract creator
- Uniswap User V2/V3

### Task
Create new assertions list as above (at least three of them)

### Related Issue

#1718 #1719 #1720 #1826 

### Acceptance criteria
What's the definition of done, if applicable?

---

### LINKED ISSUE #1971: Update assertions info
Update the assertions information based on Kayla's latest documentation.

https://docs.google.com/spreadsheets/d/1k2ZvrLK5v_KO_tsCWJVtjbs8TS8PzM_idWv64z4ynSI/edit#gid=1065809408

### LINKED ISSUE #1972: Update class of year assertion
Because the class of year can obtain the specific account creation time, after discussing with Kayla, the modified requirements are as follows: Return year in VC

`true`
```
"assertions":[
{
	"and":[
		{
			"src":"$account_created_year",
			"op":"==",
			"dst":"2015"
		}
	]
}
```

`false`
```
"assertions":[
{
	"and":[
		{
			"src":"$account_created_year",
			"op":"==",
			"dst":"invalid"
		}
	]
}
```

### LINKED ISSUE #1973: Add LIT/DOT holder assertion
Add 2 new holder assertions, as detailed in 
https://docs.google.com/spreadsheets/d/1k2ZvrLK5v_KO_tsCWJVtjbs8TS8PzM_idWv64z4ynSI/edit#gid=1065809408

* LIT holder
* DOT holder

outofxxx:
This PR mainly includes:
* Refactor&impl 11 categories of accessible assertions using `system_labels`
* Add 6 new assertions
* other small updates

resolves #1944 
resolves #1972 
resolves #1973 
resolves #1971 

---

NOTE:

1. Achainable APIs
Some APIs cannot meet the current assertion requirements, so there are some trade-offs in calling APIs. For example:
a.  the total transaction of A8 is obtained by parsing the display field to obtain txns, and 
b. the class of year is obtained by parsing metadata to obtain the year when the account was created. 
According to TDF developers, these  `unstable` fields may need to be removed in the future, so it is important to focus on them.

2. Achinable New assertions
Although there are 11 categories of assertion implemented, 6 of the PR implementations are distributed among 3 categories, and the rest currently do not implement VC content (this new part needs to be determined by Kayla later).
2.1 [API interface](https://www.notion.so/web3builders/Assertion-interface-9126ba85a925417a922f2c6ae5d62e87)
2.2 [Achainable supported names](https://www.notion.so/web3builders/Achainbale-names-160d0ba501474b4daa15193155e121ed)
2.3 [Achainbale supported chains](https://www.notion.so/web3builders/Achainable-chains-1b7c253302e14af3a7c7926c1c7c8508)
2.4 [Achainbale supported token address](https://www.notion.so/web3builders/Achainable-token-address-bd2bc18e9cad4727ab12867d8234d714)

- [x] Reconfirm with Kayla regarding the entire content of each 6 VCs
  * [assertion collection](https://www.notion.so/web3builders/Assertions-collection-06c823bd0a63411b84628048c0729dd0)


tee-worker/litentry/core/assertion-build/src/a4.rs
```diff
@@ -72,7 +72,7 @@ use std::string::ToString;
 
 const VC_A4_SUBJECT_DESCRIPTION: &str =
 	"The length of time a user continues to hold a particular token (with particular threshold of token amount)";
-const VC_A4_SUBJECT_TYPE: &str = "LIT Holding Assertion";
+const VC_A4_SUBJECT_TYPE: &str = "LIT Holding Time";
```
Kailai-Wang, line 75: Will it break IDHub or at least require their attention?

tee-worker/litentry/core/assertion-build/src/achainable/achainable_amount.rs
Kailai-Wang, line 1: I do believe when the folder is already called `achainable`, it seems a bit redundant to have file names with `achainable` prefixes.

Maybe `achainable/amount.rs` would be better?

It's not strictly required in this PR though

tee-worker/litentry/core/assertion-build/src/achainable/achainable_amounts.rs
```diff
@@ -0,0 +1,88 @@
...
+extern crate sgx_tstd as std;
+
+use crate::{achainable::request_achainable, *};
+use lc_data_providers::{
+	achainable::{Params, ParamsBasicTypeWithAmounts},
+	vec_to_string,
+};
+
+const VC_SUBJECT_DESCRIPTION: &str = "Class of year";
+const VC_SUBJECT_TYPE: &str = "ETH Class of year Assertion";
```
Kailai-Wang: This is confusing IMO - this VC is about the "class of year", but it's named `amounts.rs` :(

I can understand it's using the `Amounts` API from achainable, but still...

tee-worker/litentry/core/assertion-build/src/achainable/achainable_amounts.rs
```diff
@@ -0,0 +1,88 @@
...
+		.into_iter()
+		.flat_map(|(_, addresses)| addresses)
+		.collect::<Vec<String>>();
+
+	let _flag = request_achainable(addresses, Params::ParamsBasicTypeWithAmounts(p.clone()))?;
+
+	match Credential::new(&req.who, &req.shard) {
+		Ok(mut credential_unsigned) => {
+			credential_unsigned.add_subject_info(VC_SUBJECT_DESCRIPTION, VC_SUBJECT_TYPE);
+			// credential_unsigned.add_achainable(flag, date1, date2);
```
Kailai-Wang: please remove the line if not needed

tee-worker/litentry/core/assertion-build/src/achainable/achainable_amounts.rs
```diff
@@ -0,0 +1,88 @@
...
+		},
+	}
+}
+
+fn parse_amounts_params(param: &AchainableAmounts) -> Result<(String, String, String)> {
+	let name = param.name.clone();
+	let amount1 = param.amount1.clone();
+	let amount2 = param.amount2.clone();
+
+	let name = vec_to_string(name.to_vec()).map_err(|_| {
```
Kailai-Wang: I see `ParameterString` to `String` is widely used.
Shall we have a `From` or `TryFrom` impl for it?

tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+	let mut flag = false;
+	for address in &addresses {
+		if flag {
+			break
+		}
+
+		let is_v2 = client.uniswap_v2_user(address).unwrap_or_default();
+		let is_v3 = client.uniswap_v3_user(address).unwrap_or_default();
+
+		flag = is_v2 || is_v3
```
Kailai-Wang: It means even if `is_v2` is true, we'll send requests to query v3 result (which doesn't have an impact on the `flag` anyway), this is redundant and I doubt if the compiler would optimise this.

tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	Ok(flag)
+}
```
Kailai-Wang: What about this, is this equivalent?

```suggestion
	for address in &addresses {
		if client.uniswap_v2_user(address).unwrap_or_default() || 
	        client.uniswap_v3_user(address).unwrap_or_default() {
	        return Ok(true)
	     }
	}
	Ok(false)
}
```

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -935,67 +1016,67 @@ impl AchainableTagDeFi for AchainableClient {
...
-		let chain = "ethereum";
+		let chain: Web3Network = Web3Network::Ethereum;
 		let r1 =
-			request_basic_type_with_token(self, address, name, chain, None).unwrap_or_default();
+			request_basic_type_with_token(self, address, name, &chain, None).unwrap_or_default();
 
 		// Uniswap V2 liquidity provider
 		let name = "Uniswap V2 liquidity provider";
-		let r2 = request_basic_type_with_token(self, address, name, chain, None)?;
+		let r2 = request_basic_type_with_token(self, address, name, &chain, None)?;
```
Kailai-Wang: I think this `Ok(r1 || r2)` has the same problem as above - we could send out more requests than we actually need.

Kailai-Wang (commented):
I took a very very rough look and didn't complete the review yet, but maybe it makes sense to address some general problems first before we continue, like the goal and code style.

Is this PR aimed to convert the achainable tags to the assertions using system labels, or to implement the 6 new assertions, or both? If both, then it should have been split into _at least_ 2 PRs and we'll have to adjust the planning too (because we planned 3 - 4 days for the new assertion implementation only).

Either way I believe this PR contains too much (mixed) stuff which makes it hard to review too. I'm actually fine if a PR only contains one single category (e.g. `achainable_basics`), it seems that it's not so inseparable from others ...

tee-worker/litentry/core/data-providers/src/lib.rs
```diff
@@ -68,7 +68,7 @@ pub mod twitter_official;
 
 const TIMEOUT: Duration = Duration::from_secs(3u64);
 
-pub const LIT_TOKEN_ADDRESS: &str = "0xb59490aB09A0f526Cc7305822aC65f2Ab12f9723";
+pub const LIT_TOKEN_ADDRESS: &str = "0xb59490ab09a0f526cc7305822ac65f2ab12f9723";
```
kziemianek, line 76: ethereum addresses should be case insensitive. 
If achainable is case sensitive in this matter, we should remain address the same and instead call `to_lower_case` on it in achainable data-provider as a more complete solution. This will eliminate the risk of this happening again.

BillyWooo: Yes. This case sensitive topic should be taken care in a bigger round. We should in general check all address case sensitive topic.

BillyWooo:
After this PR got merged, these following issues need further discussion:
#1826 
#1813 
#1720 
#1719 
#1718 
#1916

BillyWooo (approved):
After our discussion and amendment, I think it looks good to me.

Kailai-Wang:
Please let's make sure:
- this PR gets into staging too
- metadata/types are updated - it's absolutely needed

PR #1988, opened 2023-08-04
Hotfix lit token address

outofxxx:
as title.


arieljbirnbaum (approved):
:ship:

kziemianek (approved):
it shouldn't be case sensitive.

If this is specific to achainable we should remain address unchanged and use `to_lowercase` to prevent such problems in the future.

kziemianek:
Do we need fix on `dev` as well ?

arieljbirnbaum:
@kziemianek the change is included in #1986 which targets `dev`

PR #1989, opened 2023-08-04
Verify web3 identity in the main thread

### LINKED ISSUE #1979: Verify web3 identities in the main thread
### Context

It seems we don't need to send the verification request over to another thread for web3 identities. We have everything we need and there's no network access required.

We should also have some cleanups about the request fields.

Tasks:
- [ ] verify web3 identities inside `trusted_call_litentry.rs`
- [ ] perform sanity checks and fail early (e.g. type mismatch between identity and validation_data)
- [ ] review the fields of `IdentityVerificationRequest`: maybe we can calculate the message before sending

---

Kailai-Wang:
### Context

resolves #1979 
I hope it improves the performance a bit


tee-worker/app-libs/stf/src/helpers.rs
```diff
@@ -146,3 +148,29 @@ pub fn get_expected_raw_message(
...
+	raw_msg: &[u8],
+	data: &Web3ValidationData,
+) -> StfResult<()> {
+	ensure!(
+		raw_msg == data.message().as_slice(),
+		StfError::LinkIdentityFailed(ErrorDetail::UnexpectedMessage)
+	);
+
+	// TODO: just to make it backwards compatible
+	//       will merge it to `VerifyWeb3SignatureFailed` after the campaign
```
arieljbirnbaum, line 163: Link to followup issue please? :)

BillyWooo (approved):
LGTM.
This commit also need to apply to `staging` branch, right? @Kailai-Wang

Kailai-Wang:
Not going to do that - especially after I see the web3 identity linking is not the bottleneck according to the stress test.

PR #1998, opened 2023-08-08
fix: allow multiple system with offset ports 

### LINKED ISSUE #1974: Support offset in binary mode
### Context
Currently, It's not possible to run the system in multiple ports  in binary because of the use `/tmp/parachain_dev` hardcoded at certain places. So if Alice starts a system with binary mode, and Bob wants to start a system with offset ports on the same machine, he will not be able to access `/tmp/parachain_dev` resulting in an error. 

### Task
We should set the folder for Parachain either via `Envrionment Variables` or `CLI Params`.  Also while starting the script should update the `path` to `genesis-state` and `genesis-wasm` in `config.ci.json` , so that scripts related to `registering parachain` can be run without a problem. 

### Acceptance criteria
Two people should be able to run the system in `local-binary` mode in the same machine. 

---

felixfaisal:
Resolves #1974

### DIFF
--- a/scripts/launch-local-binary.sh
+++ b/scripts/launch-local-binary.sh
@@ -25,7 +25,15 @@ CHAIN=$1
 POLKADOT_BIN="$2"
 PARACHAIN_BIN="$3"
 
-TMPDIR=/tmp/parachain_dev
+# Check if TMPDIR is already set
+if [ -z "$TMPDIR" ]; then
+    TMPDIR="/tmp/parachain_dev"
+    export TMPDIR
+    echo "TMPDIR was not set. Assigned value: $TMPDIR"
+else
+    echo "TMPDIR is already set to: $TMPDIR"
+fi
+
 [ -d "$TMPDIR" ] || mkdir -p "$TMPDIR"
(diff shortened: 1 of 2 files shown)


scripts/launch-local-binary.sh
```diff
@@ -25,7 +25,15 @@ CHAIN=$1
...
-TMPDIR=/tmp/parachain_dev
+# Check if TMPDIR is already set
+if [ -z "$TMPDIR" ]; then
+    TMPDIR="/tmp/parachain_dev"
+    export TMPDIR
+    echo "TMPDIR was not set. Assigned value: $TMPDIR"
+else
+    echo "TMPDIR is already set to: $TMPDIR"
+fi
+
```
Kailai-Wang: Would `TMPDIR="${TMPDIR:-/tmp/parachain_dev}"` work?

Kailai-Wang (approved):
Thanks, a small comment

scripts/launch-local-binary.sh
felixfaisal: That would be better 😅

PR #2003, opened 2023-08-08
Abort VC building in case of data provider errors

### LINKED ISSUE #1916: Abort VC building in case of data provider errors

BillyWooo:
As title,
When querying data from data provider and connection error happening, vc building process will return immediately / directly.

May also need to be cherry-picked into `staging` branch.

resolved #1916


outofxxx (approved):
looks good

Kailai-Wang (approved):
Thanks

PR #2004, opened 2023-08-08
Update Cargo.lock for `dev`

Kailai-Wang:
### Context

A minor PR to update the `Cargo.lock` in the `dev` branch - the mismatch might come from the dependabot.


Kailai-Wang:
Nvm it seems to be updated already, closing it now

PR #2005, opened 2023-08-09
Handle achainable assertion inputs

outofxxx:
1. Achainable assertion  return immediately / directly when request error occurred. related https://github.com/litentry/litentry-parachain/pull/2003 
2. When querying Uniswap users, verify the already agreed name value. If the input is incorrect, return error directly
`name: "Uniswap V2/V3 user"`


BillyWooo (approved):
thanks. LGTM

BillyWooo:
Let's also cherry-pick it to `staging` branch

PR #2008, opened 2023-08-09
Read log level from config file

outofxxx:
as title.


tee-worker/local-setup/launch.py
```diff
@@ -164,6 +167,28 @@ def setup_environment(offset, config, parachain_dir):
             for flag in config["workers"][0]["flags"]
         ]
 
+def setup_worker_log_level():
```
Kailai-Wang: Maybe it's better to take the path as parameter

tee-worker/local-setup/worker-log-level-config.toml
```diff
@@ -0,0 +1,37 @@
+[tee-worker]
```
Kailai-Wang: maybe easier section name: like `[worker]` and `[cli]`
or `[litentry-worker]` and `[litentry-cli]`

Kailai-Wang (approved):
I like this change, thanks 🙌

tee-worker/local-setup/launch.py
```diff
@@ -164,6 +167,28 @@ def setup_environment(offset, config, parachain_dir):
...
+        log_data = toml.load(f)
+
+        # Section
+        for (section, item) in log_data.items():
+            log_level_string = "";
+            indx = 0
+
+            for (k, v) in item.items():
+                if indx == 0:
+                    log_level_string += v+","
```
kziemianek, line 182: This means `RUST_LOG` should always be first ? Maybe the script could check value of the key instead of idx ?

tee-worker/local-setup/launch.py
BillyWooo: Yeah, much better to set it as parameter. @zhouhuitian

PR #2010, opened 2023-08-10
Update uniswap v2/v3 user assertion logic

outofxxx:
As discussed in [channel](https://litentry.slack.com/archives/C03KR96989G/p1691663053824849?thread_ts=1691619239.842269&cid=C03KR96989G)

```
        and: {
        {
            src:$is_uniswap_v2_user,
            op:==,
            dst:true
        },
        {
            src: is_uniswap_v3_user,
            op: ==,
            dst: true
         }
        }
value: [true]
```

Example: [uniswap-v2/v3-user](https://www.notion.so/web3builders/UniswapV2-3-user-e6cbb0bcffde4ba89c3b6863b9e799db)


BillyWooo (approved):
LGTM. We discussed together about this fix.
As time is limited here for testing, I would simply bypass branch protection this time.

Issue #1878, opened 2023-07-12
Check if manual definitions are really needed in parachain-api TS package


arieljbirnbaum:
### Context

Following up from [this discussion](https://github.com/litentry/litentry-parachain/pull/1835/files#r1260761816).
Ideally, all type definitions for the parachain and sidechain API packages should be generated from the chain metadata via the typegen scripts. However, we still seem to need manual definitions for the `identity` module in the parachain API.

### Task
- Check if the manual definitions are _really_ necessary, or if we can do away with them.
- If we cannot get rid of them, document _why_ they are needed, and set a procedure to keep these up to date
  - If this is the case, we should consider making these API packages available to clients (e.g. IDHub) so they can stay up to date as well :)

Kailai-Wang:
> we should consider making these API packages available to clients

Yes we need to start publishing packages for clients.

Btw, do we need to do this even for the metadata-generated types? I mean do clients know how to update the metadata and types, **which match a parachain release?**

Kailai-Wang:
The type usage in the current ts-test is a mess 😂 so IMO this is an important task, at least _medium_, maybe even _high_

arieljbirnbaum:
@Kailai-Wang 

> Yes we need to start publishing packages for clients.
> 
> Btw, do we need to do this even for the metadata-generated types? I mean do clients know how to update the metadata and types, **which match a parachain release?**

_In theory_ each client should be able to do this process on their own, based only on the chain metadata. In practice, it can be a big [pain](https://github.com/litentry/litentry-parachain/pull/1835) to set up, as evidenced by the time @Verin1005 and I have sunk into this just now (not to mention the work the IDHub colleagues have had to do to do their own version of this integration). Hence I think it would still be more helpful to clients if we "centralize" that pain in one location :D

Kailai-Wang:
Prioritise this - as we have more and more needs to use the correct types.

Will/IDHub colleagues asked about this. I noticed it when I wrote the code myself.

The longer we push back the issue, the more effort we'll need to take to rebuild/refactor everything when this issue is solved. 

We should start to use generated types as early as possible.

0xverin:
>Task
Check if the manual definitions are really necessary, or if we can do away with them.
If we cannot get rid of them, document why they are needed, and set a procedure to keep these up to date
If this is the case, we should consider making these API packages available to clients (e.g. IDHub) so they can stay up to date as well :)

[...]

PR #2015, opened 2023-08-12
EthereumPrettified signature for trusted calls

arieljbirnbaum:
### Context

Mitigates the issue of ["scary" garbled signature requests for trusted calls](https://litentry.slack.com/archives/C03KR96989G/p1691722509960989), by adding a new variant to the `LitentryMultiSignature` type with a suitably modified payload.

For the frontend counterpart, see https://github.com/litentry/identity-hub/pull/1050.
This is kept as a draft until both PRs are ready to go.

### Caveats

This is not a comprehensive solution, in the sense that the signed message is still not intelligible to the user (i.e., they don't _really_ know what they're signing).

### Details

The current `Ethereum` signature scheme works as follows:

- Call gets SCALE-encoded into Payload Bytes
- Payload Bytes shown in Sign dialog; user confirms
- Payload bytes (with EIP-191 prefix*) are signed

(* some wallet extensions don't add it yet)

The new `EthereumPrettified` scheme adds a slight modification:

- Call gets SCALE-encoded; *then hex-encoded* to avoid non-printable characters  
- A *fixed prefix*, `Litentry authorization token: 0x` gets prepended
  - This is displayed with the message to give the user a bit more context.
- The combined payload gets UTF-8 encoded into the Payload Bytes 
- Payload Bytes shown in Sign dialog; user confirms (as before)
- Payload bytes (with potential EIP-191) are signed (as before)

### TODO

- [ ] Pass CI
- [ ] Open followup tickets as required
- [ ] After merging, cherrypick into `staging`


arieljbirnbaum:
@Kailai-Wang should/can we enable CI runs for PRs open against `staging`?
Started a job [manually](https://github.com/litentry/litentry-parachain/actions/runs/5841662705), in any case.

Kailai-Wang:
CI won't work for staging - the gap is too big.

Is it feasible to submit a PR to `dev` and then try to cherry-pick the change to `staging`?

tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -108,23 +111,39 @@ impl LitentryMultiSignature {
...
+	sig: &EthereumSignature,
+	who: &Address20,
+	msg: &[u8],
+	signer: &Address20,
+) -> bool {
+	let digest = keccak_256(data);
+	return match recover_evm_address(&digest, sig.as_ref()) {
+		Ok(recovered_evm_address) => recovered_evm_address == who.as_ref().as_slice(),
+		Err(_e) => {
+			error!("Could not verify evm signature msg: {:?}, signer {:?}", msg, signer);
```
Kailai-Wang: I think `msg` is actually `data` here - so it's redundant

Kailai-Wang: `signer` and `who` are equivalent too

tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -108,23 +111,39 @@ impl LitentryMultiSignature {
...
-						false
-					},
-				}
+				let data = msg;
+				return verify_evm_signature(evm_eip191_wrap(data).as_slice(), sig, who, msg, signer)
+					|| verify_evm_signature(data, sig, who, msg, signer)
+			},
+			(Self::EthereumPrettified(ref sig), who) => {
+				let user_readable_message =
+					"Litentry authorization token: 0x".to_string() + &hex::encode(msg);
```
Kailai-Wang: Maybe we can consider using this: https://github.com/litentry/litentry-parachain/blob/aa12180696bb51505d98415aa32e711b7275f131/tee-worker/core-primitives/utils/src/hex.rs#L52

essentially the same but we try to keep the source of truth consistent

Kailai-Wang (commented):
Looks good, some small comments

tee-worker/litentry/primitives/src/lib.rs
arieljbirnbaum: Fixed; thanks :raised_hands: 
(do we need to log an `error!` on every unmatched signature check? we're bound to have at least some for non-eip-191-compliant wallets :thinking:)

tee-worker/litentry/primitives/src/lib.rs
arieljbirnbaum: Tried that earlier but couldn't get it to work for some reason; gave it another go and it worked like a charm this time :raised_hands: Thanks!

tee-worker/ts-tests/parachain-api/dist/src/index.js
```diff
@@ -1,17 +1,15 @@
 import "@polkadot/api/augment";
 import "@polkadot/types-augment";
 import { ApiPromise, Keyring, WsProvider } from "@polkadot/api";
-
-// Before we finalize the complete submission of the 'dist' files, we'll temporarily use the 'prepare-build' path here. Otherwise, CI can't find the files.
-import { identity } from "../../prepare-build/interfaces/definitions";
```
Kailai-Wang, line 6: This is the reason that CI failed.

According to @Verin1005 we should not touch it (or maybe manually reverted to the old version).

I had once similar problem where this file was changed automatically - I haven't figured out which step caused it

tee-worker/ts-tests/parachain-api/dist/build/interfaces/augment-api-consts.d.ts
Kailai-Wang, line 1: I believe the whole folder `dist/build/interfaces` is not checked into the repo?

Kailai-Wang (commented):
The logic looks fine, thanks

tee-worker/litentry/primitives/src/lib.rs
Kailai-Wang: Actually nope - we should log it when the we tried every variant and it still fails, but this is a minor thing :D

tee-worker/ts-tests/parachain-api/dist/src/index.js
arieljbirnbaum: That import is listed in the source for this lib; some of the test code is referencing the definitions directly (don't remember which exactly) so they should come from this package as a source of truth.

Problem was, while we relied on the `dist` folder being committed, the `dist/build` was gitignored, so without building the missing files, the committed build output is actually broken.

I want to sort this out properly along the course of #1878; in the meantime I just fixed the `.gitignore` so that if we commit `dist`, that means _all_ of `dist`.

tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -67,6 +68,8 @@ pub enum LitentryMultiSignature {
 	Ecdsa(ecdsa::Signature),
 	/// An ECDSA/keccak256 signature. An Ethereum signature. hash message with keccak256
 	Ethereum(EthereumSignature),
+	/// Same as the above, but the payload bytes are hex-encoded and prepended with a readable prefix
+	EthereumPrettified(EthereumSignature),
```
kziemianek, line 72: We could have another varaint for EIP-191 too 😂

kziemianek (commented):
Rust code looks fine 👍🏼  just one small comment because we might not be consistent.

tee-worker/litentry/primitives/src/lib.rs
arieljbirnbaum: Yeah, thought about that :joy: Was trying to keep the scope narrow here for the sake of the release; happy to follow up on it later tho.

Kailai-Wang: That's true - but let's get it over the line and launch the campaign first 😂

kziemianek: Might not be that easy as we will get nested variants Prettified EIP 🤔  ?
Approved as this is not blocking for me.

Issue #1847, opened 2023-07-05
EVM support of parachain


wangminqi:
After a careful discussion, we decide to support EVM in our parachain (at leaset from substrate aspect).
The true challenge of full compatible with Ethereum ecosystem tool is hard, and will not include in this issue for now.

We plan to have it together with a listener project before the end of this year.

hanwencheng:
I and Minqi will provide further details for the listener project later

wangminqi:
https://www.notion.so/web3builders/Contract-Event-Schema-V2-to-public-03a30886f8564f0cb7fd653fd76eeb50

https://www.notion.so/web3builders/Decentralized-Monitoring-Collator-Network-DMCN-v2-2f78eb0b0df242b7bf65a6f6eb00abbf

Kailai-Wang:
This is done

PR #2016, opened 2023-08-14
1847 evm support

wangminqi:
This PR including the implementation and integration test of EVM module for parachain (Rococo only).
Related issue: #1847 .


ts-tests/yarn.lock
wangminqi, line 1: Must use corepack yarn install to generate this file.
Otherwise, "YN0028: The lockfile would have been created by this install, which is explicitly forbidden" issue

node/src/evm_tracing_types.rs
```diff
@@ -0,0 +1,108 @@
+// Copyright 2020-2023 Litentry Technologies GmbH.
```
kziemianek: We should use new entity name here https://github.com/litentry/litentry-parachain/issues/2086

wangminqi: I see. I will update new files accordingly.

.github/workflows/ci.yml
```diff
@@ -459,9 +459,25 @@ jobs:
...
+          tail: all
+          dest: docker-logs
+
+      - name: Upload logs
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: docker-logs-${{ matrix.chain }}
+          path: docker-logs
+          if-no-files-found: ignore
```
Kailai-Wang: I think this part exists in `dev` already

node/Cargo.toml
```diff
@@ -63,6 +64,26 @@ jsonrpsee = { version = "0.16.2", features = ["server"] }
...
+pallet-ethereum = { git = "https://github.com/paritytech/frontier", branch = "polkadot-v0.9.39", default-features = false }
+pallet-evm = { git = "https://github.com/paritytech/frontier", branch = "polkadot-v0.9.39", default-features = false }
+
+# evm-tracing
+moonbeam-primitives-ext = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39" }
+moonbeam-rpc-debug = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39" }
+moonbeam-rpc-primitives-debug = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39", default-features = false }
+moonbeam-rpc-primitives-txpool = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39", default-features = false }
+moonbeam-rpc-trace = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39" }
+moonbeam-rpc-txpool = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39" }
```
Kailai-Wang, line 89: It's a bit confusing, we have forked repo from `astar-frame` but the pallets are called `moonbeam-*`.

pallets/evm-address/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+	#[pallet::storage_version(STORAGE_VERSION)]
+	#[pallet::without_storage_info]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// The overarching event type.
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+		/// A unique id representing a single reward pool
```
Kailai-Wang: I'm sure that's some copy paste ...

node/src/rpc.rs
Kailai-Wang, line 1: I firmly believe these `without_evm` versions can be merged into the original ones 🤔 

If you still prefer separate files, maybe we can do it the other way around:
- `service.rs` -> unchanged
- `service_evm.rs` -> service with evm changes

pallets/evm-address/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+		type EVMId: Default + Copy + PartialEq + core::fmt::Debug + codec::FullCodec + TypeInfo;
+	}
+
+	/// Map for PoolId <> RewardPool
```
Kailai-Wang: Same here

pallets/evm-address/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn reward_pools)]
```
Kailai-Wang, line 51: same here

pallets/evm-address/src/lib.rs
Kailai-Wang, line 1: Hmm I'm wondering why we need a pallet for this, or why we need to store the mapped address in the runtime storage.

It seems we could do it via a simple convert fn or `From<>` implementation?

runtime/litentry/Cargo.toml
```diff
@@ -95,6 +95,17 @@ pallet-extrinsic-filter = { path = "../../pallets/extrinsic-filter", default-fea
...
+# EVM/Frontier
+fp-rpc = { git = "https://github.com/paritytech/frontier", branch = "polkadot-v0.9.39", default-features = false }
+fp-self-contained = { git = "https://github.com/paritytech/frontier", branch = "polkadot-v0.9.39", default-features = false }
+pallet-ethereum = { git = "https://github.com/paritytech/frontier", branch = "polkadot-v0.9.39", default-features = false, features = ["forbid-evm-reentrancy"] }
+pallet-evm = { git = "https://github.com/paritytech/frontier", branch = "polkadot-v0.9.39", default-features = false, features = ["forbid-evm-reentrancy"] }
+
+# Moonbeam
+moonbeam-evm-tracer = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39", default-features = false }
+moonbeam-rpc-primitives-debug = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39", default-features = false }
+moonbeam-rpc-primitives-txpool = { git = "https://github.com/litentry/astar-frame", branch = "polkadot-v0.9.39", default-features = false }
```
Kailai-Wang: Don't we want to onboard them to rococo first?

scripts/launch-local-binary.sh
```diff
@@ -101,9 +101,10 @@ sleep 10
 
 # run a litentry-collator instance
 $PARACHAIN_BIN --alice --collator --force-authoring --tmp --chain $CHAIN-dev \
-  --unsafe-ws-external --unsafe-rpc-external --rpc-cors=all \
+  --unsafe-ws-external --unsafe-rpc-external --rpc-cors=all --rpc-methods=Unsafe \
```
Kailai-Wang, line 104: Does `--unsafe-rpc-external` not suffice?

node/Cargo.toml
wangminqi: This is Astar maintained moonbeam code. Moonbeam does not seperate their code as pallet, Astar maintained one for themself. And we maintain a fork of our own from Astar team.

node/src/rpc.rs
wangminqi: Sure. I will revert their name accordingly

pallets/evm-address/src/lib.rs
wangminqi: We can always use padding/truncate method when transfer evm account into substrate account/reverse vice, but if one substrate-truncated evm account is used, you can not revert into original substrate account without storage.

That's what this pallet is for, storing the existing truncated-evm address and substrate account relation.

Without this storage, we may mistakely treat a truncated-evm address as an actual evm address and using padding method to get a padding-substrate account instead of the original one.

runtime/litentry/Cargo.toml
wangminqi: Yes. For fake compiling. These pallet imports are for mimicing a fake RPC for code consistency. Their main code are never actually compiled in runtime. Only RPC require runtime proccess some of their traits (which I implement them empty).

scripts/launch-local-binary.sh
wangminqi: I do find without --rpc-methods=Unsafe, the original setup will throw error in ci-test.

.github/workflows/ci.yml
wangminqi: I am wondering in this case if I should delete this part?

wangminqi: I check, it is the exact same change regarding dev. Anyway, I will delete this part in my branch.

scripts/launch-local-binary.sh
wangminqi: By the way, --rpc-methods=Unsafe is not allowed to be used in production environment.
I think it is for different layer of control vs  --unsafe-rpc-external
https://docs.rs/sc-cli/latest/sc_cli/struct.RunCmd.html#structfield.unsafe_rpc_external

--unsafe-rpc-external is: If unsafe RPC exposed, ignore the warning.
--rpc-methods=Unsafe: Do the expose unsafe RPC job.

runtime/litentry/Cargo.toml
wangminqi: Wait a minute. Let me check if we still need fake implementation after the separating of node code.

wangminqi: You are right. Code removed now.

Kailai-Wang (approved):
Thanks!

Can you please also create an issue to merge the "with" and "without" evm node implementation?

wangminqi:
https://github.com/litentry/litentry-parachain/issues/2118
sure
> Thanks!
> 
> Can you please also create an issue to merge the "with" and "without" evm node implementation?

PR #2018, opened 2023-08-14
Set shielding key with networks

### LINKED ISSUE #2014: Add a trustedCall to allow setting Web3Network upon user shielding key setup
### Context

So far when setting the shielding key, all supported networks will be added to the prime identity's web3networks:

We want to give the client a chance to set up the network information freely. So something like `set_user_shielding_key_with_networks` which takes an additional networks parameter.

To play it safe we'll put it at the end of the IDHUB's `TrustedCall` enum

Kailai-Wang:
### Context

resolves #2014


tee-worker/app-libs/stf/src/trusted_call_litentry.rs
```diff
@@ -34,23 +37,84 @@ use litentry_primitives::{
...
+	) -> StfResult<()>
+	where
+		NodeMetadataRepository: AccessNodeMetadata,
+		NodeMetadataRepository::MetadataType: NodeMetadataTrait,
+	{
+		debug!("set_user_shielding_key, who: {}", account_id_to_string(&who));
+		let account = SgxParentchainTypeConverter::convert(
+			who.to_account_id().ok_or(StfError::InvalidAccount)?,
+		);
+		let call_index =
```
kziemianek, line 77: I think this point can be reached from two different calls but we are returning the same event for both, we should have two events as well

kziemianek (changes requested):

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
kziemianek: Ok, I see there is no such call in parachain. Looks good then 👍🏼

kziemianek (approved):

PR #2029, opened 2023-08-16
Improve move-vcregistry-snapshot script output

0xverin:
### Context

Previously, our extrinsic encode hex was being displayed in the terminal. However, when we have accumulated a significant amount of vc registry data making it difficult to use and causing terminal slowdown. see screenshot:
image

As you can see ,we have 23 batch of extrinsic encode hex in the terminal :(...
So I saving the encode hex values into JSON files.Testing has been conducted, and it has been confirmed that the saved hex values are functioning properly.

### DIFF
--- a/scripts/ts-utils/move-vcregistry-snapshot.ts
+++ b/scripts/ts-utils/move-vcregistry-snapshot.ts
@@ -33,6 +33,8 @@ async function encodeExtrinsic() {
     let txs: any[] = [];
     console.log(colors.green('vcRegistry data length'), data.length);
     let i = 0
+    const hexData = []
+
     while (data.length > 0) {
         const batch = data.splice(0, BATCH_SIZE);
         const batchTxs = batch.map((entry: any) =>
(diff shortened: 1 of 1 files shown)


scripts/ts-utils/move-vcregistry-snapshot.ts
```diff
@@ -44,15 +46,29 @@ async function encodeExtrinsic() {
             )
         );
         txs = txs.concat(batchTxs);
-
         if (data.length === 0 || txs.length >= BATCH_SIZE) {
             i++
             const extrinsics = defaultAPI.tx.utility.batch(batchTxs);
```
arieljbirnbaum, line 51: Not immediately related, but this raises a tricky issue.

If we take a snapshot _based on encoded extrinsics_ before an upgrade, the new version may no longer be compatible with the old encoding, which defeats the purpose of the backup.

For comparison, the [version currently used on staging](https://github.com/litentry/litentry-parachain/blob/48f097e42e6eea2ea010b542a4555e730b633f20/scripts/ts-utils/move-vcregistry-snapshot.ts) dumps the actual vcregistry entries in human format, so restoring from that snapshot doesn't depend on the encoding. (Type mismatch issues can still arise, but those can be inspected and addressed).

arieljbirnbaum (approved):

PR #2034, opened 2023-08-17
refactor indirect call executors

### LINKED ISSUE #1912: Merge IMP extrinsic handling and BatchAll removal
We got some dead code left referenced in this [file](https://github.com/litentry/litentry-parachain/blob/6fdae037f5e40060283973ec4b98e5938febd879/tee-worker/core/parentchain/indirect-calls-executor/src/indirect_calls/litentry/mod.rs#L17) which should be removed.

Also code in [identitiy-call-executor](indirect-call-executors) looks almost the same, maybe we can find a way to extract common parts and make them more generic.

kziemianek:
Extracts common for `IndirectCalls` that submits `TrustedCalls`

close: #1912


felixfaisal (approved):
LGTM

Kailai-Wang (approved):
Thanks!

Issue #1873, opened 2023-07-11
Verify the correctness of VC content in ts-test


Kailai-Wang:
### Context

So far we have only tested these:
- vc request works
- the vc result is in valid format
- the vc signature matches
- the vc hash matches

However, we never verify that the assertion is correctly computed - if the `value` or result of the assertion is right. This can be done by examining a few chosen accounts which match the assertion criteria or not, and comparing it against the expected result.

If having such test cases for all assertions is too much for a single PR, we can either
- do it via multiple PRs
- only do important ones first (especially those that require http responses)

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

arieljbirnbaum:
> the vc result is in valid format

In light of [this discussion](https://litentry.slack.com/archives/C03KR96989G/p1689926953000869), we should also make sure to check the precise format of each type of credential subject, especially the information-carrying fields in the respective assertions.

Kailai-Wang:
It's done, see the migrated Linear issue

PR #2035, opened 2023-08-17
vc di test

0xverin:
### Context
request vc:
- [x] A1
- [x] A2
- [x] A3
- [x] A4
- [x] A6
- [x] A7
- [x] A8
- [x] A10
- [x] A11
- [ ] Achainable

assert content:
- [x] compare vc account with signer
- [x] check vc status should be Active
- [x] compare vc payload hash with vc hash
- [x] compare vc index vc payload id
- [x] check vc signature
- [x] compare vc with jsonSchema

TODO:
checking vc content is incomplete, follow up:https://github.com/litentry/litentry-parachain/issues/1873


tee-worker/ts-tests/integration-tests/vc_ii.test.ts
```diff
@@ -0,0 +1,163 @@
+import { randomBytes, KeyObject } from 'crypto';
```
kziemianek: This should be direct invocation, not indirect invocation test right ?
If yes you will need to change the file name.

0xverin:
Regarding the implementation of achainable,follow up:https://github.com/litentry/litentry-parachain/issues/2080

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -398,3 +403,79 @@ export function assertIdentityLinkedResult(
         parseIdentity(context.sidechainRegistry, decodedLinkResult.identity, aesKey).toString()
     );
 }
+/* 
+    assert vc
+    steps:
+    1. compare vc account with signer
```
kziemianek: I think this is bad assumption.
The response `account` field holds information about `subject` not `signer`.

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -398,3 +403,79 @@ export function assertIdentityLinkedResult(
...
+    3. compare vc payload hash(blake vc payload) with vc hash
+    4. compare vc index vc payload id
+    5. check vc signature
+    6. compare vc wtih jsonSchema
+
+    TODO: This is incomplete; we still need to further check: https://github.com/litentry/litentry-parachain/issues/1873
+*/
+
+export async function assertVc(context: IntegrationTestContext, signer: Signer, data: Bytes) {
+    const vc = context.api.createType('RequestVCResult', data) as unknown as RequestVCResult;
```
kziemianek: The name of this variable might be confusing.
This is not VC, this is the **result**.

```
pub struct RequestVCResult {
	pub account: AccountId,
	pub assertion: Assertion,
	pub vc_index: H256,
	pub vc_hash: H256,
	pub vc_payload: AesOutput,
}
```

You can read vc from `vc_payload` field.

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -398,3 +403,79 @@ export function assertIdentityLinkedResult(
...
+    assert.equal(
+        vcWithoutProof.type[0],
+        'VerifiableCredential',
+        'Check Vc payload type error: vcPayload type should be VerifiableCredential'
+    );
+    assert.equal(
+        proof.type,
+        'Ed25519Signature2020',
+        'Check Vc proof type error: proof type should be Ed25519Signature2020'
+    );
```
kziemianek, line 489: Can we assert that signature is valid here ?

tee-worker/ts-tests/integration-tests/di_vc.test.ts
```diff
@@ -0,0 +1,169 @@
...
+        context = await initIntegrationTestContext(
+            process.env.WORKER_ENDPOINT!, // @fixme evil assertion; centralize env access
+            process.env.NODE_ENDPOINT!, // @fixme evil assertion; centralize env access
+            0
+        );
+        teeShieldingKey = await getTeeShieldingKey(context.tee, context.api);
+        aliceSubject = await buildIdentityFromKeypair(new PolkadotSigner(context.substrateWallet.alice), context);
+    });
+
+    it('needs a lot more work to be complete');
```
kziemianek: Do we need this ?

tee-worker/ts-tests/integration-tests/di_vc.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    it('most of the bob cases are missing');
```
kziemianek: Same as above

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
0xverin: you mean `compare vc account with subject`?

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
0xverin: Is [this signature](https://github.com/litentry/litentry-parachain/pull/2035/files/56c076f3c71c23c672d2661b882f928adc3a7f7b#diff-076e81a8b9a33ee6bdcb6cefb0fac7391fb020a4be7d0439df31685361acadcbR459) what you're talking about?

kziemianek: yes thanks.

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
kziemianek: yes

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
0xverin: thanks! fixed.

tee-worker/ts-tests/integration-tests/di_vc.test.ts
0xverin: No, removed!

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
0xverin: thanks!fixed!

kziemianek (commented):
Why do we have indentity di tests using both substrate and evm but in this case there is no such split ?

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
kziemianek: It's left untouched. You modified other function doc

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -398,3 +403,79 @@ export function assertIdentityLinkedResult(
...
+    2. check vc status should be Active
+    3. compare vc payload hash(blake vc payload) with vc hash
+    4. compare vc index vc payload id
+    5. check vc signature
+    6. compare vc wtih jsonSchema
+
+    TODO: This is incomplete; we still need to further check: https://github.com/litentry/litentry-parachain/issues/1873
+*/
+
+export async function assertVc(context: IntegrationTestContext, signer: Signer, data: Bytes) {
```
kziemianek: You should rename `signer` parameter to `subject`.

tee-worker/ts-tests/integration-tests/di_vc.test.ts
```diff
@@ -0,0 +1,166 @@
...
+            assert.equal(
+                vcIssuedEvents.length,
+                1,
+                `vcIssuedEvents.length != 1, please check the ${Object.keys(assertion)[0]} call`
+            );
+            const trustedOperationResponse = context.api.createType(
+                'TrustedOperationResponse',
+                callValue.value
+            ) as unknown as TrustedOperationResponse;
+            await assertVc(context, new PolkadotSigner(context.substrateWallet.alice), trustedOperationResponse.value);
```
kziemianek: Here You should pass `aliceSubject` not `PolkadotSigner`

0xverin: emm...why we should pass `aliceSubject`? Doesn't seem to have anything to do with the `LitentryPrimivesIdentity` type.

kziemianek: Because you want to verify subject, not the signer (maybe they both are the same) but the VC is issued for the subject, not the signer.

0xverin:
> Why do we have indentity di tests using both substrate and evm but in this case there is no such split ?

Because in the case of the `di_vc.test`, the signer being EVM or Substrate is just a difference in addresses. Adding EVM account testing case would be a bit redundant. However, for `di_evm/substrate_identity` test, we need to verify some additional information, such as `primeIdentity` and `web3Networks` verification. This requires confirming that using both EVM and Substrate addresses is indeed different in these two aspects.

tee-worker/ts-tests/integration-tests/di_vc.test.ts
0xverin: maybe just naming confusing... should change as:
`const subject = new PolkadotSigner(context.substrateWallet.alice)`
`await assertVc(context, subject, trustedOperationResponse.value)`;
:p

kziemianek: Still I don't understand why you need to use `Signer` type to assert the subject of vc. It would be more convinient to use `LitentryPrimivesIdentity` instead.

0xverin: Updates:
checking subject instead of compare account with signer.

0xverin: I didn't notice that the credentialSubject id was already the DID format.

kziemianek (approved):
Looks fine for me 👍🏼

PR #2039, opened 2023-08-18
Cleanup and refactor `deploy.sh`

Kailai-Wang:
### Context

This PR tries to simplify and improve the script. Notes:
- "Generating service files" is not extracted to a subcommand: `generate`. The service files are stored under `template/` folder for better readability
- services are now started with normal user (logged in user). It's not a good idea to always start the whole script as `root`
- common functions are extracted
- add a flag `--prod` to control how the parachain and enclave are built
- steps in`upgrade-worker` are reordered a bit, it would work on non prod only. (as we don't have the private key for prod teerex admin)
- folder structures are changed
- bug fix

Things that are tested:
- resetting the whole network: relaychain + parachain + worker
- restart worker only
- upgrade worker under dev


tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -1,736 +1,555 @@
...
+      docker pull "$DOCKER_IMAGE"
+      docker cp "$(docker create --rm $DOCKER_IMAGE):/usr/local/bin/litentry-collator" "$PARACHAIN_BASEDIR"
+    else
+      echo "Building parachain binary ..."
+      cd "$ROOTDIR" || exit
+      if [ "$PRODUCTION" = true ]; then
+        cargo build --locked --profile production
+      else
+        pwd
+        make build-node
```
felixfaisal, line 351: Shouldn't we have a flag here to build for production, I think `make build-node` only builds with `tee-dev` feature. We will need without `tee-dev` feature for production.

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -1,736 +1,555 @@
...
-  corepack yarn
-  corepack yarn upgrade-parathread 2>&1 | tee "$TMPDIR/upgrade-parathread.log"
+  corepack yarn upgrade-parathread 2>&1 | tee "$PARACHAIN_BASEDIR/upgrade-parathread.log"
   print_divider
 
-  echo "done. please check $TMPDIR for generated files if need"
-
+  echo "done. please check $PARACHAIN_BASEDIR for generated files if need"
   print_divider
+  git restore config.ci.json
```
felixfaisal, line 434: Doesn't restore remove any changes we made to the `config.ci.json` in the earlier step with `jq`.

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -1,736 +1,555 @@
...
 
-    cd $ROOTDIR/tee-worker || exit
-
-    source=$(echo "$CONFIG" | jq -r ".workers[$i].source")
-    flags=$(echo "$CONFIG" | jq -r ".workers[$i].flags[]")
-    subcommand_flags=$(echo "$CONFIG" | jq -r ".workers[$i].subcommand_flags[]")
+    for ((i = 0; i < WORKER_COUNT; i++)); do
+      worker_dir="$WORKER_BASEDIR/w$i"
+      mkdir -p "$worker_dir"
+      for f in 'key.txt' 'spid.txt' 'enclave.signed.so' 'litentry-worker'; do
```
felixfaisal, line 450: The `worker` will expect for `spid_production.txt` and `key_production.txt` in production mode.

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -1,736 +1,555 @@
...
 
-    command+=" run"
-
-    for subcommand_flag in $subcommand_flags; do
-      command+=" $subcommand_flag"
+      if [ "$PRODUCTION" = true ]; then
+        echo "Transferring balance to the enclave account $enclave_account ..."
+        cd $ROOTDIR/scripts/ts-utils/ || exit
+        yarn install
+        npx ts-node transfer.ts $enclave_account
```
felixfaisal, line 463: If we build the parachain without the `tee-dev` feature, We will also have to schedule the enclave from the first block itself, else the `register_enclave` call by the `worker` will fail.

tee-worker/scripts/litentry/release/deploy.sh
```diff
@@ -1,736 +1,555 @@
...
+
+      # wait until sidechain stalls
+      wait_for_sidechain
+      backup_workers
+      stop_worker_services
+      get_old_mrenclave
+      # TODO: actually we only need the copy-up
+      setup_working_dir
+      migrate_shard
+      remove_clean_reset
```
felixfaisal, line 186: So when we are upgrading worker we are only removing the `--clean-reset` flag, Shouldn't we also update the latest parentchain block to sync from, else it will always will be`0`

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: Just a few lines above :) Line 348

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: Yea but we need it just for the parathread upgrade, don't we? And we should be done with it here

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: Do you know where can I get it :(

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: It won't be from 0, it will continue from where it's left off. So a bunch of blocks behind the latest, but not a lot

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: I think this script can't be used for a full automated prod development anyway as we don't know the teerex admin, as written in the PR description ...

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Ah my bad 😅

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I see, makes sense.

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: Yeah I agree, but if we run it like this in Prod, The worker will not be able to broadcast the blocks and then we call the extrinsic to schedule enclave from 0 from explorer, then we restart worker. If we are okay with doing this, then it is  fine I guess.

tee-worker/scripts/litentry/release/deploy.sh
BillyWooo: `It won't be from 0, it will continue from where it's left off.` The parameter `--parentchain-start-block` is used for this feature. Why it will continue from where it's left off?

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: I'm fine with that - actually, the prod parachain should not be reset once in a while (once we move to at least rococo), so actually we only have to take care of the worker restart

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I think we keep it inside the `configs` folder inside the server, can we have a flag to specify the path for it or we just assume that it was copied to the `bin` directory.

BillyWooo (commented):
Thanks very much. Quite a nice refactor. One question following in one of the comment from Faisal above.

tee-worker/scripts/litentry/release/deploy.sh
BillyWooo: Yes. Currently it's stored on `tee-staging` server.

Kailai-Wang: > I think we keep it inside the `configs` folder inside the server, can we have a flag to specify the path for it or we just assume that it was copied to the `bin` directory.

I'm fine with whichever way you prefer.

tee-worker/scripts/litentry/release/deploy.sh
Kailai-Wang: It has the metadata stored in light_client DB IIRC

tee-worker/scripts/litentry/release/deploy.sh
felixfaisal: I think for now, let's assume we put in `bin` folder and the script copies from it.

BillyWooo (approved):

PR #2050, opened 2023-08-21
Check ts code formatting in CI

### LINKED ISSUE #2036: Run prettier check in CI for each pull request
There are npm `format` scripts that formats our js/ts code. We can use `prettier` to check wheter the code is formatted or not and block pull request from being merged untill the code is formated- the same as we do for Rust code.

kziemianek:
Add job to check ts code formatting.

closes: #2036


.github/workflows/ci.yml
```diff
@@ -134,6 +134,19 @@ jobs:
           echo "run_parachain_test=$run_parachain_test" | tee -a $GITHUB_OUTPUT
           echo "run_tee_test=$run_tee_test" | tee -a $GITHUB_OUTPUT
 
+  format-ts:
```
Kailai-Wang: How about appending it directly to `fmt` job? 😂

kziemianek: I was considering that but decided to have it separated. Would you like to have it merged ?

Kailai-Wang: Actually yes :) so that we have all-in-one place for format check.
It doesn't take too long to complete this step so it doesn't necessarily need a separate step/job

kziemianek: Ok, i'll merge. 
I was misleaded by thought  that `fmt` is specific to cargo's formatter 🙃

.github/workflows/ci.yml
```diff
@@ -167,6 +167,14 @@ jobs:
...
           cargo fmt --all -- --check
 
+      - name: Tee-worker install npm deps
+        working-directory: ./tee-worker/ts-tests
+        run: corepack yarn install
+
+      - name: Tee-worker check ts code format
+        working-directory: ./tee-worker/ts-tests
+        run: corepack yarn check-format
+
```
Kailai-Wang, line 177: I think you can do it within one step, like:
```
run: |
  corepack yarn install
  corepack yarn check-format

Kailai-Wang (approved):
Thanks!

.github/workflows/ci.yml
kziemianek: Yes, i know but decided to go with more granural aproach!

PR #2052, opened 2023-08-21
Collect docker logs for parachain ts-tests

Kailai-Wang:
### Context

As topic. A minor PR to collect docker logs and upload them when parachain ts-tests fail.
We launch the parachain in docker so `/tmp/parachain_dev` is not enough.


BillyWooo (approved):
LGTM

Issue #1995, opened 2023-08-07
Cleanup and commit the stress test code to the repo


Kailai-Wang:
### Context

Follow-up task of #1966 - the code can be inspirational/reused for the next stress test :) 

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

arieljbirnbaum:
@Kailai-Wang thanks :raised_hands: 

We did a heap of quick and dirty duplication when setting up the stress scripts just to get it going quickly; we should take a moment to refactor that before merging it in.

Also, we should add a short step in CI to exercise the code (2-3 iterations should be enough) just to make sure it keeps up with subsequent API changes. To be clear though, I _wouldn't_ add stress testing as a CI step; this is only to make sure that the stress code stays in sync so we can run it when needed.

Kailai-Wang:
Thanks for the heads-up.
No problem. This one is not so urgent, you can take your time :)

kziemianek:
@grumpygreenguy can we close this issue ?

Issue #1153, opened 2023-01-04
Restructure the integration test cases


Kailai-Wang:
### Context
We are having more and more integration tests for the tee-workers. Putting every test case into a separate YML might be overkill since it introduces a new matrix-branch in the CI.

Is there a better way to group them?

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
For the shell script we could extract the standard part (e.g. utility, helpers) to avoid code duplication

BillyWooo:
https://github.com/litentry/litentry-parachain/pull/1299
https://github.com/litentry/litentry-parachain/pull/1310

Kailai-Wang:
I think there are still a few things worth doing, like mentioned in the topics:
- group the common utilities 
- better ways of handling test case execution in parallel (other than matrix) -> it might need research

Kailai-Wang:
Closing it - we have done some improvements and so far happy with the current implementation

PR #2054, opened 2023-08-21
Script for load tests

arieljbirnbaum:
Implemented in preparation for the EVM campaign.
Adding to the repo as decided in #1995.
Duplicates code from the existing test suites to allow for modification; should eventually be refactored to have a single source of truth. Consider along #1153, #1878.

## TODO

- [x] Exercise activate/deactivate identity steps
- [x] Run on CI **with a very small number of threads/iterations** to make sure the test code continues to work as intended.


.nvmrc
arieljbirnbaum, line 1: drive-by

tee-worker/ts-tests/integration-tests/common/type-definitions.ts
arieljbirnbaum, line 1: drive-by

tee-worker/ts-tests/parachain-api/prepare-build/interfaces/identity/definitions.ts
```diff
@@ -192,7 +192,7 @@ export default {
                 A1: "Null",
                 A2: "Bytes",
                 A3: "(Bytes,Bytes,Bytes)",
-                A4: "u128",
```
arieljbirnbaum: I don't think this should be here; probably a leftover after rebasing from `staging`.

tee-worker/ts-tests/parachain-api/src/index.ts
```diff
@@ -19,7 +19,9 @@ export type {
     WorkerRpcReturnValue,
     TrustedCallSigned,
     Getter,
+    PublicGetter,
```
arieljbirnbaum, line 26: Added export

tee-worker/ts-tests/stress-runner
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/bash
+
+mkdir -p stress-output
```
arieljbirnbaum, line 3: TODO: take the number of processes and output dir as a parameter; set correct working directory.

tee-worker/ts-tests/stress/bin/stress.ts
```diff
@@ -0,0 +1,302 @@
...
+function randomEvmWallet(): ethers.Wallet {
+    return ethers.Wallet.createRandom();
+}
+function randomWallet(): Wallet {
+    return Math.random() > 0.5
+        ? { type: "substrate", keyringPair: randomSubstrateWallet() }
+        : { type: "evm", wallet: randomEvmWallet() };
+}
+
+const apiContextManager = (
```
arieljbirnbaum: [`using`](https://github.com/tc39/proposal-explicit-resource-management) at home.

The test needs to establish several connections, create new ones on the fly in case of failure, and avoid leaking resources. These `ContextManager` objects help with that.

tee-worker/ts-tests/stress/config.json
```diff
@@ -0,0 +1,11 @@
+{
+	"timerEpoch": 1692230400000,
```
arieljbirnbaum: TODO: not sure how useful this actually is in practice?

tee-worker/ts-tests/stress/config.json
```diff
@@ -0,0 +1,11 @@
...
+	"connections": 20,
+	"iterations": 10,
+	"substrateEndpoint": "ws://tee-staging.litentry.io:9944",
+	"workerEndpoint": "wss://tee-staging.litentry.io:2000",
```
arieljbirnbaum: TODO: need a way to run this against local worker on CI. Config should prob. be a command line argument?

tee-worker/docker/lit-test-stress-script.yml
```diff
@@ -0,0 +1,25 @@
...
+            context: ..
+            dockerfile: build.Dockerfile
+            target: deployed-client
+        depends_on:
+            litentry-node:
+                condition: service_healthy
+            litentry-worker-1:
+                condition: service_healthy
+            litentry-worker-2:
+                condition: service_healthy
```
arieljbirnbaum: TODO: should probably drop this one? /cc @Kailai-Wang @felixfaisal

tee-worker/ts-tests/stress/src/context-manager.ts
```diff
@@ -0,0 +1,57 @@
...
+    constructor(enter: ContextFactory<Context>) {
+        this.enter = enter;
+    }
+
+    async do<Result>(task: (context: Context) => Promise<Result>) {
+        const { context, exit } = await this.enter();
+        try {
+            return await task(context);
+        } finally {
+            exit();
```
arieljbirnbaum:
```suggestion
            await exit();
```

arieljbirnbaum: (letting the current CI run finish before I commit this)

.github/workflows/ci.yml
```diff
@@ -586,6 +586,7 @@ jobs:
           - test_name: lit-di-evm-identity-test
           - test_name: lit-parentchain-nonce
           - test_name: lit-ii-batch-test
+          - test_name: lit-test-stress-script
```
Kailai-Wang, line 590: Are we planning to trigger it for each CI run?

tee-worker/docker/lit-test-stress-script.yml
Kailai-Wang: yea if you are only sending requests to worker-1 you don't have to wait for worker-2 (it's still launched though)

tee-worker/ts-tests/stress/test/test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+function getConfig(): Config {
+    return {
+        connections: 3,
+        iterations: 2,
+        workerEndpoint: process.env.WORKER_ENDPOINT!, // @fixme evil assertion; centralize env access
+        substrateEndpoint: process.env.NODE_ENDPOINT!, // @fixme evil assertion; centralize env access
+    };
+}
+
```
0xverin, line 31: Can we add a comment here to describe the entire workflow? This could be more user-friendly for others who want to understand the stress test. :P

0xverin: To be honest, I need quite some time to digest all of this.

0xverin (commented):
Thanks!But whether to add it to the CI might need some consideration. The benefit of integrating it into CI is to ensure that this stress test remains usable, but it could increase maintenance efforts. I'm concerned that it might add complexity for other developers.

tee-worker/ts-tests/stress/test/test.ts
arieljbirnbaum: Good point; will add some :+1:

.github/workflows/ci.yml
arieljbirnbaum: Yes, to make sure that it doesn't go stale with any future changes (so that it's ready to go whenever we need it again). i.e. it's a test _of_ the stress logic rather than a stress test of the logic :)

The test starts three threads, each thread does two iterations; in total it shouldn't be a heavy burden on CI; I think :thinking: 

Then again, what's probably more expensive is just to start the container for the test, so maybe it could be grouped with some other step?

arieljbirnbaum: /cc @Verin1005 to address your [comment](https://github.com/litentry/litentry-parachain/pull/2054#pullrequestreview-1594773483); I don't think it should add much more maintenance overhead beyond keeping the _regular_ tests up to date. There is some duplication right now but I'm planning to tackle it with #1878 and #1153

Kailai-Wang: It's fine as long as it's not running extremely long 😎

tee-worker/ts-tests/stress/test/test.ts
arieljbirnbaum: @Verin1005 I added comments throughout the test to clarify the purpose of each section; does that help?

0xverin: Thanks!I think it's actually possible, because your function and variable naming is actually kind of a commentary.:p

tee-worker/ts-tests/stress-runner
arieljbirnbaum: dropped for now

0xverin (approved):
LGTM.Thanks!

PR #2064, opened 2023-08-23
Bugfix: ClassOfYear VC returns false for linked EVM addresses

### LINKED ISSUE #2044: ClassOfYear VC returns false for linked EVM addresses
### Context

It was initially reported by the community, and it occurred in my tests too.

Steps to reproduce:
- login with EVM address, set shielding key
- request this VC -> Ok
- link a bunch of EVM addresses
- request this VC -> issued by `value == false`

This must be wrong - it can't be worse than the first attempt

---

outofxxx:
fix #2044


tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
```diff
@@ -114,9 +114,9 @@ pub fn request_achainable_classofyear(
 			Error::RequestVCFailed(Assertion::Achainable(param.clone()), e.into_error_detail())
 		})?;
 
-		// If parse metadata field error, return Invalid immediately
+		// If parsing the current metadata field fails, continue to parse the next address
 		if year.parse::<u32>().is_err() {
-			return Ok((false, INVALID_CLASS_OF_YEAR.into()))
+			continue
```
Kailai-Wang, line 119: Should we abort the VC building? It's similar to https://github.com/litentry/litentry-parachain/issues/1916, isn't it?

When would parsing fail? If it's a TDF issue, they should fix it.

outofxxx: According to [TDF](https://litentry.slack.com/archives/C04914RHSAD/p1692874606687079?thread_ts=1692844891.993379&cid=C04914RHSAD), when there is no data, the metadata field is null. Thus, this place will fail to parse.

kziemianek: If there is a way to detect the `null` i would prefer to do that.
Also I suggest putting comment explaining why we are handling such special case in this way somewhere in the code.

Kailai-Wang: Could we have two possibility if we fall into this arm? Like
1. the response is invalid
2. the response is valid, but the metadata is null

Because we'll need to handle them differently: for 1 => error out; for 2 => continue.

To put it in another way, can we strictly detect the metadata is null but the response is valid?

outofxxx: Yes, only the second scenario (`the response is valid, but the metadata is null`) will enter here to continue.

Kailai-Wang: That sounds good, thanks :)

Kailai-Wang (approved):

PR #2065, opened 2023-08-23
Bugfix: DI sync response across the block boundary

### LINKED ISSUE #2063: Can't dispatch synchronous response across block boundary
### Context

This is a bug in the current implementation, and can be verified by intentionally delaying the web2 verification with e.g. `thread::sleep` in the verification thread.

There are more than one problem:

1. The `update_status_event` of first step trusted call will be called when it's included in the sidechain block. The `result.status`  would by then still contain the old status, which should be `Submitted`, thus the `matches!` case will return false:

https://github.com/litentry/litentry-parachain/blob/d4643228d41f706964e7ca861fcf4666ad080dbb/tee-worker/core/direct-rpc-server/src/rpc_responder.rs#L98-L102

2. swap_hash would fail. After the first trusted call is included in the sidechain block, the hash would be removed from the top pool, removing all watchers too. Later when swapping the hashes, the registry item with the old hash can't be found.

---

Kailai-Wang:
### Context

resolves #2063

### DIFF
--- a/tee-worker/core-primitives/top-pool/src/listener.rs
+++ b/tee-worker/core-primitives/top-pool/src/listener.rs
@@ -162,6 +162,14 @@ where
 
 	/// Litentry: swap the old hash with the new one in rpc connection registry
 	pub fn swap_rpc_connection_hash(&mut self, old_hash: H, new_hash: H) {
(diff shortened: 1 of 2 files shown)


kziemianek (approved):
Looks fine 👍🏼

PR #2023, opened 2023-08-14
Update README.md

cryptoade1:
### Context

https://github.com/litentry/litentry-parachain/issues/1434

The current repo README has many broken links/badges and outdated information. Shall we revamp and refresh it?

It doesn't need to contain every detail but should include:

the parachain logo and proper links (e.g. gitbook, CI badges)
a general introduction to parachain and tee-worker
how to build and run parachain and tee-worker - in case someone wants to play around it locally
remove outdated information (e.g. extend the leasing period, which is resolved by https://github.com/litentry/litentry-parachain/pull/1431).

PR #2066, opened 2023-08-23
fix: cleaning up tmpdir for local-binary launch

felixfaisal:
This required to make a few more changes 
- replace all instances of `TMPDIR=/tmp/parachain_dev` to `TMPDIR=${TMPDIR:-"/tmp/parachain_dev"}` 
- Initially the `launch.py` had a single positional argument which is the config, but now we have more config options and it makes more sense to use `CLI flags` for this. So I've updated this 
- I've also extended the script to automatically reassign the `parachain_dir` if the default directory or the directory they requested to is not accessible, like this 
```
Directory is not accessible, Reassigning the directory...
Directory has been reassigned to: /tmp/parachain_dev-30-08-2023
```
I realise this affects https://github.com/litentry/litentry-parachain/pull/2023 if merged before, So some changes to starting the script will be here:- 

1. `./local-setup/launch.py local-setup/github-action-config-one-worker.json` this will change to `./local-setup/launch.py --config ./local-setup/github-action-config-one-worker.json` 
2. `./local-setup/launch.py local-setup/github-action-config-one-worker.json local-binary` this will change to `./local-setup/launch.py --config ./local-setup/github-action-config-one-worker.json --parachain local-binary` 
3. `./local-setup/launch.py local-setup/development-worker.json local-binary 100` this will change to `./local-setup/launch.py --config local-setup/development-worker.json --parachain local-binary --offset 100` 

If the PR gets merged before this then I can update it myself, @cryptoade1 Check this out, I'll add comment on the Notion as well where you need to make these changes.


scripts/clean-local-binary.sh
```diff
@@ -2,7 +2,7 @@
 
 # no `set -e` here as we allow commands to fail in this script
 
-TMPDIR=/tmp/parachain_dev
+TMPDIR=${TMPDIR:-"/tmp/parachain_dev"}
```
BillyWooo: `TMPDIR=${TMPDIR:-"/tmp/parachain_dev"}` where does TMPDIR value come from when it's not default case?

Kailai-Wang (commented):
Thanks :)

I believe `clean-local-docker.sh` needs to be changed as well.

And I'm actually thinking which one is better, ask user to pass in a tmp folder, or we simply create a random one 🤔

scripts/clean-local-binary.sh
felixfaisal: It is set from `launch.py` when pass it in as a CLI parameter

felixfaisal:
So I actually thought about this, I couldn't think of a scenario wherein we need to specify the `parachain_dir` for `local-docker` setup, as most things are inside `docker` container and they will not affect if two people are running the system in the same machine. However, We need to not allow caller to set the `parachain_dir` when using `local-docker` to avoid the undefined behaviour mentioned by @grumpygreenguy

Kailai-Wang:
I mean even for the `local-binary` mode, we _could_ create the dir for the user. We just need to print out the dir name in case the user wants to check the logs in it (in most cases no as it only contains the parachain logs)

And IMO if you change `clean-local-binary`, you could change all other occurrences of `/tmp/parachain_dev` too

felixfaisal:
I made a few changes based on your feedback. Let me know what you think @Kailai-Wang

tee-worker/local-setup/launch.py
```diff
@@ -266,28 +275,38 @@ def main(processes, config_path, parachain_type, log_config_path, offset, parach
...
-        "parachain",
+        "--parachain",
         nargs="?",
         default="local-docker",
         type=str,
         help="Config for parachain selection: local-docker / local-binary / remote",
     )
     parser.add_argument(
-        "log_config_path",
+        "--log_config_path",
```
Kailai-Wang: Can we have also a short flag for it - it's a bit long 😅 

And I believe conventionally the flag name is connected using `-` instead of `_`, like `--log-config-path`

tee-worker/local-setup/launch.py
```diff
@@ -266,28 +275,38 @@ def main(processes, config_path, parachain_type, log_config_path, offset, parach
...
-        "parachain_dir", nargs="?", default="/tmp/parachain_dev", help="Parachain directory for local binary"
+        "--parachain_dir", nargs="?", default="/tmp/parachain_dev", help="Parachain directory for local binary"
     )
     args = parser.parse_args()
+    
+    if not is_directory_accessible(args.parachain_dir): 
+        print("Directory is not accessible, Reassigning the directory")
+        today = datetime.datetime.now()
+        formatted_date = today.strftime('%d-%m-%Y')
+        directory_name = f"parachain_dev-{formatted_date}"
```
Kailai-Wang: Maybe it's better not to mix `_` and `-` in the folder name, we could always use `_` in this case

tee-worker/local-setup/launch.py
Kailai-Wang, line 1: Shall we export the `parachain_dir` also for `local-docker`?

```
        # Export Parachain Directory as Global Variable
        os.environ['TMPDIR'] = parachain_dir
```

tee-worker/local-setup/launch.py
```diff
@@ -79,6 +80,14 @@ def is_port_open(port):
     except OSError:
         return False
 
+def is_directory_accessible(directory_path):
+    if os.path.exists(directory_path):
```
Kailai-Wang: I think it's better to simply reassign the directory as long as it exists.

Because even if you can access it (which probably means it's created by yourself), you don't want to overwrite it since most likely you have another instance running - but we can't be sure if you simply forget to terminate it, or you intentionally let it run

Kailai-Wang (changes requested):
Thank you, I made a few suggestions.

An alternative is to remove `parachain_dir` completely, and we just create a new tmpdir each time. The directory would only contain the parachain logs anyway (which, in most of the cases, not worth inspecting)

tee-worker/local-setup/launch.py
felixfaisal: Updated ✅

tee-worker/local-setup/launch.py
felixfaisal: Updated ✅

tee-worker/local-setup/launch.py
felixfaisal: Done ✅

tee-worker/local-setup/launch.py
felixfaisal: Updated it ✅

tee-worker/local-setup/launch.py
```diff
@@ -79,6 +80,11 @@ def is_port_open(port):
     except OSError:
         return False
 
+def is_directory_present(directory_path):
```
kziemianek: isn't this function redundant ? 
It only proxies calls to `os.path.exists`.

tee-worker/local-setup/launch.py
```diff
@@ -266,30 +273,38 @@ def main(processes, config_path, parachain_type, log_config_path, offset, parach
...
-        "parachain_dir", nargs="?", default="/tmp/parachain_dev", help="Parachain directory for local binary"
+        "--offset", nargs="?", default="0", type=int, help="offset for port"
     )
     args = parser.parse_args()
+    
+    if is_directory_present("/tmp/parachain_dev"): 
+        print("/tmp/parachain_dev already exists, Reassigning the directory")
+        today = datetime.datetime.now()
+        formatted_date = today.strftime('%d-%m-%Y')
+        directory_name = f"parachain_dev_{formatted_date}"
```
kziemianek: Why can't we just create new directory with this name, instead of trying with `/tmp/parachain_dev` ?

BillyWooo: I would say so. Just create `/tmp/parachain_dev_datetime` would be simpler. Then we don't even need to check folder existance.

tee-worker/local-setup/launch.py
Kailai-Wang: It is

tee-worker/local-setup/launch.py
Kailai-Wang: I believe it's for veterans who are used to look for `/tmp/parachain_dev`

but yea creating new dir each time is fine too

tee-worker/local-setup/launch.py
```diff
@@ -266,30 +273,38 @@ def main(processes, config_path, parachain_type, log_config_path, offset, parach
...
-        "parachain",
+        "--parachain",
         nargs="?",
         default="local-docker",
         type=str,
         help="Config for parachain selection: local-docker / local-binary / remote",
     )
     parser.add_argument(
-        "log_config_path",
+        "--log",
```
Kailai-Wang: Hmm that's actually not a so good name 😅 
I'd say for the long name we can sue `--log-config-path`, for the short name we can use `-l`

Kailai-Wang: Same for other args, we can always have an _additional_ "short" version
like `-c`, `-p`, `-l`, `-o` , it will make the user's life easier

scripts/clean-local-docker.sh
```diff
@@ -8,7 +8,7 @@ function usage() {
 
 [ $# -ne 1 ] && (usage; exit 1)
 
-TMPDIR=/tmp/parachain_dev
+TMPDIR=${TMPDIR:-"/tmp/parachain_dev"}
```
arieljbirnbaum: The default scares me tbh. Can we just fail if it's not set?

arieljbirnbaum:
Please note, there are also hardcoded references to `/tmp/parachain_dev` in `ts-tests/config.ci.json` and `ts-tests/tests/setup-bridge.ts`

tee-worker/local-setup/launch.py
felixfaisal: Not using this as always creating new_dir each time

tee-worker/local-setup/launch.py
felixfaisal: Updated ✅

tee-worker/local-setup/launch.py
felixfaisal: Updated ✅

scripts/clean-local-docker.sh
felixfaisal: I've updated this to: 
```
TMPDIR=${TMPDIR:-"/tmp/parachain_dev*"}
``` 
Since we are creating a new directory every time, If we don't specify the exact directory we want to clean-up, it will clean up all with this pattern. It will however not delete files that were created by another user since we don't have permissions for it. I don't see a scenario where this could be problematic, Let me know what do you think.

Kailai-Wang (approved):
I'm fine with it, thanks!

PR #2069, opened 2023-08-25
Force connections to wait for swap

### LINKED ISSUE #2060: Handle streamed rpc responses properly
Currently there is magic `true` encoded value hardcoded to detect whether `rpc_responder` should forcibly respond to client with `do_watch: true` so the client will keep on waiting for more values. 

https://github.com/litentry/litentry-parachain/blob/385c54864e3e45f18621466e75c22806f3dad0f1/tee-worker/core/direct-rpc-server/src/rpc_responder.rs#L91-L96

Ideally we should get rid of this trick and possibly rely on `req_ext_hash`.

kziemianek:
Add explicit flag to `ConnectionRegistry` so connections can be forced to wait for the swap.

fixes: #2060

### DIFF
--- a/tee-worker/app-libs/stf/src/trusted_call_result.rs
+++ b/tee-worker/app-libs/stf/src/trusted_call_result.rs
@@ -20,7 +20,7 @@
 
 use crate::AccountId;
 use codec::{Decode, Encode};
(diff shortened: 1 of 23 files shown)


BillyWooo (approved):
Thanks. LGTM.

PR #2070, opened 2023-08-25
make di responses minimalistic

### LINKED ISSUE #1975: DI sync response account id
Direct invocation synchronous responses contain account identifier in substrate format which is 32 byte long.
Because DI can be submitted both by substrate and evm accounts this identifier should contain original form of account id - not substrate format as it currently happens in both cases.

Pros: 
- easier for consumer to use returned id

Cons:
- not consistent with II parachain events which always hold account id in substrate format

kziemianek:
It's always good practice to keep public api as lean as possible when it comes to data transfer,  maintainability and ease of integration.

fixes #1975


Kailai-Wang (approved):
Thank you, it makes sense.

Can you please notify IDH colleagues explicitly? Just to see if there's major objection before we deploy it.

I'm also thinking how to deal with the II interface (extrinsic)

kziemianek:
@Kailai-Wang you mean applying the same changes to responses submited to parachain ?

Kailai-Wang:
> @Kailai-Wang you mean applying the same changes to responses submited to parachain ?

I'm not sure if we should.
It would need a broader discussion I guess, so let's not touch it in this PR :)

kziemianek:
Ok, If we plan to touch them, let's do this before we connect to rococo.

BillyWooo:
> Ok, If we plan to touch them, let's do this before we connect to rococo.

I think we don't need to worry about connecting to rococo. It won't affect here. 
The broader discussion is more important.

PR #2098, opened 2023-09-03
A workaround for truncated chain spec

Kailai-Wang:
### Context

The previously submitted upstream PRs(https://github.com/open-web3-stack/parachain-launch/commit/45693e2cc05b36b79ab2e96052c41bcd471c49db, https://github.com/open-web3-stack/parachain-launch/commit/eafed3cfbaf305e2b93131fd40051ee9bf7e7366) didn't solve the problem completely.

We still see truncated chain specs like:

https://github.com/litentry/litentry-parachain/actions/runs/6064157036
https://github.com/litentry/litentry-parachain/actions/runs/6063920972

I tried `stdbuf` too but it didn't work.

It's highly possible that `docker run` is doing some buffered io: https://stackoverflow.com/a/39497651/921992

As a workaround, this PR uses a [self-forked](https://github.com/open-web3-stack/parachain-launch/compare/master...Kailai-Wang:parachain-launch:use-raw-binary) `parachain-branch` where the raw binary is used instead of `docker run`. The downside is this parachain-launch version will **only work under Linux**.

For local env other than Linux, please use binary mode to launch the parachain network.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,24 +156,24 @@ jobs:
           make fmtcheck
           make taplocheck
 
-      # - name: Tee-worker fmt check
-      #   working-directory: ./tee-worker
-      #   run: |
-      #     cargo fmt --all -- --check
-      #     taplo fmt --check
+      - name: Tee-worker fmt check
+        working-directory: ./tee-worker
+        run: |
+          cargo fmt --all -- --check
+          taplo fmt --check
 
-      # - name: Enclave-runtime fmt check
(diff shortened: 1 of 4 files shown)


felixfaisal (approved):
Thanks @Kailai-Wang  
The workaround looks good to me but wondering if the fork should be moved to `Litentry` organisation.

Kailai-Wang:
> Thanks @Kailai-Wang The workaround looks good to me but wondering if the fork should be moved to `Litentry` organisation.

Thanks!
I'm fine with either way - I'll leave it in my repo first as I hope the workaround won't live too long

PR #2100, opened 2023-09-04
Achainable only accepts address as a parameter in substrate networks

### LINKED ISSUE #2082: Achainable only accepts address as a parameter in substrate networks
* If a `pubkey` is passed in, it will not prompt an error, but the returned result is incorrect.

Correct request body:

```
{
    "name": "Balance hodling {amount} since {date}",
    "address": "48DXPdgeqTPhC5zhfXqE3QJM7sCdHuxZ5ky5vbd5jAujYwyR",
    "params": {
        "chain": "litentry",
        "amount": "10",
        "date": "2023-07-01"
    }
}
```

outofxxx:
fix: #2082


Kailai-Wang (approved):
Looks good!

PR #2107, opened 2023-09-06
allow to specify worker id

kziemianek:
Allows to specify id for worker started in local setup.

the id is used in:

- log directory name
- tmp directory name

It allows to run `./local-setup/launch.py` multiple time as it will use different names for mentioned above directories.
It is useful in scenario where user wants to spin up another worker that will join existing local network.

### DIFF
--- a/tee-worker/local-setup/launch.py
+++ b/tee-worker/local-setup/launch.py
@@ -59,11 +59,12 @@ def setup_worker(work_dir: str, source_dir: str, std_err: Union[None, int, IO], 
 
 
 def run_worker(config, i: int, log_config_path):
-    log = open(f"{log_dir}/worker{i}.log", "w+")
+    id = config.get('id', i)
+    log = open(f"{log_dir}/worker-{id}.log", "w+")
     # TODO: either hard-code 'local-setup' directory, or take from input config.json
-    w = setup_worker(f"tmp/w{i}", config["source"], log, log_config_path)
+    w = setup_worker(f"tmp/w-{id}", config["source"], log, log_config_path)
 
-    print(f"Starting worker {i} in background")
+    print(f"Starting worker {id} in background")
     return w.run_in_background(
         log_file=log, flags=config["flags"], subcommand_flags=config["subcommand_flags"]
     )
--- a/tee-worker/local-setup/development-worker.json
+++ b/tee-worker/local-setup/development-worker.json
@@ -1,6 +1,7 @@
 {
   "workers": [
     {
+      "id": "dev",
       "source": "bin",
       "flags": [
         "--clean-reset",


tee-worker/local-setup/launch.py
```diff
@@ -59,11 +59,12 @@ def setup_worker(work_dir: str, source_dir: str, std_err: Union[None, int, IO],
 
 
 def run_worker(config, i: int, log_config_path):
-    log = open(f"{log_dir}/worker{i}.log", "w+")
+    id = config.get('id', i)
```
felixfaisal, line 62: Why not just use `i` which is being passed as parameter?

felixfaisal (commented):
Thanks @kziemianek . If we are going to use an additional flag for `id`, Can we have an example `json` file for it.

tee-worker/local-setup/launch.py
felixfaisal: Ah, I get it, It is for the scenario, wherein we use `launch.py` multiple times

kziemianek:
> Thanks @kziemianek . If we are going to use an additional flag for `id`, Can we have an example `json` file for it.

It wasn't there because it's totally optionall but i've added it.

tee-worker/local-setup/launch.py
kziemianek: `i` is treated as default but you can explicitly specify another value, even if you don't run the script multiple times at same time so it will be easier to find a logs for a worker if you run multiple at once.

felixfaisal (approved):

tee-worker/local-setup/launch.py
```diff
@@ -59,11 +59,12 @@ def setup_worker(work_dir: str, source_dir: str, std_err: Union[None, int, IO],
...
+    log = open(f"{log_dir}/worker-{id}.log", "w+")
     # TODO: either hard-code 'local-setup' directory, or take from input config.json
-    w = setup_worker(f"tmp/w{i}", config["source"], log, log_config_path)
+    w = setup_worker(f"tmp/w-{id}", config["source"], log, log_config_path)
```
Kailai-Wang, line 65: Hmm is it backwards compatible? We had like `w0`, `w1`...

Can we make it `w-{id}`, and if not present,  `w0`, `w1`...? 

It might affect `deploy.sh`

kziemianek: I cannot find usages of `launch.py` in other scripts, can you guide me?
If something relies on old naming, this might break it.

We can use old convention if `id` is not present.

felixfaisal: I don't think this should affect `deploy.sh` because in it, we are using `opt/worker/` for storing and not using `tmp` in the repo and not using  `launch.py`. This should only affect local development setup, but I will try setting up multiple workers using `deploy.sh` .

Kailai-Wang: I see - yes it only affects local dev

Issue #1707, opened 2023-05-17
Build a sidechain block production metrics


Kailai-Wang:
### Context

Can be a complex topic.

We'd like to measure the time consumption distribution during sidechain block production for the sake of:
- statistics
- monitoring
- better understanding of the bottleneck before further optimisation

In particular, it should include but not limited to the following:
- sidechain block import
- parachain block import, including I/O
- state pre-update, including I/O
- stf execution
- block composition
- after effect
- any others?

It would be great if we could visualise it.

If done, it will supersede the #1611

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

PR #2112, opened 2023-09-08
feat: add metrics for trusted_operation execution

felixfaisal:
This adds metrics for TrustedOperation Execution, Gives a better understanding of which block a TrustedOperation was executed and who requested for it. 
Related to #1707


tee-worker/core-primitives/stf-executor/src/executor.rs
```diff
@@ -139,7 +139,7 @@ where
 			self.node_metadata_repo.clone(),
 		) {
 			Err(e) => {
-				if let Err(e) = self.ocall_api.update_metric(EnclaveMetric::FailedTrustedOperationIncrement) {
+				if let Err(e) = self.ocall_api.update_metric(EnclaveMetric::FailedTrustedOperationIncrement(trusted_call.call.clone())) {
```
kziemianek, line 142: maybe you can pass call name here instead of cloning the whole `trusted_call` ?

felixfaisal: Actually, I was thinking of extending it to use it like an indexer for trusted_call, that is the metric would be something like this 
```
(Block Number, Trusted Operation, Caller) 
```
And I think the trusted_call has those things, but yeah I can still optimize it to pass things that i need 
Created an initial dashboard that looks like this: 
Screenshot 2023-09-08 at 6 34 20 PM

kziemianek: If the implementation is going to need it anyway then there is no need to remove it now in order to introduce it later 🙃

tee-worker/service/src/prometheus_metrics.rs
```diff
@@ -244,6 +257,74 @@ fn handle_stf_call_request(req: RequestType, time: f64) {
...
+fn format_caller(caller_option: Option<Identity>) -> String {
+	if let Some(caller) = caller_option {
+		// Note: This unwrap should be unfallable technically
+		return caller.to_account_id().unwrap().to_string()
+	}
+	"0xDeadBeef".to_string()
+}
+
+// This function will increment the metric with provided label values.
+fn inc_succesfull_trusted_operation_metric(operation: &str, block_str: &str, caller: &str) {
```
kziemianek: If it's going to be only incremented I think `counter` would fit better.

tee-worker/service/src/prometheus_metrics.rs
```diff
@@ -244,6 +257,74 @@ fn handle_stf_call_request(req: RequestType, time: f64) {
...
+}
+
+fn handle_trusted_operation<F>(call: TrustedCall, record_metric_fn: F)
+where
+	F: Fn(&str, &str, &str),
+{
+	let block_str = format!("{}", ENCLAVE_SIDECHAIN_BLOCK_HEIGHT.get());
+	match call {
+		TrustedCall::link_identity(caller, ..) => {
+			record_metric_fn("link_identity", &block_str, &format_caller(Some(caller)));
```
kziemianek: How many gauges we will end up having ?
Looks like using mixture of caller and block will make it grow pretty fast and it's hard for me to imagine the upper limit, am i right ? 🙃

felixfaisal: Yeah that is true, It would grow very quickly. I was looking into how we can prune it every x blocks or something, so that it is only stored on the prometheus database and not in the Enclave.

tee-worker/service/src/prometheus_metrics.rs
felixfaisal: Updated ✅

tee-worker/service/src/prometheus_metrics.rs
```diff
@@ -244,6 +257,56 @@ fn handle_stf_call_request(req: RequestType, time: f64) {
 	observe_execution_time(category, label, time)
 }
 
+// This function will increment the metric with provided label values.
+fn inc_succesfull_trusted_operation_counter(operation: &str) {
+	ENCLAVE_SUCCESFUL_TRUSTED_OPERATION.with_label_values(&[operation]).inc();
```
arieljbirnbaum: minor :nail_care: spelling, "successful" :)

tee-worker/service/src/prometheus_metrics.rs
felixfaisal: So yeah after thinking about this, I've decided not to go with `block` and `caller`, as each label creates a new series, and since there can x and y amounts of block and caller, the gauge would exponentially increase especially at scale. So for now the counter will be used to only identify the `call` type, allowing us to know which `TrustedOperation` is requested most and few other metrics that might be important for the business as well as for us parachain devs monitoring the system for alerts etc. We can include `block` and `caller` once we find a way to emit events using prometheus library, that is it emits and forgets and prometheus DB is responsible for storing it.

tee-worker/service/src/prometheus_metrics.rs
felixfaisal: Updated ✅

tee-worker/service/src/prometheus_metrics.rs
kziemianek: It's still a gauge

```
	static ref ENCLAVE_SUCCESFUL_TRUSTED_OPERATION: IntGaugeVec =
		register_int_gauge_vec!("litentry_worker_enclave_succesful_trusted_operation", "Litentry Succesful Trusted Operation", &["call", "block", "caller"])
			.unwrap();
```

tee-worker/core-primitives/enclave-metrics/Cargo.toml
```diff
@@ -20,7 +21,10 @@ default = ["std"]
 std = [
     "substrate-fixed/std",
     "codec/std",
+    "ita-stf/std",
+
```
Kailai-Wang: extra line

tee-worker/service/src/prometheus_metrics.rs
Kailai-Wang: Yea this value can only increase - so counter should be better

tee-worker/core-primitives/enclave-metrics/Cargo.toml
felixfaisal: Updated ✅

tee-worker/service/src/prometheus_metrics.rs
felixfaisal: My bad, Updated it

kziemianek (approved):

PR #2114, opened 2023-09-11
P-15 Limit the number of allowed workers #1225

### LINKED ISSUE #1225: Limit the number of allowed workers
### Context
related: #1163 

Even when we have a "whitelist of allowed enclaves", we might still want to limit the number of allowed workers and/or the worker accounts.

Extra workers could cause block syncing and production problems before it's thoroughly tested, even though they are running the same enclave.

---

songzhouran:
### Context

resolves #1225


pallets/teerex/src/lib.rs
```diff
@@ -216,6 +217,7 @@ pub mod pallet {
 			let sender = ensure_signed(origin)?;
 			ensure!(ra_report.len() <= MAX_RA_REPORT_LEN, <Error<T>>::RaReportTooLong);
 			ensure!(worker_url.len() <= MAX_URL_LEN, <Error<T>>::EnclaveUrlTooLong);
+			ensure!(Self::enclave_count() < MAX_ENCLAVE_NUMBER, <Error<T>>::ExceedEnclaveNumber);
```
kziemianek: `<=` ?

songzhouran: If it includes '==', after regisitering the enclave, the number will be MAX_ENCLAVE_NUMBER+1.

kziemianek:
There is another call `register_dcap_enclave` which also registers enclaves, why the condition is checked only in `register_enclave` while both calls registers enclaves ? Maybe we should have a separate counter for _non_dcap_ enclaves or check condition in both places?

songzhouran:
Condition is checked in register_enclave. @kziemianek

kziemianek:
> Condition is checked in register_enclave. @kziemianek

Yes, i've noticed that, see my comment 🙃 .
Thanks for adding missing check. 

Could you write unit tests for this ?
Can we adjust the parameter without doing runtime upgrade?

pallets/teerex/src/lib.rs
kziemianek: But the outcome of this call is not always new enclave addition, sometimes it just updates already registered enclave

```
		let enclave_idx = if <EnclaveIndex<T>>::contains_key(sender) {
			log::info!("Updating already registered enclave");
			<EnclaveIndex<T>>::get(sender)
		}
```

Does it mean the update wouldn't be possible once we hit the limit ?

kziemianek: Hi @songzhouran , could you please address this question before resolving the conversation ?

pallets/teerex/src/tests/test_cases.rs
```diff
@@ -1026,3 +1026,31 @@ fn publish_hash_with_too_much_data_fails() {
 		);
 	})
 }
+
+#[test]
+fn add_enclave_works_exceeds_number() {
+	new_test_ext().execute_with(|| {
+		crate::EnclaveCountMax::<Test>::put(0u64);
+		assert_eq!(Teerex::enclave_count_max(), 0);
```
kziemianek: If this is 0, why the test pass ?
I would expect that no enclaves can be registered.

songzhouran: Yes, now no enclaves can be registered.

pallets/teerex/src/lib.rs
songzhouran: Now already registered enclave is excluede from checking.

pallets/teerex/src/tests/test_cases.rs
songzhouran: Now 0 would fail.

pallets/teerex/src/lib.rs
```diff
@@ -122,6 +123,15 @@ pub mod pallet {
...
 	pub type EnclaveCount<T: Config> = StorageValue<_, u64, ValueQuery>;
 
+	#[pallet::type_value]
+	pub fn DefaultEnclaveCountMax() -> u64 {
+		MAX_ENCLAVE_NUMBER
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn enclave_count_max)]
+	pub type EnclaveCountMax<T: Config> = StorageValue<_, u64, ValueQuery, DefaultEnclaveCountMax>;
```
Kailai-Wang: I think normally it's called `MaxEnclaveCount` (means we are setting a maximum value)

pallets/teerex/src/lib.rs
```diff
@@ -640,6 +650,8 @@ pub mod pallet {
 		TooManyTopics,
 		/// The length of the `data` passed to `publish_hash` exceeds the limit.
 		DataTooLong,
+		/// The number of enclave exceed the upper limmit.
```
Kailai-Wang: typo: `limit`

pallets/teerex/src/lib.rs
```diff
@@ -640,6 +650,8 @@ pub mod pallet {
...
+		ExceedEnclaveNumber,
```
Kailai-Wang: Maybe `TooManyEnclaves` to be consistent with other errors, or `ExceedMaxEnclaveCount`?

pallets/teerex/src/tests/test_cases.rs
```diff
@@ -1026,3 +1026,23 @@ fn publish_hash_with_too_much_data_fails() {
...
+			Teerex::register_enclave(
+				RuntimeOrigin::signed(signer5),
+				TEST5_CERT.to_vec(),
+				URL.to_vec(),
+				None,
+				None,
+			),
+			Error::<Test>::ExceedEnclaveNumber
+		);
+	})
```
Kailai-Wang, line 1107: I think we also have to make sure the enclave registration succeeds if we set the `MaxEnclaveCount` to 1 or higher

Kailai-Wang (commented):
The logic looks fine, I only have some complaints about naming and incomplete test cases :)

Kailai-Wang (commented):
It looks good after the change, thanks! : )

I think we only miss a setter extrinsic (there's no way to update/set this number during runtime), after that it's good to go

Kailai-Wang (approved):

Issue #2087, opened 2023-08-31
Typescript packages for client API (parachain+sidechain)


arieljbirnbaum:
### Context

Downstream projects have a hard time keeping up with API changes, especially changes in the runtime and RPC data types. We seek to alleviate that by publishing these in appropriate packages and setting up our build pipeline to keep these up to date.

Note: ideally, we would be able to generate these automatically from the parachain+worker code without any manual intervention. However, there are two limitations:

* The metadata (which we need for type generation) isn't generated statically on build; it has to be retrieved via a live query. That means we need to start a process, and then fetch the metadata from that process in order to proceed. Since the process may be running on non-default ports, we need to make sure these ports are passed to the generation process in order to get the correct metadata version.
* Even worse, the types for RPC methods are [_not_ exported with the metadata](https://github.com/litentry/litentry-parachain/issues/1878) (or anywhere else). As long as we have custom RPC calls, the types for these need to be maintained manually :sadpanda: 

### Tasks

This project involves several steps; individual issues are linked below:

- #2088 
- #2089
- #2090 
- #2091
- #2092
- #2093
- #2095
- #2096

### Acceptance criteria

This issue will be considered done once we have the first successful integration of the published packages with a downstream application.

---

kziemianek:
@Verin1005 @grumpygreenguy is there anything to be done here or this can be closed ?

0xverin:
> can be closed

Thanks.Can be closed.

PR #2122, opened 2023-09-13
ts-tests: make compatible with Node 20 #1886

### LINKED ISSUE #1886: ts-tests: make compatible with Node 20
### Context

As tested by @kziemianek, the `typegen` scripts throw the dreaded `Unknown file extension ".ts"` message under Node 20 (the script returns "successfully", but the build breaks later on :roll_eyes:).

It currently works OK with Node 16 (and we can [ensure](https://github.com/litentry/litentry-parachain/issues/1885) that the scripts run with it), but this version is [reaching end-of-life in September](https://github.com/nodejs/release#release-schedule), so we need to make sure to switch before that. Version 20 will enter LTS in October this year, so we should target that version already (rather than 18).

### Task

- [x] Make sure the build and test scripts run cleanly under Node 20
- [x] Update the nvmrc (assuming #1885 is done)
- [x] Update the docker files accordingly

songzhouran:
### Context
resolves #1886

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/hydrogen
+v20.3.1
--- a/tee-worker/build.Dockerfile
+++ b/tee-worker/build.Dockerfile
@@ -100,7 +100,7 @@ FROM ubuntu:22.04 AS runner
 RUN apt update && apt install -y libssl-dev iproute2 curl
 
 ## ts-tests
-RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash
+RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash
 RUN apt-get install -y nodejs jq
 RUN corepack enable
 RUN corepack prepare yarn@3.6.1 --activate


0xverin:
These modifications are only for the Docker environment, right? AFAIK, ts-tests doesn't yet support Node version 20.x.
So, this might not be considered a resolution for #1886.

kziemianek:
> These modifications are only for the Docker environment, right? AFAIK, ts-tests doesn't yet support Node version 20.x. So, this might not be considered a resolution for #1886?

Yes, the `build` script fails for `parachain-api` workspace

```
kziemianek@kziemianek:~/projects/litentry/litentry-parachain/tee-worker/ts-tests$ node --version
v20.1.0
kziemianek@kziemianek:~/projects/litentry/litentry-parachain/tee-worker/ts-tests$ corepack yarn workspace parachain-api build
...
src/index.ts:30:5 - error TS2305: Module '"../build/interfaces"' has no exported member 'LinkIdentityResult'.

30     LinkIdentityResult,
       ~~~~~~~~~~~~~~~~~~


Found 952 errors in 6 files.

Errors  Files
    16  build/interfaces/augment-api-consts.ts:328
   338  build/interfaces/augment-api-events.ts:35
   190  build/interfaces/augment-api-query.ts:42
   394  build/interfaces/augment-api-tx.ts:55
     1  build/interfaces/index.ts:4
    13  src/index.ts:7
```

So the changes in PR doesn't look like the solution for described problem.

arieljbirnbaum (changes requested):
The Node upgrade affects the build of the chain API; let's complete #2087 first before this to avoid additional delay (and also to have the API build integrated into CI, which will properly gate the upgrade if we break compatibility).

Note that we've already moved from Node 16 to Node 18, which is now entering maintenance mode (as opposed to end-of-life); we should still upgrade to 20 but we're not forced to do it now.

PR #2133, opened 2023-09-14
Make tee-prod build less resource intensive

Kailai-Wang:
### Context

We have problems building the `tee-prod` docker image of parachain:
https://github.com/litentry/litentry-parachain/actions/runs/6173744153/job/16767091897

The error log is:
```
lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```

It's most likely CPU/RAM starvation.

This PR tries to use [thin LTO](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) to reduce the resource usage.
It looks fine in the manual test: https://github.com/litentry/litentry-parachain/actions/runs/6185784593

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -326,7 +326,8 @@ jobs:
 
       - name: Build docker image
         if: |
-          needs.set-condition.outputs.rebuild_parachain == 'true'
+          needs.set-condition.outputs.rebuild_parachain == 'true' &&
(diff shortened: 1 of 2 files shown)


BillyWooo (approved):
Thanks. New things to learn.

kziemianek:
@Kailai-Wang is this short or long term solution ?
I think we should try to find a way to keep well optimized build.
Would increasing runner resources help?

PR #2138, opened 2023-09-16
Remove req ext hash from di sync responses

### LINKED ISSUE #2068: Assert  sync responses `value` and `req_ext_hash` in DI tests
DI ts tests are missing result assertion step, they only check if the trusted operation is included in sidechain block but the responses hold return `value` specific to submitted trusted call and `req_ext_hash`.

Example:

```
export function assertIdentityLinkedResult(
    context: IntegrationTestContext,
    requestIdentifier: string,
    expectedIdentity: LitentryPrimitivesIdentity,
    returnValue: WorkerRpcReturnValue
) {
    const decodedRes = context.api.createType(
        'TrustedOperationResponse',
        returnValue.value
    ) as unknown as TrustedOperationResponse;
    assert.equal(decodedRes.req_ext_hash.toHex(), requestIdentifier);

    const decodedLinkResult = context.api.createType(
        'LinkIdentityResult',
        decodedRes.value
    ) as unknown as LinkIdentityResult;

    assert.equal(
        expectedIdentity.toString(),
        parseIdentity(context.sidechainRegistry, decodedLinkResult.identity, aesKey).toString()
    );
}
```

### LINKED ISSUE #2131: Remove `req_ext_hash` from `TrustedOperationResponse`
Request/response correlation can be done at JSON-RPC 2.0 protocol level, so we dont need `req_ext_hash` in direct invocation's sync responses anymore. 

Scope:
- remove `req_ext_hash` from di sync responses
- use JSON-RPC 2.0 req/resp `id` in ts-tests to match requests and responses

kziemianek:
This PR:

- removes `req_ext_hash` from DI sync responses
- adds logic in ts-tests code to correlate req/resp by `id` field
- adds missing di sync response assertions

closes: #2131 
closes: #2068


tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -368,38 +367,35 @@ export async function assertIdentity(
...
     assert.isNotNull(decodedLinkResult.id_graph);
 }
+
+export function assertSetUserShieldingKeyResult(context: IntegrationTestContext, returnValue: WorkerRpcReturnValue) {
+    const decodedLinkResult = context.api.createType(
+        'SetUserShieldingKeyResult',
+        returnValue.value
+    ) as unknown as SetUserShieldingKeyResult;
+
+    assert.isNotNull(decodedLinkResult.id_graph);
```
0xverin: Just checking for `not null` here might not be sufficient. What we might be looking for is a check against the expected content inside the idgraph. If you're unsure how to do this here, you can add a TODO for now, and we can come back to it later to fill in the details.Thanks!

0xverin: The same issue applies to assertIdentityLinkedResult as well.:)

tee-worker/ts-tests/integration-tests/di_evm_identity.test.ts
```diff
@@ -529,15 +469,24 @@ describe('Test Identity (evm direct invocation)', function () {
...
-            context.tee,
-            context.api,
-            context.mrEnclave,
-            teeShieldingKey,
-            deactivateIdentityCall
-        );
+        const res = await sendRequestFromTrustedCall(context, teeShieldingKey, deactivateIdentityCall);
         assert.isTrue(res.do_watch.isFalse);
         assert.isTrue(res.status.asTrustedOperationStatus[0].isInvalid);
+        assertWorkerError(
```
0xverin, line 486: Can we standardize the assert functions, for example, by adding `export async function ...` and consistently using `await` when calling them? It's just for uniformity, without any specific implication. :p

Just feel that mixing "await" and not using it inconsistently can be a bit confusing

kziemianek: I don't see a point in doing that really.
But I can add extra `async` and `await` keywords here and there but I would like to hear more voices on that, maybe @grumpygreenguy ?

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
kziemianek: True, I will try to add more precise assertions in this PR.

0xverin: You can use [assertInitialIdGraphCreated](https://github.com/litentry/litentry-parachain/blob/remove-req-ext-hash/tee-worker/ts-tests/integration-tests/common/utils/assertion.ts#L59) as a reference.

kziemianek: Thanks 👍🏼

BillyWooo (commented):
could this be a breaking change to IDHub?

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
kziemianek: I've added assertions

kziemianek:
> could this be a breaking change to IDHub?

Potentially yes, but in practice AFAIK they haven't started adjusting yet. I've added label, thanks 👍🏼

0xverin (approved):
Thanks!LGTM.

BillyWooo (approved):
LGTM

tee-worker/ts-tests/integration-tests/di_evm_identity.test.ts
arieljbirnbaum: The main issue I see here is the risk of having an async assertion that we're not awaiting, e.g. 

```
    const events = await eventsPromise;
    assertFailedEvent(...); // oops, forgot await
```

If we wanted to standardize, I'd actually push for making all assertions synchronous :P
It'll take some dependency injection to get it to work tho

kziemianek: Thanks for your voice @grumpygreenguy .

PR #2145, opened 2023-09-19
Use accountNextIndex instead of system.account.nonce

### LINKED ISSUE #1425: Use `accountNextIndex` instead of `system.account.nonce`
### Context

See https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-take-the-pending-tx-pool-into-account-in-my-nonce

Current `get_nonce_of`:

https://github.com/litentry/litentry-parachain/blob/d9da36695553531b10209e8f6667e4e526c639aa/tee-worker/core-primitives/node-api/api-client-extensions/src/account.rs#L34-L36

We could also keep the original one 

---

outofxxx:
fix #1425 
* Add `get_account_next_index`


BillyWooo (approved):
LGTM

PR #2153, opened 2023-09-22
Merging of HOLDER VCs

### LINKED ISSUE #1813: Merging of HOLDER VCs

outofxxx:
fix: #1813 

Merge `a4/a7/a10/a11` assertions into `HoldingTime`

---
Add `HoldingTime` type
HoldingTime(AmountHoldingTimeType, ParameterString),    // (AmountHoldingTimeType, minimum_amount)

```
pub enum AmountHoldingTimeType {
	LIT,
	DOT,
	WBTC,
	ETH,
}
```


Kailai-Wang:
Actually I thought we would keep the interface with the client, which means we still have `a4/a7/a10/a11` while changing the internal implementation so that we make the interface backwards compatible.

It's fine to have a unified interface but we would need to communicate with IDH - have you done that?

BillyWooo (dismissed):
Nice work. I only review the RUST part. This will be a breaking change and deserves more people to review. 
The backwards compatibility is important as Kai mentioned.

outofxxx:
Replace this with another pr

**Please do not delete this branch at this time**

PR #2155, opened 2023-09-22
Intentionally fail sidechain block production

### LINKED ISSUE #1621: Add ways to intentionally fail sidechain block production in test
### Context

It can be useful if we can fail sidechain block production on purpose in the test: then we can test the handling in various phases during block production: e.g.
- before aura
- on_slot
- after_aura
- send and broadcast block

Additional switches need to be added to the code to make it possible.
It's a nice-to-have feature request.

---

kziemianek:
Adds a way to intentionally fail sidechain at configured slot.

The fail can happen:
- before the worker composes a block
- after the worker composed a block but before broadcasting it

fixes: #1621


BillyWooo (approved):
LGTM

PR #2156, opened 2023-09-22
Apply caching for enclave CI build 

### LINKED ISSUE #2140: (Re-)Apply cache in CI builds
### Context

We have been suffering from the long building cycle of parachain/enclave build in CI - even with a minute change in source code we need to go through the whole compilation process.

We had a few attempts to accelerate the whole process but they weren't so successful:
- https://github.com/litentry/litentry-parachain/issues/259
- https://github.com/litentry/litentry-parachain/issues/659
- https://github.com/litentry/litentry-parachain/issues/935

But things might have changed and a retry is worth it. Ideas:
- use buildx + cache mounts (with sccache) with docker, or
- use plain sccache with binary build

Priority is set to `high` because it affects (benefits) all upcoming builds

---

Kailai-Wang:
### Context

resolves #2140 

This PR applies the caching mechanism to the enclave builds in CI after some research:
- as written in the issue, buildx GHA cache + cache mount won't work, as the cache mount won't be considered as docker layers and thus can't be persisted via buildx cache
- [buildkit-cache-dance](https://github.com/reproducible-containers/buildkit-cache-dance) doesn't work reliably, sometimes I get empty sccache folder - see the comments in `ci.yml`
- [cargo-chef](https://github.com/LukeMathWalker/cargo-chef) is proven to be unsuitable for all cases

Here's an informative summary of the problem and its background, including possible solutions(workarounds): https://hackmd.io/@kobzol/S17NS71bh

In this PR, a plain copy in/out of cached data to/from docker image is performed and persisted via the official `action/cache` GHA. Things that are cached:
- `$CARGO_HOME/registry/index`
- `$CARGO_HOME/registry/cache`
- `$CARGO_HOME/git/db`
- `sccache` artifacts

Apart from that, this PR uses a few more techs to make it possible/more efficient:
- free up disk space
- a local-stash to store the caching only
- local registry pushing
- compressed docker export

Working example: this is a re-run with small changes to the original code:
https://github.com/litentry/litentry-parachain/actions/runs/6274181997/job/17039285320

Compilation was done within 15min, and this is the sccache stats:
image

There are still a few TODOs, I'll create follow-up issues.


.github/workflows/ci.yml
```diff
@@ -312,15 +308,11 @@ jobs:
...
-        run: |
-          sudo rm -rf /usr/local/lib/android
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /opt/ghc
-          sudo rm -rf /usr/local/share/boost
-          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: true
+          swap-storage: false
```
kziemianek, line 315: Have you considered setting `large-packages` to `false` ?
We can exchange few GB for run minutes.

I've set it to false, here are the results:

Comparsion:
old: https://github.com/litentry/litentry-parachain/actions/runs/6296965077/job/17095165858
new: https://github.com/litentry/litentry-parachain/actions/runs/6296504239/job/17091914474


parachain-build-dev:
old: 9m 41s
new: 1m 44s

parachain-build-tee-prod:
old: 3m 7s
new: 16s

tee-build:
old: 2m 9s
new: 19s

**Total time saved: 12m 38s**

kziemianek (approved):
Great work. In general it looks fine to me, just one small comment which can be resolved in separate PR if really needed..

.github/workflows/ci.yml
BillyWooo: woo. impressive.

BillyWooo (approved):
👍 great work.

.github/workflows/ci.yml
Kailai-Wang: Thanks! I kept that because I fell short in disk after building the docker (once only 2-3 G left), but it seems to suffice in your case.

I'll change it in my next PR

PR #2159, opened 2023-09-25
feat: use threadpools for STF Tasks 

### LINKED ISSUE #1958: A thread pool for the STF task handler
### Context

Currently, some STF tasks are handled asynchronously, there's one extra thread backed up which receives tasks from the main thread (where the top pool resides), processes them and posts the result back to the main thread.

However, users might still experience a big delay if they are e.g. requesting the _holder_ VCs simultaneously. And one http request latency will still affect the subsequent requests.

Shall we extend it to a thread-pool-like structure? So that we can have multiple worker threads that parallelise the task processing.

---

felixfaisal:
Resolves #1958 
Utilizing a threadpool offers a notable advantage over executing tasks sequentially, particularly in scenarios that are I/O-bound or involve a combination of I/O and CPU operations. One of the primary strengths of a threadpool is its ability to distribute tasks effectively. When faced with tasks of varying durations, a threadpool can evenly distribute the workload across its threads, ensuring that no single thread is disproportionately burdened.

However, this concurrency presents a challenge: when multiple tasks attempt to submit trusted calls, nonce conflicts can arise.
To address this challenge, I've implemented an MPSC channel. By routing submissions through a single receiver thread, we can maintain the asynchronicity of task completions while ensuring that trusted call submissions don't encounter nonce conflicts. By having tasks only signal their completion through these channels, rather than submitting trusted calls directly, we've effectively introduced a layer that ensures ordered submissions, resolving the nonce conflict issue.


tee-worker/litentry/core/stf-task/receiver/src/handler/mod.rs
```diff
@@ -17,16 +17,28 @@
...
 	type Result;
-	fn start(&self) {
+	fn start(&self, sender: std::sync::mpsc::Sender<(ShardIdentifier, H256, TrustedCall)>) {
 		match self.on_process() {
-			Ok(r) => self.on_success(r),
-			Err(e) => self.on_failure(e),
+			Ok(r) => self.on_success(r, sender),
+			Err(e) => self.on_failure(e, sender),
 		}
+		// sender.send(0_i32).unwrap();
```
kziemianek: Can be removed.

tee-worker/litentry/core/stf-task/receiver/src/lib.rs
```diff
@@ -165,35 +167,60 @@ pub fn run_stf_task_receiver<K, A, S, H, O>(
...
 	let receiver = stf_task_sender::init_stf_task_sender_storage()
 		.map_err(|e| Error::OtherError(format!("read storage error:{:?}", e)))?;
 
+	let (sender, to_receiver) = std::sync::mpsc::channel::<(ShardIdentifier, H256, TrustedCall)>();
+
+	// Spawn thread to handle received tasks
+	let context_for_thread = context.clone();
+	std::thread::spawn(move || loop {
+		if let Ok((shard, hash, call)) = to_receiver.recv() {
+			error!("Submitting trusted call to the pool");
```
kziemianek: Why this is an `error` ?

kziemianek (approved):
Looks ok. 
Would be good to have one more approval on this one.

Kailai-Wang (approved):
Looks good from my side, thanks!

I'm only thinking if we should have some tests for it - e.g. having a `ThrottledRequestType` where the handling logic simply sleeps for X seconds before it returns.

But that doesn't belong to the scope of this PR ofc

tee-worker/litentry/core/stf-task/receiver/src/handler/mod.rs
felixfaisal: Removed ✅

tee-worker/litentry/core/stf-task/receiver/src/lib.rs
felixfaisal: Updated to `info`

kziemianek: @felixfaisal can this be `debug` or `trace` or maybe even removed?

Imho this log entry doesn't provide any useful information and will only spawn logs.

PR #2101, opened 2023-09-04
Extract api definitions

0xverin:
### Context

resolves #2088 

Note: Although this PR involves many file modifications, it's because there were folder movements.Please pay particular attention to the files `src/index.ts` and `dist/src/index.d.ts` in the` parachain-api/sidechain-api` folder.

PR #2164, opened 2023-09-26
Replace yarn with pnpm

arieljbirnbaum:
### Context

Needed to unblock #2101
Implements P-141

### Notable Changes

- `corepack enable pnpm` is used to add `pnpm` to the `PATH` so it can be invoked directly (i.e. without `corepack pnpm`).
- Looks like `only-allow pnpm` does _not_ prevent `npm install` from running :( `yarn install` OTOH is already blocked by the `packageManager` prop. Hence, dropping the `preinstall` hook.
- `pnpm run` doesn't automatically find scripts in workspace (i.e. like `yarn run` does if the script name includes a colon); we need to pass the workspace explicitly (with the `--filter` option)

### TODO

- [ ] Regenerate API defs (if needed)


scripts/ts-utils/move-vcregistry-snapshot.ts
arieljbirnbaum, line 1: Just updated the comment on top; prettier took over after that :D

Kailai-Wang (approved):
I'm fine with it as long as CI is happy

Feel free to resolve the conflict and merge

.github/workflows/ci.yml
```diff
@@ -475,6 +478,8 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
+          build-contexts: local-stash=docker-image://localhost:5000/local-stash:latest
+          load: true
```
Kailai-Wang, line 482: We don't need it anymore - but let's delete it in the next PR :)

0xverin: same as line 492&&493

PR #2166, opened 2023-09-27
Cleanup deps

kziemianek:
Some dependencies are not needed anymore, some can be trated as dev dependencies only, for example if they are used in tests only.

This PR tries to tidy `Cargo.toml` files a little bit which might eventually lead to:
* better developer experience
* shorten build times

I've applied [rustc lint rule](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-crate-dependencies) to modules one by one to see what deps are not used.

This rule cannot be toggled on because it might produce false-positives depending on specified build features.


Kailai-Wang:
Thank you! I'm wondering if there's any tool for it? Not necessarily integrated in CI but we can use it locally at least

kziemianek:
> Thank you! I'm wondering if there's any tool for it? Not necessarily integrated in CI but we can use it locally at least

Yes 🙃 previously I tried with rustc lint rule but I found out [machete](https://crates.io/crates/cargo-machete) to be really acurate. I had also exerimented with [udeps](https://crates.io/crates/cargo-udeps) but without a success.

tee-worker/core/parentchain/indirect-calls-executor/Cargo.toml
```diff
@@ -61,8 +53,8 @@ itp-test = { path = "../../../core-primitives/test" }
 itp-top-pool-author = { path = "../../../core-primitives/top-pool-author", features = ["mocks"] }
 itc-parentchain-test = { path = "../../../core/parentchain/test" }
 
-# litentry
-hex = "0.4.2"
+# scs/integritee
+substrate-api-client = { git = "https://github.com/scs/substrate-api-client.git", branch = "polkadot-v0.9.39", default-features = false }
```
BillyWooo, line 57: @kziemianek did you manually switch this line or by the tool?

kziemianek: Manually.

This module is referenced only in tests.

Kailai-Wang (approved):
Looks good from my side 👍

PR #2170, opened 2023-09-28
Use ubuntu 20 as the docker base image #2124

### LINKED ISSUE #2124: Use ubuntu 20 as the docker base image
### Context

We are currently using [ubuntu 22.04](https://github.com/litentry/litentry-parachain/blob/33bb70e7d46f695ac92d7da15a31562dd52b9cf3/docker/Dockerfile#L24) as the base image for packaging the parachain binary.

It works fine in CI but we have problems when:
- we launch the parachain on dev server 
- we deploy the production

The builder env needs to double-checked too (`rust:latest`)

As in both cases we copy the binary from the docker image and run it locally. We'll get GLIBC error when running on a ubuntu20 host (via `make launch-docker-rococo`):
```
stderr: "generated-rococo/litentry-collator: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by generated-rococo/litentry-collator)
```

---

songzhouran:
### Context

resolves https://github.com/litentry/litentry-parachain/issues/2124

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,7 +6,7 @@ ARG PROFILE
 # https://hub.docker.com/_/rust
 # our host is based on bullseye/sid => rust:latest
 # ==========================
-FROM rust:1.20.0 as builder
+FROM rust:1.40.0 as builder
 
 WORKDIR /litentry
 COPY . /litentry
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,7 +6,7 @@ ARG PROFILE
 # https://hub.docker.com/_/rust
 # our host is based on bullseye/sid => rust:latest
 # ==========================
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG PROFILE
 # https://hub.docker.com/_/rust
 # our host is based on bullseye/sid => rust:latest
 # ==========================
-FROM rust:latest as builder
+FROM rust:1.64 as builder
```
Kailai-Wang: It's not a good idea to use older rust toolchain (base image), I think `rust:bullseye` is enough, see `Image Variants` in https://hub.docker.com/_/rust

Let me change it directly :)

kziemianek (approved):
Looks fine!

PR #2182, opened 2023-10-08
P-75 : Merging of HOLDER VCs

### LINKED ISSUE #1813: Merging of HOLDER VCs

outofxxx:
fix: #1813 

Fixes P-75


BillyWooo:
Somewhere we had a discussion about keeping the backwards compatibility (keep the old A* logic as well).

outofxxx:
> Somewhere we had a discussion about keeping the backwards compatibility (keep the old A* logic as well).

Keep the old logic? It is precisely because their logic is the same that they need to be merged, currently retaining the old interface.

BillyWooo:
> > Somewhere we had a discussion about keeping the backwards compatibility (keep the old A* logic as well).
> 
> Keep the old logic? It is precisely because their logic is the same that they need to be merged, currently retaining the old interface.

My mistake. What I mean is only about the compatibility.

tee-worker/litentry/core/assertion-build/src/holding_time.rs
```diff
@@ -0,0 +1,350 @@
...
+// There's an issue for this: https://github.com/litentry/litentry-parachain/issues/1655
+//
+// There is a problem here, because TDF does not support mixed network types,
+// It is need to request TDF 2 (substrate+evm networks) * 14 (ASSERTION_FROM_DATE) * addresses http requests.
+// If TDF can handle mixed network type, and even supports from_date array,
+// so that ideally, up to one http request can yield results.
+fn do_build(
+	identities: Vec<(Web3Network, Vec<String>)>,
+	htype: &AmountHoldingTimeType,
+	q_min_balance: &String,
```
Kailai-Wang: Normally we use `&str` instead of `&String`

Kailai-Wang (approved):
Looks good, thanks! I didn't go line by line but I assume it should be copied from the old code

But we are missing the regression test to tell if it works correctly or not :(

tee-worker/litentry/core/assertion-build/src/holding_time.rs
outofxxx: Could those [testcase](https://github.com/litentry/litentry-parachain/blob/bbcaaabf07f8fa80c6942d67189c01b7134e737c/tee-worker/litentry/core/assertion-build/src/holding_time.rs#L234-L290) be used as regression tests?

PR #2186, opened 2023-10-10
P-150 feat: unit test for threadpool behaviour

### LINKED ISSUE #2165: Create artificial lag for testing threadpool behavior
### Context
This was mentioned in https://github.com/litentry/litentry-parachain/pull/2159
We need to introduce artificial lag onto the system and see if behaves correctly based on threadpools, One way would be  having a `ThrottledRequestType` where the handling logic simply sleeps for X seconds before it returns.

felixfaisal:
Resolves #2165 

Some notes: 
1. I had to implement our own AuthorApiMock, so that `watch_top` sends the trusted operation via a channel and we can use this channel to assert the order of operations. 
2. I've introduced a hardcoded sleep into assertion `A1` under the `test` feature to create artificial lag during testing so that I can demonstrate the expected behaviour of threadpool 
3. This can be made even more modular to perform more tests but it could be a different issue


Kailai-Wang:
Can you try to link the Linear issue too :)

tee-worker/litentry/core/stf-task/receiver/src/test.rs
```diff
@@ -0,0 +1,62 @@
...
+				if let TrustedCall::request_vc_callback(_, _, assertion, ..) =
+					trusted_call_signed.call
+				{
+					assert_eq!(expected_output.remove(0), assertion);
+				},
+			_ => {
+				// Do nothing
+			},
+		}
+		if expected_output.len() == 0 {
```
kziemianek, line 58: If something goes wrong, i bet this may not be reached and the test might get stuck in the while loop, would it be timedout ?

felixfaisal: Yeah this is possible if something has been changed drastically, I think on the safer side, I will add a timeout, The while loop exits within 60 seconds(or less) else we fail with timeout.

felixfaisal: Done ✅

Kailai-Wang (commented):
Thanks!

The logic looks fine, I've got two questions:
- can we simply implement `watch_top` in AuthorApiMock? We are free to change it :) - to avoid repeating the code
- can we have a test case where we have more than `n_workers` tasks (can be delayed A1) and see them returning in batch? Like the `n_workers + 1`th task should be processed only after we have vacant thread

tee-worker/core-primitives/top-pool-author/src/mocks.rs
```diff
@@ -39,15 +39,23 @@ use itp_types::ShardIdentifier;
 use jsonrpc_core::{futures::future::ready, Error as RpcError};
 use lazy_static::lazy_static;
 use sp_core::{blake2_256, H256};
+#[cfg(not(feature = "std"))]
```
kziemianek: Sgx deps should be enabled under `sgx` feature like this

```
#[cfg(feature = "sgx")]
use std::sync::SgxMutex as Mutex;
```

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
```diff
@@ -57,8 +57,13 @@ where
 		// create the initial credential
 		// TODO: maybe we can further simplify this
 		let mut credential = match self.req.assertion.clone() {
-			Assertion::A1 => lc_assertion_build::a1::build(&self.req),
-
+			Assertion::A1 => {
+				#[cfg(test)]
+				{
+					std::thread::sleep(core::time::Duration::from_secs(2));
```
kziemianek: Maybe such sleep can be performed in `AchainablePost` mock for example?

kziemianek (commented):
Looks good in general. I left two comments. I think the one regarding feature toggle needs to be addresed at least.

Kailai-Wang:
We need to resolve the conflict and then we are good to go I guess?

BillyWooo (approved):
Let's solved the conflicts and merge.

PR #2188, opened 2023-10-11
Clean up disk between clippys in CI

Kailai-Wang:
### Context

As topic, see https://github.com/litentry/litentry-parachain/actions/runs/6480387549

Please note `jlumbroso/free-disk-space@main` doesn't make sense because the job is run inside a container.
There seems to be no built-in way to run steps before containerised jobs: https://github.com/actions/runner/issues/812

Another option would be using `docker run` manually


.github/workflows/ci.yml
```diff
@@ -265,6 +265,17 @@ jobs:
           cargo clippy --release --features evm -- -D warnings
           echo "::endgroup::"
 
+      - name: Clean up disk
+        working-directory: ./tee-worker
+        run: |
+          echo "::group::Show disk usage"
+          df -h .
+          echo "::endgroup::"
+          cargo clean --profile release
```
BillyWooo, line 274: why we need `--profile release` here? Can we simply clean all?

Kailai-Wang:
A test run is here: https://github.com/litentry/litentry-parachain/actions/runs/6482833287/job/17603195260

.github/workflows/ci.yml
Kailai-Wang: That's actually the same - we have release profile only anyway

BillyWooo (approved):

PR #2189, opened 2023-10-12
Add `id-graph` command in CLI

Kailai-Wang:
### Context

As topic, it's a getter to retrieve the IDGraph from CLI, only under non-prod.

See usage example in `id_graph.rs`

The byte display will be tackled along with P-161


BillyWooo (commented):
Didn't figure out how this command is only enabled with non-prod mode. Or do you mean 'cli' by default is only for non-prod usage?

Kailai-Wang:
> Didn't figure out how this command is only enabled with non-prod mode. Or do you mean 'cli' by default is only for non-prod usage?

It's set on one level higher, see the sig verification in `getter.rs`

kziemianek:
It's only cli command, the worker implementation remains unchanged.
Currently Alice can get any id graph in non prod mode.

Looks fine 👍🏼

kziemianek (approved):

PR #2190, opened 2023-10-12
P 160 callback metrics for enclave

felixfaisal:
This adds metrics to calculates the time it takes to receive a callback


tee-worker/service/src/prometheus_metrics.rs
```diff
@@ -209,6 +219,77 @@ impl ReceiveEnclaveMetrics for EnclaveMetricsReceiver {
...
 }
 
+fn handle_callback_calls(call: &TrustedCall) {
+	if let Ok(mut hashmap) = STF_REQUEST_EXT_HASH.lock() {
+		match call {
+			TrustedCall::link_identity(_, _, _, ValidationData::Web2(_), _, _, hash) => {
+				hashmap.insert(*hash, std::time::Instant::now());
+			},
+			TrustedCall::link_identity_callback(_, _, identity, _, hash) => {
+				if let Some(time) = hashmap.remove(hash) {
```
kziemianek, line 229: I think this will result it stale entries if we run multiworker setup with synced trusted operation pool.
The map will only grow larger and larger because callbacks can be eventually processed by different workers.

kziemianek: This should be ultimately cleaned up [here](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/sidechain/consensus/aura/src/block_importer.rs#L126) if we want to really measure that. Given that TOPs can be processed by different workers I'm not sure this metric will produce valuable insights.

felixfaisal: But does our current implementation support another worker processing the STF Tasks?

kziemianek: No, but I think we should take that into account. What do you think?

felixfaisal: I don't really have a preference here, but I'd argue we wait for concrete implementation that we can test with multiple workers and have a separate issue for it. We still have to somehow measure how much time it takes for a request to be completed.

Kailai-Wang (commented):
I can see the motivation and idea behind this.

What @kziemianek said makes sense, this is (almost) guaranteed to be broken once we launch the multi-worker setup :)

Another thing is it doesn't necessarily fulfill the purpose "_to get a more accurate sense of the time it takes to fulfill a user's request_" - even for single worker, as the callback relies on the  `SuccessfulTrustedOperationIncrement` callback. While it for most of the cases means the user will get response soon, it's not always the case. STF execution is just one step in the block production: imagine the STF is executed OK but for whatever reason the block isn't composed in the end, the top will be returned to the pool to be picked up in the next round.

If you want to measure the time slot that reflects on the UE, I think the time difference between a top enters and gets removed from the pool is more accurate. But it requires a bigger change.

Maybe we should put this issue into the icebox and come back to it when we have more resource 🤔

felixfaisal:
Closing this PR and moving the Issue to `ice box` or `tech-debt`

PR #2192, opened 2023-10-12
chore: inheriting proxy env vars from host when building with docker

AlseinX:
Modified the build script to pass `HTTP_PROXY` environment variables from the host machine to the build container, so that developers under GFW would use their custom proxy to build the project.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,5 @@
 # global args that are used across multiple stages
 ARG PROFILE
-ARG HTTP_PROXY
 
 # ==========================
 # stage 1: build
@@ -8,10 +7,6 @@ ARG HTTP_PROXY
 # our host is based on bullseye/sid (ubuntu 20)
 # ==========================
 FROM rust:bullseye as builder
-ENV HTTP_PROXY=$HTTP_PROXY
-ENV HTTPS_PROXY=$HTTP_PROXY
-ENV http_proxy=$HTTP_PROXY
-ENV https_proxy=$HTTP_PROXY
 
 WORKDIR /litentry
 COPY . /litentry
@@ -27,10 +22,6 @@ RUN cargo build --locked --profile $PROFILE $BUILD_ARGS
 # stage 2: packaging
 # ==========================
 FROM ubuntu:20.04
-ENV HTTP_PROXY=$HTTP_PROXY
-ENV HTTPS_PROXY=$HTTP_PROXY
-ENV http_proxy=$HTTP_PROXY
-ENV https_proxy=$HTTP_PROXY
 LABEL maintainer="Trust Computing GmbH <info@litentry.com>"
 
 ARG PROFILE
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,6 @@
 # global args that are used across multiple stages
 ARG PROFILE
+ARG HTTP_PROXY
 
 # ==========================
 # stage 1: build
@@ -7,6 +8,10 @@ ARG PROFILE
 # our host is based on bullseye/sid (ubuntu 20)
 # ==========================
 FROM rust:bullseye as builder
+ENV HTTP_PROXY=$HTTP_PROXY
+ENV HTTPS_PROXY=$HTTP_PROXY
+ENV http_proxy=$HTTP_PROXY
+ENV https_proxy=$HTTP_PROXY
 
 WORKDIR /litentry
 COPY . /litentry
@@ -22,6 +27,10 @@ RUN cargo build --locked --profile $PROFILE $BUILD_ARGS
 # stage 2: packaging
 # ==========================
 FROM ubuntu:20.04
+ENV HTTP_PROXY=$HTTP_PROXY
+ENV HTTPS_PROXY=$HTTP_PROXY
+ENV http_proxy=$HTTP_PROXY
+ENV https_proxy=$HTTP_PROXY
 LABEL maintainer="Trust Computing GmbH <info@litentry.com>"
 
 ARG PROFILE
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -22,6 +27,10 @@ RUN cargo build --locked --profile $PROFILE $BUILD_ARGS
 # stage 2: packaging
 # ==========================
 FROM ubuntu:20.04
+ENV HTTP_PROXY=$HTTP_PROXY
+ENV HTTPS_PROXY=$HTTP_PROXY
+ENV http_proxy=$HTTP_PROXY
+ENV https_proxy=$HTTP_PROXY
```
Kailai-Wang: Do we need that in the packaging? (for `apt-get`?)

scripts/build-docker.sh
```diff
@@ -51,6 +51,9 @@ echo "Building ${GITUSER}/${GITREPO}:${TAG} docker image ..."
 docker build ${NOCACHE_FLAG} --pull -f ./docker/Dockerfile \
     --build-arg PROFILE="$PROFILE" \
     --build-arg BUILD_ARGS="$ARGS" \
+    --build-arg HTTP_PROXY="${HTTP_PROXY//localhost/host.docker.internal}" \
+    --add-host=host.docker.internal:host-gateway \
+    --network host \
```
Kailai-Wang, line 60: These seem to be redundant, is `--network host` not enough? I haven't tested it though

Kailai-Wang (commented):
It's great that you proactively submit PRs to tackle the problem you met 👍 

Have you tried this? 
https://docs.docker.com/network/proxy/

Like using a customised docker config for it - as it doesn't apply to everyone, maybe it's a good idea to keep the change local too

Kailai-Wang:
It also seems that even without definition inside Dockerfile, the env HTTP_RPOXY can be recognised:

```
docker build --no-cache --progress=plain  --build-arg HTTP_PROXY="http://proxy.example.com:3128" -t tmp -<<EOF
FROM alpine
RUN env | grep -i _PROXY
EOF

#0 building with "desktop-linux" instance using docker driver

#1 [internal] load .dockerignore
#1 transferring context: 2B done
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 74B done
#2 DONE 0.0s

#3 [internal] load metadata for docker.io/library/alpine:latest
#3 DONE 0.7s

#4 [1/2] FROM docker.io/library/alpine@sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978
#4 CACHED

#5 [2/2] RUN env | grep -i _PROXY
#5 0.160 http_proxy=http://proxy.example.com:3128
#5 0.160 HTTP_PROXY=http://proxy.example.com:3128
#5 DONE 0.2s

#6 exporting to image
#6 exporting layers 0.0s done
#6 writing image sha256:61266719edeac635a4e8a606d67321e8484c20224296759db620c7f617e108c2 done
#6 naming to docker.io/library/tmp done
#6 DONE 0.0s
```

docker/Dockerfile
AlseinX: Yes, the default debian source is extremely slow under the GFW
```
Step 5/19 : RUN apt-get update && apt-get install -yq clang libclang-dev cmake protobuf-compiler
 ---> Running in 5220bc661d7c
Get:1 http://deb.debian.org/debian bullseye InRelease [116 kB]
Get:2 http://deb.debian.org/debian-security bullseye-security InRelease [48.4 kB]
Err:2 http://deb.debian.org/debian-security bullseye-security InRelease
  Connection timed out [IP: 2a04:4e42:8c::644 80]
Err:3 http://deb.debian.org/debian bullseye-updates InRelease
  Connection failed [IP: 2a04:4e42:8c::644 80]
```

AlseinX:
> It also seems that even without definition inside Dockerfile, the env HTTP_RPOXY can be recognised:

I performed a test removing the definitions inside Dockerfile, within the server under GFW.
It fails to recognise the proxy settings and seems not to work exactly as documented.

scripts/build-docker.sh
AlseinX: I had got the same concern and tested. It would stop working without `--add-host=host.docker.internal:host-gateway`.

Kailai-Wang:
> I performed a test removing the definitions inside Dockerfile, within the server under GFW.
It fails to recognise the proxy settings and seems not to work exactly as documented

This is utterly strange.
It should at least get those envs assigned inside your docker, what's output if you add `RUN env | grep _PROXY` in Dockerfile?

Whether it can actually access host proxy is another story.

scripts/build-docker.sh
Kailai-Wang: The following worked for me locally:

1. start a simple http server locally at 8000

2. create a Dockerfile:
```
FROM ubuntu:22.04
RUN apt-get update && apt-get install -yq curl
RUN env | grep -i _PROXY || true
RUN curl -I host.docker.internal:8000
```

3. build the docker:
```
docker build --no-cache --progress=plain --network host -t tmp .
```

It will show `curl` succeeds:

```
#7 0.072 HTTP/1.0 200 OK
#7 0.072 Server: SimpleHTTP/0.6 Python/3.11.6
#7 0.072 Date: Fri, 13 Oct 2023 08:18:42 GMT
#7 0.072 Content-type: text/html; charset=utf-8
#7 0.072 Content-Length: 474
```

Is that also the case from your side?

AlseinX:
I tried and found that `--build-arg HTTP_PROXY="..."` only sets exact `HTTP_PROXY`, and others are unset, therefore it does not recognise, which means I have to pass all of these env vars with `--build-arg` within the build script
I have recommitted to remove modification on dockerfile.

Kailai-Wang (approved):
Thanks!

Kailai-Wang:
In my test `--network=host` is enough for the container to access services on the host via e.g. `localhost:<port-number>`, but anyway it's not that important

PR #2199, opened 2023-10-17
Allow requesting VCs from cli

Kailai-Wang:
### Context

This PR adds a `request-vc` command to CLI, see `request_vc.rs` for more usage examples.

Please note the command doesn't print the generated VC in the stdout, this is due to old code in CLI trusted operation handling.
It will be addressed in a follow-up issue P-177

### DIFF
--- a/tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
+++ b/tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
@@ -1,0 +1,74 @@
+// Copyright 2020-2023 Trust Computing GmbH.
+// This file is part of Litentry.
+//
+// Litentry is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Litentry is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{
+	get_layer_two_nonce,
+	trusted_cli::TrustedCli,
+	trusted_command_utils::{get_identifiers, get_pair_from_str},
+	trusted_operation::perform_trusted_operation,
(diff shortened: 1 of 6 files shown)


tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
```diff
@@ -0,0 +1,349 @@
...
+				to_para_str(&arg.guild_id),
+				to_para_str(&arg.channel_id),
+				to_para_str(&arg.role_id),
+			),
+			Command::A4(arg) => Assertion::A4(to_para_str(&arg.minimum_amount)),
+			Command::A6 => Assertion::A6,
+			Command::A7(arg) => Assertion::A7(to_para_str(&arg.minimum_amount)),
+			Command::A8(arg) => {
+				let networks: Vec<Web3Network> =
+					arg.networks.iter().map(|n| n.as_str().try_into().unwrap()).collect();
```
kziemianek: It would be good to improve error handling here. If user provides wrong network it panics like this:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', cli/src/trusted_base_cli/commands/litentry/request_vc.rs:241:71
```

Consider other places too.

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
```diff
@@ -0,0 +1,349 @@
...
+				AchainableCommand::Token(arg) =>
+					Assertion::Achainable(AchainableParams::Token(AchainableToken {
+						name: to_para_str(&arg.name),
+						chain: arg.chain.as_str().try_into().unwrap(),
+						token: to_para_str(&arg.token),
+					})),
+			},
+		};
+
+		println!("assertion: {:?}", assertion);
```
kziemianek: This can be probably removed.

kziemianek (approved):
Checked it locally, works fine.

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
Kailai-Wang: Thanks - I removed it

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
Kailai-Wang: Thanks - I changed it to `expect(...)` to give more error info

PR #2200, opened 2023-10-18
Retrieve mrenclave automatically in cli

felixfaisal:
Some notes: 
1. I've first made the `Mrenclave` as an `Option<String>` in the Cli arguments 
2. In order to retrieve the `Mrenclave` from a running worker, I'm creating the `DirectApi` client and using it to retrieve the MRENCLAVE 
3. In order to initialise `DirectApi` client, We need `Cli` arguments, So I've updated most functions to include this argument 

TODO: 
- [ ] Perform E2E test for all commands with MRENCLAVE provided and without MRENCLAVE

### DIFF
--- a/tee-worker/cli/src/benchmark/mod.rs
+++ b/tee-worker/cli/src/benchmark/mod.rs
@@ -118,10 +118,10 @@ impl BenchmarkCommand {
 			self.random_wait_before_transaction_min_ms,
 			self.random_wait_before_transaction_max_ms,
 		);
-		let store = LocalKeystore::open(get_keystore_path(trusted_args), None).unwrap();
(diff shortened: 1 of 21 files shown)


kziemianek (approved):
I wasn't able to test it locally due to reported problems with parachain launch scripts but the code looks fine. 

Very nice improvement ❤️

Kailai-Wang (approved):
Looks good - thanks! 

Do you plan to adjust CLI shell scripts in this PR or another? If the latter let's create a follow-up issue :)

Another small thing is maybe we can remove `P 163` from the title - the branch name is sufficient to link the linear issue. If you really want to include the issue number in the title, you can use `[P-163]` or `P-163 | ` or the like

PR #2204, opened 2023-10-19
Fix TMPDIR usage

arieljbirnbaum:
### Context

Addressing the temp directory issues reported by @AlseinX.
Our scripts have been overloading the system variable `TMPDIR`; this had worked until now, but it's clashing with `pnpm` since the latter relies on `os.tmpdir()` working properly.

### Fix

- Use a specific name for the variable that designates the installation directory.
  - Drive-by: make sure the directory is actually set (i.e. no defaults)
- In the few cases where `TMPDIR` is actually intended to be the catch-all temporary location, set up its creation and cleanup

### TODO

- [x] Make sure CI passes (some steps may fail if they're relying on the default value rather than setting the variable)

### DIFF
--- a/scripts/clean-local-binary.sh
+++ b/scripts/clean-local-binary.sh
@@ -2,9 +2,9 @@
 
 # no `set -e` here as we allow commands to fail in this script
 
-TMPDIR=${TMPDIR:-"/tmp/parachain_dev*"}
+LITENTRY_PARACHAIN_DIR=${LITENTRY_PARACHAIN_DIR:?}
 
-# for f in $(ls $TMPDIR/*.pid 2>/dev/null); do
+# for f in $(ls $LITENTRY_PARACHAIN_DIR/*.pid 2>/dev/null); do
 #   echo "Killing $f ..."
 #   kill -9 $(cat "$f")
 # done
@@ -14,6 +14,6 @@ TMPDIR=${TMPDIR:-"/tmp/parachain_dev*"}
 killall polkadot
 killall litentry-collator
 
-rm -rf "$TMPDIR"
+rm -rf "$LITENTRY_PARACHAIN_DIR"
 
 echo "cleaned up."
--- a/scripts/clean-local-docker.sh
+++ b/scripts/clean-local-docker.sh
@@ -8,7 +8,7 @@ function usage() {
 
 [ $# -ne 1 ] && (usage; exit 1)
 
-TMPDIR=${TMPDIR:-"/tmp/parachain_dev*"}
+LITENTRY_PARACHAIN_DIR=${LITENTRY_PARACHAIN_DIR:?}
 CHAIN=$1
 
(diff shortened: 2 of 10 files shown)


scripts/clean-local-binary.sh
```diff
@@ -2,9 +2,9 @@
 
 # no `set -e` here as we allow commands to fail in this script
 
-TMPDIR=${TMPDIR:-"/tmp/parachain_dev*"}
+LITENTRY_PARACHAIN_DIR=${LITENTRY_PARACHAIN_DIR:?}
```
felixfaisal, line 5: Shouldn't we have the default/fallback path if we don't set it? This is particularly useful when we call it via the `make` like `make clean-docker-rococo`.

scripts/clean-local-docker.sh
```diff
@@ -8,7 +8,7 @@ function usage() {
 
 [ $# -ne 1 ] && (usage; exit 1)
 
-TMPDIR=${TMPDIR:-"/tmp/parachain_dev*"}
+LITENTRY_PARACHAIN_DIR=${LITENTRY_PARACHAIN_DIR:?}
```
felixfaisal, line 11: Same

0xverin:
I tried it locally, and it works:
```
Setting up worker in tmp/w-0
Copying files from bin
Copying source files to working directory
bin/key.txt does not exist, this is fine, but you can't perform remote attestation with this.
bin/spid.txt does not exist, this is fine, but you can't perform remote attestation with this.
Initialized worker.
Starting worker 0 in background
worker command is: ['./litentry-worker', '--clean-reset', '-P', '2000', '-w', '2001', '-r', '3443', '-h', '4545', '--running-mode', 'mock', '--enable-mock-server', '--parentchain-start-block', '0', '--enable-metrics', 'run', '--skip-ra', '--dev']
```

scripts/clean-local-binary.sh
arieljbirnbaum: No, that's yet another way how errors end up getting masked because things work by accident until they don't :P
In t-shirt form, "defaults are bug eggs"

0xverin (approved):
LGTM.Thanks!

Kailai-Wang:
Unfortunately this PR breaks the case where only the parachain is launched without worker, because `LITENTRY_PARACHAIN_DIR` can only be set from py.

e.g.
`make launch-docker-rococo` will cause the generated files to be written to `/tmp/` directly
`make clean-docker-rococo` will error out due to unset env - and I don't know how to set this var even if I want to fix it - as the artefacts are located under root /tmp/ :(

```
make clean-docker-rococo
./scripts/clean-local-docker.sh: line 11: LITENTRY_PARACHAIN_DIR: parameter null or not set
make: *** [Makefile:158: clean-docker-rococo] Error 1
```

I think we need to get it working.

Other things to check:
- binary mode should work too
- multiple copies of worker + parachain should work at the same host

PR #2206, opened 2023-10-23
feat: reuse macro & solve P-137 via reuse

### LINKED ISSUE #2118: Merge evm/non-evm node implementation
Follow-up of https://github.com/litentry/litentry-parachain/pull/2016

We should be able to merge the node (client) implementation with and without evm:
- `service{_evm}.rs`
- `rpc{_evm}.rs`

AlseinX:
This PR introduced an attribute macro `reuse` that uses a mod definition as a template and emits mods for each cases. The usage is documented within the macro definition.

Closes #2118


primitives/core/macros/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "litentry-macros"
+version = "0.1.0"
+edition = "2021"
+
```
BillyWooo, line 7: Please take other Cargo.toml file as example to add some more fields here.

Kailai-Wang (commented):
Thanks!

I'm highly delighted with the fact you came up with a good idea and completed the implementation within a short time, very impressive! 👍 

I can see the idea behind it but I'm also a bit concerned that it doesn't diminish maintenance overhead a lot: imagine the upstream changes the implementation of `build_import_queue` - do we need to adjust both versions? I had expected that we could have only one `build_import_queue` which comprises most of the duplicate code - I'm fine with the conditional compilation guarded with features too. 

This is only an example tho, I'm not even sure if that's feasible, or maybe this is the best we can get already.

How do you feel about it? @wangminqi @kziemianek

AlseinX:
Actually I have tried wasting my time on merging the implementations with traits.
Rust's type system is not so expressive to make abstraction on trait bounds, and my attempt to work around this is too complicated, with a significantly larger amount of code than normally copied implementations, and any further change would require modification on the abstraction and both implementations. Thus I do think that defeats the purpose of abstraction, and making a trait abstraction on only 2 certain implementations is an overkill.
Plus, the rust compiler is still too buggy on type system that sometimes makes overflows on bound checking and false negatives when extra bounds on associated types of generic parameters are too complicated.
Therefore I believe attribute macro, the code analysis of which is supported as well as normal rust code, a simple and efficient solution of this case.

As for `build_import_queue`, the two versions share the same function body and have different signatures (precisely, on trait bounds and return type). The code looks like this:
```rust
#[no_evm]
fn build_import_queue(/*ARGS*/) -> /*RETURN TYPE*/
where
    /*BOUNDS*/
{
    __  // This double underscore as a placeholder is handled by the macro, which means it reuses the function body of the next function
}

#[evm]
fn build_import_queue(/*ARGS*/) -> /*RETURN TYPE*/
where
    /*BOUNDS*/
{
    /*IMPLEMENTATION*/
}
```

So when we need to adjust this function, only one implementation is to be modified.
As for the signatures, although it looks too long and duplicated, it is impossible to merge these 2 versions in either way, since rust supports neither trait abstractions nor attaching attributes on trait bounds.

felixfaisal (commented):
I really like this PR ❤️  
This expands my knowledge of macros and more stuff we can do with it. 
Thanks @AlseinX

primitives/core/macros/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+Expands to:
+```
+mod x {
+	fn a() -> i32 {
+		let p = 1;
+		p + 1
+	}
+	fn u() {
+		println!("hello world!");
+	}
```
kziemianek, line 71: @AlseinX would you mind explaining how

```
	#[x]  // This item would only appear on case `x`
	fn u() {
		__
	}
```

expanded to


```
	fn u() {
		println!("hello world!");
	}
```
?

AlseinX: As I have commented, leaving a function body with a double underscore placeholder indicates that it reuses the body of the next function.
This make it possible to give a function different signatures by cases.
It is a little bit like typescript's function overloading syntax, that a function may have multiple signatures, but in order to satisfy rust's syntax parser, which does not accept a standalone function signature, there has to be a placeholder function body.

kziemianek: Thanks, I probably missed the part that it's the body of the next function.

primitives/core/macros/src/reuse.rs
```diff
@@ -0,0 +1,404 @@
...
+
+pub fn handle_reuse(args: Args, m: ItemMod) -> Result<TokenStream> {
+	let mut result = TokenStream::new();
+	let cases = args.cases.iter().map(ToString::to_string).collect::<Vec<_>>();
+	for current in 0..args.cases.len() {
+		let mut visitor = Visitor { cases: &cases, current, error: None };
+		let mut m = m.clone();
+		m.ident = args.cases[current].clone();
+		visitor.visit_item_mod_mut(&mut m);
+		if let Some(e) = visitor.error.take() {
```
felixfaisal, line 52: I have a question here, I don't see anywhere else the `error` parameter is being updated, is it safe to assume it is always going to be `None` since it was set to it when `Visitor` struct was created in `line 32`

AlseinX: There's currently no error to be emitted, but the visitor may be extended in the future for futher cases, eg. attribute on raw functions rather than `mod`.
I think it necessary to always keep the ability to emit error on a clear `Span` for humanised compiling errors, rather than direct panics. Therefore I keep everywhere in proc macro returning `Result`.
Since visitor's definition does not allow returning `Result`, I use this error field as workaround, regardless of whether it would be used.

felixfaisal: That makes sense

kziemianek:
@Kailai-Wang I prefer to keep the old version even if the code is duplicated

primitives/core/macros/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+#![allow(clippy::tabs_in_doc_comments)]
```
kziemianek, line 17: please add license header

AlseinX: done

Kailai-Wang:
> @Kailai-Wang I prefer to keep the old version even if the code is duplicated

Tough call. 

Either way would actually work, and I can see both benefits and setbacks for each.
It does lower the code readability, but the trade-off is it removes / mitigates the need of maintaining duplicate code.

I'm inclined to merging it.

Later even when we don't need to separate evm/non-evm (= all runtimes include evm) we might still find the `reuse` macro useful

Kailai-Wang (approved):

PR #2209, opened 2023-10-24
Use AES to encrypt the trustedcall payload

### LINKED ISSUE #2071: Use AES to encrypt the TrustedCall payload
### Context

This issue supersedes https://github.com/litentry/litentry-parachain/issues/2046

Similar to what [we did](https://github.com/litentry/litentry-parachain/issues/828) to user shielding key, the TrustedCallSigned can exceed the length of RSA encryption limit.

A long-term solution would be using AES to encrypt the payload, using TEE's shielding key to encrypt the AES key and send it along. So the steps are:
- client generates a random, ephemeral AES key
- construct the TrustedCallSigned payload
- encrypt the AES key with TEE's shielding key
- encrypt the TrustedCallSigned payload with AES key
- send both encrypted data to the worker as a request

This is going to be a big change. Please take these into consideration:
- request construction
- handling of requests in TEE
- II handling
- Should this new request co-exist with the old pattern, or shall we only accept the new request pattern?
- Is (trusted)-Getter affected?
- All matching tests

---

Kailai-Wang:
### Context

resolves #2071 
resolves P-115

This PR:
- **adds** another way of encrypting the DI request - see `AesRequest`
- renames the old request struct to `RsaRequest` to make the distinction
- renames the RPC interfaces to `author_submitAndWatchRsaRequest` and `author_submitAndWatchAesRequest`, respectively
- adds a `DecryptableRequest` trait to unify the processing logic in top pool
- adjusts tests: now DI requests are all constructed using the new encryption mechanism

Responses to some points raised in the original issue:

> Should this new request co-exist with the old pattern, or shall we only accept the new request pattern?

They co-exist. From protocol level we should support both - it's quite possible that some requests can only go through the parachain in the future

> II handling

II will construct the request in the old way: `RsaRequest`, the AesKey is set to `None`

> Is (trusted)-Getter affected?

Trusted getter uses the `RsaRequest` too - things might change in **P-183**.

The removal of the shielding key setup will be followed up soon.


Kailai-Wang:
cc @jonalvarezz FYI

tee-worker/litentry/primitives/src/aes_request.rs
```diff
@@ -0,0 +1,69 @@
...
+
+	fn decrypt<T: Debug>(
+		&mut self,
+		enclave_shielding_key: Box<dyn ShieldingCryptoDecrypt<Error = T>>,
+	) -> core::result::Result<Vec<u8>, ()> {
+		let aes_key: UserShieldingKeyType = enclave_shielding_key
+			.decrypt(&self.key)
+			.map_err(|_| ())?
+			.try_into()
+			.map_err(|_| ())?;
```
Kailai-Wang, line 66: Any better way of writing this?

tee-worker/core-primitives/types/src/lib.rs
```diff
@@ -91,11 +90,35 @@ impl Encode for OpaqueCall {
 	}
 }
 
-// Note in the pallet teerex this is a struct. But for the codec this does not matter.
+// Litentry: re-defined due to orphan rule - it needs to be kept identical to that in teerex-primitives
```
arieljbirnbaum: Can that be enforced by a compiler check and/or unit test? e.g. encoding from one struct and decoding from the other should yield the same information? :thinking:

Kailai-Wang: That's a good idea, thanks @grumpygreenguy  - I'll look into it but maybe a macro declaration should be enough

Kailai-Wang: Updated : )

tee-worker/core-primitives/types/src/lib.rs
```diff
@@ -91,11 +90,26 @@ impl Encode for OpaqueCall {
...
+
+	fn payload(&self) -> &[u8] {
+		self.payload.as_slice()
+	}
+
+	fn decrypt<T: Debug>(
+		&mut self,
+		enclave_shielding_key: Box<dyn ShieldingCryptoDecrypt<Error = T>>,
+	) -> core::result::Result<Vec<u8>, ()> {
+		enclave_shielding_key.decrypt(self.payload.as_slice()).map_err(|_| ())
```
arieljbirnbaum, line 111: Not sure I follow, what error is being swallowed here?

arieljbirnbaum (approved):
LGTM :+1:

tee-worker/core-primitives/types/src/lib.rs
Kailai-Wang: The `T` here : ) - we don't care about this error, as in the end they will be converted to the top_pool decryption error anyway

primitives/teerex/src/lib.rs
```diff
@@ -202,12 +202,26 @@ pub type Pcesvn = u16;
...
 pub type SidechainBlockNumber = u64;
 
-#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, sp_core::RuntimeDebug, TypeInfo)]
-pub struct Request {
-	pub shard: ShardIdentifier,
-	pub cyphertext: Vec<u8>,
+// Litentry: use the name `RsaRequest` to differentiate from `AesRequest` (see aes_request.rs in
+// tee-worker) `Rsa` implies that the payload is RSA-encrypted (using enclave's shielding key)
+#[macro_export]
+macro_rules! decl_rsa_request {
```
kziemianek, line 208: Maybe we can get rid of teerex's `call_worker` ?

kziemianek (approved):

primitives/teerex/src/lib.rs
Kailai-Wang: Hmm then there's no way to call arbitrary trusted call for II.

We didn't use `call_worker` because of https://github.com/litentry/tee-worker/issues/58

In general I think it still makes sense to keep it, maybe we want to rewrite it someday

PR #2210, opened 2023-10-24
Reintroduce default temp dir

arieljbirnbaum:
### Context

Scripts fail on CI without the default values removed in #2204.
Setting the env variable in `ci.yml` still leads to runtime failure; attempts to locate the root cause have been so far unsuccessful.


Kailai-Wang (approved):
Thanks!

PR #2211, opened 2023-10-24
TOP pool synchronization

### LINKED ISSUE #1935: Linear processing of DI requests
### Context

Following up on the discussion in https://github.com/litentry/litentry-parachain/pull/1920#discussion_r1272208657

Imagine there are 100 workers and we send a request to worker0, worker0 happens to just finish the its block production so the newly submitted request won't be included in the block for now. Since we have an aura authorship mechanism, the request can be processed after 100 blocks at the earliest, this sounds terrifying.

The root cause is the tx pool of sidechain is not synchronised among nodes, so a DI request can only be stored and processed by the node, to which the request was submitted.

It's a critical issue for multi-worker setup, but I don't think we have time to solve it soon, hence the medium prior.

---

kziemianek:
### Context

resolves: #1935 
resolves: P-100

This PR attempts to add TOP pool synchronization through direct trusted rpc server. Trusted calls should be eventually broadcasted to other registered workers and optimistically picked and executed by next block producer.

Changes: 

- `author_submitAndWatchExtrinsic` JSON-RPC method broadcasts trusted call to other peers
- adds `author_submitAndWatchBroadcastedExtrinsic` JSON-RPC method which acts as `author_submitAndWatchExtrinsic` but doesn't broadcast trusted calls
- adds `TrustedOperationStatus::TopExecuted` which is fired upon TOP execution (**breaking change**)
- adds ocall `get_trusted_peers_urls` in order pull trusted peers urls from worker service to enclave runtime

 animated


tee-worker/core-primitives/ocall-api/src/lib.rs
```diff
@@ -135,6 +135,8 @@ pub trait EnclaveSidechainOCallApi: Clone + Send + Sync {
 		maybe_until_block_hash: Option<BlockHash>,
 		shard_identifier: ShardIdentifier,
 	) -> SgxResult<Vec<SignedSidechainBlock>>;
+
+	fn get_trusted_peers_urls(&self) -> SgxResult<Vec<String>>;
```
Kailai-Wang, line 146: How is it different from `get_untrusted_peer_url_of_shard`? Is it using a cached result? Like using the `peer_regsitry` that is populated when searching peers

tee-worker/core/peer-top-broadcaster/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+				},
+				Err(e) => log::error!("Could not connect to peer {}, reason: {:?}", peer, e),
+			}
+		}
+
+		std::thread::spawn(move || {
+			while let Ok((id, rpc_return_value)) = responses_receiver.recv() {
+				match rpc_return_value.status {
+					DirectRequestStatus::TrustedOperationStatus(status, _) => {
+						// Some((response.0, status, response.1.do_watch))
```
Kailai-Wang: we need to remove this

tee-worker/core-primitives/top-pool/src/watcher.rs
```diff
@@ -114,6 +114,11 @@ where
 		self.send(TrustedOperationStatus::Broadcast)
 	}
 
+	/// The extrinsic hsa been executed.
```
Kailai-Wang: typo: `has`

tee-worker/core/direct-rpc-client/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+		let addrs = ws_server_url.socket_addrs(|| None).unwrap();
+		let stream = TcpStream::connect(&*addrs)
+			.map_err(|e| format!("Could not connect to {:?}, reason: {:?}", &addrs, e))?;
+
+		let (mut socket, _response) =
+			client_tls_with_config(ws_server_url, stream, None, Some(connector))
+				.map_err(|e| format!("Could not open websocket connection: {:?}", e))?;
+
+		//it fails to perform handshake in non_blocking mode so we are setting it up after the handshake is performed
+		Self::switch_to_non_blocking(&mut socket);
```
Kailai-Wang, line 152: Hmm why do we set to non_blocking mode?

As far as I read https://github.com/snapview/tungstenite-rs/issues/113, non-blocking mode normally requires customised wait/resume handling. But we seem to do a manual poll here, wouldn't it result in high CPU load (well it sleeps for 10ms)?

I mean while we don't use synchronous reading?

Kailai-Wang (commented):
I haven't completed the review yet - only took a rough look. Will go into more details later

The animation is well made btw!

tee-worker/core-primitives/ocall-api/src/lib.rs
kziemianek: 1. we use trusted direct rpc to establish ws connection
2. we need a list of all enclaves

kziemianek: Currently the implementation is using cached result. 
The cache is updated during sidechain block proposal.

tee-worker/core/direct-rpc-client/src/lib.rs
kziemianek: The websocket connection is established between a pair of peers and multiple request / responses are going through it.
Both writing (sending a request) and reading (receiving a response) requires mutable reference to the socket. 
In order to have an ability to use a socket for both reading and writing we have to:
1. Use Mutex in order to manage the mutable access to socket.
2. Set non-blocking mode in order perform reads/writes alternately. Without it set, the thread would be blocked on reading data from socket, making other threads waiting for a lock in order to perform writes.

This has been described in the issue you mentioned:
 _Instead, you generally combine non-blocking read_message with your own blocking waiting. This differs from blocking read_message in that you have full control over blocking waiting and are able to interrupt it for i.e. sending._

In my tests with ws connections to 3 peers I didn't noticed high CPU load.

kziemianek: I've changed the logic a little bit and managed to get rid of mentioned `Mutex`

tee-worker/core/direct-rpc-client/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+impl DirectRpcClient {
+	pub fn new(url: &str, responses_sink: SyncSender<Response>) -> Result<Self, Box<dyn Error>> {
+		let ws_server_url =
+			Url::from_str(url).map_err(|e| format!("Could not connect, reason: {:?}", e))?;
+		let mut config = rustls::ClientConfig::new();
+		// we need to set this cert verifier or client will fail to connect with following error
+		// HandshakeError::Failure(Io(Custom { kind: InvalidData, error: WebPKIError(UnknownIssuer) }))
+		config.dangerous().set_certificate_verifier(Arc::new(IgnoreCertVerifier {}));
```
arieljbirnbaum, line 139: The knee-jerk reaction here would be that we'd rather want to install the relevant cert in the chain here to make the TLS handshake work; however I don't see a way to make it work (esp. on the browser!) without allowing a malicious worker operator to present a cert from _outside_ the enclave :disappointed: 

If that's the case tho, I think it's a tradeoff we'd have to document in big bold flaming letters (both in the code and in relevant docs) :thinking:

tee-worker/core/direct-rpc-client/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+		&self,
+		_presented_certs: &[rustls::Certificate],
+		_sni: Option<&DNSName>,
+	) -> Result<rustls::ClientCertVerified, rustls::TLSError> {
+		Ok(rustls::ClientCertVerified::assertion())
+	}
+}
+
+pub trait RpcClientFactory {
+	type RpcClient: RpcClient;
```
Kailai-Wang: Slight name confusion here - as `RpcClient` is used both as type and as trait. Maybe something like `RpcClientTrait`?

tee-worker/core/direct-rpc-client/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+		let connector = Connector::Rustls(Arc::new(config));
+		let addrs = ws_server_url.socket_addrs(|| None).unwrap();
+		let stream = TcpStream::connect(&*addrs)
+			.map_err(|e| format!("Could not connect to {:?}, reason: {:?}", &addrs, e))?;
+
+		let (mut socket, _response) =
+			client_tls_with_config(ws_server_url, stream, None, Some(connector))
+				.map_err(|e| format!("Could not open websocket connection: {:?}", e))?;
+
+		let (request_sender, requests_receiver) = channel();
```
Kailai-Wang: A minor naming mismatch: singular vs plural
Maybe we should always use the singular form? Same applies to `request/response_sink`

tee-worker/core/direct-rpc-client/src/lib.rs
Kailai-Wang: Is it possible to actually verify the dest server's cert?

If it's too much for this PR we can have a separate issue

tee-worker/core/direct-rpc-client/src/lib.rs
Kailai-Wang: Thanks - I misread - I thought the reading and writing were in different threads

Kailai-Wang (approved):
Overall looks good to me, nicely done!

tee-worker/core/direct-rpc-client/src/lib.rs
kziemianek: I'll change to use singular form 👍🏼

tee-worker/core/direct-rpc-client/src/lib.rs
kziemianek: `RpcClient` -> `Client`

tee-worker/core/direct-rpc-client/src/lib.rs
kziemianek: I think yes but it would require:

* publishing generated cert to enclave registry so it can be added to client's config root store before establishing TLS connection
* adjusting generated cert common name to match the hostname of published trusted rpc endpoint url (currently hardcoded to `Enclave`)


How do you feel @Kailai-Wang , @grumpygreenguy does it make sense ?

arieljbirnbaum: Not sure it's worth the trouble tbh.

To prevent an MITM from a malicious node operator, we'd have to make sure that "our" CA cert is _the only one_ in the store when establishing the connection --- otherwise, the operator can present any valid cert for the same domain name, and the TLS logic will be none the wiser.

That _might_ be doable in the (Rust) client, but AFAICT there's no way to do it in the browser (and I'd be more than happy to be proven wrong on that one) --- that means, we still cannot simply rely on TLS there to provide the needed authentication and encryption. If that's the case, I think it just makes sense to stick with the same solution for both.

tee-worker/core-primitives/ocall-api/src/lib.rs
clangenb: This is good, the untrusted rpc server is legacy, and we hope to get rid of it.

tee-worker/core/direct-rpc-client/src/lib.rs
clangenb: Yeah, our loose concept we had for this was the the enclave's signing key should be part of the certificate, which can be checked against the public registry. This custom approach is of course doable in rust, but browsers obviously don't have such a feature. We have been thinking about a browser extension that would take care of this check though.

But then again how do you explain that to the enduser? To many people don't understand https and the lock XD, we are once again ahead of time.

clangenb (commented):
Hey, we definitely want to gossip the top pool. I had a superficial look and this PR looks very nice and clean!

I noticed that you gossip the executed TOP to peers, but this is probably not the mechanism to remove it from the synchronized pool? It is done via the sidechain block import, isn't it?

Maybe @brenzi also wants to chime in here.

kziemianek:
> Hey, we definitely want to gossip the top pool. I had a superficial look and this PR looks very nice and clean!
> 
> I noticed that you gossip the executed TOP to peers, but this is probably not the mechanism to remove it from the synchronized pool? It is done via the sidechain block import, isn't it?
> 
> Maybe @brenzi also wants to chime in here.

Thanks @clangenb for your time and taking a look at this 👍🏼 .

Yes, this change doesn't affect how TOP pool is managed (when TOPs are added or removed). It basically only gossips trusted calls to other known peers and routes back selected responses to the caller. All the rest remains the same. 

RPC client could submit trusted call to all peers in the network separately and the efect would be pretty the same.

brenzi:
I think you should take care not to gossip getters or inherents.

Top pool holds tops (TrustedOperation) which can be getters, indirect or direct TrustedCalls

Also, a trusted call can be an inherent/intrinsic(signed by enclave itself) i.e. shield_funds(). These could be gossipped in principle, but we need to think about edge cases

kziemianek:
> I think you should take care not to gossip getters or inherents.
> 
> Top pool holds tops (TrustedOperation) which can be getters, indirect or direct TrustedCalls
> 
> Also, a trusted call can be an inherent/intrinsic(signed by enclave itself) i.e. shield_funds(). These could be gossipped in principle, but we need to think about edge cases

Hi @brenzi, thanks for your advices, here are my thoughts:

* Only trusted operations submitted through json-rpc method `author_submitAndWatch*` are gossiped - this means that indirect calls submitted by indirect executor directly to `AuthorApi` wouldn't be gossiped
* Only succesfully submitted trusted operations are gossiped. This means that as long as we have correct `Author`'s filter (`CallsOnlyFilter`) applied getters wouldn't be gossiped.

We can add filtering logic to `DirectRpcBroadcaster` similar to the one in `Author` but that would require going through the whole ceremony of decrypting/decoding or move gossiping to `Author` - might be worth to make sure only direct calls are broadcasted. What do you think @Kailai-Wang ?

kziemianek:
I've moved broadcasting triggering point from json-rpc method to `Author`, it gives following benefits:
* broadcasting is encapsulated in Author so other components don't need to be aware about how broadcasting should be performed
* possibility to filter TrustedCalls so only direct calls are broadcasted

Kailai-Wang:
> I've moved broadcasting triggering point from json-rpc method to `Author`, it gives following benefits:
> 
> * broadcasting is encapsulated in Author so other components don't need to be aware about how broadcasting should be performed
> * possibility to filter TrustedCalls so only direct calls are broadcasted

Thank you - I like the incorporation of `BroadcastedRequest`

And yes I don't think we need to broadcast all kinds of trusted operations:
- direct trusted call => sure
- trusted getters => theoretically we should be able to submit getters via `author_submitAndWatch*` but practically we don't do that. Maybe we should simply disallow submitting getters from this rpc method - or apply filtering. Anyways it makes little sense to gossip it
- indirect calls => we don't need to gossip it, all workers will get them by syncing the parachain

So I'd say for now we only broadcast direct trusted calls.

Kailai-Wang (approved):
Looks good, let's merge it.

`BroadcastedTopFilter` is passed in as `DirectCallsOnlyFilter` to initialisation AFAIU, isn't it

kziemianek:
> Looks good, let's merge it.
> 
> `BroadcastedTopFilter` is passed in as `DirectCallsOnlyFilter` to initialisation AFAIU, isn't it

Yes, this will ensure only direct calls are broadcasted to peers.

kziemianek:
> > I've moved broadcasting triggering point from json-rpc method to `Author`, it gives following benefits:
> > 
> > * broadcasting is encapsulated in Author so other components don't need to be aware about how broadcasting should be performed
> > * possibility to filter TrustedCalls so only direct calls are broadcasted
> 
> Thank you - I like the incorporation of `BroadcastedRequest`
> 
> And yes I don't think we need to broadcast all kinds of trusted operations:
> 
> * direct trusted call => sure
> * trusted getters => theoretically we should be able to submit getters via `author_submitAndWatch*` but practically we don't do that. Maybe we should simply disallow submitting getters from this rpc method - or apply filtering. Anyways it makes little sense to gossip it
> * indirect calls => we don't need to gossip it, all workers will get them by syncing the parachain
> 
> So I'd say for now we only broadcast direct trusted calls.

Trusted getters are not added to TOP pool due to filters applied :
```
#[cfg(feature = "sidechain")]
pub type AuthorTopFilter = crate::top_filter::CallsOnlyFilter;
```

PR #2214, opened 2023-10-26
Fix: change sidehcain metadata update path

0xverin:
### Context
When I was testing this [action](https://github.com/litentry/litentry-parachain/actions/runs/6649339905/job/18067796042), I noticed that the sidechain metadata I obtained was consistently different from what's local. Upon inspecting the code, I discovered that the update path being used is incorrect. The correct path should be `prepare-build`, as updating to `build` would be overwritten. Unfortunately, this incorrect path has been in place for three months, rendering the sidechain metadata updates over the past three months completely ineffective in `sidechain-api`.

### DIFF
--- a/tee-worker/client-api/sidechain-api/package.json
+++ b/tee-worker/client-api/sidechain-api/package.json
@@ -5,7 +5,7 @@
 	"main": "dist/src/index.js",
 	"scripts": {
 		"clean": "rm -rf dist build node_modules",
-		"update-metadata": "../../bin/litentry-cli print-sgx-metadata-raw > build/litentry-sidechain-metadata.json",
+		"update-metadata": "../../bin/litentry-cli print-sgx-metadata-raw > prepare-build/litentry-sidechain-metadata.json",
 		"prepare-dir": "mkdir -p build && cp -rf prepare-build/* build",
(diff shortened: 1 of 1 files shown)


Kailai-Wang:
Good catch, thanks! Does it mean that we have been using the wrong metadata? 😲 And why is it working?

arieljbirnbaum (approved):

arieljbirnbaum:
> Good catch, thanks! Does it mean that we have been using the wrong metadata? 😲 And why is it working?

If there haven't been changes in the sidechain _types_ (or those changes don't affect the test code), then we wouldn't notice any breakage. Keep in mind that the runtime metadata is downloaded fresh upon connection; the "static" one only affects type generation.

0xverin:
> Good catch, thanks! Does it mean that we have been using the wrong metadata? 😲 And why is it working?

The types primarily used in our sidechain-api are related to identity. If these haven't changed, it can actually still work. Since we didn't use the new metadata for type generation and then for using the new types, we wouldn't have noticed this issue...

Then, I've been continuously testing on this [branch](https://github.com/litentry/litentry-parachain/tree/2092-rebuilding-api-package) and found that the obtained metadata is no longer compilable. After local testing from I and @felixfaisal , it's confirmed that the new metadata indeed cannot be compiled. :(

0xverin:
@grumpygreenguy If you run update metadata after this pr merged, you'll find that sidehcian-api no longer build.

Issue #2093, opened 2023-08-31
Add package building step to CI


arieljbirnbaum:
### Context

Part of #2087 
:warning: Blocked by #2092 

### Task

Instead of having build outputs committed to the repo, we should rebuild them when needed to run the tests.

- If possible: add one CI step before the ts-tests to build the packages and pass the build outputs as [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts)
- If the above is not possible, add the build as needed before running each ts-test (:sadpanda: because it makes the tests longer/heavier)

- Remove the committed metadata and build outputs from the repo (and .gitignore them)

### Acceptance criteria

- Build outputs and intermediate files are gone from the repo
- CI passes

0xverin:
>Remove the committed metadata and build outputs from the repo (and .gitignore them)

Do we need to temporarily retain the metadata commits because the CI requires fmt checks(client-api needs to build first)?

PR #2216, opened 2023-10-27
Rebuilding api package on CI

### LINKED ISSUE #2092: Automate API package rebuilding
### Context

Part of #2087 
:warning: Blocked by #2088 

### Task

- Create a script (Makefile step? Yarn script? Shell script? TBD) that rebuilds the API packages based on:
  - the current runtime metadata from a _running_ worker/parachain, and
  - the current (manually defined) RPC type definitions
  - NOTE: the script needs to take the ports to reach the correct processes. NO DEFAULTS --- otherwise we run the risk of hitting the wrong process and building with the wrong metadata :P
  - NOTE: the RPC type definitions, as well as the build outputs, can live in predefined locations in the working tree.
- Add a clean command to clean up build outputs and intermediate files.

### Acceptance criteria

- On a clean clone, build the packages with the script and start the ts-tests successfully
- Check that the clean command removes the generated files

0xverin:
### Context

resolves: #2092
resolves: P-126

We have only the `latest` tag for `litentry/litentry-worker` and `litentry/litentry-cli` on Docker Hub. Do we have plans for a `staging` tag? @grumpygreenguy @BillyWooo 

**Follow up**:

~~The sidechain metadata obtained through the CLI cannot be built locally, similar to the CI. However, it works fine when retrieved through wss and decoded (after testing, this way provides the same sidechain metadata as the existing one). I suggest creating a new PR to address this issue separately .~~

~~1. Investigate why the CLI method is failing to compile and make the necessary fixes.~~

~~2. Temporarily create a Ts script to fetch the sidechain metadata via WebSocket (wss), and then decode and use it.~~

The follow up already fixed.

Please feel free to point out any issues with the workflows, thanks!


BillyWooo:
About `staging` tag, I think we should have something like that. I would suggest to address it once we have the release branch ready.

0xverin:
> About `staging` tag, I think we should have something like that. I would suggest to address it once we have the release branch ready.

Okay, let's focus on the dev tag first. :p

arieljbirnbaum:
Re tags for other mainlines; that's probably best done with/after https://linear.app/litentry/issue/P-193 /cc @Kailai-Wang

.github/workflows/release-ts-api-package.yml
```diff
@@ -1,13 +1,159 @@
...
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: litentry-parachain-dev
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: litentry-tee
+
+      - name: Load docker image
```
Kailai-Wang: I'm not sure if you need all of these.
It seems you just have to pull the docker images as you don't have to build anything locally

tee-worker/cli/release_ts_api_package.sh
```diff
@@ -0,0 +1,59 @@
...
+echo "update parachain metadata"
+
+# Why the metadata fetched by cli cannot be built now. :(
+cd  /client-api/sidechain-api
+${CLIENT} print-sgx-metadata-raw > prepare-build/litentry-sidechain-metadata.json
+echo "update sidechain metadata"
+
+cd /client-api
+pnpm install
+pnpm run build
```
Kailai-Wang: Do we need to update types? Like `update-build`?

.github/workflows/release-ts-api-package.yml
Kailai-Wang, line 1: It seems this GHA can only be triggered manually now - we'll definitely need to connect it to the releasing process later

Kailai-Wang (commented):
I failed to understand the goal of this PR/GHA.

AFAICT it starts a parachain + enclave network and tries to re-generate the metadata - but that's it, the metadata is neither uploaded nor merged into any branch. It seems to be disregarded

Edit: I see there's a `publish` section, but again the metadata generated in the previous step isn't persisted between jobs I guess?

0xverin:
> I failed to understand the goal of this PR/GHA.

The purpose of this GitHub Action (GHA) is as follows:

1. Start the parachain and worker.

2. Update metadata.

3. Generate types.

4. Execute the `release` part.

> AFAICT it starts a parachain + enclave network and tries to re-generate the metadata - but that's it, the metadata is neither uploaded nor merged into any branch. It seems to be disregarded

We won't upload metadata, but we will upload package outputs. Please take a look at this [issue](https://github.com/litentry/litentry-parachain/issues/2093). The goal is that in the future, all our developers won't have to face the fear of making so many file changes after running typegen locally.

> Edit: I see there's a `publish` section, but again the metadata generated in the previous step isn't persisted between jobs I guess?

`update-metadata` step has already generated new type outputs, and `publish` is meant to publish these outputs.

tee-worker/cli/release_ts_api_package.sh
0xverin: ~~It has already been updated [here](https://github.com/litentry/litentry-parachain/pull/2216/files/e3a816e3ad14649e6f7e1a77c4438192a0e25287#diff-16bfcf5319ea98ea537262555928586aa7bc9bbb825c63db5fd1e3a24fd947d9R48).~~

Rewrote it.

.github/workflows/release-ts-api-package.yml
0xverin: Yes, we don't want to perform any compilation operations in this GHA. The goal is to pull the latest existing Docker images and tags, and then proceed to publish their TypeScript API packages.

.github/workflows/release-ts-api-package.yml
```diff
@@ -1,16 +1,133 @@
...
+      - name: Save docker image
+        run: |
+          docker save litentry/litentry-parachain:latest | gzip > litentry-parachain-dev.tar.gz
+
+      - name: Upload docker image
+        uses: actions/upload-artifact@v3
+        with:
+          name: litentry-parachain-dev
+          path: litentry-parachain-dev.tar.gz
+          if-no-files-found: error
```
Kailai-Wang: I still fail to understand why we need to upload and download it 😓

.github/workflows/release-ts-api-package.yml
```diff
@@ -1,16 +1,133 @@
...
+
+      - name: Enable corepack and pnpm
+        run: corepack enable && corepack enable pnpm
+
+      - name: Generate parachain artefacts
+        run: |
+          ./tee-worker/scripts/litentry/generate_parachain_artefacts.sh
+          ls -l docker/generated-rococo/
+          ls -l tee-worker/docker/litentry/
+          shasum tee-worker/docker/litentry/*.json
```
Kailai-Wang: are they for debugging only?

tee-worker/cli/lit_ts_api_package_build.sh
```diff
@@ -0,0 +1,58 @@
...
+sudo chmod a+x /usr/local/bin/websocat
+websocat --version
+
+cd /client-api/parachain-api
+echo '{"id":1,"jsonrpc":"2.0","method":"state_getMetadata","params":[]}' | /usr/local/bin/websocat -n1 -k -B 99999999 $NODEURL:$NPORT > prepare-build/litentry-parachain-metadata.json
+echo "update parachain metadata"
+
+cd  /client-api/sidechain-api
+${CLIENT} print-sgx-metadata-raw > prepare-build/litentry-sidechain-metadata.json
+echo "update sidechain metadata"
```
Kailai-Wang: What prevents us from running `pnpm update` directly?

Kailai-Wang (dismissed):
And my previous question remains:

> I failed to understand the goal of this PR/GHA.

> AFAICT it starts a parachain + enclave network and tries to re-generate the metadata - but that's it, the metadata is neither uploaded nor merged into any branch. It seems to be disregarded

> Edit: I see there's a publish section, but again the metadata generated in the previous step isn't persisted between jobs I guess?

.github/workflows/release-ts-api-package.yml
0xverin: I've modified this part of the steps.

tee-worker/cli/lit_ts_api_package_build.sh
0xverin: Changes have been made.

.github/workflows/release-ts-api-package.yml
0xverin: If I don't include it here, it will result in an error: 

`unable to prepare context: path "/home/runner/work/litentry-parachain/litentry-parachain/tee-worker/docker/litentry" not found.`
:p

tee-worker/cli/lit_ts_api_package_build.sh
0xverin: I rolled back this part of the code because running 'litentry-cli' directly in the 'package.json' was causing this [CI error](https://github.com/litentry/litentry-parachain/actions/runs/6770473168/job/18398946938) whether using the `../../bin` or `/usr/local/bin` path...

tee-worker/cli/lit_ts_api_package_build.sh
```diff
@@ -0,0 +1,52 @@
...
+
+CLIENT="${CLIENT_BIN} -p ${NPORT} -P ${WORKER1PORT} -u ${NODEURL} -U ${WORKER1URL}"
+
+echo "Using client binary $CLIENT_BIN"
+echo "Using node uri $NODEURL:$NPORT"
+echo "Using trusted-worker uri $WORKER1URL:$WORKER1PORT"
+echo ""
+
+cd /client-api/parachain-api
+curl -s -H "Content-Type: application/json" -d '{"id": "1", "jsonrpc": "2.0", "method": "state_getMetadata", "params": []}' http://litentry-node:9912 > prepare-build/litentry-parachain-metadata.json
```
arieljbirnbaum: The url there `http://litentry-node:9912` should also be a param, no? Other than that lgtm :+1:

arieljbirnbaum (approved):

.github/workflows/release-ts-api-package.yml
Kailai-Wang: I mean these three lines:
```
          ls -l docker/generated-rococo/
          ls -l tee-worker/docker/litentry/
          shasum tee-worker/docker/litentry/*.json
```

.github/workflows/release-ts-api-package.yml
```diff
@@ -1,16 +1,55 @@
-name: Release Ts API package
+# Reference: ci.yml
+name: Release Ts API Package
+
 on:
   workflow_dispatch:
-
+      inputs:
+        tag:
+          description: 'Publish tag'
```
Kailai-Wang: Maybe a better desc would be `release docker image tag`

.github/workflows/release-ts-api-package.yml
```diff
@@ -1,16 +1,55 @@
...
+
+      - name: Set env variables
+        run: echo "RELEASE_IMAGE_VERSION=${{ inputs.tag }}" >> "$GITHUB_ENV"
+
+      - name: Pull litentry image optionally
+        run: |
+          docker pull parity/polkadot
+          docker pull litentry/litentry-worker:${{ inputs.tag }}
+          docker pull litentry/litentry-cli:${{ inputs.tag }}
+          docker pull litentry/litentry-parachain:${{ inputs.tag }}
```
Kailai-Wang, line 29: It seems to require that all these tags are in sync.
I'm not sure if that's always the case

cc @grumpygreenguy

.github/workflows/release-ts-api-package.yml
```diff
@@ -23,25 +62,53 @@ jobs:
...
+          else
+            echo "Parachain API dist and build files do not exist. Publishing failed."
+            exit 1
+          fi
+
+          npm publish --tag ${{ inputs.tag }}
 
           echo "------------------------parachain API published------------------------"
 
           cd ../sidechain-api
```
Kailai-Wang: You can remove the duplicate code by:

```
for c in parachain sidechain; do
   cd "../$c-api"
   ...
done
```

tee-worker/cli/lit_ts_api_package_build.sh
Kailai-Wang: I think when launching docker the default port is 9912

.github/workflows/release-ts-api-package.yml
0xverin: Thanks!
Removed.

.github/workflows/release-ts-api-package.yml
0xverin: Modified!

.github/workflows/release-ts-api-package.yml
0xverin: Done.

tee-worker/cli/lit_ts_api_package_build.sh
0xverin: Done. 
Passed it [here](https://github.com/litentry/litentry-parachain/pull/2216/files#diff-17c6a820206756c0f4a88fcd583d78432174966cba8bff08f16a911bc2e08d1dR20-R21)

PR #2217, opened 2023-10-29
Split TEE release into worker and enclave

Kailai-Wang:
### Context

This PR:

- split the TEE release into two parts: worker and enclave
- reworked the release type
- reworded the release building script, I think `enclave.signed.so` and `litentry-worker` are the only things to release (= made public)
- applied proper version for worker and enclave:
   - worker starts from `0.0.1`
   - enclave starts from `100`

We should bump the enclave runtime version for any change that results in a different MRENCLAVE, even if it's not a direct sgx-pallet/runtime change.

TODO:
- maybe to have a GHA to detect mismatch of enclave version and MRENCLAVE (e.g. the MRENCLAVE changes but the version is not bumped) in case we forget it
- `SGX_SIGN_PASSFILE` is ignored for now, we need to figure out how/if we should use it => an issue will follow

A trial run that builds all 4 artefacts:
- parachain client
- parachain runtime
- TEE worker
- TEE enclave

https://github.com/litentry/litentry-parachain/releases/tag/untagged-e92681d128065981fdb1


felixfaisal (approved):
LGTM

kziemianek (approved):
I haven't tried it out but it looks fine.

PR #2218, opened 2023-10-30
P-179 Add credentials-priority-1-and-2

outofxxx:
Fixes P-179


tee-worker/client-api/parachain-api/dist/build/interfaces/identity/types.d.ts
Kailai-Wang, line 1: Why is the format of these files changed (again)?

There must be a reason - we need to solve this as it affects (and will continuously affect) the review experience :(
@Verin1005 @grumpygreenguy

0xverin: If we resolve this [issue](https://github.com/litentry/litentry-parachain/issues/2093), I believe we won't have this trouble anymore.

tee-worker/litentry/core/credentials/src/litentry_profile/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+pub mod holding_amount;
+pub mod mirror;
+pub mod token_balance;
+
+fn match_balance(range: Vec<(&str, &str)>, balance: &str) -> usize {
+	let balance = balance.parse::<f64>().unwrap_or_default();
+	let mut r_index = range.len();
+	for (index, item) in range.iter().enumerate() {
+		let low = item.0.parse::<f64>().unwrap();
+		let high = item.1.parse::<f64>().unwrap();
```
Kailai-Wang: Can we avoid `unwrap()`? There're other places too

tee-worker/litentry/core/credentials/src/litentry_profile/token_balance.rs
```diff
@@ -0,0 +1,144 @@
...
+	("0", "5"), // FALSE
+	("5", "10"),
+	("10", "50"),
+	("50", "100"),
+	("100", "150"),
+	("150", "300"),
+	("300", "500"),
+	("500", "800"),
+	("800", "1200"),
+];
```
Kailai-Wang: I think these can be dramatically simplified by maintaining a vector of thresholds like: `[0, 5, 10, 50, 100, 150, 300, 500, 800, 1200]` and only use index to denote the interval that it falls into.

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -740,6 +779,69 @@ impl AchainableAccountTotalTransactions for AchainableClient {
...
+
+pub trait HoldingAmount {
+	fn holding_amount(&mut self, addresses: Vec<String>, param: Params) -> Result<String, Error>;
+}
+impl HoldingAmount for AchainableClient {
+	fn holding_amount(&mut self, addresses: Vec<String>, param: Params) -> Result<String, Error> {
+		let mut total_balance = 0_f64;
+		addresses.iter().for_each(|address| {
+			let body = ReqBody::new(address.into(), param.clone());
+			debug!("Holding amount body: {:?}", body);
```
Kailai-Wang: I assume this will print addresses too? We should avoid that in enclave (or use something similar to Kasper's impl) - there're other places too

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -740,6 +779,69 @@ impl AchainableAccountTotalTransactions for AchainableClient {
...
+			let balance =
+				self.post(SystemLabelReqPath::default(), &body).and_then(Self::get_balance);
+			total_balance += balance.unwrap_or_default();
```
Kailai-Wang: I think we agreed that if http request fails we should error out - otherwise the `total_balance` will be simply wrong => VC is wrong

Please check other places too

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -625,6 +643,27 @@ impl Default for ParamsBasicTypeWithDatePercent {
...
+
+	pub chain: String,
+
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub post_quantity: Option<String>,
+}
+
+impl ParamsBasicTypeWithMirror {
+	pub fn new(name: String, network: &Web3Network, post_quantity: Option<String>) -> Self {
+		let chain = web3_network_to_chain(network);
```
Kailai-Wang: I don't think we need `web3_network_to_chain` anymore, it has derived from `IntoStaticStr`

tee-worker/litentry/core/data-providers/src/achainable_names.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl GetAchainableName for AchainableNameBalance {
+	fn name(&self) -> &'static str {
+		match self {
+			AchainableNameBalance::BalanceUnderAmount => "Balance under {amount}",
+			AchainableNameBalance::BalanceOverAmount => "Balance over {amount}",
+		}
+	}
+}
```
Kailai-Wang, line 71: I'm really not a fan of doing comparisons of strings everywhere - we should use enums instead.

Using strings as parameter might seem convenient to write the top-level interface at first glance, but you'll need countless sanity checks to make sure you cover all corner cases. And it's really error-prone.

We should ask for the IDHub's opinion as well (as they are the client).

I'm not sure if it's too much to fix this in this PR, if not please create an issue and fix it in a follow-up PR, thanks!

tee-worker/litentry/core/assertion-build/src/achainable/amount.rs
```diff
@@ -103,13 +108,30 @@ pub fn build_amount(req: &AssertionBuildRequest, param: AchainableAmount) -> Res
...
+	let bname = AchainableNameBalance::from(achainable_param.name()).map_err(|e| {
+		Error::RequestVCFailed(
+			Assertion::Achainable(achainable_param.clone()),
+			e.into_error_detail(),
+		)
+	})?;
+	let mut balance = "".to_string();
+	let mut flag = false;
+	if bname == AchainableNameBalance::BalanceUnderAmount {
+		balance = request_achainable_balance(addresses, achainable_param.clone())?;
```
Kailai-Wang: I can't really tell from the names what's the difference between `request_achainable_balance` and `request_achainable` - their distinct return types make it even more mysterious.

Maybe they are either sloppily named or implemented :(

Kailai-Wang (commented):
Thank you! I haven't really tried it so I can't tell if the vc building logic is correct or not.

I've left comments purely by reading the code

In general I prefer having smaller PRs with thoughtful design over bigger ones that work (= realise the functionality) but feebly implemented.

I suggest we put more thoughts when writing it, even when we are under time pressure : )

tee-worker/client-api/parachain-api/dist/build/interfaces/identity/types.d.ts
Kailai-Wang: I kind of disagree - that PR only moves the building steps to CI (instead of locally), but it **doesn't** explain the root cause.

tee-worker/litentry/core/assertion-build/src/achainable/amount.rs
outofxxx: `request_achainable` is the standard achainable API interface, which only requires the API for `true/false` results, and later added `_balance` and others require special handling of the results returned by achainable, such as parsing balance from the `text` field. Perhaps the naming of this is too simple and crude.

tee-worker/litentry/core/data-providers/src/achainable_names.rs
outofxxx: ok, i will create a issue to track this.

tee-worker/client-api/parachain-api/dist/build/interfaces/identity/types.d.ts
0xverin: solved in #2241

BillyWooo (approved):
LGTM. Please resolve the conflict, make sure CI pass. Then we merge.

PR #2219, opened 2023-10-30
feat: encrypt ws server outputs

AlseinX:
This allows the websocket client to send a pub key at the beginning of the stream to enable encrypting the following server outputs.
It is optional for the client to send a pub key, directly sending normal text messages or an empty key would make it work in legacy non-encrypting mode, so **it would not be a breaking change for clients**.
A unit test has been modified to use encrypting mode.


tee-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -189,14 +218,16 @@ where
 				);
 			},
 			Message::Binary(_) => {
-				warn!("received binary message, don't have a handler for this format");
+				unreachable!()
```
felixfaisal: I have a question here, Do you think it makes sense to panic here?

AlseinX: Ah good question, I forgot to change it back since I wanted to hard panic here when debugging.

tee-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -127,20 +127,49 @@ where
...
 				"Read is possible for connection {}: {}",
 				self.connection_token.0,
 				web_socket.can_read()
 			);
 			match web_socket.read_message() {
-				Ok(m) =>
+				Ok(Message::Binary(key)) if !is_inited => {
+					debug!("server reads binary message as after established, encryption enabled");
+					let mut state = StreamState::Invalid;
+					std::mem::swap(&mut self.stream_state, &mut state);
```
felixfaisal, line 147: Why can't we just use 
```
std::mem::replace(&mut self.stream_state, StreamState::Invalid) 
```
Instead of having to initialise `state` and then swapping with it. I'm just curious although it does the same thing, i feel `replace` is more idiomatic.

kziemianek (commented):
Have you considered using handshake's HTTP headers for passing the key ?

arieljbirnbaum:
> It is optional for the client to send a pub key, directly sending normal text messages or an empty key would make it work in legacy non-encrypting mode, so it would not be a breaking change for clients.

Note that if we don't get TLS to work as needed, we'll have to remove the non-encrypted option before going to production --- otherwise the traffic is completely naked :thinking: 
Do we have a way to emit deprecation warnings if the non-encrypted option is used?

Kailai-Wang:
> Do we have a way to emit deprecation warnings if the non-encrypted option is used?

I think we can. So basically:
- in prod: we only accept requests with a key
- in dev: we accept both

tee-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -127,20 +127,49 @@ where
 	///
 	/// Returns a boolean 'connection should be closed'.
 	fn read_or_initialize_websocket(&mut self) -> WebSocketResult<bool> {
-		if let StreamState::EstablishedWebsocket(web_socket) = &mut self.stream_state {
+		if let (StreamState::EstablishedWebsocket(web_socket), _, is_inited)
```
Kailai-Wang, line 135: `is_inited` is a bit confusing, can we make it `is_initialized`?

tee-worker/core/tls-websocket-server/src/connection.rs
Kailai-Wang: I believe that's because `state` is used in the following line

tee-worker/core/tls-websocket-server/src/encrypt.rs
```diff
@@ -0,0 +1,87 @@
+use core::num::NonZeroU32;
+#[cfg(feature = "sgx")]
+use std::vec::Vec;
+
+use log::warn;
+use rsa::{Pkcs1v15Encrypt, RsaPublicKey};
+
+pub struct Encryptor {
+	key: RsaPublicKey,
```
Kailai-Wang: I remember we wanted to use AES key?

tee-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -127,20 +127,49 @@ where
...
+					debug!("server reads binary message as after established, encryption enabled");
+					let mut state = StreamState::Invalid;
+					std::mem::swap(&mut self.stream_state, &mut state);
+					if let StreamState::EstablishedWebsocket(ws) = state {
+						if key.is_empty() {
+							self.stream_state = StreamState::AllReady(ws, None);
+							return Ok(false)
+						}
+
+						if let Some(encryptor) = Encryptor::new(&key) {
```
Kailai-Wang: Is the `key` actually sent with clear text? I expect it to be encrypted with enclave's shielding key at least 🤔

Kailai-Wang:
> Have you considered using handshake's HTTP headers for passing the key ?

This is also a good option IMO

Kailai-Wang (commented):
Thanks! I get the idea behind it, but I've still got questions regarding the key passing.

Apart from the inline comments, I believe we need to:
- change `TrustedCallResult` to clear text if I'm not mistaken
- at least provide an example how to send the requests via JS

tee-worker/core/tls-websocket-server/src/connection.rs
AlseinX: I remember that shielding key is said to be removed?

tee-worker/core/tls-websocket-server/src/encrypt.rs
AlseinX: AES is symmetric, so the key sent to the enclave has to be encrypted, while RSA does not need it.
Since the enclave is guaranteed to be real, I believe it safe to send a public key with clear text.

AlseinX:
> > It is optional for the client to send a pub key, directly sending normal text messages or an empty key would make it work in legacy non-encrypting mode, so it would not be a breaking change for clients.
> 
> Note that if we don't get TLS to work as needed, we'll have to remove the non-encrypted option before going to production --- otherwise the traffic is completely naked 🤔 Do we have a way to emit deprecation warnings if the non-encrypted option is used?

Some queries that is not so sensitive may not require an encryption, maybe we should leave it to the user to opt-in or not.

tee-worker/core/tls-websocket-server/src/connection.rs
arieljbirnbaum: > I remember that shielding key is said to be removed?

I think that's a different (albeit similarly named) key.

The "enclave shielding key" refers to a fixed public key owned by the enclave; this one's still needed to send messages securely to the worker (in particular, keep them from interception by the node operator).

The (user) "shielding key" is a user-specific encryption key used to encode payloads for that user (including the VCs published on-chain); this is the piece of state that we're aiming to remove.

felixfaisal:
I will close this issue, We can open it again once we have a proper issue and assignee to continue working on this. Feel free to re-open it if you think otherwise.

PR #2225, opened 2023-10-31
Fix idh 329 Add sora quiz VC

outofxxx:
as title.


Kailai-Wang (approved):
The code logic looks fine, thanks!

I didn't try it though - need someone who knows what it _should be_ to test it.

Another thing is please update litentry-cli tool accordingly, either in this PR or in a follow-up PR.

tee-worker/litentry/core/credentials/src/sora/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+		SoraQuizType::Master => "$is_master",
+	}
+}
+
+fn get_sora_assertion_info(qtype: &SoraQuizType) -> (&'static str, &'static str) {
+	match qtype {
+		SoraQuizType::Attendee => VC_LITENTRY_SORA_QUIZ_INFOS[0],
+		SoraQuizType::Master => VC_LITENTRY_SORA_QUIZ_INFOS[1],
+	}
+}
```
wli-pro, line 64: Is this gonna be essentially one VC, but accepting any discord roleId and discord handler, to check if that user has the role?

How does F/E pass the qtype?

If qtype is passed I was thinking we don't need F/E to pass role id insteand enclave will decide the roleId for qtype, because we don't want F/E to pass any other role id other than the two we intended here I guess?

kziemianek: I think we are following A3 assertion design here which also has this flaw 😞

wli-pro: We will remove roleId as parameter

PR #2231, opened 2023-11-03
feat: Isolate VC issuance from block production

felixfaisal:
I've resolved all the conflicts and refactored based on the latest changes in the `dev` branch. This PR implements the feature `isolate vc issuance from block production`, and at its current state, it serves as an MVP/POC. Let's discuss the implementation:

1. A new method called `author_submitVCRequest` has been added to the JSONRPC server, which sends the `RsaRequest` to the independent VC thread.
2. The VC thread handles the `assertion` request, responds with the encrypted VC payload, and submits the extrinsic to the chain.
3. The initialization of this independent thread is similar to that of the `stf-task` thread, and I've followed a similar design.
4. I have opted not to remove the current method of requesting VC and believe it should be phased out gradually.

Now, let's address some issues:

#### Nonce Issue
Since the request is sent to an independent thread, verifying the nonce is possible, but mutating the nonce would require altering the sidechain storage from this thread, which might be challenging.

#### Accessing Sidechain Storage
Currently, I am cloning the entire sidechain storage and then using `execute_with` to access it:

```rust
let (mut state, _) = match context.state_handler.load_cloned(&req.shard) {
    // ...
};
state.execute_with(|| {
    // ...
});
```

I'm cloning because `execute_with` requires mutable access to the state. However, there is a way to read sidechain storage without cloning or requiring mutable access:

```rust
fn get(&self, key: &[u8]) -> Option<&Vec<u8>> {
    self.state.get(key)
}
```

Here, `key` refers to the encoded version `concat(pallet+storage_name)`, similar to what we do on a parachain. I plan to improve upon this in a separate PR.

#### Sending Extrinsics to Parachain
To send extrinsics to the parachain, I am currently using this:

```rust
self.context.ocall_api.send_to_parentchain(xt, &ParentchainId::Litentry);
```

I'm not certain this is the best method, but it is functional.

#### Concurrency/Parallelism
At present, I am not using scoped threads for parallelism. I want to ensure my logic is sound for handling one request at a time before implementing it. Although the code has been written with this in mind, implementing it should be simple.

Additionally, in the RPC Handler, I'm using a `oneshot` sender, which is useful for receiving the results of asynchronous tasks. However, I am also using a `while` loop:

```rust
while let Ok(response) = receiver.try_recv() {
    if let Some(Ok(response)) = response {
        // ...
    }
}
```

This can be refined by using `.add_method` instead of `.add_sync_method`. `.add_method` expects a lambda that resolves to a future, allowing the use of `.await` rather than a `while` loop. I plan to address this in another PR. The `oneshot` sender is quite handy for parallel tasks.

#### Final Thoughts
The PR still contains several comments and `unwrap()` instances that I'm working to clean up. I'm also aiming to refactor the code to reduce redundancy. I would appreciate feedback on any aspects I may have overlooked. My comments are based on my understanding, and I'd like to hear your thoughts.

#### Future Issues to Address:
- Implement a nonce mechanism (or alternative).
- Optimize access to sidechain storage.
- Use thread pools to process requests in parallel.
- Write example code in TypeScript.
- Conduct load testing.
- Modify the `indirect executor`.
- Phase out the existing mechanism for `request_vc`.

### Testing
To test this feature, I've added a command to `cli` that you can use:

```
./litentry-cli trusted -d request-vc-direct did:litentry:substrate:0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 a8 litentry,litmus
```

Thank you for allowing me to work on this feature; it has been an enjoyable experience 😄


Kailai-Wang:
Thanks for the expressive summary, without(before) taking a deeper look I'd like to share my thoughts on these points raised:
- Implement a nonce mechanism (or alternative)
  => why do we even need a nonce? It should be a client/server model - if you want to distinguish between requests, we have id the JSONRPC request 🤔 
- Optimize access to sidechain storage
   I remember there was some old code doing that - back when we were experimenting with aysnc handling. Need to find that
- Use thread pools to process requests in parallel.
   Yes please
- Modify the indirect executor.
   I thinking to gradually take down VC-request via II, especially after P-174 where the vc payload will **not** be available in   the callback extrinsic

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/Cargo.toml
```diff
@@ -0,0 +1,103 @@
...
+
+# no_std dependencies
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
+lazy_static = { version = "1.1.0", features = ["spin_no_std"] }
+log = { version = "0.4", default-features = false }
+sp-core = { default-features = false, features = ["full_crypto"], git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.39" }
+
+# internal dependencies
+ita-stf = { path = "../../../../app-libs/stf", default-features = false }
+# ita-sgx-runtime = { default-features = false, path = "../../sgx-runtime" }
```
kziemianek, line 30: to be removed?

kziemianek: to be removed?

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc_direct.rs
```diff
@@ -0,0 +1,239 @@
...
+//
+// oneblock VC:
+// ./bin/litentry-cli trusted -m <mrencalve> -d request-vc-direct \
+//   did:litentry:substrate:0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 oneblock completion
+//
+// achainable VC:
+// ./bin/litentry-cli trusted -m <mrencalve> -d request-vc-direct \
+//   did:litentry:substrate:0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 achainable amount-holding a litentry 1 2014-05-01
+
+// fn to_para_str(s: &str) -> ParameterString {
```
kziemianek: Please remove

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -68,6 +68,18 @@ pub(crate) fn perform_trusted_operation(
...
 }
 
+pub(crate) fn perform_direct_operation(
+	cli: &Cli,
+	trusted_args: &TrustedCli,
+	top: &TrustedOperation,
+) -> TrustedOpResult {
+	match top {
+		TrustedOperation::indirect_call(_) => send_indirect_request(cli, trusted_args, top),
+		TrustedOperation::direct_call(_) => send_direct_vc_request(cli, trusted_args, top),
```
kziemianek: This is a little bit confusing that this function supports:

- all types of indirect calls
- all types of getters
- only `RequestVc` in case of `direct_call` 

Have you considered supporting only `request_vc` ?

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -313,6 +325,86 @@ fn send_direct_request(
...
 
+//
+fn send_direct_vc_request(
+	cli: &Cli,
+	trusted_args: &TrustedCli,
+	operation_call: &TrustedOperation,
+) -> TrustedOpResult {
+	let encryption_key = get_shielding_key(cli).unwrap();
+	let shard = read_shard(trusted_args, cli).unwrap();
+	let jsonrpc_call: String = get_vc_json_request(shard, operation_call, encryption_key);
```
kziemianek: If i'm correct this will accept any `trusted_call` not only `request_vc`. Is this fine? Maybe it would be better to fail earlier.

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -313,6 +325,86 @@ fn send_direct_request(
...
+			Ok(response) => {
+				debug!("received response");
+				let response: RpcResponse = serde_json::from_str(&response).unwrap();
+				if let Ok(return_value) = RpcReturnValue::from_hex(&response.result) {
+					debug!("successfully decoded rpc response: {:?}", return_value);
+					match return_value.status {
+						DirectRequestStatus::Error => {
+							debug!("request status is error");
+							if let Ok(value) = String::decode(&mut return_value.value.as_slice()) {
+								println!("[Error] {}", value);
```
kziemianek: Can you use `error` macro here?

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+		let (vc_response, sender) = response_receiver.recv().unwrap();
+		if let Err(e) = vc_response.clone() {
+			send_rpc_error(format!("Failed to generate credential: {:?}", e), sender);
+		} else {
+			vc_handler.request_vc_callback(vc_response.clone().unwrap(), sender);
+		}
+	});
+}
+
+pub fn handle_jsonrpc_request<K, A, S, H, O>(
```
kziemianek: I think we are not dealing with jsonrpc requests here, `req` is just an application type. I would suggest removing the dependency on `jsonrpc_core` from `lc-vc-task-sender` and all mentions about jsonrpc because it's only used for error handling and such errors can be mapped in another layer.

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+	sender: VCMPResponseSender,
+) where
+	K: ShieldingCryptoDecrypt + ShieldingCryptoEncrypt + Clone + Send + Sync + 'static,
+	A: AuthorApi<Hash, Hash> + Send + Sync + 'static,
+	S: StfEnclaveSigning + Send + Sync + 'static,
+	H: HandleState + Send + Sync + 'static,
+	H::StateT: SgxExternalitiesTrait,
+	O: EnclaveOnChainOCallApi + EnclaveMetricsOCallApi + 'static,
+{
+	// let shielding_key = match self.shielding_key_repo.retrieve_key() {
```
kziemianek: It look like you've forgot to remove it.

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+	// let request_vec = match request.decrypt(Box::new(shielding_key)) {
+	// 	Ok(req) => req,
+	// 	Err(_) => return Box::pin(ready(Err(ClientError::BadFormatDecipher.into()))),
+	// };
+	// // decode call
+	// let trusted_operation = match TrustedOperation::decode(&mut request_vec.as_slice()) {
+	// 	Ok(op) => op,
+	// 	Err(_) => return Box::pin(ready(Err(ClientError::BadFormat.into()))),
+	// };
+	// TODO: Refactor based on AES Request
```
kziemianek: Is this todo going to be addresed in this PR ?

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+	let (sender, response_receiver) =
+		channel::<(Result<VCResponse, VCMPError>, oneshot::Sender<Result<Vec<u8>, RpcError>>)>();
+
+	// TODO: Use a builder pattern here
+	let vc_callback_handler =
+		VCCallbackHandler::new(context.clone(), extrinsic_factory, node_metadata_repo);
+	let vc_callback_handler = Arc::new(vc_callback_handler);
+	start_response_handler(vc_callback_handler, response_receiver);
+
+	loop {
```
kziemianek: There is another way which I find more cleaner:
```
	while let Ok(req) = receiver.recv() {
		handle_jsonrpc_request(context.clone(), req, sender.clone());
	}
```

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/vc_handling.rs
```diff
@@ -0,0 +1,154 @@
...
+impl<K, A, S, H, O> VCRequestHandler<K, A, S, H, O>
+where
+	K: ShieldingCryptoDecrypt + ShieldingCryptoEncrypt + Clone,
+	A: AuthorApi<Hash, Hash>,
+	S: StfEnclaveSigning,
+	H: HandleState,
+	H::StateT: SgxExternalitiesTrait,
+	O: EnclaveOnChainOCallApi,
+{
+	pub fn process(self) -> Result<VCResponse, VCMPError> {
```
kziemianek, line 43: Is it possible to make it less VCMP aware ?

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/vc_callback.rs
```diff
@@ -0,0 +1,133 @@
...
+					return
+				},
+			};
+			match self.context.ocall_api.send_to_parentchain(xt, &ParentchainId::Litentry) {
+				Ok(_) =>
+					if let Err(e) = sender.send(Ok(res.encode())) {
+						log::warn!("Unable to send response jsonrpc handler: {:?}", e);
+					},
+				Err(e) => {
+					send_rpc_error(
```
kziemianek: If you follow my advice about decoupling other logic from RPC then I think you can appli it here aswell

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/vc_handling.rs
```diff
@@ -0,0 +1,154 @@
...
+		let mut credential = match self.req.assertion.clone() {
```
kziemianek, line 44: 40 duplicated LOC - can we have a single function to perform such mapping ?

arieljbirnbaum:
> I thinking to gradually take down VC-request via II, (...)

Following that thought, wouldn't we rather aim for dropping II altogether? From an application dev point of view, I honestly struggle to see when I'd prefer using II over DI if given the choice :thinking:

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/Cargo.toml
felixfaisal: This is removed

felixfaisal:
> Following that thought, wouldn't we rather aim for dropping II altogether? From an application dev point of view, I honestly struggle to see when I'd prefer using II over DI if given the choice 🤔

Please open a slack thread or a notion page for this discussion, Out of scope for the PR 😉

tee-worker/cli/src/trusted_operation.rs
felixfaisal: Updated this ✅

tee-worker/cli/src/trusted_operation.rs
felixfaisal: It fails in `perform direct operation` ✅

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: Removed it ✅  It was for myself

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: I removed the `jsonrpc_core` and doing as you mentioned

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: Removed ✅

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/vc_handling.rs
felixfaisal: I didn't get you 😅

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/vc_callback.rs
felixfaisal: Done ✅

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/vc_handling.rs
felixfaisal: Yeah, We have to eventually remove this mapping on the other side. So until we do that, this redundancy is fine according to me. Let me know what you think.

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: Done

tee-worker/cli/src/trusted_operation.rs
felixfaisal: Done

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc_direct.rs
felixfaisal: Done

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/vc_handling.rs
kziemianek: I was thinking about removing the usage of `VCMPError` from here, this code could be completly decoupled from the pallet imho but I see `tee-worker/litentry/core/assertion-build/src/lib.rs` also uses this type so it would require additional refactor 😞

Kailai-Wang:
> > I thinking to gradually take down VC-request via II, (...)
> 
> Following that thought, wouldn't we rather aim for dropping II altogether? From an application dev point of view, I honestly struggle to see when I'd prefer using II over DI if given the choice 🤔

For now the identity linking/status change via II would still work - except you don't get the updated IDGraph by the callback extrinsic, but you can still tell from it whether the request succeeds or not.

VC-request won't make sense - so we can take it down for now (but we might want to revive it when later VC has other ways of transmission)

In general, I do think we still want to keep II even if it's not the primary way of sending requests.. some benefits it gives:
- apply parachain tokenomics
- more decentralised as it will "randomise" the enclave that executes it
- mutate parachain state before getting executed

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -313,6 +328,85 @@ fn send_direct_request(
...
+	shard: ShardIdentifier,
+	operation_call: &TrustedOperation,
+	shielding_pubkey: sgx_crypto_helper::rsa3072::Rsa3072PubKey,
+) -> String {
+	let operation_call_encrypted = shielding_pubkey.encrypt(&operation_call.encode()).unwrap();
+
+	// compose jsonrpc call
+	let request = RsaRequest { shard, payload: operation_call_encrypted };
+	RpcRequest::compose_jsonrpc_call("author_submitVCRequest".to_string(), vec![request.to_hex()])
+		.unwrap()
```
Kailai-Wang: We need to use AesRequest now :)

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -82,6 +84,36 @@ where
 		Ok(json!(json_value))
 	});
 
+	let author_submit_extrinsic_name: &str = "author_submitVCRequest";
```
Kailai-Wang: You might want to give it a name other than `author_submit_extrinsic_name`

tee-worker/litentry/core/vc-issuance/lc-vc-task-sender/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+impl Default for VcRequestSender {
+	fn default() -> Self {
+		Self::new()
+	}
+}
+
+impl SendVcRequest for VcRequestSender {
+	fn send_vc_request(&self, request: VCRequest) -> Result<(), String> {
```
Kailai-Wang: Hmm it looks like all of these can be done via a pub fn, why do we need to have a struct and trait here?

Kailai-Wang (approved):
Overall it looks good to me, thanks!

We need to use AesRequest now btw :)

tee-worker/litentry/core/vc-issuance/lc-vc-task-sender/src/lib.rs
felixfaisal: I followed the same pattern as `stf-sender`, a `pub fn` is sufficient actually.

tee-worker/litentry/core/vc-issuance/lc-vc-task-sender/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+		Self::new()
+	}
+}
+
+impl VcRequestSender {
+	pub fn send_vc_request(&self, request: VCRequest) -> Result<(), String> {
+		debug!("send vc request: {:?}", request);
+
+		// Acquire lock on extrinsic sender
+		// TODO: Can we optimise using RwLock
```
kziemianek: Please do it or create issue with `tech-debt` tag

kziemianek (approved):

tee-worker/litentry/core/vc-issuance/lc-vc-task-sender/src/lib.rs
felixfaisal: I tried doing it, but I don't think it works. So it's safe to remove it i guess

PR #2232, opened 2023-11-03
fix: .env.local is updated correctly

felixfaisal:
So the problem is that this line 
```
data.replace(":9944", ":" + os.environ.get("CollatorWSPort", "9944"))
``` 
does not mutate the data in place and instead returns a new string, because of which `.env.local` always has `9944` and `2000`


AlseinX:
It looks like the former version was a typo 😆

AlseinX (approved):

Kailai-Wang (approved):
Thanks 😂 good catch

PR #2233, opened 2023-11-03
Distinguish file types in `pre-commit` githooks

Kailai-Wang:
### Context

As topic, tested locally and it looked good.


BillyWooo (approved):
nice refactor. 
much more clear 👍

PR #2236, opened 2023-11-06
Efface user shielding key setup completely

Kailai-Wang:
### Context

As topic.

Changes:
- shielding key setup is completely removed, now the IDGraph (including the prime identity) will be created upon the first identity linking attempt
- shielding key storage is removed from sidechain storage
- renamed `UserShieldingKeyType` to `RequestAesKey`
- the identity verification message calculation is changed to 
```
blake2_256(<sidechain nonce> + <primary account> + <identity-to-be-linked>)
```
- II extrinsic callbacks contain public information only, no encrypted information. They behave more like real "events". Related: P-184
- the limitation of returning x identities in IDGraph is waived in `get_id_graph`, we had it to avoid too long extrinsic parameters, but it's not passed back via extrinsic anymore
- adjustments to ts-test

VC-request via II will be gradually phase out, you could theoretically still request it via parachain extrinsic, but it makes less sense if the generated vc can't be parsed


tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -745,41 +686,40 @@ where
...
-				Ok(TrustedCallResult::RequestVC(res))
+
+				if let Some(key) = maybe_key {
+					Ok(TrustedCallResult::RequestVC(RequestVCResult {
+						vc_index,
+						vc_hash,
+						vc_payload: aes_encrypt_default(&key, &vc_payload),
+					}))
+				} else {
+					Ok(TrustedCallResult::Empty)
```
felixfaisal, line 721: I just have a question here, Isn't it wasteful to process a `request-vc` request, when `maybe_key` is `None` since the Result returned is empty. I might be missing something here, Let me know what you think.

Kailai-Wang: Yes, it makes little sense to request a vc via parachain now, as you simply don't get the vc content 😂 

as mentioned in the PR:
> VC-request via II will be gradually phase out, you could theoretically still request it via parachain extrinsic, but it makes less sense if the generated vc can't be parsed

tee-worker/app-libs/stf/src/getter.rs
```diff
@@ -205,10 +203,7 @@ impl ExecuteGetter for TrustedGetterSigned {
 					None
 				},
 			// litentry
-			TrustedGetter::user_shielding_key(who) =>
-				IdentityManagement::user_shielding_keys(&who).map(|key| key.encode()),
-			TrustedGetter::id_graph(who) =>
-				Some(IdentityManagement::get_id_graph(&who, usize::MAX).encode()),
+			TrustedGetter::id_graph(who) => Some(IdentityManagement::get_id_graph(&who).encode()),
```
kziemianek, line 206: We might need paginiation in the future.

tee-worker/client-api/parachain-api/build/interfaces/augment-api-tx.ts
```diff
@@ -1507,29 +1506,21 @@ declare module "@polkadot/api-base/types/submittable" {
...
-                    identity: CorePrimitivesKeyAesOutput | { ciphertext?: any; aad?: any; nonce?: any } | string | Uint8Array,
-                    reqExtHash: H256 | string | Uint8Array
-                ) => SubmittableExtrinsic<ApiType>,
-                [AccountId32, CorePrimitivesKeyAesOutput, H256]
+                (account: AccountId32 | string | Uint8Array, reqExtHash: H256 | string | Uint8Array) => SubmittableExtrinsic<ApiType>,
+                [AccountId32, H256]
             >;
+            /**
+             * ---------------------------------------------------
+             * The following extrinsics are supposed to be called by TEE only
```
kziemianek: Isn't `identityActivated` and `identityDeactivated` also supposed to be called  by TEE only ?

kziemianek (commented):
Maybe we could rename UserShieldingKeyType?

Aren't primitives like `UserShieldingKeyType`, `AesOutput` valid only in context of sidechain so their definitions can be moved to tee-worker after this change ?

tee-worker/client-api/parachain-api/build/interfaces/augment-api-tx.ts
Kailai-Wang: I think the generated ts just copy comments from .rs files, there is where the comment comes from:
https://github.com/litentry/litentry-parachain/blob/e327c377bb0be2bf1893879be377176dd1594bf3/pallets/identity-management/src/lib.rs#L235-L237

And it's meant to say: starting from this line, all following extrinsics should be called by TEE

tee-worker/app-libs/stf/src/getter.rs
Kailai-Wang: Yes that's a good point

Kailai-Wang: We have a max length of 100 tho - maybe that's still OK for a rpc response - I mean I definitely saw very very long rpc response

kziemianek: True,
maybe such requirement will come from one of api consumers in the future.

Kailai-Wang:
> Aren't primitives like `UserShieldingKeyType`, `AesOutput` valid only in context of sidechain so their definitions can be moved to tee-worker after this change ?

Indeed - I'll move it to tee-worker/ scope
`UserShieldingKeyTpe` renamed to something like `RequesEnctKey`?

kziemianek:
> > Aren't primitives like `UserShieldingKeyType`, `AesOutput` valid only in context of sidechain so their definitions can be moved to tee-worker after this change ?
> 
> Indeed - I'll move it to tee-worker/ scope `UserShieldingKeyTpe` renamed to something like `RequesEnctKey`?

Can be `RequesEnctKey` or `EncryptionKey`. Just to avoid confusion in the future because there should be no such thing like shielding key anymore.

Kailai-Wang:
I'll do `RequestAesKey`

`RequestEncKey` is fine too but that key is actually used for decryption too 😓

tee-worker/ts-tests/integration-tests/di_evm_identity.test.ts
```diff
@@ -340,8 +271,6 @@ describe('Test Identity (evm direct invocation)', function () {
         }
 
         assert.equal(deactivatedIdentityEvents.length, 2);
```
0xverin, line 273: Is asserting just the length sufficient here? If it's not enough, you can add a "todo" to assign it to the development team responsible for ts-tests.

Kailai-Wang: Yes for parachain events it's enough

0xverin (commented):
Thanks! Ts-tests part LGTM!

Kailai-Wang:
Note:
there could be leftovers in the stress-test part as I didn't run it myself - it might need your adjustments later when we start to use it again 😅 @grumpygreenguy @Verin1005

kziemianek (approved):

0xverin:
> Note: there could be leftovers in the stress-test part as I didn't run it myself - it might need your adjustments later when we start to use it again 😅 @grumpygreenguy @Verin1005

Ok,I'll adjust it.

PR #2226, opened 2023-10-31
Add package building step to CI

0xverin:
### Context

resolves: #2093
resolves: #2095
resolves: p-127
resolves: p-129

**Note:**
- This PR needs the  #2216  to be merged before it can be reviewed.

- After this PR is merged, there will no longer be a situation where client-api has a bunch of file changes and commits. Additionally, the hassle of constant formatting changes has been reduced accordingly.

PR #2241, opened 2023-11-07
Add prettier and editor rules

0xverin:
### Context

The `client-api`, due to the absence of formatting rules, results in numerous file modifications for everyone after building and formatting. This PR aims to fix that. Although this [PR](https://github.com/litentry/litentry-parachain/pull/2226) deprecates this functionality, we need it for reference.

Also configured in ts-tests.


arieljbirnbaum (approved):
LGTM. tbh it'd probably be better to exclude generated files from format checking altogether, but like you said #2226 makes the point moot so :shrug:

PR #2243, opened 2023-11-09
P-217 Spike domain based VC data provider

outofxxx:
1. Add `nodereal` data provider
2. Add `bnb domain holding amount` VC
3. Add `bnb 999 & 10k domain amount` VC


primitives/core/src/assertion.rs
```diff
@@ -213,6 +217,8 @@ impl Assertion {
 			Self::Achainable(a) => vec![a.chain()],
 			// Oneblock Assertion
 			Self::Oneblock(..) => vec![Web3Network::Polkadot, Web3Network::Kusama],
+			// SPACEID Assertions
+			Self::SPACEIDBnbDomainHolding => vec![Web3Network::Bsc],
```
wli-pro: Is there any downside to remove the SPACEID, so we could make this as generic as possible?
i.e. there is other partner wanted bnbdomainholding, we can reuse this.

outofxxx: No, thanks. I will remove it.

tee-worker/litentry/core/assertion-build/src/nodereal/spaceid/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+		}
+
+		Ok(Domains { domains })
+	}
+
+	pub fn size(&self) -> usize {
+		self.domains.len()
+	}
+
+	pub fn non_expired_domains(&self) -> Result<Vec<&DomainInfo>> {
```
Traf333: it reads like the function should return a list of domains. What do you think about renaming this function to 
```suggestion
	pub fn non_expired_domain_infos(&self) -> Result<Vec<&DomainInfo>> {
```

tee-worker/litentry/core/data-providers/src/nodereal.rs
```diff
@@ -0,0 +1,204 @@
...
+impl NoderealHttpMethos for NoderealClient {
+	fn post(
+		&mut self,
+		path: NoderealServiceReqPath,
+		body: &SpaceIDReqBody,
+	) -> Result<serde_json::Value, Error> {
+		let response = self
+			.client
+			.post_capture::<NoderealServiceReqPath, SpaceIDReqBody, serde_json::Value>(path, body);
+		response.map_err(|e| Error::AchainableError(format!("Nodereal response error: {}", e)))
```
Traf333, line 127: is the client also refers to Achainable provider? or Achainable and Nodereal are different providers? I was thinking that if they are different and we want to share common error, then the name might mislead and it may deserve to be named like `DataProviderClientError` or something similar, what do you think?

Traf333 (commented):
Really enjoyed to review this PR, as it seems clear and I can digest what's going on there 😅 . I put a couple of comments only for my curiosity and learning purpose

tee-worker/litentry/core/assertion-build/src/nodereal/spaceid/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+};
+use serde::{Deserialize, Serialize};
+use std::string::ToString;
+
+pub struct SPACEIDLogic;
+pub trait SPACEIDLogicInterface {
+	fn get_bnb_domain_holding_amount(addresses: &[String]) -> Result<usize>;
+}
+
+impl SPACEIDLogicInterface for SPACEIDLogic {
```
wli-pro: Similar here as well as everywhere else where we have spaceid folder/file/name, I am thinking we should try to remove SPACEID everywhere, except maybe in some of the assertion product does need us add spaceid (but I did check and didn't see any of that in the VC requirement/assertions etc), then our new VC is simply a domain holding VC, that's completely not limited to SPACEID. Make sense? @zhouhuitian

tee-worker/litentry/core/data-providers/src/nodereal.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl NoderealHttpMethos for NoderealClient {
+	fn post(
+		&mut self,
+		path: NoderealServiceReqPath,
+		body: &SpaceIDReqBody,
+	) -> Result<serde_json::Value, Error> {
+		let response = self
+			.client
+			.post_capture::<NoderealServiceReqPath, SpaceIDReqBody, serde_json::Value>(path, body);
```
wli-pro, line 126: Could we add some exponential back off retry here if we see nodereal specific rate limiting error

primitives/core/src/assertion.rs
```diff
@@ -185,6 +185,10 @@ pub enum Assertion {
 
 	// Sora Quiz
 	SoraQuiz(SoraQuizType, ParameterString),  // (sora_quiz_type, guild_id)
+
+	// ----- begin SPACEID -----
```
Kailai-Wang, line 221: I assume we are going to have multiple SPACEID VCs?

tee-worker/litentry/core/assertion-build/src/nodereal/spaceid/mod.rs
Kailai-Wang: I'll completely leave it on your device to figure out if we need nested directory structure as you are the most familiar with the VC feature request (and potentially incoming new VCs) :) @zhouhuitian @higherordertech

tee-worker/litentry/core/data-providers/src/nodereal.rs
```diff
@@ -0,0 +1,204 @@
...
+
+		let mut headers = Headers::new();
+		headers.insert(CONNECTION.as_str(), "close");
+		let client = build_client(api_url.as_str(), headers);
+
+		NoderealClient { api_key, api_url, client }
+	}
+}
+
+pub trait NoderealHttpMethos {
```
Kailai-Wang: typo?

tee-worker/litentry/core/assertion-build/src/nodereal/spaceid/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+	fn get_bnb_domain_holding_amount(owners: &[String]) -> Result<usize> {
+		let config = DataProviderConfigReader::read()
+			.map_err(|e| Error::RequestVCFailed(Assertion::BnbDomainHolding, e))?;
+		let mut client = NoderealClient::new(&config);
+		let response = client.by_owners(owners).map_err(|e| {
+			Error::RequestVCFailed(Assertion::BnbDomainHolding, e.into_error_detail())
+		})?;
+
+		let owned_domains: Domains = Domains::from_value(&response)
+			.map_err(|_e| Error::UnclassifiedError(ErrorDetail::ParseError))?;
```
Kailai-Wang: We should avoid using `UnclassifiedError` when possible, as commented in the enum:

> 	// should be unreached, but just to be on the safe side
	// we should classify the error if we ever get this

tee-worker/litentry/core/assertion-build/src/nodereal/spaceid/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct Domains {
+	domains: Vec<Domain>,
+}
+
+impl Domains {
+	pub fn from_value(value: &serde_json::Value) -> Result<Domains> {
```
Kailai-Wang: Hmm I'm wondering if it can be auto-deserialised to `Domains` struct

tee-worker/litentry/core/assertion-build/src/nodereal/spaceid/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	pub fn size(&self) -> usize {
+		self.domains.len()
+	}
+
+	pub fn non_expired_domains(&self) -> Result<Vec<&DomainInfo>> {
+		let mut ts: Vec<&DomainInfo> = vec![];
+
+		for domain in self.domains.iter() {
```
Kailai-Wang: maybe this can be simplified by `map_filter` or the like?

tee-worker/litentry/core/data-providers/src/nodereal.rs
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	pub fn is_expired(&self) -> Result<bool, Error> {
+		// e.g. "2014-11-28T12:45:59.324310806Z"
+		let now = now();
+
+		// "expires": "2032-08-24T00:15:56Z"
+		let expired =
+			NaiveDateTime::parse_from_str(&self.expires, "%Y-%m-%dT%H:%M:%SZ").map_err(|e| {
+				Error::AchainableError(format!("Nodereal parse domain expires date error: {:?}", e))
```
Kailai-Wang: Why is it AchainableError? 😂

tee-worker/litentry/core/credentials/src/spaceid/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+pub fn match_domain_amount<T>(range: &[T], amount: T) -> Option<usize>
+where
+	T: std::cmp::PartialOrd + Display + Debug,
+{
+	let mut index = None;
+
+	for (idx, item) in range.iter().enumerate() {
+		if amount < *item {
+			index = Some(idx - 1);
```
Kailai-Wang: What if `idx` == 0 here? So the first item in `range` is already greater than `amount`

tee-worker/litentry/core/credentials/src/spaceid/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+pub mod bnb_domain_holding_amount;
+
+pub fn match_domain_amount<T>(range: &[T], amount: T) -> Option<usize>
+where
+	T: std::cmp::PartialOrd + Display + Debug,
+{
+	let mut index = None;
+
+	for (idx, item) in range.iter().enumerate() {
```
Kailai-Wang: Can we use [position](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position)?

tee-worker/litentry/core/data-providers/src/nodereal.rs
outofxxx: fixed.

primitives/core/src/assertion.rs
outofxxx: Yes, there will have 3 `SPACEID` VCs this time.

tee-worker/litentry/core/assertion-build/src/nodereal/spaceid/mod.rs
outofxxx: sure, i will re-adjust the structure.

tee-worker/litentry/core/data-providers/src/nodereal.rs
outofxxx: At present, the rate is sufficient, but there may be a need to upgrade the account in the future, so I don't think we need to add this feature for now

tee-worker/litentry/core/data-providers/src/nodereal.rs
outofxxx: Thanks, fixed now ,this is a copy-paste error...

Kailai-Wang (approved):
Looks good, thanks! Didn't test it locally though

wli-pro (approved):

PR #2245, opened 2023-11-10
`ErrorDetail` variants docs

### LINKED ISSUE #2177: Document the `ErrorDetail` type
### Context

The documentation is the current [ErrorDetail](https://github.com/litentry/litentry-parachain/blob/585f332c5d6a0cbaed680be411e1db2330c940c5/primitives/core/src/error.rs#L30) type is insufficient - we often hear feedbacks from the client/users that they don't understand the reasons of failure just by looking at the error name (e.g. `UnexpectedMessage`).

We should amend the doc - particularly, we'd better name a few classical examples where a specific error can occur.

Optional:
- add the doc as new chapters to gitbook (developer's guide)
- explore an automatic way to generate such docs (will `cargo doc` work?)

---

kziemianek:
Adds missing docs to `ErrorDetail` variants to help understand circumstances under which given error happens.


primitives/core/src/error.rs
```diff
@@ -42,13 +42,18 @@ pub enum ErrorDetail {
 	ParseError,
 	// errors when communicating with data provider, e.g. HTTP error
 	DataProviderError(ErrorString),
-	InvalidIdentity,
+	// error when tee-worker detects that identity verification data is linked to different web2
+	// account than expected
 	WrongWeb2Handle,
```
jonalvarezz: My understanding was that it errors when the provided proof doesn't match the account to be linked. 

The description sounds like "IdentityAlreadyLinked"

kziemianek: @jonalvarezz, maybe I should replace word _linked_ and add an example, what do you think about:

	// error when tee-worker detects that identity verification data is related to other web2
	// account than expected, for example wrong tweet id was provided

jonalvarezz: @kziemianek sounds good

jonalvarezz (approved):

Kailai-Wang:
Resolve the conflict and we are good to go

PR #2247, opened 2023-11-11
add scale codec index to enums

kziemianek:
Adds explicit enum variant index for enums that are scalec encoded/decoded. This should decrease the number of breaking changes when adding/removing/reorganizing enum variants on parachain node/tee-worker side.

It only introduces breaking change because of `TrustedCall` and `TrustedGetter`, for example previously `set_user_shielding_key` had an index 4 because evm trusted calls where ignored but I decided to assigned them indexes according to definition order. If you feel it's worth it we can adjust it in a way that this  will be breaking change free.


felixfaisal (approved):
I like this PR, an important change for maintenance.

Kailai-Wang (approved):
Thanks, looks good!

jonalvarezz:
This is high value, thanks @kziemianek !

PR #2249, opened 2023-11-13
Adjust the order of ts-test's fmt-check

0xverin:
### Context
resolves p-232


Kailai-Wang:
Thanks - looks good but CI needs to pass

0xverin:
> Thanks - looks good but CI needs to pass

Fixed!

Kailai-Wang (approved):

PR #2261, opened 2023-11-16
missing configs for new VCs added recently

BillyWooo:
minor fix, following up #2231


felixfaisal (approved):
Looks good, This is something we have to do until `request-vc` is phased out from `DI` and TrustedCalls.

kziemianek (approved):

PR #2244, opened 2023-11-10
Handling minor issues

0xverin:
### Context

resolves #1853 
resolves p-182
resolves p-81

And split `parachain-api/interfaces/identity/definitions` into `vc` `identity` `trustedCalls` for easier use and management.

PR #2262, opened 2023-11-17
Release new client api(new exported types)

0xverin:
### Context

Since #2244 has been merged, adding exports for three new types (identity, vc, trusted_operations), we need to release a new version of the package.


0xverin:
Regarding the changes for this version, I'd like to have a discussion:

- Should we consider using a dedicated branch(i.e.`release-latest`, `release-dev`, `release-staging`) for releasing the  package? With this approach, we won't need PRs; we can simply publish after modifying the version and type-definitions.

cc @jonalvarezz @grumpygreenguy

0xverin:
> Regarding the changes for this version, I'd like to have a discussion:
> 
> * Should we consider using a dedicated branch(i.e.`release-latest`, `release-dev`, `release-staging`) for releasing the  package? With this approach, we won't need PRs; we can simply publish after modifying the version and type-definitions.
> 
> cc @jonalvarezz @grumpygreenguy

I've come up with a solution that doesn't require submitting PRs or creating a dedicated branch for this. When manually executing the [workflow](https://github.com/litentry/litentry-parachain/actions/workflows/release-ts-api-package.yml), we can query the package version, make the necessary changes (manually input version), and then proceed with the release.

arieljbirnbaum:
@Verin1005 

> I've come up with a solution that doesn't require submitting PRs or creating a dedicated branch for this. When manually executing the [workflow](https://github.com/litentry/litentry-parachain/actions/workflows/release-ts-api-package.yml), we can query the package version, make the necessary changes (manually input version), and then proceed with the release.

Sounds good for now; ultimately I think we should update the different release workflows to include this :thinking: Should be a separate ticket tho

Kailai-Wang:
A dumb question: if `0.0.1` is not used by anyone, can we simply replace it?

arieljbirnbaum:
@Kailai-Wang 

> A dumb question: if `0.0.1` is not used by anyone, can we simply replace it?

Nope. For security reasons, publishing the same version twice is not allowed.

arieljbirnbaum:
Then again, the versions of the API packages should probably follow the versions of the respective clients, no? Otherwise we'd have to add them to the release tag and then it'll be a horrendous mess :joy:

0xverin:
https://linear.app/litentry/issue/P-266/add-version-modification-to-release-client-api-workflows

0xverin:
> Then again, the versions of the API packages should probably follow the versions of the respective clients, no? Otherwise we'd have to add them to the release tag and then it'll be a horrendous mess 😂

It should be.

0xverin:
After some consideration, no matter which approach we take, we still need a PR to handle the client API's version.  So we won't be modifying the workflow. directly modify the version via a PR. I've discussed this with @Kailai-Wang , and we both agree that a PR is necessary.

kziemianek:
Just to add my two cents here: we are not releasing anything important for the client, it's just a refactor/restructure of the code. 
I just doubt whether it need a release.

0xverin:
> Just to add my two cents here: we are not releasing anything important for the client, it's just a refactor/restructure of the code. I just doubt whether it need a release.

We have re-defined the file structure, it might be worth doing a release. :p
However, I will wait to release until after this [PR](https://github.com/litentry/litentry-parachain/pull/2275/files#diff-f5f0c9b0e57eac75b915382530a5cd4b938ded86e6f53aca92755be1b9f151b2R25) is merged.

0xverin:
one of the version change is covered in https://github.com/litentry/litentry-parachain/pull/2273

PR #2263, opened 2023-11-17
block import time metrics

### LINKED ISSUE #1707: Build a sidechain block production metrics

kziemianek:
Adds more detailed metrics to montior parentchain/sidechain block import time and sidechain block production.

resolves: P-50
resolves: P-60
resolves: #1707


tee-worker/core-primitives/enclave-metrics/src/lib.rs
```diff
@@ -42,7 +44,8 @@ pub enum EnclaveMetric {
 	StfTaskExecutionTime(Box<RequestType>, f64),
 	SuccessfulTrustedOperationIncrement(TrustedCall),
 	FailedTrustedOperationIncrement(TrustedCall),
-	// OracleMetric(OracleMetric<MetricsInfo>),
+	ParentchainBlockImportTime(Duration),
+	SidechainBlockImportTime(Duration), // OracleMetric(OracleMetric<MetricsInfo>),
```
Kailai-Wang: The comment is bit confusing

tee-worker/core-primitives/test/src/mock/sidechain_ocall_api_mock.rs
```diff
@@ -23,18 +23,18 @@ use std::sync::RwLock;
...
 use codec::{Decode, Encode};
 use core::marker::PhantomData;
-use itp_ocall_api::EnclaveSidechainOCallApi;
+use itp_ocall_api::{EnclaveMetricsOCallApi, EnclaveSidechainOCallApi};
 use itp_types::{BlockHash, ShardIdentifier};
 use sgx_types::{sgx_status_t, SgxResult};
 use std::vec::Vec;
 
-pub struct SidechainOCallApiMock<SignedSidechainBlockType> {
+pub struct OCallApiMock<SignedSidechainBlockType> {
```
Kailai-Wang: Maybe we can leave it unchanged? It's not that important and we could keep it in sync with upstream

Kailai-Wang (commented):
Thanks, looks good!
However, I don't think it resolves P-50 completely.

Ideal goal of P-50 is to be able to visualize the breakdown of a sidechian block frame

kziemianek:
> Thanks, looks good! However, I don't think it resolves P-50 completely.
> 
> Ideal goal of P-50 is to be able to visualize the breakdown of a sidechian block frame

Thanks @Kailai-Wang , regarding more granular metrics we have another ticket P-60 - I've noticed it once I finished this implementation hovewer I can continue topic in this pull request.

tee-worker/core-primitives/test/src/mock/sidechain_ocall_api_mock.rs
BillyWooo: I think what you mean is: try to keep sync with upstream with least effort, right? @Kailai-Wang 
I would suggest not to touch it, unless it blocks us from any development/functionality.

kziemianek:
@Kailai-Wang can we merge this PR and continue P-60 in follow up ?

BillyWooo (commented):
I would suggest let's revert the name change: `SidechainOCallApi -> OCallApi`.

tee-worker/core-primitives/enclave-metrics/src/lib.rs
kziemianek: moved to next line, actually this is commented out variant

tee-worker/core-primitives/test/src/mock/sidechain_ocall_api_mock.rs
kziemianek: changed

kziemianek:
> I would suggest let's revert the name change: `SidechainOCallApi -> OCallApi`.

Renamed

BillyWooo (approved):
Thank you. LGTM.
Please resolve the conflicts.

PR #2268, opened 2023-11-20
Add `remove identity` feature for  dev purpose

Traf333:
### Context

in order to test some functionality with linking accounts and preventing resetting whole state, we decided to introduce `remove identity` functionality which potentially saves dev time for long run

### Testing Evidences

attached consistent commands execution which shows the id graph changes

Screenshot 2023-11-23 at 10 25 20


tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -178,6 +178,9 @@ pub enum TrustedCall {
 	handle_vcmp_error(Identity, Option<Identity>, VCMPError, H256),
 	#[codec(index = 17)]
 	send_erroneous_parentchain_call(Identity),
+	#[cfg(not(feature = "production"))]
```
Kailai-Wang: IMO It's OK that we keep it in enum even for prod - otherwise we'll need to have two different client-api packages.

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -579,6 +584,21 @@ where
 					Ok(TrustedCallResult::Streamed)
 				}
 			},
+
+			TrustedCall::remove_identity(signer, who, identity) => {
```
Kailai-Wang: I think we shall add the production guard here, like error out for prod directly.

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -579,6 +584,21 @@ where
...
+				debug!("remove_identity, who: {}", account_id_to_string(&who));
+
+				Self::remove_identity_internal(
```
Kailai-Wang: Maybe we don't need the extra `*internal` function as it's a bit too verbose - just to increase the readability :)

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -231,6 +235,26 @@ pub mod pallet {
...
+		#[pallet::call_index(5)]
+		#[pallet::weight({15_000_000})]
+		pub fn remove_identity(
+			origin: OriginFor<T>,
+			who: Identity,
+			identity: Identity,
+		) -> DispatchResult {
+			T::ManageOrigin::ensure_origin(origin)?;
+
+			ensure!(LinkedIdentities::<T>::contains_key(&identity), Error::<T>::IdentityNotExist);
```
Kailai-Wang: Additionally, we need to check the identity belongs to `who`'s IDGraph

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -231,6 +235,26 @@ pub mod pallet {
...
+			ensure!(identity != who, Error::<T>::RemovePrimeIdentityDisallowed);
```
Kailai-Wang: It's for dev only, so I think we should allow the removal of prime identity.

We could even go one step further and change the function args (and CLI args) to:
```
  who: identity,
  identities: Option<Vec<Identity>>
```

So we remove a vector of identities in the `who`'s IDGraph, and if `identities` is `None`, we simply remove the whole IDGraph. How does that sound?

Kailai-Wang (commented):
Thank you! I left some comments.
Something that is still missing is the unittest :)

You need to get the CI (clippy) working too

tee-worker/app-libs/stf/src/trusted_call.rs
kziemianek: Since we explicitly set indices we can have this variant only for non prod mode while having single client-api package. If the client try to use it on prod mode the sidechain will fail on deconding. imho it would be ok. If we follow `__Unused` pattern in ts definition this variant wouldn't be even reflected in generated types so it would not be present in generated API.

kziemianek: `tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts`
We should add this varaint to `TrustedCall` definition in this file ^ as `_Unused_*`, `link_identity_callback`, `request_vc_callback`,  `handle_imp_error`, `handle_vcmp_error`, `send_erroneous_parentchain_call` should be added too - otherwise we will have problems if we add new variant with index 19 that should be used by `client-api`.

tee-worker/litentry/pallets/identity-management/src/lib.rs
BillyWooo: It's also nice to support `remove [who]` directly -> remove the whole IDGraph.

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -796,6 +816,7 @@ where
 			TrustedCall::balance_shield(..) => debug!("No storage updates needed..."),
 			// litentry
 			TrustedCall::link_identity(..) => debug!("No storage updates needed..."),
+			TrustedCall::remove_identity(..) => debug!("No storage updates needed..."),
```
BillyWooo, line 959: maybe same guard here.

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: Either way would work - I'd argue having it not guarded with prod feature has better maintainability and is less error-prone, while having it guarded makes sure we won't have such trusted call in prod whatsoever.

So your choice :) @Traf333

Traf333: For now I'm not too confident to push code which might leak to production 😅 and in addition what @kziemianek said sounds like voodoo magic for me, so I want to choose the way with learning about `__Unused` pattern to understand it better. 

By the way, where can I read more about `codec(index...)` stuff @kziemianek ? Does it just reffers to `trusted_operations/definitions.ts`?

kziemianek: @Traf333 check this [issue](https://linear.app/litentry/issue/P-230/explicitly-define-index-of-scale-encoded-enums) and linked sources. If you have any questions feel free to reach out to me 👍🏼

tee-worker/app-libs/stf/src/trusted_call.rs
BillyWooo: Make use of `if_production_or`. Maybe simply give some message out that it's not supported in production mode.

Traf333: Sounds good 👍

tee-worker/core-primitives/stf-primitives/src/error.rs
```diff
@@ -67,6 +67,9 @@ pub enum StfError {
 	InvalidAccount,
 	#[codec(index = 15)]
 	UnclassifiedError,
+	#[codec(index = 15)]
```
kziemianek: You shouldn't assign already assigned index, change it to 16 :)

Kailai-Wang:
Would you like to resolve all conversations that have been resolved? :)
And normally we have to make CI pass before the review

Traf333:
> Would you like to resolve all conversations that have been resolved? :)
> And normally we have to make CI pass before the review

Thanks for pointing that out. Yes, I'm waiting to get all tests green and then will ask for the re-review.

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -133,6 +134,9 @@ pub enum TrustedCall {
 	handle_vcmp_error(Identity, Option<Identity>, VCMPError, H256),
 	#[codec(index = 24)]
 	send_erroneous_parentchain_call(Identity),
+	#[cfg(not(feature = "production"))]
+	#[codec(index = 18)]
```
Kailai-Wang: Any reason to use 18 instead of 25? :) Maybe you aren't aware that the codec index has changed in the past commits

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -676,6 +682,27 @@ where
 					Ok(TrustedCallResult::Streamed)
 				}
 			},
+			TrustedCall::remove_identity(signer, who, identities) => {
```
Kailai-Wang, line 686: I don't think this will even compile under production, under which `remove_identity` wouldn't even exist

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
```diff
@@ -38,6 +38,12 @@ export default {
                 activate_identity: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, H256)",
                 request_vc: "(LitentryIdentity, LitentryIdentity, Assertion, Option<RequestAesKey>, H256)",
                 set_identity_networks: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, Vec<Web3Network>, H256)",
+                __Unused_link_identity_callback: "Null",
+                __Unused_request_vc_callback: "Null",
+                __Unused_handle_imp_error: "Null",
+                __Unused_handle_vcmp_error: "Null",
+                __Unused_send_erroneous_parentchain_call: "Null",
+                __Unused_remove_identity: "Null",
```
Kailai-Wang: This will not work, `link_identity_callback` starts from index 20 😂

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -91,6 +93,8 @@ pub mod pallet {
 		IdentityNotExist,
 		/// creating the prime identity manually is disallowed
 		LinkPrimeIdentityDisallowed,
+		/// removing the prime identity manually is disallowed
+		RemovePrimeIdentityDisallowed,
```
Kailai-Wang: I don't think this one is used, plus we should actually allow removing the prime identity (which removes the whole IDGraph)

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
Kailai-Wang: If you don't need them in ts-test, you don't need to add it

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: Actually `index = 5` would be better, calls starting from index 20 are normally callbacks

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -231,6 +235,42 @@ pub mod pallet {
...
+					IDGraphs::<T>::remove(&who, identity);
+					LinkedIdentities::<T>::remove(identity);
+				}
+			} else {
+				// removing prime identity with all linked identities
+				IDGraphs::iter_prefix(&who)
+					.collect::<IDGraph<T>>()
+					.iter()
+					.for_each(|(identity, _context)| LinkedIdentities::<T>::remove(identity));
+				let _ = IDGraphs::<T>::clear_prefix(&who, 1, None);
```
Kailai-Wang: I think we should use something like `u32::MAX` instead of `1`

Kailai-Wang (changes requested):
The overall logic looks fine! I do have a few comments tho

tee-worker/app-libs/stf/src/trusted_call.rs
Traf333: wow, it seems index changes ever dev branch merge. I'm a bit lost now on what todo with @kziemianek 's recommendation on updating `./tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts`

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
Traf333: yep now I'm a bit lost on using __Unused definitions

tee-worker/litentry/pallets/identity-management/src/lib.rs
Traf333: good catch, thanks 🙌

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
kziemianek: Use it if you want to keep indices unchanged while removing invariant or if invariant is feature toggled.
Otherwise generated ts types wouldn't match indices on server side causing integration problems.

Kailai-Wang: You could refer to https://docs.substrate.io/reference/scale-codec/ for more info

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: The index is related to the scale codec, while rust has the syntax to start from a specific integer, ts package doesn't have such. That's why we have all those `__Unused` fields to fill the gaps

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
Traf333: yes, that's what I understood from your initial proposition, but what if `link_identity_callback` starts on index = 20? shall we insert 14 empty records to sync the index?

Kailai-Wang: for ts yes :)
those callbacks are supposed to be called internally (e.g. from another trusted call), so we don't expect them to be called by ts directly.

kziemianek: If we don't want to reflect some calls in generated types because we don't expect consumers to dispatch them then it's a good reason to use `__Unused` too as Kai described.

Traf333: hm, from what I understood we also don't plan to call `remove_identity` by ts, isn't it? Also what about original integritee trusted calls, starting from index 50, are they also called internally?

Kailai-Wang: both answers would be no.

However, `remove_identity` is more like triggerable from externals (rust-cli or ts), so conventionally it should follow `set_identity_networks` - that's why I suggested `index 5`

Traf333:
> Something that is still missing is the unittest :)

I wish to have this message at the beginning of the task, 😃  since I couldn't find any tests initially and thought you guys just ignore them in order to use integration tests. Anyway, thanks for raising this point - addressed ✅

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
Traf333: > original integritee trusted calls, starting from index 50, are they also called internally - no

Does it mean we should fill empty indices with `__Unused: 'Null'`? @kziemianek

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
```diff
@@ -38,6 +38,8 @@ export default {
                 activate_identity: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, H256)",
                 request_vc: "(LitentryIdentity, LitentryIdentity, Assertion, Option<RequestAesKey>, H256)",
                 set_identity_networks: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, Vec<Web3Network>, H256)",
+                __Unused_remove_identity: "Null",
```
kziemianek, line 41: This is ok.
It has index 5 in Rust code `#[codec(index = 5)]`
And it matches ts definition:


`link_identity` - 0
`deactivate_identity` - 1
`activate_identity` - 2
`request_vc` - 3
`set_dentity_networks` - 4 
`__Unused_remove_identity` - 5

So in the future if we add another call with index 6 like 
```
#[codec(index = 6)]
new_trusted_call_with_index_6
```

We will end up having such ts definition: 

```
        TrustedCall: {
            _enum: {
                link_identity:
                    "(LitentryIdentity, LitentryIdentity, LitentryIdentity, LitentryValidationData, Vec<Web3Network>, Option<RequestAesKey>, H256)",
                deactivate_identity: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, H256)",
                activate_identity: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, H256)",
                request_vc: "(LitentryIdentity, LitentryIdentity, Assertion, Option<RequestAesKey>, H256)",
                set_identity_networks: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, Vec<Web3Network>, H256)",
                __Unused_remove_identity: "Null",
               new_trusted_call_with_index_6: "(LitentryIdentity)",
            },
        },
```
Because of filling the gap with `__Unused` it will have index 6 so it will match index on Rust side, otherwise there would be a mismatch. This is the main reason for using `__Unused` pattern.

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
kziemianek: Current code is ok, I explained it a bit in this [comment](https://github.com/litentry/litentry-parachain/pull/2268/files#r1415166487)

Kailai-Wang (approved):
Looks good, congratz!

Traf333:
The words can not describe how happy I am
200w

PR #2270, opened 2023-11-20
feat: add threadpools for isolated vc issuance

felixfaisal:
As the title suggests, This PR implements thread pool so that vc issuance thread can handle requests concurrently


Kailai-Wang (approved):
Looks good to me, thanks!

PR #2274, opened 2023-11-23
Fix block import error for EVM node client

wangminqi:
This is referring to https://github.com/paritytech/frontier/issues/603


wangminqi:
The sync issue "Runtime Ethereum block not found, rejecting!" is now recovered.
The client side needs to upgrade first since we introduce new HostFunction.

Kailai-Wang (approved):
Looks good, thanks.

Tested OK locally.

PR #2275, opened 2023-11-23
Add all vc di case

### LINKED ISSUE #2080: Define the correct Achainable type
### Context
Currently, we are unable to define the achainable type correctly. It seems that manual modifications are required after building in order to use it, which is not feasible. We need to discuss this.

Additionally, I've examined the code in IDHub:

1. They manually created the [achainable](https://github.com/litentry/identity-hub/blob/bcb04d90a7e2b42a65284e077f21044505098344/libs/chain-types/src/lib/interfaces/vcManagement/types.ts#L49) type. 

2. Here, [CorePrimitivesAssertion](https://github.com/litentry/identity-hub/blob/bcb04d90a7e2b42a65284e077f21044505098344/libs/chain-types/src/lib/interfaces/trustedCalls/definitions.ts#L66C59-L66C82) is declared, but it is changed to [Assertion](https://github.com/litentry/identity-hub/blob/bcb04d90a7e2b42a65284e077f21044505098344/libs/chain-types/src/lib/interfaces/trustedCalls/types.ts#L63) manually after building.

In any case, achainable should either be manually defined in the definitions or directly generated by the parachain for use, rather than being modified after building.

0xverin:
### Context

As the topic, and removed some duplicate VC cases, keeping only cases with different types and functionalities.

also:
    resolvse p-119 
    resoves #2080
    resolves p-292


0xverin:
~~[Reason](https://linear.app/litentry/issue/P-286/achainable-assertion-failure) for CI error.~~

tee-worker/ts-tests/integration-tests/di_vc.test.ts
```diff
@@ -88,4 +67,46 @@ describe('Test Vc (direct invocation)', function () {
...
+                new PolkadotSigner(context.substrateWallet.alice),
+                aliceSubject,
+                context.api.createType('Assertion', assertion).toHex(),
+                context.api.createType('Option<RequestAesKey>', aesKey).toHex(),
+                requestIdentifier
+            );
+
+            const res = await sendRequestFromTrustedCall(context, teeShieldingKey, requestVcCall);
+            assert.isTrue(res.do_watch.isFalse);
+            assert.isTrue(res.status.asTrustedOperationStatus[0].isInvalid);
```
jonalvarezz: I'm not sure I am following: so, we test that it fails?

0xverin: This will definitely fail here, as indicated in the [comments](https://github.com/litentry/litentry-parachain/pull/2275/files/bc17c91cb992f1877fa4e8b5d323c8e6b4c52320#diff-803593e3a8b9e787b7d0c1864c9da5b19e7bc805f93a9ebd8615cc5bf4823641R150). Therefore, the purpose here is not to test for failure, but to test if DI can properly retrieve the response and trigger an event

tee-worker/ts-tests/integration-tests/di_vc.test.ts
```diff
@@ -13,33 +13,13 @@ import type { IntegrationTestContext } from './common/type-definitions';
 import { aesKey } from './common/call';
 import { LitentryPrimitivesIdentity } from 'sidechain-api';
 import { subscribeToEventsWithExtHash } from './common/transactions';
+import { basicAssertions, extraAssertions } from './common/utils/vc-helper';
+
```
arieljbirnbaum: The names `basicAssertions` and `extraAssertions` aren't telling me much about what these are _supposed_ to be or why they are split that way; could you give them more descriptive names please?

arieljbirnbaum (changes requested):

tee-worker/ts-tests/integration-tests/di_vc.test.ts
0xverin: Renamed, and with the help of the comments [here](https://github.com/litentry/litentry-parachain/pull/2275/files#diff-803593e3a8b9e787b7d0c1864c9da5b19e7bc805f93a9ebd8615cc5bf4823641R157), can it be understood more clearly?

Kailai-Wang: You probably need to re-request the review, otherwise @grumpygreenguy won't know it's ready to review :)

tee-worker/ts-tests/integration-tests/di_vc.test.ts
arieljbirnbaum: OK, I _think_ I get it now -- in that case, please rename the variable to `unconfiguredAssertions` (to make it explicitly that they should be configured but are currently not); then add a [pending test](https://mochajs.org/#pending-tests) and a Linear issue to tackle this at a later point :pray:

arieljbirnbaum (changes requested):

tee-worker/ts-tests/integration-tests/di_vc.test.ts
0xverin: > OK, I _think_ I get it now -- in that case, please rename the variable to `unconfiguredAssertions` (to make it explicitly that they should be configured but are currently not); then add a [pending test](https://mochajs.org/#pending-tests) and a Linear issue to tackle this at a later point 🙏

Thanks! https://linear.app/litentry/issue/P-328/follow-up-make-unconfiguredassertions-pass-the-test

arieljbirnbaum (approved):

PR #2278, opened 2023-11-24
Skip `update-client-api` conditionally in CI

Traf333:
### Context

this PR speeds up the automation for the case when worker or parachain should not rebuild and the `update-client-api` job shouldn't be triggered


.github/workflows/ci.yml
```diff
@@ -602,6 +608,8 @@ jobs:
     needs:
       - parachain-build-dev
       - tee-build
+      - set-condition
+    if: needs.set-condition.outputs.run_update_client_api
```
0xverin: Will this work here? I feel like it won't work. Should we` if: needs.set-condition.outputs.run_update_client_api== 'false'`? :p

0xverin (commented):
We might need to add something: 

1. When rebuilding the `parachain` and `tee`, we need to upload the `update-client-api` build outputs at that time. Maybe upload it to Docker Hub or somewhere. 

2. When the `parachain` and `tee` are not rebuilt, we can simply download client-api and use it from there. 

I believe this PR is meant to implement all these functionalities.

.github/workflows/ci.yml
Traf333: yeah, good catch missed that one but with `true` flag 😄

Traf333:
> 2. When the parachain and tee are not rebuilt, we can simply download client-api and use it from there.

@Verin1005 @Kailai-Wang might there be a case when client-api might be different?

For example:
1. DeveloperA works on the task with only ts-tests and in this case everything looks evident. 
- no rebuild
- download client-api bundle from somewhere
- check tests
2. DeveloperB works on some huuuge task with bunch of changes on parachain and worker side
- build parachain/worker
- build client-api bundle and upload somewhere (<--- this is where developerA might struggle since stored bundle changes)
- check tests


the main question is - are we going to share client-api build file globally or just for the following builds within the same PR?

.github/workflows/ci.yml
```diff
@@ -91,6 +91,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          lfs: true
```
BillyWooo, line 94: why need `lfs` here?

BillyWooo:
>  just for the following builds within the same PR

I think this is the answer. 

I would NOT recommend to upload these CI build output to be committed/pushed into origin repo. 
Besides, if:
1. user needs to user client-api, they should install it. -> that's the reason why make a release, right?
2. user changes a lot in parachain/worker, which result in changes to client-api, then use the build output from current PR. that's enough.

Kailai-Wang:
> the main question is - are we going to share client-api build file globally or just for the following builds within the same PR?

There's exactly such precedent in the docker image handling - you may check how `litentry/litentry-parachain:latest` is handled in the current CI.

In short, the client-api package (if updated) should be pushed to wherever it's stored only for `dev` branch. For anyone who is working on the feature branch and updates the client-api, the change should be kept local

Traf333:
Ok, got it. Thanks. I'll keep consistency and implement it with docker

.github/workflows/ci.yml
```diff
@@ -596,15 +596,53 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: client-api
+          path: client-api.tar.gz
+          if-no-files-found: 'warn'
+
+      - name: Early exit if bundle exists and parachain/worker reubild is not required
+        run: |
+          # Check if either rebuild_parachain or rebuild_tee is true
+          if [[ "${{ needs.set-condition.outputs.rebuild_parachain }}" = "false" && "${{ needs.set-condition.outputs.rebuild_tee }}" = "false" && -f "client-api.tar.gz" ]]; then
```
BillyWooo: why not simply moving the client-api download steps into this if branch?

.github/workflows/release-ts-api-package.yml
```diff
@@ -2,31 +2,37 @@
...
 jobs:
-        
   update-publish:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
 
       - name: Set env variables
-        run: echo "RELEASE_IMAGE_VERSION=${{ inputs.tag }}" >> "$GITHUB_ENV"
+        run: echo "RELEASE_IMAGE_VERSION=${{ inputs.release }}" >> "$GITHUB_ENV"
```
BillyWooo: Here `inputs.release` used. Line 83 `inputs.tag` used. Can you explain a bit?

Traf333: The separation of the inputs helps to pick `parachain/worker` with `release` input, build the stuff and release it with specific tag. For example with latest `parachain/worker` we can publish `dev/alpha` release of `client-api`. Before it was possible to publish only the same versions.

.github/workflows/ci.yml
```diff
@@ -675,12 +688,46 @@ jobs:
...
+          echo "//npm.pkg.github.com/:_authToken=\${{ secrets.RELEASE_TS_API_PACKAGE_TOKEN }}" > .pnpmrc
+          pnpm config set @litentry:registry https://npm.pkg.github.com
+          pnpm i -w @litentry/parachain-api @litentry/sidechain-api
+
+      - name: Compress packages
+        run: |
+          tar -zcvf client-api.tar.gz tee-worker/client-api/node_modules/@litentry
+
+      - name: Upload client-api as artifact
+        if: needs.set-condition.outputs.rebuild_client_api == 'false'
```
BillyWooo: why need it again? already one at line 694.

.github/workflows/ci.yml
```diff
@@ -675,12 +688,46 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: client-api
+          path: client-api.tar.gz
+          if-no-files-found: 'warn'
+
   tee-test:
     runs-on: ubuntu-latest
+    if: always()
```
BillyWooo: this `if: always()` might not be necessary.

.github/workflows/ci.yml
Traf333: good catch. That's a leftover. Thanks 🙏

.github/workflows/ci.yml
Traf333: it is necessary since one of the jobs from `needs` will be skipped. Here is an example without `always()`
https://github.com/litentry/litentry-parachain/actions/runs/7021920077?pr=2278

.github/workflows/ci.yml
```diff
@@ -596,91 +596,13 @@ jobs:
...
-          if-no-files-found: ignore
-          retention-days: 3
-      
   tee-test:
     runs-on: ubuntu-latest
+    if: needs.set-condition.outputs.run_tee_test == 'true'
     needs:
       - set-condition
-      - update-client-api
+      - tee-build
```
BillyWooo, line 605: I don't think we need `tee-build` here. e.g.: only ts-tests changes, so only need to run tee-test.

Traf333: we still need it as in tee-build there is a step to upload new build or docker build and only after this step we should trigger tee-test
this step `.github/workflows/ci.yml:465`

Traf333: in the other words
parachain-build and tee-build jobs both will be triggered every time but in one case they will build and upload zip file and in the other case they will pull docker image and then upload zip file

.github/pull_request_template.md
```diff
@@ -15,6 +15,6 @@ Please apply following PR-related labels when appropriate:
 
 ### Testing Evidences
 
-Please attach any relevent evidences if applicable
+Please attach any relevant evidences if applicable
```
arieljbirnbaum, line 18: <3

tee-worker/docker/lit-di-evm-identity-test.yml
```diff
@@ -17,7 +17,10 @@ services:
                 condition: service_healthy
         networks:
             - litentry-test-network
-        entrypoint: "bash -c '/usr/local/worker-cli/lit_ts_test.sh test-di-evm-identity 2>&1' "
+        entrypoint: |
+            /usr/local/worker-cli/lit_ts_api_package_build.sh -p 9912 -u ws://litentry-node -W http://litentry-node -V wss://litentry-worker-1 -A 2011 -C /usr/local/bin/litentry-cli 2>&1
+            && \
+            /usr/local/worker-cli/lit_ts_test.sh test-resuming-worker 2>&1
```
arieljbirnbaum: Hm, maybe it'd be cleaner to wrap it into the `lit_ts_test.sh` script itself? ^_^

arieljbirnbaum (approved):
lgtm but I think it could be refactored a bit; not a blocker in any case.

tee-worker/docker/lit-di-evm-identity-test.yml
BillyWooo: Looks like all copy-paste error within these xxx-test.yml files. 
Besides, it's better to keep `bash -C` from the old file.

.github/workflows/ci.yml
```diff
@@ -743,16 +655,14 @@ jobs:
         run: |
           cd tee-worker/docker
           docker compose -f litentry-parachain.build.yml build
-      
+
       - name: Integration Test ${{ matrix.test_name }}
-        if: needs.set-condition.outputs.run_tee_test == 'true'
```
BillyWooo, line 748: why need to remove this line?

tee-worker/docker/lit-di-evm-identity-test.yml
Traf333: Thanks, valid comment. Initially, I thought the same and later came to the conclusion that running tests and preparing setup for them might be two different intentions. Like if there is already generated metadata. Though I agree that there is a space for refactoring

.github/workflows/ci.yml
Traf333: I moved to whole job instead of skipping only step. As it seems redundant to trigger job without running tests in it. By the way, was there specific case why we need the job to be completed even if we skip tests?

PR #2280, opened 2023-11-27
Remove client-api type definitions in ts-tests

0xverin:
### Context
resolves p-289


0xverin:
In the ts-tests, all types that were defined in common/type-definitions for client-api have been deleted and replaced with types directly imported from client-api . Only [a few types](https://github.com/litentry/litentry-parachain/pull/2280/files#diff-8842a3ba8102cede4a70b418b32fe28c2e187cee438832e66b9e2b6e2b28e753) that cannot be used directly have been retained.
cc @Kailai-Wang

Kailai-Wang (commented):
Thank you!
What about types in `parachain-api/prepare-build/interfaces/definitions.ts`, do we need them all?

BillyWooo:
> Thank you! What about types in `parachain-api/prepare-build/interfaces/definitions.ts`, do we need them all?

Yeah. I think we need that file to keep the types.

BillyWooo (approved):
LGTM

Kailai-Wang:
E.g. `LitentryIdentity` in identity/definitions.ts => we have `LitentryPrimitivesIdentity` already, don't we?

0xverin:
> E.g. `LitentryIdentity` in identity/definitions.ts => we have `LitentryPrimitivesIdentity` already, don't we?

`LitentryIdentity` is used in definitions, such as in `trusted_operations/definitions`, `sidechain/definitions`, so it's more like a necessary variable - we use `LitentryIdentity` to define the structure of `LitentryPrimitivesIdentity` for use in other definitions. Therefore, while their content is the same, LitentryIdentity cannot be deleted.

PR #2281, opened 2023-11-27
Update upstream commit base `922e8733`

Kailai-Wang:
### Context

As topic, it includes a significant amount of changes and adjustments.

Notes:
- some adjustments in scripts are done as well
- I see some changes in `build_and_test.yml` (e.g. separate worker and cli build) but I consider them less suitable for our case, so I didn't apply it to our CI
- in `pallets/teerex/src/lib.rs` the `block_number` in `confirm_processed_parentchain_block` is not compact-encoded anymore - to be consistent with upstream
- there's a new `noop` trusted call - so the **codec has changed** - the metadata/types need to be re-generated
- `data-dir` in development config is removed, to avoid directory conflict for multiple users on the same host


tee-worker/.github/workflows/publish-docker-sidechain.yml
```diff
@@ -0,0 +1,43 @@
...
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+      - name: Load Worker & Push
+        env:
+          DOCKER_BUILDKIT: 1
+        run: |
+            docker image load --input integritee-worker-sidechain.tar.gz
+            docker images --all
+            docker push integritee/sidechain:${{ github.event.release.tag_name }}
```
kziemianek, line 36: Is this tag correct ?

tee-worker/app-libs/parentchain-interface/src/integritee/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+	Invoke(InvokeArgs),
+	// Litentry
+	#[codec(index = 2)]
+	LinkIdentity(LinkIdentityArgs, Option<MultiAddress<AccountId32, ()>>, H256),
+	#[codec(index = 3)]
+	DeactivateIdentity(DeactivateIdentityArgs, Option<MultiAddress<AccountId32, ()>>, H256),
+	#[codec(index = 4)]
+	ActivateIdentity(ActivateIdentityArgs, Option<MultiAddress<AccountId32, ()>>, H256),
+	#[codec(index = 5)]
+	RequestVC(RequestVCArgs, Option<MultiAddress<AccountId32, ()>>, H256),
```
kziemianek, line 64: `tee-worker/app-libs/parentchain-interface/src/integritee/mod.rs`

What do you think about renaming this file or creating separate one for litentry parentchain and removing all litentry related stuff from this?

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -44,38 +58,36 @@ pub use litentry_primitives::{
...
-
 pub type IMTCall = ita_sgx_runtime::IdentityManagementCall<Runtime>;
 pub type IMT = ita_sgx_runtime::pallet_imt::Pallet<Runtime>;
 
 #[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
 #[allow(non_camel_case_types)]
 pub enum TrustedCall {
 	#[codec(index = 0)]
-	balance_set_balance(Identity, AccountId, Balance, Balance),
+	noop(Identity),
```
kziemianek: Can we skip this TrustedCall or add it after `send_erroneous_parentchain_call` with `index=18`?

tee-worker/cli/test_auto_shielding_with_transfer_bob.sh
```diff
@@ -0,0 +1,140 @@
...
+            CLIENT_BIN=$OPTARG
+            ;;
+        *)
+            echo "invalid arg ${OPTARG}"
+            exit 1
+    esac
+done
+
+# Using default port if none given as arguments.
+INTEGRITEE_RPC_PORT=${INTEGRITEE_RPC_PORT:-9944}
```
kziemianek, line 37: LITENTRY ?

kziemianek (approved):
I briefly checked file by file. Looks fine!

tee-worker/.github/workflows/publish-docker-sidechain.yml
BillyWooo: I think we don't check these files within `tee-worker/.github`.  They are all upstream. Not relevant to us.

Kailai-Wang: It's not used/triggered - this file is automatically created when applying the patch

tee-worker/app-libs/parentchain-interface/src/integritee/mod.rs
Kailai-Wang: Thanks, I'll create a `litentry` mod next to `integritee` (and revert all changes in the current `integritee` mod)

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: I was struggling with that, the problem is putting `noop` in the middle of many meaningful trustedCalls seems uncommon.

At the same time I can see integritee trying to extend their trusted call collection too in the future.
How about I move all litentry fields to the beginning and start integritee fields from a large index (like 50)? In ts definitions we only list litentry trusted calls. How does that sound? 
```
pub enum TrustedCall {
  #[codec(index = 0)]
  link_identity(...),
  ...
  #[codec(index = 50)]
  noop(..),
  #[codec(index = 51)]
  ...
}
```

tee-worker/cli/test_auto_shielding_with_transfer_bob.sh
Kailai-Wang: This file is not used in tests either, so I didn't bother changing it

tee-worker/app-libs/stf/src/trusted_call.rs
BillyWooo: As 50 already mentioned, why not start from 100?
But I would vote for this solution.

tee-worker/cli/src/trusted_base_cli/commands/set_balance.rs
```diff
@@ -48,7 +51,7 @@ impl SetBalanceCommand {
 
 		let (mrenclave, shard) = get_identifiers(trusted_args, cli);
 		let nonce = get_layer_two_nonce!(signer, cli, trusted_args);
-		let top: TrustedOperation = TrustedCall::balance_set_balance(
+		let top: TrustedOperation<TrustedCallSigned, Getter> = TrustedCall::balance_set_balance(
```
BillyWooo, line 54: Noticed some part of the cli code using `let xxx = TrustedOperation<x,x>(Trustedxxx::xx())`. But here using `let xxx: TrustedOperation<x,x> = Trustedxxx:xx()`. Should keep the consistency? 
e.g. [tee-worker/cli/src/evm/commands/evm_read.rs](https://github.com/litentry/litentry-parachain/pull/2281/files#diff-a0657c51fc79b2f626549e863cb50069d30886c4a23e565b1c4a00f581bc7885)

tee-worker/app-libs/stf/src/trusted_call.rs
kziemianek: Do we want to let users dispatch such calls ? If not maybe we can skip them ?

Kailai-Wang: @kziemianek do you mean dispatching integritee calls? Unlikely.
They are kept untouched to not make "too much" code difference which hopefully facilitate the maintenance. If we follow this principle (= remove code that we don't need), then a lot of code will be removed 😂

tee-worker/cli/src/trusted_base_cli/commands/set_balance.rs
Kailai-Wang: ~~Good catch - I'll change it~~

No actually they are different impls - here `into_trusted_operation` is used

tee-worker/app-libs/stf/src/trusted_call.rs
kziemianek: @Kailai-Wang yes, but I was thinking that we can make this trade off and ignore not needed changes in enums that are part of public contract.

Using offset indices sounds good to me too.

Kailai-Wang: I'll use offset indices for now.

Just had a rough check - if we remove these enums, then quite some code will be affected including the imported fns/crates.. that would probably result in a larger change scope :(

tee-worker/app-libs/parentchain-interface/src/integritee/mod.rs
Kailai-Wang: During the refactoring process I realised there's more than that - including many enclave initialisation process.

I'll create a separate issue for it, and maybe have a discussion beforehand if we should rename all

Kailai-Wang: Issue created: P-294, currently assigned to @felixfaisal

Issue #1133, opened 2022-12-29
Review the enclave registration process


Kailai-Wang:
### Context
The last modification regarding enclave registration in worker-main won't change the logic, because `trusted_url` differs each time although `MRENCLAVE` is identical, is this intended? https://github.com/litentry/litentry-parachain/blob/a56a56b05f309d29a9df51ec72ad50f46fb4dd34/tee-worker/service/src/main.rs#L544-L554

<img width="860" alt="image" src="https://user-images.githubusercontent.com/7630809/210019265-e5cc414d-6614-4f76-b1a5-8f6c235d1fbd.png">

We might need to ask Integritee why such a registration process is implemented, it seems more intuitive to have MRENCLAVE as key and a list of workers as (sub)-values.

---
:heavy_check_mark: Please set appropriate **labels** and **assignees** if applicable.

Kailai-Wang:
Please note if we want to change it we have to consider some implicit consequences, e.g. when `search_peers` it looks for all registered enclaves (even with duplicated `MRENCLAVE`) to find all worker-URLs: https://github.com/litentry/litentry-parachain/blob/a56a56b05f309d29a9df51ec72ad50f46fb4dd34/tee-worker/service/src/worker.rs#L150-L161

felixfaisal:
Hey @Kailai-Wang 
Could this issue be related to this - https://github.com/litentry/litentry-parachain/issues/1820

PR #2287, opened 2023-11-28
Return encoded extrinsics without padding

### LINKED ISSUE #2168: Worker always fails to recognize itself as primary validateer
After a startup the worker registers itself in teerex's enclave registry and tries to determine whether he's a primary validateer (the first to be registered) or not but from my experience it looks like it always says it's not.

I've checked the campaign instances logs and it seems the problem existed there too.

```
[2023-09-27T21:26:23Z ERROR litentry_worker] ExtrinsicFailed ExtrinsicNotFound
[<] Extrinsic got finalized. Hash: None
```

In short, the implementation could not determine the block at which extrinsic was finalized so it fallbacks to current block. Then based on current block it check if any registered enclave existed before. The answer is yes, thus it pretends as it's not the primary while in fact it is.

```
if we_are_primary_validateer {
	println!("[+] We are the primary validateer");
} else {
	println!("[+] We are NOT the primary validateer");
}
```

Upstream ticket: https://github.com/scs/substrate-api-client/issues/624

Kailai-Wang:
### Context

It should fix the `ExtrinsicNotFound` error during enclave registration.

With the padded version (post-padded with spaces `0x20`) the extrinsic submission would succeed since the decoding goes through even with postpositioned (unused) bytes. However, the substrate-api-client wouldn't find the hash with the prolonged extrinsic that we pass in. 

This is also the reason why CI of #2281 failed

fixes #2168 
fixes P-151

Related: https://github.com/scs/substrate-api-client/issues/624


Kailai-Wang:
Reported upstream too: https://github.com/integritee-network/worker/pull/1505

kziemianek:
@Kailai-Wang does it fix #2168 ?

Kailai-Wang:
> @Kailai-Wang does it fix #2168 ?

It should.

However, (I was about to say to you) the registration process needs to be reviewed after this one and #2281 get merged: https://github.com/litentry/litentry-parachain/issues/1133

BillyWooo (approved):
Thank you for the explanation. I think it's good to go.

PR #2292, opened 2023-11-29
feat: P-278 add WeirdoGhostGang holder vc

wli-pro:
### Context

1 added nodereal jsonrpc client
2 added WeirdoGhostGang holder VC
3 restructured bit nodereal based VC 

### Labels

### Testing Evidences

Please attach any relevent evidences if applicable


wli-pro:
Case 1: 
Generate VC successfully: 
image


Case 2:
Unable to generate VC due to data provider issue (created by passing wrong api key to simulate data provider non-2xx reponse):
image

primitives/core/src/assertion.rs
```diff
@@ -225,6 +225,9 @@ pub enum Assertion {
 	#[codec(index = 14)]
 	GenericDiscordRole(GenericDiscordRoleType),  // (generic_discord_role_type)
 
+	#[codec(index = 15)]
+	WeirdoGhostGangHolder,
+
```
Kailai-Wang: Maybe we should append the new VC to the end of the list, otherwise VCs previously generated with Index 15 would be decoded into this one and you'll see weird errors or even silently succeed but that's undesired - I think it's better to let it fail explicitly (like can't decode at all)

tee-worker/client-api/parachain-api/prepare-build/interfaces/vc/definitions.ts
```diff
@@ -22,6 +22,7 @@ export default {
                 A20: "Null",
                 Oneblock: "OneBlockCourseType",
                 GenericDiscordRole: "GenericDiscordRoleType",
+                WeirdoGhostGangHolder: "Null",
```
Kailai-Wang: Please beware we should apply https://github.com/litentry/litentry-parachain/pull/2275/files#diff-f5f0c9b0e57eac75b915382530a5cd4b938ded86e6f53aca92755be1b9f151b2 if we change the index of `WeirdoGhostGangHolder` to 18

tee-worker/litentry/core/assertion-build/src/nodereal/weirdo_ghost_gang_holder.rs
Kailai-Wang, line 1: why bnb_domain has its dedicated folder/mod while this one doesn't?

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
```diff
@@ -0,0 +1,343 @@
...
+		let api_key = GLOBAL_DATA_PROVIDER_CONFIG.write().unwrap().nodereal_api_key.clone();
+		let api_retry_delay = GLOBAL_DATA_PROVIDER_CONFIG.write().unwrap().nodereal_api_retry_delay;
+		let api_retry_times = GLOBAL_DATA_PROVIDER_CONFIG.write().unwrap().nodereal_api_retry_times;
+		let api_url = GLOBAL_DATA_PROVIDER_CONFIG
+			.write()
+			.unwrap()
+			.nodereal_api_chain_network_url
+			.clone();
+		let base_url =
+			api_url.replace("{chain}", chain.to_string()).replace("{network}", "mainnet");
```
Kailai-Wang: I don't see `{network}` can have values other than `mainnet` 🤔

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
```diff
@@ -0,0 +1,343 @@
...
+				Ok(response) => return Ok(response),
+				Err(err) => {
+					let req_err: Error =
+						Error::RequestError(format!("Nodereal enhanced api error: {}", err));
+					match err {
+						HttpError::HttpError(code, _) =>
+							if code == 429 {
+								// Too Many Requests
+								// exponential back off
+								thread::sleep(base_delay * 2u32.pow(retries as u32));
```
Kailai-Wang, line 173: This will potentially block the whole VC issuance thread for 32s - this can easily be an attacking interface without threadpool and VC-Isolation. 

Even if that's not intentional - people could spam the requests if they don't get requests for half a minute which would make it even worse 😂 

How likely will this 429 happen? Can we only retry for 1-2 times and then error out (i.e. ask the user to retry it later)

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
```diff
@@ -0,0 +1,343 @@
...
+		})
+	}
+}
+
+#[derive(Serialize, Debug)]
+pub struct GetNFTHoldingsParam {
+	// The address of the account in hex format
+	pub account_address: String,
+	// (optional) Please specify the type of token you query for, e.g. "ERC721", "ERC1155", etc.
+	pub token_type: String,
```
Kailai-Wang, line 200: If it's optional then it's better to use `Option<>` IMO

tee-worker/litentry/core/assertion-build/src/nodereal/weirdo_ghost_gang_holder.rs
```diff
@@ -0,0 +1,163 @@
...
+			Err(e)
+		},
+	}
+}
+
+pub fn build(req: &AssertionBuildRequest) -> Result<Credential> {
+	debug!("WeirdoGhostGang holder");
+
+	let mut has_nft = false;
+	let mut client = NoderealJsonrpcClient::new(NoderealChain::Eth);
```
Kailai-Wang: Then why do we need to have `NoderealChain` enums? Is it rather preparation for the upcoming requirements?

tee-worker/litentry/core/assertion-build/src/nodereal/weirdo_ghost_gang_holder.rs
```diff
@@ -0,0 +1,163 @@
...
+			},
+		}
+	}
+
+	if !has_nft {
+		if let Some(e) = error {
+			return Err(Error::RequestVCFailed(
+				Assertion::WeirdoGhostGangHolder,
+				ErrorDetail::DataProviderError(ErrorString::truncate_from(
+					format!("{e:?}").as_bytes().to_vec(),
```
kziemianek: Why only last error ?

tee-worker/litentry/core/data-providers/src/discord_official.rs
```diff
@@ -133,7 +133,7 @@ mod tests {
 		let message_id = "1";
 
 		let data_provider_config = DataProviderConfigReader::read().unwrap();
-		let mut client = DiscordOfficialClient::new(&data_provider_config);
+		let mut client: DiscordOfficialClient = DiscordOfficialClient::new(&data_provider_config);
```
kziemianek: You don't need that type here :)

tee-worker/enclave-runtime/src/stf_task_handler.rs
```diff
@@ -56,7 +56,12 @@ pub unsafe extern "C" fn run_stf_task_handler(dpc: *const u8, dpc_size: usize) -
 			dpc.set_sora_quiz_master_id(data_provider_config.sora_quiz_master_id);
 			dpc.set_sora_quiz_attendee_id(data_provider_config.sora_quiz_attendee_id);
 			dpc.set_nodereal_api_key(data_provider_config.nodereal_api_key);
+			dpc.set_nodereal_api_retry_delay(data_provider_config.nodereal_api_retry_delay);
+			dpc.set_nodereal_api_retry_times(data_provider_config.nodereal_api_retry_times);
```
Kailai-Wang, line 60: Shall we hardcode these numbers into enclave?

Kailai-Wang (commented):
A very rough review by me - we need to have another eye (who is more familiar with it) on the detailed request/build logic.. @BillyWooo / @zhouhuitian

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
```diff
@@ -0,0 +1,343 @@
...
+}
+
+impl ReqPath {
+	pub fn new(api_key: &str) -> Self {
+		Self { path: "v1/".to_string() + api_key }
+	}
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ReqBody<P>
```
kziemianek: We already have this type in `itp-rpc`, have you considered reusing it ?

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
kziemianek: If we always use "mainnet" why replace it ?

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
kziemianek, line 200: Consider using `Option` type for values that are optional.

tee-worker/enclave-runtime/src/stf_task_handler.rs
wli-pro: The optimal value of these parameters won't be easily decided, and keeping it in config makes it can be tuned easily without needing to recompile

primitives/core/src/assertion.rs
BillyWooo: Yes. We should only append from tail.

tee-worker/litentry/core/assertion-build/src/nodereal/weirdo_ghost_gang_holder.rs
higherordertech: Do we want to join all the errors instead?

tee-worker/litentry/core/assertion-build/src/nodereal/weirdo_ghost_gang_holder.rs
wli-pro: Yeah preparation for expending the vc to different network, e.g.  NoderealChain::Bsc etc later

tee-worker/litentry/core/assertion-build/src/nodereal/weirdo_ghost_gang_holder.rs
```diff
@@ -0,0 +1,163 @@
...
+		.collect::<Vec<String>>();
+
+	let mut error: Option<DataProviderError> = None;
+	for address in addresses {
+		match check_has_nft(&mut client, address.as_str()) {
+			Ok(res) => {
+				has_nft = res;
+				break
+			},
+			Err(err) => {
```
outofxxx: If error occurs, we should return to client directly.

tee-worker/litentry/core/assertion-build/src/nodereal/weirdo_ghost_gang_holder.rs
wli-pro: The logic is that if one account has multiple evm address, we are gonna call each address to see if it has this NFT token, till one has, or exhausted all of the evm address.

Along this journey, example:

You have 2 evm address under idgraph: evm1 and evm2
1 pass evm1 to data provider api, received error with response code 500
2 then pass evm2 to data provider api,  also received error with response code 500
So current it returns the last step i.e. step 2 's error response

Or we can join all of the errors for all steps?

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
```diff
@@ -105,11 +105,16 @@ where
 			Assertion::GenericDiscordRole(role_type) =>
 				lc_assertion_build::generic_discord_role::build(&self.req, role_type),
 
+			Assertion::WeirdoGhostGangHolder =>
+				lc_assertion_build::nodereal::weirdo_ghost_gang_holder::build(&self.req),
+
```
BillyWooo: when you shift the order, maybe also move this branch to the end.

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/vc_handling.rs
```diff
@@ -81,11 +81,16 @@ where
 			Assertion::GenericDiscordRole(role_type) =>
 				lc_assertion_build::generic_discord_role::build(&self.req, role_type),
 
+			Assertion::WeirdoGhostGangHolder =>
+				lc_assertion_build::nodereal::weirdo_ghost_gang_holder::build(&self.req),
```
BillyWooo: same here. maybe also move this branch to the end.

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
BillyWooo: Yes, 1-2 times would be enough. It's better to ask user to retry it from IDHub.

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
BillyWooo: Actually don't get where the "mainnet" come from ...

tee-worker/litentry/core/credentials/src/nodereal/weirdo_ghost_gang_holder.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright 2020-2023 Trust Computing GmbH.
```
BillyWooo: the same question here: it's better to keep same structure as bnb_domain module.

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
kziemianek: This will make this more complicated but what about not blocking thread while waiting for retry ? 
Can this request go back to the queue ?

tee-worker/litentry/core/assertion-build/src/nodereal/weirdo_ghost_gang_holder.rs
kziemianek: Yes, I see 🙃 - thats why I ask why you treat last error better than previous 😂 
If we decide to bubble up this error I would vote for joining them all.

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
wli-pro: This is a good idea, I will refactor it in seperate PR.

tee-worker/enclave-runtime/src/stf_task_handler.rs
kziemianek: I would opt for making these values same for all enclaves.
This retry strategy is still subject of change, see https://github.com/litentry/litentry-parachain/pull/2292/files#r1408881954

tee-worker/litentry/core/assertion-build/src/nodereal/bnb_domain/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+					"nodeHash": "0xr4b0bf28adfcee93c5069982a895785c9231c1fe",
+					"bind": "0xr4b0bf28adfcee93c5069982a895785c9231c1fe",
+					"name": "000",
+					"expires": "2024-10-30T18:40:51Z"
+				},
+				{
+					"nodeHash": "0xr4b0bf28adfcee93c5069982a895785c9231c1fe",
+					"bind": "0xr4b0bf28adfcee93c5069982a895785c9231c1fe",
+					"name": "999",
+					"expires": "2024-03-19T18:16:59Z"
```
kziemianek, line 258: This hardcoded date will make `non_expired_domain_infos_works` test case fail in 4 months. Use `Instant::now()` with offset to make it work at any time in future.

tee-worker/litentry/core/credentials/src/nodereal/nft_holder/weirdo_ghost_gang_holder.rs
```diff
@@ -0,0 +1,38 @@
...
+// VC type / info
+const WEIRDO_GHOST_GANG_HOLDER_INFOS: (&str, &str) =
+	("WeirdoGhostGang Holder", "You are WeirdoGhostGang NFT holder");
+
+pub trait WeirdoGhostGangHolderAssertionUpdate {
+	fn update_weirdo_ghost_gang_holder_assertion(&mut self, value: bool);
+}
+
+impl WeirdoGhostGangHolderAssertionUpdate for Credential {
+	fn update_weirdo_ghost_gang_holder_assertion(&mut self, value: bool) {
```
kziemianek, line 31: Have you considered including block at which this check was performed (EVM block for example), currently we pass `latest` in query but NFTs can be transferred, or maybe it doesn't matter ?

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
```diff
@@ -0,0 +1,348 @@
...
+		let mut headers = Headers::new();
+		headers.insert(CONNECTION.as_str(), "close");
+		let client = build_client(base_url.as_str(), headers);
+
+		NoderealJsonrpcClient { api_key, api_retry_delay, api_retry_times, client }
+	}
+
+	// https://docs.nodereal.io/docs/cups-rate-limit
+	// add retry functionality to handle situations where requests may surpass predefined constraints.
+	fn retry<A>(&mut self, action: A) -> Result<RpcResponse, Error>
```
kziemianek, line 141: I would love to see test for this.

kziemianek (approved):
I left few comments which may be addressed in separate PR if you find it relevant.
The code looks fine, except retry mechanism which we decided to improve in follow up PR.

Nice 🙃

BillyWooo (approved):
Thanks for the refinement. LGTM.
Let's not forget comments from @kziemianek

tee-worker/litentry/core/assertion-build/src/nodereal/bnb_domain/mod.rs
wli-pro: @zhouhuitian FYI

PR #2293, opened 2023-11-29
feat: application level rate limiter 

felixfaisal:
Some notes for the PR 
- I've created a new struct instead of using `TrustedCall` and `TrustedOperation` for the P-159 feature 
- By employing a new struct, We can use the `AES` key provided in the request, instead of having two `AES` keys in a request. (Update: We are still using two `AES` keys)
- Used a `rwlock` to perform rate limiting


tee-worker/cli/src/trusted_operation.rs
```diff
@@ -74,17 +75,10 @@ pub(crate) fn perform_trusted_operation<T: Decode + Debug>(
 pub(crate) fn perform_direct_operation<T: Decode + Debug>(
 	cli: &Cli,
 	trusted_args: &TrustedCli,
-	top: &TrustedOperation<TrustedCallSigned, Getter>,
+	top: &TrustedVCRequestSigned,
```
BillyWooo, line 79: So `perform_direct_operation` will be limited only for VC request operations? If yes, `perform_direct_operation` can be renamed to `perform_direct_vc_request_operation`. 
I just think the scenario would be different. There might be other operation in the future which also needs direct operation. How do you think so?

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -41,20 +41,57 @@ use itp_stf_executor::traits::StfEnclaveSigning;
...
+	requests: RwLock<HashMap<String, Instant>>,
+}
+
+impl RateLimiter {
+	fn should_allow(&self, request_key: String) -> Result<(), String> {
+		let requests = self.requests.read().unwrap();
+		log::error!("Request key: {:?}", request_key);
+		if let Some(&last_instant) = requests.get(&request_key) {
+			if last_instant.elapsed() < Duration::from_secs(10) {
+				return Err("Request limit reached".to_string()) // Reject if within 5 minutes
```
BillyWooo: 5 minutes? vs `Duration::from_secs(10)` ?

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -298,11 +292,11 @@ fn send_direct_request<T: Decode + Debug>(
 	let (sender, receiver) = channel();
 	direct_api.watch(jsonrpc_call, sender);
 
-	debug!("waiting for rpc response");
+	println!("waiting for rpc response");
```
BillyWooo: would `info!` be better?

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -298,11 +292,11 @@ fn send_direct_request<T: Decode + Debug>(
...
 	loop {
 		match receiver.recv() {
 			Ok(response) => {
-				debug!("received response");
+				println!("received response");
```
BillyWooo: `info!`?

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -369,7 +363,7 @@ fn send_direct_request<T: Decode + Debug>(
 fn send_direct_vc_request<T: Decode + Debug>(
 	cli: &Cli,
 	trusted_args: &TrustedCli,
-	operation_call: &TrustedOperation<TrustedCallSigned, Getter>,
+	operation_call: &TrustedVCRequestSigned,
```
BillyWooo, line 358: The same argument: this will limit the interface only for `VC request`. Should we do it?

tee-worker/litentry/core/vc-issuance/lc-vc-task-sender/src/lib.rs
```diff
@@ -47,6 +48,48 @@ pub struct VCRequest {
...
+	pub vc_request: TrustedVCRequest,
+	pub signature: LitentryMultiSignature,
+}
+
+impl TrustedVCRequestSigned {
+	pub fn new(vc_request: TrustedVCRequest, signature: LitentryMultiSignature) -> Self {
+		TrustedVCRequestSigned { vc_request, signature }
+	}
+
+	pub fn verify_signature(&self, mrenclave: &[u8; 32], shard: &ShardIdentifier) -> bool {
```
BillyWooo: where this 'verify_signature` is used?

Kailai-Wang:
It makes sense to have a dedicated type.
While we are already here, does it even make more sense to give it a more general name and enum? We'll have more calls like VcRequest for sure, so for calls which require a signature but don't go into the tx pool.

So basically adding a new member in `TrustedOperation` - I don't know a good name tho, maybe `TrustedOperation::native`?

We'll probably need a dedicated port (or reuse the existing one) to accept such requests.

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -41,20 +41,57 @@ use itp_stf_executor::traits::StfEnclaveSigning;
...
+		log::error!("Writing to the hashmap");
+		let mut requests = self.requests.write().unwrap();
+		requests.insert(request_key, Instant::now()); // Update with current Instant
+		Ok(())
+	}
+}
+
+// Global instance using lazy_static
+lazy_static! {
+	static ref GLOBAL_RATE_LIMITER: RateLimiter =
```
kziemianek: Is it possible to avoid using globals and pass `RateLimiter` as an argument to functions/object's constructors ?

Kailai-Wang:
I didn't check the code very carefully yet, but seeing the description:
> By employing a new struct, We can use the AES key provided in the request, instead of having two AES keys in a request

I'm just wondering if we should keep it consistent with other requests.
We intentionally embed the AesKey parameter in the request itself, because:
1. it needs to be async handled and routed to other workers too (maybe we don't need to consider this in vc-request case)
2. it allows the requestor to use a key that is different from request-encryption-key to encrypt the result: https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/app-libs/stf/src/trusted_call.rs#L82-L83

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: Updated ✅

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: Yeah we can do this, just curious on why you would want to avoid globals. Let me know what you think.

kziemianek: I find it easier to undestand the code and function/object's dependencies when all they come in with api arguments rather being bound to globals. 

It's also easier to test the code because you can pass whatever instance you want to object being tested. 

If you find it easier to share object through globals then it's ok, but I would like to promote using it only when really needed.

felixfaisal: I concur, updated to not use globals

tee-worker/cli/src/trusted_operation.rs
felixfaisal: I would say for now I will change it to `perform_direct_vc_request_operation` and create a separate issue to move this operation to `TrustedOperation`.

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -71,20 +72,13 @@ pub(crate) fn perform_trusted_operation<T: Decode + Debug>(
...
 ) -> TrustedOpResult<T> {
-	match top {
-		TrustedOperation::direct_call(call) => match call.call {
-			TrustedCall::request_vc(..) => send_direct_vc_request(cli, trusted_args, top, key),
-			_ => Err(TrustedOperationError::Default { msg: "Only request vc allowed".to_string() }),
-		},
-		_ =>
-			Err(TrustedOperationError::Default { msg: "Only Direct Operation allowed".to_string() }),
-	}
+	send_direct_vc_request(cli, trusted_args, top, key)
```
BillyWooo, line 82: TBH, I still in favor of `perform_direct_operation` to be more general here. Otherwise we just need keep one of `perform_direct_vc_operation` or `send_direct_vc_request`. That's enough. 
But the more general purpose will left the arch for easy extension.

felixfaisal: Yeah I think I understand, I didn't want to do it in this PR, The plan is to move it to TrustedOperation and include it as `native_call` and have an enum for it similar to how we have `DirectCall`. This is the issue - https://linear.app/litentry/issue/P-352/move-request-vc-parameters-to-trustedoperation

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -72,20 +73,13 @@ pub(crate) fn perform_trusted_operation<T: Decode + Debug>(
 	}
 }
 
-pub(crate) fn perform_direct_operation<T: Decode + Debug>(
+pub(crate) fn perform_direct_vc_operation<T: Decode + Debug>(
```
Kailai-Wang, line 76: Maybe we don't even need this fn? It seems to be a wrapper only

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -39,24 +38,56 @@ use itp_sgx_crypto::{ShieldingCryptoDecrypt, ShieldingCryptoEncrypt};
...
+		let requests = self.requests.read().unwrap();
+		if let Some(&last_instant) = requests.get(&request_key) {
+			if last_instant.elapsed() < Duration::from_secs(10) {
+				return Err("Request limit reached".to_string()) // Reject if within 10 seconds
+			}
+		}
+		drop(requests); // Drop read lock
+
+		let mut requests = self.requests.write().unwrap();
+		requests.insert(request_key, Instant::now()); // Update with current Instant
```
Kailai-Wang, line 86: So if I keep retrying like every 9 sec, I'll never get access to it?

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -39,24 +38,56 @@ use itp_sgx_crypto::{ShieldingCryptoDecrypt, ShieldingCryptoEncrypt};
...
+
+#[cfg(feature = "sgx")]
+use std::sync::SgxRwLock as RwLock;
+
+pub struct RateLimiter {
+	requests: RwLock<HashMap<String, Instant>>,
+}
+
+impl RateLimiter {
+	fn new() -> Self {
```
Kailai-Wang, line 72: Ideally the cooldown period should be passed in as constructor parameter

Kailai-Wang: and 10 might be too strict :)

Kailai-Wang (commented):
Thanks, in general it looks good

Can you wait a bit until I submit a PR for p-83? There I removed the need of `enclave_account` in BuildRequest, as it was solely for A13 which is a) deprecated now b) we'll only support DI for vc request anyway 

After that this problem isn't supposed to exist anyway and you could revert it to old impl :)

Edit: P-83 is ready to be reviewed

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: Yeah Ideally I want a bit more sophisticated cooldown period, For example, If someone's request returns a specific  error then the cooldown period is depending on that type of error, I think it would take some time to implement such a mechanism. So yeah we need a consensus on the arbitrary number, How does a 2 second period sound like for all cases for now?

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: No, if you retry every 9 seconds, The first one will return Err `Request Limit Reached` and the second one will succeed. We are not writing to `rwlock` when it has been called within 10 seconds and are returning an `Err`.

tee-worker/cli/src/trusted_operation.rs
felixfaisal: Yeah, We don't need this, I'll just remove this

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
Kailai-Wang: I see, thanks - I misread

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
Kailai-Wang: Yes we should avoid using globals when possible (and use [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection) instead)

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: I think this will also depend on the implementation for this issue - https://linear.app/litentry/issue/P-298/implement-an-abortstrategy-for-assertion-building

Kailai-Wang: Yes that takes some time to implement. 2s is fine as the initial configuration

The constructor parameter doesn't mean a static one either, it's just the initial one :)

felixfaisal:
I will raise this PR after rebasing with `dev` to accomodate new changes and adjust accordingly

PR #2294, opened 2023-11-29
clone only filtered tops

kziemianek:
I found this during working on other tasks. It simply reduces the amount of TOPs cloned during retrieval from top pool.


Kailai-Wang (commented):
Now that we are super close, would you like to use `filter_map` to chain them? :D

kziemianek:
> Now that we are super close, would you like to use `filter_map` to chain them? :D

I think I prefer 

```
			.filter(|o| matches!(o.data(), TrustedOperation::get(Getter::trusted(_))))
			.map(|o| o.data().clone())
```

over 

```
			.filter_map(|o| {
				if matches!(o.data(), TrustedOperation::get(Getter::trusted(_))) {
					Some(o.data().clone())
				} else {
					None
				}
			})
```

What do you think @Kailai-Wang ?

Kailai-Wang:
I actually reckon this has a better readability:
```
.filter_map(|o| match o.data() {
  TrustedOperation::get(Getter::trusted(_)) => Some(o.data().clone()),
  _ => None
})
.collect()
```
But yea it's rather a flavor's problem

Kailai-Wang (approved):
Either way I'll approve it

PR #2295, opened 2023-11-29
do not run mock server in production build

kziemianek:
There is no practical need to run mock server even if it's possible.
Maybe it would be possible to remove this cli argument if we use [clap derive](https://docs.rs/clap/latest/clap/_derive/_tutorial/chapter_0/index.html) but currently our cli args definition is placed in `yml` file.

fixes: P-269


BillyWooo (approved):
Nice

Kailai-Wang (approved):
Thanks!

PR #2296, opened 2023-11-30
Hotfix of tee-test dependency in CI

Kailai-Wang:
### Context

As topic, otherwise the tee-test will be launched too early:

image

~~I added some comments to further explain the usage in our case, because~~

~~# [3] please beware that if a job in `needs` is skipped, its dependent job will also be skipped,~~
~~# see https://github.com/actions/runner/issues/491~~

No we can't use it in job level, because if the matrix is completely skipped, the status check will be pending forever, that was why I forced it to always run:

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -599,9 +599,12 @@ jobs:
 
   tee-test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Traf333 (approved):
I was curious why didn't we skip the whole job and now it's clear. Thanks @Kailai-Wang

PR #2298, opened 2023-11-30
fix evm build

kziemianek:
Our evm build broke recently.
I noticed it because I run `scripts/pre-commit.sh` before each push.


Kailai-Wang (approved):
Good catch, thanks!
Maybe other features (like teeracle) are broken too as we don't use/compile them 😂

kziemianek:
`pre-commit.sh` script checked:

* sidechain
* evm
* teeracle
* offchain-worker

PR #2302, opened 2023-12-01
Verify data provider certs

kziemianek:
Adds twitter server's certifiacte verification step to ensure enclaves are talking to correct service.

fixes: P-301


Kailai-Wang:
Thank you! Since it's a root CA, does it fix other endpoints too, or you need more intermediate certs?
Also I'm wondering if there's any known rust crate already providing the hardcoded CAs - like what browser does. Maybe `webpki`?

kziemianek:
We don't need intermediate certs as they are derived from root certs.  Intermediate certs may change more frequently so it's better to rely on root certs.

We will need to extend list of certs, for example VIP3 uses cert issued by Amazon RSA 2048 M01.
We will also need to maintain them and add new ones in case one of them expires. 

I couldn't find ready to use crate containing those certs 😿 .

Kailai-Wang:
If there's no rust crate doing this, we have to maintain it ourselves.
It's fine that we have multiple root CAs to cover different data providers.

Regarding intermediate CAs - what if the data provider's cert isn't signed by root CA and when following the cert chain, it finds the intermediate CAs missing? I haven't checked the data provider's certs so I don't know if that's the case, just thinking

Kailai-Wang (approved):
Thanks!

Only one small request: can you please add a `certificates/README.md` to note down how/where is each cert downloaded? In case someone else wants to maintain them when you are not available?

kziemianek:
@Kailai-Wang I changed the code so it's using root certs from `webpki-roots` crate. I'm still not sure why upstream used `empty_root_store`.

Kailai-Wang:
> @Kailai-Wang I changed the code so it's using root certs from `webpki-roots` crate. I'm still not sure why upstream used `empty_root_store`.

My guess is they just want to start with empty store and add the [required certs](https://github.com/integritee-network/worker/tree/a36ae4b283e17b25130b95dede87e90d2282c900/app-libs/oracle/src/certificates) (in oralce) explicitly

Kailai-Wang (approved):
Looks good, thanks!

Maybe we can change the PR title to be more accurate :)
Does it solve the certs for all current endpints?

kziemianek:
All certs except tee microservice.

PR #2304, opened 2023-12-04
Record IDGraph hash on chain

### LINKED ISSUE #1955: Sychronization of client side state of IDGraph
### Context
What's the context and background of this issue?

It's mainly about a consistent way of handling the IDGraph synchronisation. 

User might have multiple devices connecting to IDH which cached the local version of the IDGraph.

Now if one device tried to mutate the IDGraph, how would the other devices know the IDGraph has been mutated?

There's no centralised cloud service to do the synchronisation and notification, yet the clients have the needs to know if the local copy is valid or not.

### Task
What needs to be done here?

There might be two ways to solve that:
1 the sidechain includes the IDGraphs (or its hash) in the response of each mutation (IDGraph write operation)
2 upon each mutation, the sidechain publishes the hash to parachain, so there's a "idgraph prime identity -> hash of idgraph" mapping.
Precondition is the hash is deterministic.

I prefer No. 2. The idea is similar to etag in http. Both the client (local browser) and sidechain keep a copy of IDGraph, and use hash to determine if the local copy is outdated or not.

So IDhub will need to determine when to check those hashes (e.g. when the user logs into a new browser). If the hash doesn't match, IDHub knows the IDGraph is out of sync, then it needs to fetch it from sidechain again.

### Acceptance criteria
What's the definition of done, if applicable?

---

### LINKED ISSUE #2053: Sync IDGraph wihout passing full IDGraph around
### Context

- clients can have similar requirements to #2051 (imagine the user links the identity on another browser and goes back to the old browser)
- In some responses we return the full IDGraph, e.g. `set_user_shielding_key`, `link_identity`. Is it an overkill? Can we simply return the hash of IDGraph **after** the state mutation? Similar to the etag concept, the client can mutate the IDGraph locally and by comparing the hash it knows if the IDGraph is out of sync.

Again the registry to hold a mapping is only an initial idea, better suggestions are welcome.

---

Kailai-Wang:
### Context

As topic.

This PR:
- provides an `update_id_graph_hash` extrinsic for explicit population (e.g. for existing IDGraphs)
- adds an `id_graph_hash` field to all IDGraph mutation extrinsics, theoretically we could have separated it into two events (e.g. `IdentityLinked` and `IDGraphHashUpdated`) for a bit easier indexing, but considering they are always emitted together so IMO it still makes sense to consolidate them to look like an "atomic" operation.
- adds an `identity_networks_set` callback extrinsic
- adjusts ts-tests
- removes IMP and VCMP from Litmus as they're unused

fixes #1955
fixes #2053 

TODO (P-320):
adds a CLI to dump all account <> IDGraphHashes so that we can populate the initial on-chain storage once we go online


pallets/identity-management/src/lib.rs
```diff
@@ -264,14 +312,34 @@ pub mod pallet {
...
+			IDGraphHash::<T>::insert(account.clone(), id_graph_hash);
+			Self::deposit_event(Event::IdentityNetworksSet {
+				account,
+				id_graph_hash,
+				req_ext_hash,
+			});
+			Ok(Pays::No.into())
+		}
+
+		#[pallet::call_index(35)]
```
kziemianek, line 342: Is this harmless update or breaking ?

kziemianek (approved):

pallets/identity-management/src/lib.rs
Kailai-Wang: It's a (runtime) breaking change, but it has as much effect on the runtime as the addition of extrinsic parameters. So you'll need to update the parachain runtime anyway

PR #2307, opened 2023-12-05
Fix: Add worker tag input

0xverin:
### Context

Previously, when using Docker images, the worker and parachain used the same tag consistently. It's necessary to separate them now, as otherwise, errors similar to those discussed in [Slack](https://litentry.slack.com/archives/C05328L0GAF/p1701704472234529) will occur.

### DIFF
--- a/.github/workflows/release-ts-api-package.yml
+++ b/.github/workflows/release-ts-api-package.yml
@@ -25,9 +25,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set env variables
-        run: echo "RELEASE_IMAGE_VERSION=${{ inputs.worker-tag }}" >> "$GITHUB_ENV"
-
       - name: Pull litentry image optionally
         run: |
           docker pull parity/polkadot
--- a/.github/workflows/release-ts-api-package.yml
+++ b/.github/workflows/release-ts-api-package.yml
@@ -13,7 +13,7 @@ on:
           required: true
           default: 'latest'
         release-tag:
-          description: 'Ts api release tag'
+          description: 'Client-api release tag'
           required: true
           default: 'latest'
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-ts-api-package.yml
```diff
@@ -18,14 +26,14 @@ jobs:
...
 
       - name: Pull litentry image optionally
         run: |
           docker pull parity/polkadot
-          docker pull litentry/litentry-worker:${{ inputs.tag }}
-          docker pull litentry/litentry-cli:${{ inputs.tag }}
-          docker pull litentry/litentry-parachain:${{ inputs.tag }}
+          docker pull litentry/litentry-worker:${{ inputs.worker-tag }}
+          docker pull litentry/litentry-cli:${{ inputs.worker-tag }}
+          docker pull litentry/litentry-parachain:${{ inputs.parachain-tag }}
```
Kailai-Wang, line 33: I assume you haven't carefully verified it. This will not work.

Later you have
```
docker compose -f docker-compose.yml -f lit-ts-api-package-build.yml up --no-build --exit-code-from lit-ts-api-package-build lit-ts-api-package-build
```
but `docker-compose.yml` is hardcoded with `latest` tags all over the place.

So you basically pulled the images but they are never used.
Same applies to the `generate_parachain_artefacts.sh` step

Kailai-Wang (changes requested):

.github/workflows/release-ts-api-package.yml
Kailai-Wang: I think an easy solution is to (re)tag them locally with `docker tag`

0xverin: Thank you. I have re-tagged the image, and by looking at [docker images --all](https://github.com/litentry/litentry-parachain/actions/runs/7098761623/job/19321757527), we can see that `latest` points to `v0.9.17-1`.

Verin1005:
Waiting for P-330 to finish together.

Kailai-Wang (approved):
Looks good, thanks!
It doesn't work (yet) as we don't build tee-worker images when releasing

0xverin:
> Looks good, thanks! It doesn't work (yet) as we don't build tee-worker images when releasing

see https://github.com/litentry/litentry-parachain/pull/2313

PR #2308, opened 2023-12-05
Run tests in multiworker setup in CI

kziemianek:
This PR adds GHA job to run tests in multiworker setup. 
Let's run it for a while with `continue-on-error: true` to not fail pipelines.
Later we can remove this rule or and remove single worker tests or keep them both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -700,7 +700,6 @@ jobs:
           - test_name: lit-di-evm-identity-multiworker-test
           - test_name: lit-di-substrate-identity-multiworker-test
           - test_name: lit-di-vc-multiworker-test
-          - test_name: lit-id-graph-stats-multiworker
           - test_name: lit-ii-batch-test-multiworker
           - test_name: lit-ii-identity-multiworker-test
           - test_name: lit-ii-vc-multiworker-test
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -710,7 +792,7 @@ jobs:
     needs:
       - set-condition
       - parachain-ts-test
-      - tee-test
+      - tee-single-worker-test
```
BillyWooo, line 795: is it necessary to add `tee-multi-worker-test` dependency here?

BillyWooo (approved):
LGTM.
There are still old `tee-test` in the task list. Maybe github error?

kziemianek:
> LGTM. There are still old `tee-test` in the task list. Maybe github error?

Yes, I'm totally confused - I renamed them.

.github/workflows/ci.yml
Kailai-Wang: Yea I believe you need to add `tee-multi-worker-test` to the `needs` here

Kailai-Wang (approved):
Thanks - as discussed I'm fine with using this as the starting point.

The tests still talk to single worker tho, we'll need follow-up issues for it

.github/workflows/ci.yml
kziemianek: I didn't add it here because `tee-multi-worker-test` has `continue-on-error: true` - the output of that job shouldn't affect final outcome of workflow. 

Once we remove `continue-on-error` we will add it here. What do you think?

Kailai-Wang: That works too

PR #2309, opened 2023-12-05
Provide a cli command to dump idgraph hashes

Kailai-Wang:
### Context

as topic.

This PR adds a `dump-id-graph-hash` subcommand to the CLI with default shard.
It retrieves the IDGraph hash map via the `PublicGetter:: all_id_graph_hash`


BillyWooo (approved):
All looks good to me.

tee-worker/app-libs/stf/src/getter.rs
```diff
@@ -99,6 +99,8 @@ pub enum PublicGetter {
 	some_value,
 	#[codec(index = 1)]
 	nonce(Identity),
+	#[codec(index = 2)]
+	all_id_graph_hash,
```
kziemianek, line 103: Can we have this getter in two variants ? 

`MY|ALL` ?

tee-worker/app-libs/stf/src/getter.rs
```diff
@@ -278,6 +280,8 @@ impl ExecuteGetter for PublicGetter {
 				} else {
 					None
 				},
+			PublicGetter::all_id_graph_hash =>
+				Some(IdentityManagement::all_id_graph_hash().encode()),
```
kziemianek, line 284: May this cause problems if somebody starts to flood node with such getters ?

Kailai-Wang: Then it's not (only) a problem of this one: it applies to any RPC interface: getter, trusted getter or trusted calls. It also applies to any relay/parachains - you could spam costly requests like `state_getKeys`

kziemianek (approved):

PR #2310, opened 2023-12-05
Deterministic IDGraph sorting

Traf333:
### Context

this PR helps to enhance sorting function `IMT::get_id_graph(&who)` in case there are prime identity and to-be-linked identity have the same `link_block` number

### DIFF
--- a/tee-worker/litentry/pallets/identity-management/src/lib.rs
+++ b/tee-worker/litentry/pallets/identity-management/src/lib.rs
@@ -45,6 +45,7 @@ use frame_system::pallet_prelude::*;
 pub use litentry_primitives::{
 	all_evm_web3networks, all_substrate_web3networks, Identity, ParentchainBlockNumber, Web3Network,
 };
+use sp_core::{blake2_256, H256};
 use sp_std::{vec, vec::Vec};
 
 pub type BlockNumberOf<T> = <T as frame_system::Config>::BlockNumber;
@@ -311,6 +312,12 @@ pub mod pallet {
 			});
 
 			id_graph
+		}
+
+		pub fn all_id_graph_hash() -> Vec<(Identity, H256)> {
+			IDGraphLens::<T>::iter_keys()
+				.map(|k| (k.clone(), H256::from(blake2_256(&Self::get_id_graph(&k).encode()))))
+				.collect()
 		}
 
 		// get count of all keys account + identity in the IDGraphs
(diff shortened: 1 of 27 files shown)


tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -293,7 +294,22 @@ pub mod pallet {
...
 		pub fn get_id_graph(who: &Identity) -> IDGraph<T> {
 			let mut id_graph = IDGraphs::iter_prefix(who).collect::<IDGraph<T>>();
-			id_graph.sort_by(|a, b| Ord::cmp(&a.1.link_block, &b.1.link_block));
+
+			// Initial sort to ensure a deterministic order with case where prime identity `link_block` might be equal another identity-to be linked
+			id_graph.sort_by(|a, b| {
+				let a_is_prime_identity = &a.0 == who;
+				let b_is_prime_identity = &b.0 == who;
+
+				match (a_is_prime_identity, b_is_prime_identity) {
```
Kailai-Wang: Thank you!

Maybe we want a more general one - _theoretically_ two non-prime identities could be linked in the same block still, imagine you have two back-to-back identity linking requests, there's a chance that they finally get fit into one sidechain block.

So just checking if it's prime identity is not sufficient.
We need to have some rules that can be applied to any identity.

It can be simple as long as it's deterministic, e.g. using string alphabetic order of identity.to_did() if their `link_block`s are equal

tee-worker/litentry/pallets/identity-management/src/tests.rs
```diff
@@ -422,16 +422,53 @@ fn get_id_graph_works() {
...
+
+		let mut identities = vec![
+			(alice_twitter_identity(1), vec![]),
+			(alice_substrate_identity(), vec![Web3Network::LitentryRococo]),
+			(alice_evm_identity(), vec![Web3Network::Ethereum]),
+			(bob_substrate_identity(), vec![Web3Network::Litentry]),
+		];
+
+		let mut rng = rand::thread_rng();
+		identities.shuffle(&mut rng);
```
kziemianek: Why such shuffle is needed ? Maybe you could declare identities in desired order ?
I think the outcome of shuffle can be the same order as before.

Traf333: The intention was to show that any order is possible. And initially I've implemented it with two different orders for the identities, and then I realised that I could refactor it in a more dynamic way. Do you think it's unnecessary to have such approach?

BillyWooo: I think the fix order is enough and solid to verify whether you get the expected order later.

BillyWooo (approved):
LGTM. 
Suggest to only use fix order for testing. 
Make sure the CI passes.

tee-worker/litentry/pallets/identity-management/src/tests.rs
kziemianek: If all random orders are equal just pick one, if not then write as many test cases as comes to your mind.
Having static inputs for tests results in deterministic outputs.

Kailai-Wang (approved):
Thanks!

PR #2311, opened 2023-12-06
Hotfix holding time VC query

outofxxx:
The issue from the community is that the query for WBTC holdings held in November failed because currently only holdings before July 1st, 2023 will be queried. So according to the requirements of the product team, add the time point of December 1st, 2023 to temporarily solve the problem of holding time query false.

And partially fix [IDH-413](https://linear.app/litentry/issue/IDH-413/holding-time-vc-hot-fix)

### DIFF
--- a/primitives/core/src/assertion.rs
+++ b/primitives/core/src/assertion.rs
@@ -293,9 +293,8 @@ pub const ASSERTION_FROM_DATE: [&str; ASSERTION_DATE_LEN] = [
 	"2022-07-01",
 	"2023-01-01",
 	"2023-07-01",
-
-	// In order to address the issue of the community encountering a false query for WBTC in November, 
-	// the product team feels that adding this date temporarily solves this problem.
+	// In order to address the issue of the community encountering a false query for WBTC in
+	// November, the product team feels that adding this date temporarily solves this problem.
 	"2023-12-01",
 ];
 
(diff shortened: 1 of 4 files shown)


BillyWooo (approved):
LGTM
Make sure CI passes.

primitives/core/src/assertion.rs
```diff
@@ -292,6 +293,9 @@ pub const ASSERTION_FROM_DATE: [&str; 14] = [
 	"2022-07-01",
 	"2023-01-01",
 	"2023-07-01",
+	// In order to address the issue of the community encountering a false query for WBTC in
+	// November, the product team feels that adding this date temporarily solves this problem.
+	"2023-12-01",
```
jonalvarezz, line 298: Hi @zhouhuitian, 

I am not following whether we are certain about this fix. 

is it a shot in the dark, or are we positive it will address the issue?

BillyWooo: As I understand: the best fit is `2024-01-01`. But API doesn't support a timestamp from future. I would expect it to be a temp solution, but will be updated to `2024-01-01` later. @zhouhuitian right?

outofxxx: Yes, The product team prefer this temp solution to solve the problem asap.

PR #2312, opened 2023-12-06
Allow to run `launch.py` without workers.

kziemianek:
`launch.py` expects at least on worker definition in config file. This PR allows to run local setup without workers.


Kailai-Wang (approved):
Thanks! 
What's the use case btw? 😂 only launch the parachain?

kziemianek:
> Thanks! What's the use case btw? 😂 only launch the parachain?

Yes

kziemianek:
I decided to use this script to launch only parachain for testing `resuming-worker.ts` and that's how I discovered this bug.

PR #2315, opened 2023-12-07
bring back slot time check

kziemianek:
Brings back slot time check only in case there are more than one worker registered in teerex registry.


tee-worker/enclave-runtime/src/top_pool_execution.rs
```diff
@@ -271,8 +271,17 @@ where
 		)
 		.with_claim_strategy(SlotClaimStrategy::RoundRobin);
 
+	let is_single_worker = match ocall_api.get_trusted_peers_urls() {
```
kziemianek, line 317: We only check if there are more workers registered, which might not really mean they are online and syncing sidechain state but that should be enough for now.

tee-worker/sidechain/consensus/slots/src/lib.rs
```diff
@@ -558,6 +556,7 @@ impl<ParentchainBlock: ParentchainBlockTrait, T: SimpleSlotWorker<ParentchainBlo
 		&mut self,
 		slot_info: SlotInfo<ParentchainBlock>,
 		shards: Vec<Self::ShardIdentifier>,
+		single_worker: bool,
```
Kailai-Wang: Could you make this variable consistent with previously declared one? (i.e. `is_single_worker`)

tee-worker/enclave-runtime/src/top_pool_execution.rs
Kailai-Wang: Could you add this into code comment? :)

Kailai-Wang (approved):
Thanks, I like the unit tests

PR #2316, opened 2023-12-07
Manual tests of removing the second worker during running

kziemianek:
This PR adds back `resuming-worker.ts` to CI pipeline.
The test is moved to separate workspace because it doesn't rely on `client-api` (generation of that packages requires worker instance to be run).


.github/workflows/ci.yml
```diff
@@ -708,6 +706,7 @@ jobs:
           - test_name: lit-ii-batch-test-multiworker
           - test_name: lit-ii-identity-multiworker-test
           - test_name: lit-ii-vc-multiworker-test
+          - test_name: lit-resume-worker
```
kziemianek, line 710: Failure of this job doesn't fail whole action so we will run it for several days to see whether it still causes the problems.

I've performed more than 10 runs locally and it never failed.

tee-worker/ts-tests/worker/package.json
```diff
@@ -0,0 +1,49 @@
+{
+    "name": "stress",
+    "license": "ISC",
+    "type": "module",
+    "scripts": {
+        "format": "pnpm exec prettier --write .",
+        "check-format": "pnpm exec prettier --check .",
+        "run-script": "NODE_TLS_REJECT_UNAUTHORIZED=0 pnpm exec ts-node bin/stress.ts",
```
0xverin: Are these still scripts related to stress testing? Shouldn't be related to the worker?

tee-worker/ts-tests/worker/.eslintrc.json
```diff
@@ -0,0 +1,40 @@
+{
```
0xverin, line 1: Can we move this ESLint file to the `ts-tests` directory? I'm not entirely sure if it will apply globally, but it should work.

tee-worker/ts-tests/worker/package.json
```diff
@@ -0,0 +1,49 @@
...
+        "@types/ws": "^8.5.3",
+        "@typescript-eslint/eslint-plugin": "^5.60.0",
+        "@typescript-eslint/parser": "^5.60.0",
+        "chai": "^4.3.6",
+        "cross-env": "^7.0.3",
+        "dotenv": "^16.3.1",
+        "eslint": "^8.43.0",
+        "ethers": "^5.7.2",
+        "mocha": "^10.1.0",
+        "parachain-api": "../../client-api/parachain-api",
```
0xverin: As you described, the worker doesn't depend on the client-api, we can remove it.

0xverin: Also, any unused dependencies should be removed to reduce installation time. Thanks.

tee-worker/ts-tests/worker/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+        "declaration": true,
+        "strict": true,
+        "skipLibCheck": true,
+        "esModuleInterop": true,
+        "allowSyntheticDefaultImports": true,
+        "resolveJsonModule": true,
+        "baseUrl": "."
+    },
+    "paths": {
+        "parachain-api/*": ["../parachain-api/dist/*", "../parachain-api/build/*"],
```
0xverin: same as above, should be removed.

tee-worker/ts-tests/worker/package.json
kziemianek: Hmm it looks like my local changes weren't included in this commit. I'll take a look

Kailai-Wang:
The `resuming-worker` test is failing though, would you be able to check it?

kziemianek:
> The `resuming-worker` test is failing though, would you be able to check it?

Yes @Kailai-Wang , I plan to fix it. I'll do that once I find a while.

tee-worker/ts-tests/worker/.eslintrc.json
kziemianek: Let's do this in separate PR.

kziemianek:
> The `resuming-worker` test is failing though, would you be able to check it?

@Kailai-Wang the test finally pass.

Kailai-Wang (approved):
Thank you!

PR #2320, opened 2023-12-11
feat: optimise sidechain read

felixfaisal:
This PR removes the need to clone the entire sidechain state everytime we need to read from it, This would have been a bottleneck as the state of the sidechain grew over a period of time. 

Some notes: 
- While obtaining a value using encoded keys is simple, but performing `iter_prefix` was not as straightforward 
- I created a function `iter_prefix` and included it as part of the `Externalities` trait, It is generic and can be used in the future 
- We have the `primary key` or (prefix key), We need to iter for all the suffix keys and also obtain their original value, Since we use `Blake2_128Concat`, The hasher concatenates the original value to the hash and the structure of the key is like this : `[key2 hash][key2 value]` 
- So once we iter the suffix keys, We just need to extract the original key value that was hashed, and the value it holds and this is how we do it here:
```rust
	let suffix_start = key_prefix.len() + HASH_PART_SIZE;
	if encoded_key.len() > suffix_start {
			let suffix = &encoded_key[suffix_start..];
			let decoded_key = K::decode(&mut &suffix[..]).ok();
```

### DIFF
--- a/tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
+++ b/tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
@@ -110,7 +110,7 @@ pub trait SgxExternalitiesTrait {
 	/// Get the next key in state after the given one (excluded) in lexicographic order.
 	fn next_storage_key(&self, key: &[u8]) -> Option<Vec<u8>>;
 
-	/// Iter prefix
+	/// Reads all keys and values under given prefix
 	fn iter_prefix<K: Decode + Debug, V: Decode + Debug>(
 		&self,
 		key_prefix: &[u8],
--- a/tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
+++ b/tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
@@ -27,7 +27,7 @@ use derive_more::{Deref, DerefMut, From, IntoIterator};
 use itp_hashing::Hash;
 use serde::{Deserialize, Serialize};
 use sp_core::{hashing::blake2_256, H256};
-use std::{collections::BTreeMap, vec, vec::Vec};
+use std::{collections::BTreeMap, fmt::Debug, vec, vec::Vec};
 
 pub use scope_limited::{set_and_run_with_externalities, with_externalities};
 
@@ -110,6 +110,12 @@ pub trait SgxExternalitiesTrait {
 	/// Get the next key in state after the given one (excluded) in lexicographic order.
 	fn next_storage_key(&self, key: &[u8]) -> Option<Vec<u8>>;
 
+	/// Iter prefix
+	fn iter_prefix<K: Decode + Debug, V: Decode + Debug>(
+		&self,
+		key_prefix: &[u8],
+	) -> Option<Vec<(K, V)>>;
+
 	/// Clears all values that match the given key prefix.
 	fn clear_prefix(&mut self, key_prefix: &[u8], maybe_limit: Option<u32>) -> u32;
 
@@ -173,6 +179,40 @@ where
 
 	fn prune_state_diff(&mut self) {
 		self.state_diff.clear();
+	}
+
+	fn iter_prefix<K: Decode + Debug, V: Decode + Debug>(
+		&self,
+		key_prefix: &[u8],
+	) -> Option<Vec<(K, V)>> {
+		// The size of the hash part in Blake2_128Concat (16 bytes for blake2_128)
+		const HASH_PART_SIZE: usize = 16;
+
+		let key_values = self
+			.state
+			.range::<[u8], _>((Bound::Included(key_prefix), Bound::Unbounded))
+			.take_while(|(k, _)| k.starts_with(key_prefix))
+			.filter_map(|(encoded_key, encoded_value)| {
+				let suffix_start = key_prefix.len() + HASH_PART_SIZE;
+				if encoded_key.len() > suffix_start {
+					let suffix = &encoded_key[suffix_start..];
+					let decoded_key = K::decode(&mut &suffix[..]).ok();
+					let decoded_value = V::decode(&mut &encoded_value[..]).ok();
+					match (decoded_key, decoded_value) {
+						(Some(key), Some(value)) => Some((key, value)),
+						_ => None,
+					}
+				} else {
+					None
+				}
+			})
+			.collect::<Vec<_>>();
+
+		if key_values.is_empty() {
+			None
+		} else {
+			Some(key_values)
+		}
 	}
 
 	fn clear_prefix(&mut self, key_prefix: &[u8], _maybe_limit: Option<u32>) -> u32 {
--- a/tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
+++ b/tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
@@ -181,6 +181,7 @@ where
 		self.state_diff.clear();
 	}
 
+	// Note: This implementation only works for keys encoded with Blake2_128Concat
 	fn iter_prefix<K: Decode + Debug, V: Decode + Debug>(
 		&self,
 		key_prefix: &[u8],
@@ -303,6 +304,7 @@ impl Encode for EncodeOpaqueValue {
 pub mod tests {
 
 	use super::*;
+	use itp_storage::{storage_double_map_key, storage_map_key, StorageHasher};
 
 	#[test]
 	fn mutating_externalities_through_environmental_variable_works() {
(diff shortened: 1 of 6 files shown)


tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -137,86 +140,102 @@ where
...
+					state.iter_prefix::<Identity, IdentityContext<Runtime>>(&prefix_key).unwrap();
+				id_graph.sort_by(|a, b| {
+					let order = Ord::cmp(&a.1.link_block, &b.1.link_block);
+					if order == Ordering::Equal {
+						// Compare identities by their did formated string
+						Ord::cmp(&a.0.to_did().ok(), &b.0.to_did().ok())
+					} else {
+						order
+					}
+				});
```
Kailai-Wang: Hmm, what prevents us from using `IMT::get_id_graph` here?

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -137,86 +140,102 @@ where
...
+					.collect();
+
+				identities
+			})
+			.map_err(|e| format!("Failed to fetch sidechain data due to: {:?}", e))?;
+
+		let signer = match signer.to_account_id() {
+			Some(s) => s,
+			None => return Err("Invalid signer account, failed to convert".to_string()),
+		};
```
Kailai-Wang: Is it possible to use `.ok_or` here for better readability? Other places too

tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
```diff
@@ -110,6 +110,12 @@ pub trait SgxExternalitiesTrait {
 	/// Get the next key in state after the given one (excluded) in lexicographic order.
 	fn next_storage_key(&self, key: &[u8]) -> Option<Vec<u8>>;
 
+	/// Iter prefix
+	fn iter_prefix<K: Decode + Debug, V: Decode + Debug>(
```
Kailai-Wang, line 114: This is low-level change, I'd highly recommend having a unit test for it if you want to add this fn

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: To use `IMT::get_id_graph`, we can run it only under the externalities i.e. `execute_with` and that function requires mutable access to the state, So previously, I cloned the state and made it mutable for it to be used with `execute_with`. I couldn't find any other way to use `IMT::get_id_graph`.

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: I initially tried using `unwrap_or` but I had some problems, So decided to use `match`, I can try it again.

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
Kailai-Wang: I see, thanks!
Would you like to pack the common sorting code into a fn? AFAIK it's going to change soon (again)

felixfaisal: Updated with common code ✅

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: `.ok_or` works like a charm, updated ✅

tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
felixfaisal: Added a unit test covering this function ✅

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -119,104 +121,108 @@ where
...
-				})
-				.collect();
-
-			let signer = match signer.to_account_id() {
+		let key = maybe_key.ok_or_else(|| "User shielding key not provided".to_string())?;
+		let identities: Vec<IdentityNetworkTuple> = context
+			.state_handler
+			.execute_on_current(&shard, |state, _| {
+				let prefix_key = storage_map_key(
+					"IdentityManagement",
```
kziemianek, line 144: Can you remove this hardcoded values and rely on generated ones ?

```
					pallet_identity_management_tee::pallet::IDGraphs::<Runtime>::module_prefix(),
					pallet_identity_management_tee::pallet::IDGraphs::<Runtime>::storage_prefix(),
```

kziemianek: Hmm maybe this can be further simplified to this:

```
let prefix_key = pallet_identity_management_tee::pallet::IDGraphs::<Runtime>::storage_double_map_final_key1(&who);
```

Kailai-Wang (approved):
OK from my side, but maybe you want to address @kziemianek 's comments :)

tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
```diff
@@ -175,6 +181,41 @@ where
 		self.state_diff.clear();
 	}
 
+	// Note: This implementation only works for keys encoded with Blake2_128Concat
+	fn iter_prefix<K: Decode + Debug, V: Decode + Debug>(
+		&self,
+		key_prefix: &[u8],
+	) -> Option<Vec<(K, V)>> {
```
kziemianek, line 188: I find it simpler to return just empty vector instead of `Option` in such cases.

tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
```diff
@@ -110,6 +110,12 @@ pub trait SgxExternalitiesTrait {
 	/// Get the next key in state after the given one (excluded) in lexicographic order.
 	fn next_storage_key(&self, key: &[u8]) -> Option<Vec<u8>>;
 
+	/// Iter prefix
```
kziemianek: Consider changing doc to `Reads all keys and values under given prefix`

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
felixfaisal: I am slightly confused here, 
`module_prefix()` and `storage_prefix()` are generated for `IDGraphs` ? I don't find these functions anywhere could you share me some documentation I can look up at? And can these functions be called outside of the externalities environment? 

Also I don't understand the function `storage_double_map_final_key`, Do you want to implement this function for the storage item `IDGraphs`? 🤔

tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
felixfaisal: I prefer `Option` here, The other functions also use `Option` making it consistent and returning an empty vector is not very ergonomic.

tee-worker/core-primitives/substrate-sgx/externalities/src/lib.rs
felixfaisal: Done ✅

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
kziemianek: This function is available on trait `frame_support::storage::generator::StorageDoubleMap` . The trait impl is added by `#[pallet::storage]` macro expansion.


The trait itself is hidden from docs so you need to look at frame-support source code to find out.

```
#[doc(hidden)]
pub mod generator;
```

It's documented as follows:
```
	/// Generate the first part of the key used in top storage.
	fn storage_double_map_final_key1<KArg1>(k1: KArg1) -> Vec<u8>
```

kziemianek: You should be able to call this function regardless of context. You don't need to add any impl.

felixfaisal: It compiles, but I get the following error when I perform e2e test: 
```
thread '<unnamed>' panicked at 'storage cannot be called outside of an Externalities-provided environment.', /home/faisal/litentry-parachain/tee-worker/core-primitives/substrate-sgx/sp-io/src/lib.rs:131:9
```
As I mentioned I don't think we can call this outside of externalities

kziemianek: You are right, I though it's just a static string - don't know yet why it fails 😿

kziemianek (approved):

PR #2321, opened 2023-12-11
Bring consistency to di responses

Kailai-Wang:
### Context

As topic.

Now for the requests that mutate the IDGraph, the response result will contain at least:
  - `mutated_id_graph`: the mutated IDGraph, which is the subset of the whole IDGraph
  - `id_graph_hash`: the new IDGraph hash after the mutation

Currently it applies to the following DI requests:
  - `link_identity`
  - `activate_identity`
  - `deactivate_identity`
  - `set_identity_networks`


tee-worker/app-libs/stf/src/trusted_call_result.rs
```diff
@@ -51,9 +60,37 @@ impl StfExecutionResult for TrustedCallResult {
...
+/// - `id_graph_hash`: the new IDGraph hash after the mutation
+///
+/// Currently it applies to the following DI requests:
+/// - `link_identity`
+/// - `activate_identity`
+/// - `deactivate_identity`
+/// - `set_identity_networks`
 #[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
 pub struct LinkIdentityResult {
-	pub id_graph: AesOutput,
```
jonalvarezz, line 56: I added the breaking change label because of this.

jonalvarezz (approved):
lgtm

BillyWooo (approved):

tee-worker/app-libs/stf/src/trusted_call_result.rs
Kailai-Wang: Yes, thanks!

PR #2324, opened 2023-12-12
P-314 add lit holding amount

outofxxx:
As title.

Interface & example: [here](https://www.notion.so/web3builders/NewYear-LIT-holding-amount-9104fb6188a74828bde37414e317963a).


BillyWooo (approved):
LGTM.
As discussed, we may need some general refactor about VC generation part.

PR #2331, opened 2023-12-15
Add support for BTC account linking

Kailai-Wang:
### Context

As topic.

The raw pubkey (`[u8; 33]`) is used as the bitcoin identity handle, thus we require that the wallet API provides such interface (to retrieve an account's pubkey).

Otherwise parachain needs to support [all 5 address formats](https://github.com/rust-bitcoin/rust-bitcoin/blob/8afa379f3946f9b7605a278b59cd103dbd5020b1/bitcoin/src/address/mod.rs#L62-L73) instead of a unified pubkey. This not only brings extra work, it also makes it inconsistent with other web3 networks (substrate/EVM) where the raw bytes are used as Identity Handle.

The message signing and verification is yet to be tested with the real wallet extension. It **might not work** as the wallet could massage the message freely. In this implementation, a standard ECDSA/SECP256k1 keypair with a recoverable signature is used.

This impl **should not** break the old `Identity` or `IDGraph` structures, or the sidechain storages - but we'll need to test it out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -620,6 +620,7 @@ jobs:
           - test_name: lit-ii-identity-test
           - test_name: lit-di-substrate-identity-test
           - test_name: lit-di-evm-identity-test
+          - test_name: lit-di-bitcoin-identity-test
           - test_name: lit-di-vc-test
           - test_name: lit-parentchain-nonce
           - test_name: lit-ii-batch-test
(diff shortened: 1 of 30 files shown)


BillyWooo (commented):
I only reviewed RUST code. LGTM

BillyWooo (approved):
Let's move it in. and test.

PR #2336, opened 2023-12-16
Use bitcoin crate for signature verification

Kailai-Wang:
### Context

The bitcoin message is a bit different from the standard ecdsa signature (e.g. IRS instead of RSV format, magic hash of the signing message etc).

The PR uses `bitcoin` rs crate for all the actions so that it complies with the bitcoin message signing and wallet actions. The ts-tests are adjusted too.

Note:
- the bitcoin signing message is expected to a string, in our case it's a blake2_256 hash, so we sign the hex-encoded version of it (without `0x` prefix)
- the signature is by default base64-encoded, we need to convert it to byte arrays before passing it to the enclave.

Please refer to the test code.


BillyWooo (approved):
Mainly review RUST code. LGTM. 
Let's roll it out and test.

tee-worker/ts-tests/integration-tests/common/di-utils.ts
```diff
@@ -94,9 +94,21 @@ export const createSignedTrustedCall = async (
...
+
+    let signature;
+
+    // for bitcoin signature, we expect a hex-encoded `string` without `0x` prefix
+    // TODO: any better idiomatic way?
+    if (signer.type() === 'bitcoin') {
+        const payloadStr = u8aToHex(payload).substring(2);
+        signature = parachainApi.createType('LitentryMultiSignature', {
+            [signer.type()]: u8aToHex(await signer.sign(payloadStr)),
+        });
```
jonalvarezz, line 106: so, if the u8a is signed, it will fail?

Kailai-Wang: Yes, it's incompatible with the interface that unisat-wallet (and the underlying bitcore-lib) uses. It expects a `string` as signing message, unlike other signing mechanism where typically a u8a is used

Kailai-Wang: And our signing message is a `[u8; 32]`, so we need to convert it to string before signing

PR #2338, opened 2023-12-18
Add missing bitcoin match arm

Kailai-Wang:
### Context

As topic, found this bug when implementing P-363

Question to `a20`: why did it even work without `get_supported_web3networks` implementation for A20? (so no identity should participate in a20 calculation actually

### DIFF
--- a/tee-worker/litentry/core/assertion-build/src/lib.rs
+++ b/tee-worker/litentry/core/assertion-build/src/lib.rs
@@ -97,6 +97,11 @@ pub fn transpose_identity(
 					addresses.push((address, n));
 					networks_set.insert(n);
 				},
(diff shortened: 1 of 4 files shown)


tee-worker/litentry/core/assertion-build/src/lib.rs
```diff
@@ -97,6 +97,11 @@ pub fn transpose_identity(
 					addresses.push((address, n));
 					networks_set.insert(n);
 				},
+				Identity::Bitcoin(address) => {
```
BillyWooo, line 100: The `address` here is still the pubkey, not the real BTC address, right?

BillyWooo (approved):

PR #2339, opened 2023-12-18
Parachain staking: reduce awaiting time

Traf333:
### Context

we want to properly test whole flow of parachain staking functionality with smart contract, but some functions require too much time to wait, which prevents us from testing the whole flow with integration tests. This PR tends to extend parachain staking presetup with `prod_or_fast` macro to allow us to test the contract functions: 

- scheduleDelegatorBondLess
- scheduleRevokeDelegation
- executeDelegationRequest


pallets/parachain-staking/src/delegation_requests.rs
Kailai-Wang, line 1: Why these changes? Are they from the formatting?

runtime/rococo/src/lib.rs
```diff
@@ -856,19 +856,19 @@ impl pallet_parachain_staking::Config for Runtime {
 	type Currency = Balances;
 	type MonetaryGovernanceOrigin = EnsureRootOrAllCouncil;
 	/// Minimum round length is 2 minutes (10 * 12 second block times)
-	type MinBlocksPerRound = ConstU32<{ 2 * MINUTES }>;
+	type MinBlocksPerRound = ConstU32<{ prod_or_fast!(2 * MINUTES, 2) }>;
```
Kailai-Wang, line 858: Can it be too fast to test?
Not saying it's NOK, just a reminder :D

Makefile
```diff
@@ -32,7 +32,7 @@ build-all:
 
 .PHONY: build-node ## Build release node with `tee-dev` feature
 build-node:
-	cargo build --locked -p $(call pkgid, $(NODE_BIN)) --release --features=tee-dev
+	cargo build --locked -p $(call pkgid, $(NODE_BIN)) --release --features=tee-dev,fast-runtime
```
Kailai-Wang, line 35: This will have an impact on local build commands only, I don't think CI is affected where docker build is applied.

Traf333: This is really valuable comment thanks. Shall we build/push/pull specific docker build for parachain ts tests or from now on we will build fast-run in general first CI?

pallets/parachain-staking/src/delegation_requests.rs
Traf333: My bad, it's a leftover after debugging stuff. I'll revert it within next commit

runtime/rococo/src/lib.rs
Traf333: To be honest I'm not sure which amount of time blocks is sufficient to wait. I made it to wait 2 blocks instead of 10 as you recommended to make it faster. 😊
What time you do you think should be reasonable for this case?

Kailai-Wang: I can't tell before seeing the real tests :D If you don't find a problem with 2 blocks, just go for it

Makefile
Kailai-Wang: We are safe to apply this flag to general CI.

The release docker image is ensured to be built with different flags.

runtime/rococo/src/lib.rs
```diff
@@ -855,19 +855,19 @@ impl pallet_parachain_staking::Config for Runtime {
...
 	/// Blocks per round
-	type DefaultBlocksPerRound = ConstU32<{ 2 * MINUTES }>;
+	type DefaultBlocksPerRound = ConstU32<{ prod_or_fast!(2 * MINUTES, 2) }>;
```
felixfaisal, line 860: I am just curious here, Does changing the blocks per round affect the rewards being paid out?

Kailai-Wang: Potentially yes if `BlocksPerRound` setting is overriden on chain - but this flag will not be applied in prod anyway

Kailai-Wang (approved):
Thanks, looks fine from my side

PR #2345, opened 2023-12-19
Use public getter for IDGraphHash

Kailai-Wang:
### Context

See the linked issue.

This PR:
- dismisses the on-parachain IDGraph storage in favor of a public getter
- adjusts ts-test
- updates some crates to supersede dependabot PRs

It's a breaking change for the client.

PR #2346, opened 2023-12-20
Hotfix Achainable response

outofxxx:
For accounts without transaction records, when querying, Achainable returns `Address has no [token] balance`, causing ERC20 related VC to return a ParsingError error.

Now, for such accounts, return 0 balance directly for processing.


BillyWooo:
Looks like these two test failures come from [here](https://github.com/litentry/litentry-parachain/pull/2345).
As this hot-fix is needed in urgent, we will merge this PR now. And fix the test failures separately.

PR #2347, opened 2023-12-21
IDGraph should be created in more actions than linking

Kailai-Wang:
### Context

As topic.

Requesting VC with an empty IDGraph is tackled in #2348 already

Additionally, I don't find a reason why deactivating the prime identity should be disallowed - deactivating is nothing more than excluding it from VC building, it can still be used to manage the IDGraph. So I removed this restriction. If you see problems, please let me know.


Kailai-Wang:
It seems we can't trigger IDGraph creation upon VC request, as it would introduce sidechain block production.
We'll simply include the prime identity to the eligible identities manually @felixfaisal

felixfaisal:
> It seems we can't trigger IDGraph creation upon VC request, as it would introduce sidechain block production.
We'll simply include the prime identity to the eligible identities manually @felixfaisal

This would depend on the type of VC right? Like for example if someone trying to use Substrate Identity to create a VC for Ethereum, it would just fail right because this prime identity can only support probably `Polkadot` or `Litentry` network. However, I was wondering if we could have two kinds of VC requests if we really want to skip the `link-identity` step, 
1. This is current VC request, A user requests for a VC, we iter through the IDGraph, and build the assertion 
2. This is another type of VC Request, Wherein a user hasn't linked/created an IDGraph, and just wants to create a VC Request using either prime or another identity, wherein make sure the signatures and everything are correct. 

Again. I think I am talking from the context of isolated vc issuance, However the 2nd type will not be a good idea because Alice could create a VC using an X identity, and then so can Bob. It may work for only the `prime identity`, however I think we need to put some more thought on this ,as I am not sure 😅 

Also, Another question, the client at all times has the `IDGraph`(or access to it) does it make sense to we even allow someone to request for a VC when the `IDGraph` is empty? wdyt?

Kailai-Wang:
> Also, Another question, the client at all times has the IDGraph(or access to it) does it make sense to we even allow someone to request for a VC when the IDGraph is empty? wdyt?

@felixfaisal this is what I talked about, we should allow it.
Imagine you log into IDH with your evm account, but you can't request a vc until you link an identity - this doesn't make sense 😂

felixfaisal:
> @felixfaisal this is what I talked about, we should allow it.
Imagine you log into IDH with your evm account, but you can't request a vc until you link an identity - this doesn't make sense 😂

Ah, I got confused 😅 . I think I get it now.

felixfaisal (approved):
LGTM

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -276,6 +259,31 @@ pub mod pallet {
 	}
 
 	impl<T: Config> Pallet<T> {
+		// try to create an IDGraph if there's none - `who` will be the prime identity
+		// please note the web3networks for the prime identity will be all avaiable networks
+		fn maybe_create_id_graph(who: &Identity) -> Result<(), DispatchError> {
```
kziemianek, line 264: Will this newly created identity show up in DI response ?

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -184,8 +163,8 @@ pub mod pallet {
 			identity: Identity,
 		) -> DispatchResult {
 			T::ManageOrigin::ensure_origin(origin)?;
+			Self::maybe_create_id_graph(&who)?;
```
kziemianek, line 166: how is it possible that we reached this point without id graph being created before?

tee-worker/litentry/pallets/identity-management/src/lib.rs
Kailai-Wang: It will.

Imagine you link an identity A to your prime identity, then both your prime identity and A will show up in the `mutated_id_graph` because that's the exactly the IDGraph change (it was empty before)

tee-worker/litentry/pallets/identity-management/src/lib.rs
Kailai-Wang: > how is it possible that we reached this point without id graph being created before?

You could deactivate the prime identity.
As written in the PR description, I didn't find a reason not to allow it.

tee-worker/litentry/pallets/identity-management/src/lib.rs
kziemianek: I see so this means id graph is created but prime identity is deactivated - thanks.

tee-worker/litentry/pallets/identity-management/src/lib.rs
kziemianek: I see so this means id graph is created but prime identity is deactivated - thanks.

kziemianek: Hmm but deactivating identity is nothing more than changing its status. 
It will still be present in `IDGraphs` map, am I right  ?

Kailai-Wang: Exactly, but "status" only exists in the context of an IDGraph. So if you deactivates the (prime) identity without linking any other identities, you kind of have to create the IDGraph (that has only 1 element which is the prime identity) first

tee-worker/litentry/pallets/identity-management/src/lib.rs
kziemianek: But in case of `set_identity_networks` trusted call there is only one identity pushed to `mutated_id_graph` while this `set_identity_networks` call may alter two identities sometimes.

```
				let mut mutated_id_graph = IDGraph::<Runtime>::default();
				if let Some(identity_context) = IMT::id_graphs(&who, &identity) {
					if let Some(key) = maybe_key {
						mutated_id_graph.push((identity, identity_context));
						return Ok(TrustedCallResult::SetIdentityNetworks(
							SetIdentityNetworksResult {
								mutated_id_graph: aes_encrypt_default(
									&key,
									&mutated_id_graph.encode(),
								),
								id_graph_hash,
							},
						))
					}
				} else {
					// if should not happen, so we just log the error here
					error!("failed to get identity_context for {:?}, {:?}", &who, &identity);
				}
```

Kailai-Wang: > this set_identity_networks call may alter two identities sometimes

How could this happen? 😂

kziemianek: Create prime identity [here]( https://github.com/litentry/litentry-parachain/pull/2347/files/c45491a137d12cf9894127d18b4d81eaff74f606#diff-0673e79aa46adac84174d588afcb0816746db60d4b560b4cdd01097951b322b0R209)
Mutate identity [here](https://github.com/litentry/litentry-parachain/pull/2347/files/c45491a137d12cf9894127d18b4d81eaff74f606#diff-0673e79aa46adac84174d588afcb0816746db60d4b560b4cdd01097951b322b0R212)

? 😕

Kailai-Wang: If you are setting the networks of the prime identity for the first time - the IDGraph must have not existed;
if you are setting the networks of a non-prime identity, then either the IDGraph must have existed or it fails because identity doesn't exist (for an empty IDGraph)

So in either case you can't alter two identities?

Kailai-Wang: But I think what's missing is for the second case, should the mutated_id_graph contain the prime identity? The extrinsic in tee-pallet is unfortunately not atomic

kziemianek: Adding `edit_block` to `IdentityContext` would make it possible to detect all changes inside idgraph at `trusted_call` level and have common and accurate method for producing idgraph diff.

kziemianek: > If you are setting the networks of the prime identity for the first time - the IDGraph must have not existed; if you are setting the networks of a non-prime identity, then either the IDGraph must have existed or it fails because identity doesn't exist (for an empty IDGraph)
> 
> So in either case you can't alter two identities?

Ok I see - it's not clear by looking just at this code.

Kailai-Wang: > Adding edit_block to IdentityContext

Do you mean look for any entry whose `edit_block` equals to the current block number?
Changing IdentityContext struct might need a storage migration (or maybe not if you append it)

Can we do a simple diff? - it's keyed by the prime identity so the computation shouldn't be that costly

kziemianek: > Do you mean look for any entry whose edit_block equals to the current block number?

Exactly 🙃 

> Can we do a simple diff? - it's keyed by the prime identity so the computation shouldn't be that costly

Should do the job also without extra storage space.

Kailai-Wang: Implemented - re-request the review now @kziemianek :)

kziemianek: Thanks 👍🏼

kziemianek (approved):

PR #2348, opened 2023-12-21
Filter out web3 identities with empty networks

Kailai-Wang:
### Context

As topic - they should not participate in the assertion building as it doesn't make sense.

Web2 identities are not affected.

This PR addresses the issue in https://github.com/litentry/litentry-parachain/pull/2347#issuecomment-1866277442, otherwise CI won't pass. It allows Vc to be requested without any IDGraph mutation.

It also avoids using `_` in matching arms to be more explicit about member fields.

~~This should go to `staging` too~~ => maybe we are not in such hurry - @BillyWooo can make a decision


tee-worker/app-libs/stf/src/trusted_call_litentry.rs
```diff
@@ -172,16 +172,23 @@ impl TrustedCallSigned {
...
+					//   care about web2 identities, this step will empty `IdentityContext.web3networks`
+					// - it helps to reduce the request size a bit
+					networks.retain(|n| assertion_networks.contains(n));
+					if networks.is_empty() && item.0.is_web3() {
+						None
+					} else {
+						Some((item.0, networks))
+					}
+				} else {
+					None
```
felixfaisal: I have a question here, I think I had the same question in my mind when I was implementing the isolated vc issuance, Does it make sense to create an `AssertionBuildRequest` when the `identities` vector is an empty vector? We can reject the request right here instead of letting it to go through with the `StfRequest`, wdyt?

Kailai-Wang: Thanks @felixfaisal - that makes sense, if `identities` is empty, there's no point moving on.

I'll add that

Kailai-Wang: Added :)

kziemianek: It would be good to have a test case covering this error.

kziemianek (approved):

felixfaisal (approved):
LGTM

Kailai-Wang:
I extracted the common code and added a unit test, thus the re-review

felixfaisal (approved):
Thanks Kai! Always good to have a unit test for coverage, I don't see any changes required here. LGTM

Kailai-Wang:
Sorry @felixfaisal I need to re-request the review as I added the code to address https://github.com/litentry/litentry-parachain/pull/2347#issuecomment-1866277442 - otherwise CI wouldn't pass

BillyWooo:
I think it's fine to only stay with `dev` branch. `staging` is already too busy for the moment. Let's decide after Christmas/New Year vacation season.

Kailai-Wang:
Ready to be reviewed @litentry/parachain :)

primitives/core/src/assertion.rs
```diff
@@ -298,7 +300,6 @@ impl Assertion {
 			// BRC20 Holder
 			Self::BRC20AmountHolder => vec![Web3Network::BitcoinP2tr],
```
BillyWooo, line 301: how to you think to add back the other four btc types here, and do a combination of filtering of `input network type` against `vc supported type` ?

BillyWooo (approved):
LGTM. 
It's good timing now to merge to dev. Then we make a new release ASAP for client-sdk and idhub development.

kziemianek (approved):

PR #2349, opened 2023-12-21
Extend bitcoin networks to all address types

Kailai-Wang:
### Context

As topic.

Theoretically the client only needs to rename `Bitcoin` to `BitcoinP2tr` and that's all

### DIFF
--- a/primitives/core/src/network.rs
+++ b/primitives/core/src/network.rs
@@ -70,9 +70,17 @@ pub enum Web3Network {
 	#[codec(index = 8)]
 	Bsc,
 
-	// btc
+	// btc, see https://github.com/rust-bitcoin/rust-bitcoin/blob/9ea3e29d61569479b7b4618c8ae1992612f3d01a/bitcoin/src/address/mod.rs#L64-L75
 	#[codec(index = 9)]
-	Bitcoin,
+	BitcoinP2tr,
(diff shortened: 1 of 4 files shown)


jonalvarezz (approved):
looking good to me

primitives/core/src/network.rs
```diff
@@ -70,9 +70,17 @@ pub enum Web3Network {
 	#[codec(index = 8)]
 	Bsc,
 
-	// btc
+	// btc, see https://github.com/rust-bitcoin/rust-bitcoin/blob/9ea3e29d61569479b7b4618c8ae1992612f3d01a/bitcoin/src/address/mod.rs#L64-L75
 	#[codec(index = 9)]
-	Bitcoin,
+	BitcoinP2tr,
```
kziemianek, line 75: Have you considered using nested enum? I think there is only one bitcoin network 🙃

Kailai-Wang: there's only one mainnet but these are used to distinguish between different address types, so theoretically one pubkey can be used for any one of the address types. From this sense it's similar to EVM where one universal pubkey for all networks. Wdyt

Kailai-Wang: I'll add a comment later

PR #2350, opened 2023-12-22
Expand underscore expression in Assertion Type

outofxxx:
Expand underscore expression. 
If a new Assertion is forgotten to be added, there will be a build error instead of the default `vec! []`


Kailai-Wang (approved):
Makes sense, thanks!

wli-pro (approved):
Great thanks

PR #2351, opened 2023-12-22
Adjust the assertion codec index

BillyWooo:
very minor change as title. 
But I think it's worth it.

### DIFF
--- a/primitives/core/src/assertion.rs
+++ b/primitives/core/src/assertion.rs
@@ -249,10 +249,10 @@ pub enum Assertion {
 	#[codec(index = 20)]
 	LITStaking,
 
-	#[codec(index = 22)]
+	#[codec(index = 21)]
 	EVMAmountHolding(EVMTokenType),  // (evm_token_type)
 
-	#[codec(index = 23)]
+	#[codec(index = 22)]
 	BRC20AmountHolder,
 }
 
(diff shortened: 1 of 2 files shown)


Kailai-Wang (approved):
It's fine as long as we haven't started to use it yet and nothing occupied 21 before

cc @jonalvarezz too

Kailai-Wang:
Btw it's still safer to adjust `vc/definitions.ts` too, although I don't know if @Verin1005 is using it or not (if not, why not delete them?)

BillyWooo:
> Btw it's still safer to adjust `vc/definitions.ts` too, although I don't know if @Verin1005 is using it or not (if not, why not delete them?)

This order is adjusted with 'vc/definitions.ts` already.

Kailai-Wang:
> > Btw it's still safer to adjust `vc/definitions.ts` too, although I don't know if @Verin1005 is using it or not (if not, why not delete them?)
> 
> This order is adjusted with 'vc/definitions.ts` already.

This is even worse :D as it means we made some mistakes ...

jonalvarezz (approved):
IDH uses indirect calls, so the impact is low and constraint to type changes only.

Kailai-Wang:
> IDH uses indirect calls, so the impact is low and constraint to type changes only.

IDH uses direct vc request (which requires `assertion` parameter), doesn't it

This PR is on dev branch anyway so it's targeted at DI definitely

I'll still mark it as "breaking" and merge it

PR #2352, opened 2023-12-22
Allow empty web3networks when linking identities

Kailai-Wang:
### Context

This is a workaround to reduce the encoded call size.

We allow the client to pass in empty `web3networks` when linking identities and restore it to full network types internally.
Please note `set_user_shielding_key` is not affected - the client still has to provide valid network types.

### DIFF
--- a/tee-worker/litentry/core/data-providers/src/geniidata.rs
+++ b/tee-worker/litentry/core/data-providers/src/geniidata.rs
@@ -20,15 +20,12 @@ use crate::sgx_reexport_prelude::*;
 #[cfg(all(not(feature = "std"), feature = "sgx"))]
 extern crate sgx_tstd as std;
 
-use crate::{
-	build_client,  Error as DataProviderError,
-	GLOBAL_DATA_PROVIDER_CONFIG,
-};
+use crate::{build_client, Error as DataProviderError, GLOBAL_DATA_PROVIDER_CONFIG};
 use http::header::ACCEPT;
 use http_req::response::Headers;
 use itc_rest_client::{
 	error::Error as RestClientError,
-	http_client::{HttpClient, DefaultSend},
+	http_client::{DefaultSend, HttpClient},
 	rest_client::RestClient,
 	RestGet, RestPath,
 };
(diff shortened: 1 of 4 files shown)


tee-worker/litentry/core/data-providers/src/geniidata.rs
```diff
@@ -82,9 +79,15 @@ impl GeniidataClient {
 	pub fn new() -> Self {
 		let mut headers = Headers::new();
 		headers.insert(ACCEPT.as_str(), "application/json");
-		headers.insert("api-key", GLOBAL_DATA_PROVIDER_CONFIG.read().unwrap().geniidata_api_key.as_str());
+		headers.insert(
+			"api-key",
+			GLOBAL_DATA_PROVIDER_CONFIG.read().unwrap().geniidata_api_key.as_str(),
+		);
```
jonalvarezz, line 85: [off-topic] do you have an automatic syntax formatter running? how do I set up mine? 😄

tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -170,13 +171,28 @@ pub mod pallet {
...
+			let mut adjusted_web3networks = web3networks;
+
+			// To reduce the encoded call size, the client has a workaround on `staging` branch to
+			// pass in empty `web3networks`, we'll need to restore it to full network types
+			if identity.is_web3() && adjusted_web3networks.is_empty() {
+				adjusted_web3networks = match identity {
+					Identity::Substrate(..) => all_substrate_web3networks(),
+					Identity::Evm(..) => all_evm_web3networks(),
+					Identity::Bitcoin(..) => all_bitcoin_web3networks(),
+					_ => Vec::new(),
```
jonalvarezz, line 183: shouldn't it throw here and panic?

tee-worker/litentry/pallets/identity-management/src/tests.rs
```diff
@@ -121,6 +121,29 @@ fn link_identity_with_wrong_network_fails() {
...
+			RuntimeOrigin::signed(ALICE),
+			who.clone(),
+			alice_evm_identity(),
+			web3networks,
+		));
+		assert_eq!(
+			IMT::id_graphs(who.clone(), alice_evm_identity()).unwrap(),
+			IdentityContext {
+				link_block: 1,
+				web3networks: vec![Web3Network::Ethereum, Web3Network::Bsc].try_into().unwrap(),
```
jonalvarezz: can we do something like this so it doesn't break on new evm networks? 
```suggestion
				web3networks: all_evm_web3networks().try_into().unwrap(),
```

jonalvarezz (approved):
Sweet! thank you, Kai!

tee-worker/litentry/core/data-providers/src/geniidata.rs
Kailai-Wang: There's github hook `.githooks/pre-commit`

tee-worker/litentry/pallets/identity-management/src/lib.rs
Kailai-Wang: Normally we don't panic inside pallet, it's better to return an error here - but since it's on staging, I didn't even bother adding another error type (which requires refreshment of sidechain metadata) as I know it's unreachable.

tee-worker/litentry/pallets/identity-management/src/tests.rs
Kailai-Wang: Actually I wrote this intentionally since we used `all_evm_web3networks` in the impl, let me change it back

PR #2363, opened 2023-12-26
Fix truncated chain spec in parachain-launch

### LINKED ISSUE #2099: Fix truncated chain spec in parachain-launch
### Context

Follow-up of #2081

#2098 is still a workaround, see the PR description.

We'll still need a way to solve it cleanly

---

Traf333:
this PR tends to remove workaround (#2098)  in order to keep things simple and consistent without supporting forked packages

Fixes: #2099


Traf333:
In addition, I've triggered whole system build to be sure the CI will work in case we touch some rust code
https://github.com/litentry/litentry-parachain/actions/runs/7330479550

Kailai-Wang:
Could you explain a bit why this PR fixes the issue? I don't get it 😂

Traf333:
> Could you explain a bit why this PR fixes the issue? I don't get it 😂

As from description I understood that we got some tests failing in CI and we introduced workaround https://github.com/litentry/litentry-parachain/pull/2098 

Maybe I understood the issue in a wrong way 🤔  

if it's not the case, can you help me to understand what's the expected result? @Kailai-Wang

Kailai-Wang:
> As from description I understood that we got some tests failing in CI and we introduced workaround #2098 

This is correct, but simply reverting this PR won't solve the problem :(

The issue is aimed to find the root cause and really "fix" it instead of working around it

Traf333:
This PR includes not only reverting your PR, but
- updated some packages
- preventing build for the `@open-web3/parachain-launch` which might cause the issue

and I was thinking if the tests are green - it means the issue is solved 😄 
since I reverted the workaround, probably now we should be able to see the issue, right? do you know how can I reproduce it?

Kailai-Wang:
> do you know how can I reproduce it?

This is the hardest part, the issue was a race condition that (I believe) was related to the I/O buffering. While it worked most of the time, it did fail sometimes.

You could read the links (to upstream PRs/issues, which are linked again to other issues) in the original PR thoroughly.

We can also revert it (but then maybe you want a higher version of the package) and see how it goes, but that's not a legit "fix" :)

0xverin:
>do you know how can I reproduce it?

As we briefly discussed in the meeting, this part is challenging to reproduce, which is also why I mentioned "not sure." :p

Traf333:
> but then maybe you want a higher version of the package

Not sure there are any major changes which might be related to our case, even packages update are only touched dev packages
https://github.com/open-web3-stack/parachain-launch/compare/v1.4.2...v2.0.0

Thanks for the input. 🙏  I'll re-read related issue and PR carefully and will continue to debug

Kailai-Wang:
Thanks!

Feel free to jump on any other tasks that have a higher priority tho :)

Traf333:
closing it for now as it stale and may pick up later

PR #2364, opened 2023-12-27
Add additional fields to VC

Kailai-Wang:
### Context

This PR:
- adds three additional fields to VC:
   - `parachainBlockNumber`
   - `sidechainBlockNumber` - these two numbers represent the chain height on which the chain state is used to calculate the VC
   - `assertionText` - the assertion text generated from the default `std::fmt::Debug` for a specific assertion. It's subject to change, e.g. we might want to use `Display` impl for customised printing

How does it look?
Do we still need `description` and `type` fields? 

Example a8 VC:
```
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/security/suites/ed25519-2020/v1"
  ],
  "id": "0xb50f79f4f3687482f399828d180371552b985b45de6af9d9811e13d5a3febd8d",
  "type": [
    "VerifiableCredential"
  ],
  "credentialSubject": {
    "id": "did:litentry:substrate:0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48",
    "description": "Gets the range of number of transactions a user has made for a specific token on all supported networks (invalid transactions are also counted)",
    "type": "EVM/Substrate Transaction Count",
    "assertionText": "A8(BoundedVec([Litentry, Litmus], 128))",
    "assertions": [
      {
        "and": [
          {
            "src": "",
            "op": ">=",
            "dst": "50"
          },
          {
            "src": "",
            "op": "<",
            "dst": "100"
          },
          {
            "or": [
              {
                "src": "",
                "op": "==",
                "dst": "Litmus"
              },
              {
                "src": "",
                "op": "==",
                "dst": "Litentry"
              }
            ]
          }
        ]
      }
    ],
    "values": [
      true
    ],
    "endpoint": "http://localhost:9933"
  },
  "issuer": {
    "id": "did:litentry:substrate:0xf7dc5214709ff407024b6a9d4987d900fc73e6d8d75a3b9493b8b1222e039b18",
    "name": "Litentry TEE Worker",
    "mrenclave": "CEacnAF5CeuBTFawk5UtSAsC1p7ZxTpsUw5s34Q8tCuY"
  },
  "issuanceDate": "2023-12-27T11:14:18.043912210+00:00",
  "parachainBlockNumber": 40,
  "sidechainBlockNumber": 75,
  "proof": {
    "created": "2023-12-27T11:14:18.044243691+00:00",
    "type": "Ed25519Signature2020",
    "proofPurpose": "assertionMethod",
    "proofValue": "e4e4708f660ed91829b3d1cbc312809b067c778ab6895c0e79382a9f80614a209757a7b514337b18dc7c95536ea5c1fcaae0319b25694607eb3f6914b4ba540c",
    "verificationMethod": "0xf7dc5214709ff407024b6a9d4987d900fc73e6d8d75a3b9493b8b1222e039b18"
  }
}
```


tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -208,19 +231,19 @@ where
...
+
 		// This internally fetches nonce from a Mutex and then updates it thereby ensuring ordering
-		let xt = extrinsic_factory
-			.create_extrinsics(&[call], None)
-			.map_err(|e| format!("Failed to construct extrinsic for parentchain: {:?}", e))?;
-		context
-			.ocall_api
-			.send_to_parentchain(xt, &ParentchainId::Litentry, false)
-			.map_err(|e| format!("Unable to send extrinsic to parentchain: {:?}", e))?;
+		// Only log the error message even if it fails as `res` is already available to the client
```
felixfaisal: We are not awaiting for the extrinsic to be included in a block, that is why the `false` in `send_to_parentchain(xt, &ParentchainId::Litentry, false)`.I think this begs the question of how important is sending an extrinsic to the parachain  and what are the consequence of not sending an extrinsic to parachain, and if none, I'm wondering why is there even a need to submit an `extrinsic`. What do you think?

Kailai-Wang: The conclusion was drawn from [this doc](https://www.notion.so/web3builders/Collection-of-open-discussions-2023-08-2299815065a24b3fbc9043194e5738cd#c5c395f513c64818b79ae536e1d096fa).

Basically we use these events for tracking and statistic purpose. Concretely, we can index the `vc_issued` events to know e.g. the requester count distribution for each assertion type. So it's basically acceptable that we are (somehow) missing these events randomly, but if there's a large amount of event failure, we should look into it.

`send_to_parentchain` can fail early if e.g. we have the wrong call_index even with `false` in parameter, I tried to be more lenient with such cases

felixfaisal: So every time we create an extrinsic(in vc-issuance) we utilize a Mutex to get the nonce(of worker), and this slows down the response time because every request essentially waits for this mutex to get unlocked. I was just wondering if it is inconsequential, it would be good to move away from it. (or Alternatively we don't have to send one extrinsic at a time and do batch). This might be out of scope for the PR, but just want to hear your thoughts on this.

Kailai-Wang: Yes I agree it would be good to remove it or do it aysnc
Will create an issue later

jonalvarezz (commented):
Thank you for including one example!

> Do we still need description and type fields?

Yes, definitely. I find it significant from a user point of view. I thought it was settled on this discussion https://linear.app/litentry/issue/P-142/enable-use-of-jsonschema-in-vc#comment-77d1adce

> How does it look?

I'm not very convinced about `assertionText`: replicating `A8[Litentry, Litmus] 128`  from the type may not be easy in other languages. Any value on having the size?

Can we achieve something like: 
```
"assertion": {
    id: "a8",
    params: ["litentry", "litmus"]
}
```

Or (inspired by litentry-cli's request-vc),
```
"assertion": "a8 litentry,litmus"
```

Kailai-Wang:
Thanks @jonalvarezz 

I reviewed "description" and "type" - "description" makes sense to me, however, "type" seems to be too generic to be used - do we even use it? Thus the question (again).

> Can we achieve something like:

1. We can, that's what's mentioned in the PR description: "use Display impl for customised printing". It's just not the most scalable - we'll have to manually customise the printing of each assertion type, which is OK, but we need to agree on the text.

This is kind of against the initial motivation of this PR though, the situation won't change much if parchain <> client-sdk / IDHub still reach agreements via offline discussions (agree on some expected strings and make comparison from client's side). So the more I think about it, the stronger I believe it should be done in a scalable way. Encoded assertion way is such a direction - although not human-readable. What do you think?

2. we have the json key `assertions` already, so `assertion` might create confusion - that's why I tried to use a more distinct word 😂

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
Kailai-Wang: one sec - I think nonce mutex acquiring should be quite light-weighted, and I don't see any time-consuming job during the lock holding period, so it _should_ be fine most of the time. It's still better to optimise it away though 🤔

Kailai-Wang: Created: P-393

jonalvarezz:
> however, "type" seems to be too generic to be used - do we even use it? Thus the question (again).

IDH uses it to help identify the kind of assertion. `type` is misleading though, maybe rename to `title`? i think it is nice to have in any case, same as `description`

> Encoded assertion way is such a direction - although not human-readable

I see your motivation but I don't think we need to reach an agreement with clients for the encoding as long as it is predictable and human readable. The `litentry-cli request-vc` is a good example, perhaps we can follow that one? like one should be able to copy and paste that on the CLI and get the same assertion. 

> we have the json key assertions already, so assertion might create confusion - that's why I tried to use a more distinct word 

🥲

Kailai-Wang:
> I see your motivation but I don't think we need to reach an agreement with clients for the encoding as long as it is predictable and human readable. 

The problem is it's not predictable (for now), remember this could change over the runtime upgrade where we add/remove assertions. Encoding doesn't need an agreement because the mechanism is public and firm.

In fact, avoiding an offline agreement was the motivation to include such an additional field. It tries to solve the problem: given a VC how can we tell which assertion it is about? - in a scalable way.

If we need human-readable information, "description" and "type" exist 🤔

BillyWooo (commented):
I am OK to include these new fields. But still don't get the point how it can scale VC out?
Besides, "type" is used for identify/ differentiate VCs.  Yes. It's not a good name and design.

Kailai-Wang:
> I am OK to include these new fields. But still don't get the point how it can scale VC out? Besides, "type" is used for identify/ differentiate VCs. Yes. It's not a good name and design.

The assertionText is simply the debug text of the assertion, so you don't have to define it manually (each time when you add a new assertion

BillyWooo (approved):
LGTM

felixfaisal (approved):
LGTM, Would be great if we can get this merged as it blocks one of my PRs, Thanks!

PR #2366, opened 2023-12-28
Support prettified bitcoin signing message

Kailai-Wang:
### Context

As topic.

This PR adds a preliminary implementation of prettified bitcoin message similar to prettified ethereum message:  `Litentry authorization message: ` + `0x` prefixed hex-encoded message.


tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
```diff
@@ -91,6 +91,7 @@ export default {
                 Ethereum: "EthereumSignature",
                 EthereumPrettified: "EthereumSignature",
                 Bitcoin: "BitcoinSignature",
+                BitcoinPrettified: "BitcoinSignature",
```
jonalvarezz, line 94: Any strong reason to keep both?

jonalvarezz (approved):
sweet!

tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
Kailai-Wang: Analog to ethereum sig,  we keep both the standard and prettified version

jonalvarezz: ethereum standard isn't used by IDH, so I thought of keeping the prettified version only. Enforcing prettified version for clients sounds good to me. we want all users to be able to read what clients ask them to sign.

cryptic signatures (standard versions) could be abused by clients we don't control.

kziemianek (approved):

tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
Kailai-Wang: >  cryptic signatures (standard versions) could be abused by clients we don't control.

I don't get this one

IMO enforcing prettified version doesn't help much here, the client could still wrap it with prefix and trick users into signing it. Incidentally, the prettified version is only related to the identity verification, in trusted service invocation later it accepts standard crypto sig only.

PR #2367, opened 2023-12-29
Fix positional args in request-vc CLI

Kailai-Wang:
### Context

This PR tries to fix the positional args problem in request-vc `amount-token` CLI. The root cause is the terrifying combination of "positional args" with `num_args` + required arg + optional arg, which causes a big problem in clap parser.

The best solution I've found so far is to use the flag explicitly, but with the "equals" form, see the comment and example in the code.

Minimal repro-case:
```
use clap::Parser;

#[derive(Parser, Debug)]
struct Cli {
    #[clap(
        num_args = 1..,
        value_delimiter(','),
        required(true),
    )]
    chain: Vec<String>,
    amount: String,
    token: Option<String>,
}

fn main() {
    let cli = Cli::parse();
    println!("cli: {:#?}", cli);
}
```

Good luck with these trial runs:
```
cargo run -- op1,op2 3
cargo run -- op1,op2 3 token
cargo run -- op1 op2 3 token
```

See how it behaves differently without the optional `token`.
Even with the flag the space form won't work:
```
cargo run -- -c op2 op3 3
cargo run -- -c op2 op3 3 token
```

### DIFF
--- a/tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
+++ b/tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
@@ -74,7 +74,7 @@ pub struct RequestVcCommand {
 }
 
 // see `assertion.rs`
-#[derive(Subcommand)]
+#[derive(Subcommand, Debug)]
 pub enum Command {
 	A1,
 	A2(A2Arg),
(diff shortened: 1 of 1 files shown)


felixfaisal (approved):
Thanks Kai for the quick PR, This looks good from my side.

jonalvarezz (approved):
wouldn't changing the order of the args solve the issue too? anyway I'm ok with it if it works 😄 – thanks Kai for the quick fix

Kailai-Wang:
> wouldn't changing the order of the args solve the issue too? anyway I'm ok with it if it works 😄 – thanks Kai for the quick fix

It won't, unfortunately, because of the optional arg

PR #2372, opened 2024-01-02
Update lit staking response result field to Option<String>

outofxxx:
Resolve the situation of returning null.

```json
{
    "jsonrpc": "2.0",
    "result": null,
    "id": 1
}
```


Kailai-Wang (approved):
Thanks!

PR #2373, opened 2024-01-02
Replace `T::AccountId` with `Identity` in callback extrinsics

Kailai-Wang:
### Context

As topic - it's a breaking change for the client.


pallets/identity-management/src/lib.rs
```diff
@@ -93,22 +93,22 @@ pub mod pallet {
 		// TODO: what if the event is triggered by an extrinsic that is included in a batch call?
 		//       Can we retrieve that extrinsic hash in F/E?
 		IdentityLinked {
-			account: T::AccountId,
+			identity: Identity,
```
kziemianek: This might be misleading - it doesn't hold information about the identity that was linked, but idgraph's root identity.
What do you think about renaming it ? `primary_identity` or `idgraph_root_identity`  ?

Kailai-Wang: Good catch, thanks - I'll rename it (and other occurrences) to `prime_identity`

Kailai-Wang: renamed

BillyWooo (approved):
I only checked rust part. LGTM

kziemianek (approved):

PR #2382, opened 2024-01-05
fix: add storage item for feature control

felixfaisal:
This PR updates the code to make use of storage item instead of feature-guard for `skip-scheduled-enclave-check` 
- [x] Generate Weights for new extrinsic 
So I initially thought that I would need to benchmark this new extrinsic, but I also see that not all extrinsics are benchmarked, and I'm using the same weights as one of the extrinsics, I think the operation of this extrinsic fairly simple as it performs a single write, So I also don't think we need to perform actually benchmarking as there is no worst case complexity. 
- [x] Unit test to test this PR 
So I've updated the unit tests to make use of the storage item instead of feature guards. 

I'm actually not sure if this PR does all the necessary changes 😅  
- So we can also do the same `skip-ias-dev` but I want to do it another PR and not in this 
- We have to update `deploy.sh` however I'm not sure if we have to `Alice` as root in production environment or we should manually call this extrinsic


pallets/teerex/src/lib.rs
```diff
@@ -132,6 +132,10 @@ pub mod pallet {
 	#[pallet::getter(fn enclave_count)]
 	pub type EnclaveCount<T: Config> = StorageValue<_, u64, ValueQuery>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn schedule_enclave)]
+	pub type ScheduleEnclave<T: Config> = StorageValue<_, bool, ValueQuery>;
```
Kailai-Wang: We should have a better name.

We have `ScheduledEnclave` already which can be easily mixed up with this one.

How about `ScheduledEnclaveCheck` or on the contrary, `SkipScheduledEnclaveCheck`? Please take the default value (`false`) into account - maybe the latter is better, as we don't want to skip this check by default in prod.

Kailai-Wang: Alternatively we should put it to genesis

pallets/teerex/src/lib.rs
```diff
@@ -637,6 +645,20 @@ pub mod pallet {
...
+
+		/// This extrinsic is used to set ScheduleEnclave storage item
+		/// This storage item is used to perform feature control during register_enclave
+		#[pallet::call_index(31)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal, Pays::No))]
+		pub fn set_schedule_enclave(
+			origin: OriginFor<T>,
+			schedule_enclave: bool,
+		) -> DispatchResultWithPostInfo {
+			ensure_root(origin)?;
```
Kailai-Wang: I think the admin account origin check is better

pallets/teerex/src/lib.rs
```diff
@@ -637,6 +645,20 @@ pub mod pallet {
...
+
+			<ScheduleEnclave<T>>::set(schedule_enclave);
```
Kailai-Wang: We should emit an event IMO

Kailai-Wang (changes requested):
Thanks!

> We have to update deploy.sh however I'm not sure if we have to Alice as root in production environment or we should manually call this extrinsic

In prod we have parachains already running, so we have to either rely on the default value of this storage, or set it manually (which is hard for automation)

pallets/teerex/src/lib.rs
felixfaisal: > How about ScheduledEnclaveCheck or on the contrary, SkipScheduledEnclaveCheck? Please take the default value (false) into account - maybe the latter is better, as we don't want to skip this check by default in prod.

I agree, I will update the code

pallets/teerex/src/lib.rs
felixfaisal: Updated to use `TeerexAdmin`

pallets/teerex/src/lib.rs
felixfaisal: Emitting event now

pallets/teerex/src/lib.rs
felixfaisal: So I did the above and updated the scripts for `local-setup` and `deploy.sh` and not using `genesis`.

pallets/teerex/src/lib.rs
```diff
@@ -637,6 +647,23 @@ pub mod pallet {
 			// Do not pay a fee
 			Ok(Pays::No.into())
 		}
+
+		/// This extrinsic is used to set ScheduleEnclave storage item
+		/// This storage item is used to perform feature control during register_enclave
+		/// Can only be called by the Teerex Admin
+		#[pallet::call_index(31)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal, Pays::No))]
+		pub fn set_scheduled_enclave_check(
```
Kailai-Wang: Sorry for being pernickety, the name is still misleading..  you "set the scheduled-enclave check" to _true_, but in fact, the check is skipped (because you actually set the SkipScheduledEnclaveCheck storage)

How about:
```
pub fn set_skip_scheduled_enclave_check(origin, should_skip)
```

Kailai-Wang (commented):
Looks good now! Only a small thing

kziemianek:
What do you think about letting primary validateer (first worker to register in teerex's enclave registry) to set scheduled mrenclave to it's actual mrenclave in `dev` ?

Would we still need `skip-scheduled-enclave-check` feature/fag ?

Kailai-Wang:
> What do you think about letting primary validateer (first worker to register in teerex's enclave registry) to set scheduled mrenclave to it's actual mrenclave in `dev` ?
> 
> Would we still need `skip-scheduled-enclave-check` feature/fag ?

We could do that but it will still require scheduled-enclave-check for subsequent registration. So it saves the step to set the first / genesis mrenclave from external

kziemianek:
> > What do you think about letting primary validateer (first worker to register in teerex's enclave registry) to set scheduled mrenclave to it's actual mrenclave in `dev` ?
> > Would we still need `skip-scheduled-enclave-check` feature/fag ?
> 
> We could do that but it will still require scheduled-enclave-check for subsequent registration. So it saves the step to set the first / genesis mrenclave from external

You mean if code changes ... hmm maybe we can override it everytime.

felixfaisal:
So I made a few changes 
 - Instead of calling an extrinsic during local-setup, I've included this in the genesis spec for `rococo-dev` chain 
 - This was because I had a few CI problems, and I felt this was also better than calling an extrinsic during the launch.

pallets/teerex/src/lib.rs
felixfaisal: Updated ✅

scripts/launch-standalone.sh
```diff
@@ -28,4 +28,12 @@ $PARACHAIN_BIN --dev --unsafe-ws-external --unsafe-rpc-external \
...
   &> "$LITENTRY_PARACHAIN_DIR/para.alice.log" &
 
+echo "Setting skip-schedule-enclave-check flag ..." 
+cd "$ROOTDIR/ts-tests"
+if [[ -z "${NODE_ENV}" ]]; then
+    echo "NODE_ENV=ci" > .env
+else
+    echo "NODE_ENV=${NODE_ENV}" > .env
+fi
+
```
Kailai-Wang: I don't think it actually sets it

Kailai-Wang (approved):
Looks good, thanks! Just one small thing

scripts/launch-standalone.sh
felixfaisal: Updated ✅

Kailai-Wang:
Hmm there seems to be CI error

PR #2393, opened 2024-01-09
fix: double encoding

felixfaisal:
This is a simple PR - It removes double encoding, The RPC handler does not need to encode the payload again as it receives the encoded payload already.


Kailai-Wang (approved):
Thanks!

We don't have relevant tests, do we? That's why we haven't discovered it

felixfaisal:
Yeah we haven't included CI tests yet for direct request vc else it would have been caught by it.

PR #2398, opened 2024-01-10
Hotfix - IDH-488 Fix lit staking parse error

outofxxx:
Filter out web2 identity.


Kailai-Wang (commented):
Thank you! 

Does it happen to other assertion building as well? Shall we have such a filtering in one level higher?

outofxxx:
> Thank you!
> 
> Does it happen to other assertion building as well? Shall we have such a filtering in one level higher?

No other issues have been found yet, but here I used the identity method `to_account_id` cause ParseError. Filtering out web2 here can solve the problem.

Kailai-Wang (approved):

PR #2399, opened 2024-01-10
refactor: remove substratetestnet

felixfaisal:
I think this PR is good for review, It's a simple PR that removes the `SubstrateTestnet` and based on @Verin1005 said, I don't have to generate types manually as it is done in the CI.

### DIFF
--- a/tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
+++ b/tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
@@ -41,7 +41,6 @@ export default {
                 "Litmus",
                 "LitentryRococo",
                 "Khala",
-                "SubstrateTestnet",
                 "Ethereum",
                 "Bsc",
                 "BitcoinP2tr",
--- a/tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
+++ b/tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
@@ -41,7 +41,7 @@ export default {
                 "Litmus",
                 "LitentryRococo",
                 "Khala",
-                __Unused6: "Null",
+                "Null",
                 "Ethereum",
                 "Bsc",
                 "BitcoinP2tr",
--- a/tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
+++ b/tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
@@ -41,6 +41,7 @@ export default {
                 "Litmus",
                 "LitentryRococo",
                 "Khala",
+                __Unused6: "Null",
                 "Ethereum",
                 "Bsc",
                 "BitcoinP2tr",
--- a/primitives/core/src/network.rs
+++ b/primitives/core/src/network.rs
@@ -70,9 +70,7 @@ pub enum Web3Network {
 	LitentryRococo,
 	#[codec(index = 5)]
 	Khala,
-	#[codec(index = 6)]
-	SubstrateTestnet, // when launched it with standalone (integritee-)node
-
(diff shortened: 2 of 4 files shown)


0xverin:
> Generate client types, Need help

Did yesterday's discussion help you?

felixfaisal:
> > Generate client types, Need help
> 
> Did yesterday's discussion help you?

Yeah I think I got it, It is automatically generated in CI, So I don't think I have to do anything in that area

tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
```diff
@@ -41,7 +41,6 @@ export default {
                 "Litmus",
                 "LitentryRococo",
                 "Khala",
-                "SubstrateTestnet",
```
kziemianek, line 44: You need to fill the gap with `__Unused6: "Null"`

felixfaisal: Thank @kziemianek! Updated ✅

0xverin (approved):
LGTM.Thanks!

kziemianek (approved):

tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
```diff
@@ -41,7 +41,7 @@ export default {
...
+                __Unused6: "Null",
```
Kailai-Wang: The syntax is wrong tho 😂 

Btw is `Web3Network` actually used? Shouldn't we use `CorePrimitivesNetworkWeb3Network` ubiquitously? @Verin1005 I thought this problem was fixed already

0xverin: I checked, and we are still using the `Web3Network` at the moment.

Kailai-Wang: Hmm why is that? I thought it had been mentioned multiple times that we'd need to use auto-generated types :(

0xverin: We can indeed replace the introduced [type](https://github.com/litentry/litentry-parachain/blob/p-92-remove-substratetestnet-from-web3network/tee-worker/ts-tests/integration-tests/di_evm_identity.test.ts#L28) with `CorePrimitivesNetworkWeb3Network`.

But this PR update will have an impact on these [definitions](https://github.com/litentry/litentry-parachain/blob/p-92-remove-substratetestnet-from-web3network/tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts#L36C105-L36C116) , so `Web3Network` actually used.

Kailai-Wang: Thanks for checking, why can't we use `CorePrimitivesNetworkWeb3Network` in the definitions? Or is there a way to do that?

0xverin: I have made updates [here](https://litentry.slack.com/archives/C05328L0GAF/p1703884176997769), and the conclusion is negative. @jonalvarezz 's approach is not feasible in our case; we cannot manually import types every time.

Kailai-Wang: Hmm so we can't define the source like types from registry or lookup?

0xverin: As of now, it seems not possible; it cannot be automatically imported. I've also referenced Moonbeam repo and haven't seen them do something similar at the moment. :(

PR #2400, opened 2024-01-11
P-62 Connect the worker to litentry-rococo parachain

BillyWooo:
- bug fix: compose teerex::register_enclave extrinsics; 
- use intel 'dev' server for attestation temporarily; 
- add one example config file

### DIFF
--- a/tee-worker/local-setup/rococo_one_worker.json
+++ b/tee-worker/local-setup/rococo_one_worker.json
@@ -1,0 +1,29 @@
+{
+  "workers": [
+    {
+      "source": "bin",
+      "flags": [
+        "--clean-reset",
+        "--ws-external",
+        "-P",
+        "2000",
+        "-w",
+        "2001",
+        "-r",
+        "3443",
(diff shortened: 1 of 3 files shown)


tee-worker/local-setup/rococo_one_worker.json
```diff
@@ -0,0 +1,29 @@
...
+        "-r",
+        "3443",
+        "-h",
+        "4545",
+        "-u",
+        "wss://rpc.rococo-parachain.litentry.io",
+        "-p",
+        "443",
+        "--running-mode",
+        "mock",
```
Kailai-Wang, line 21: We don't need this I believe

Kailai-Wang (approved):
Thanks

tee-worker/local-setup/rococo_one_worker.json
BillyWooo: But we need to keep 'mock' in example file. Change it to real config file when deploying.

PR #2401, opened 2024-01-11
P-54 Binary search for holding time assertion

arieljbirnbaum:
### Context

Reduce the number of data provider requests needed to implement the "holding time" assertion.

Resolves [P-54](https://linear.app/litentry/issue/P-54/reduce-number-of-requests-in-holder-assertion-building)

### How (Optional)

The basic idea is to do a binary search over the range of relevant dates, testing every relevant address on each iteration, and discarding any addresses that prove to be irrelevant, that is, where we know for sure that the longest holding time is _not_ held by that address.

An address is proven irrelevant if:
- The data provider query returns `false` for that address on a given date
- The data provider query returns `true` for _another_ address on _the same_ date

### Note on Error Handling

Since the goal is to find the earliest holding date (and not necessarily to identify the corresponding address), it is OK to ignore certain errors as long as the search can proceed. In a nutshell, the search can only get "stuck" (i.e. we don't know whether to choose the earlier or the later half of the date range) if:

- No data provider query returns `true`
- At least one query returned an error.

In that case, the actual value of the failed query may have been `true` or `false`, so we don't know in which direction the search should continue.

If at least one query returned `true`, we know that the search continues in the earlier half; at worst we are failing to remove some irrelevant addresses that will be discarded in the subsequent step when the query returns `false` for them.

### TODO

- [ ] Pass CI


tee-worker/litentry/core/assertion-build/src/holding_time.rs
```diff
@@ -81,112 +81,243 @@ pub fn build(
 ) -> Result<Credential> {
 	debug!("Assertion A4 build, who: {:?}", account_id_to_string(&req.who));
 
-	let q_min_balance = pre_build(&htype, &min_balance)?;
-	let identities = transpose_identity(&req.identities);
-	let (is_hold, optimal_hold_index) = do_build(identities, &htype, &q_min_balance)
+	let q_min_balance = prepare_min_balance(&htype, &min_balance)?;
+	let accounts = prepare_accounts(&req.identities, &htype);
+
+	// TODO: handle empty `identities`
```
Kailai-Wang: `identities` can't be empty as otherwise it should have aborted with `NoEligibleIdentity` error (before entering the assertion build logic)

tee-worker/litentry/core/assertion-build/src/holding_time.rs
```diff
@@ -81,112 +81,243 @@ pub fn build(
...
+	);
+	return client.is_holder(account.address.as_str(), holding).map_err(|e| {
+		error!("Assertion HoldingTime request error: {:?}", e);
+		e.into_error_detail()
+	})
+}
+
+// Check against the data provider whether any of the given accounts has been holding since the given date.
+// If at least one positive outcome is found, the accounts that yielded a (conclusive) negative outcome are eliminated.
+fn holding_time_search_step<'r>(
```
Kailai-Wang: I don't think you need lifetime here

tee-worker/litentry/core/assertion-build/src/holding_time.rs
```diff
@@ -81,112 +81,243 @@ pub fn build(
...
+) -> (QueryOutcome, Vec<Account>) {
+	// Check all remaining identities on the given date
+	let outcomes: Vec<QueryOutcome> = accounts
+		.iter()
+		.map(|account| account_is_holding(client, q_min_balance, &account, date))
+		.collect();
+
+	// If any positive result is found:
+	//   - Discard all identities that yielded a _negative_ result
+	//     - KEEP the ones that yielded error; they may still be relevant!
```
Kailai-Wang: indention :)

tee-worker/litentry/core/assertion-build/src/holding_time.rs
```diff
@@ -81,112 +81,243 @@ pub fn build(
...
+	//   - Return the remaining identities with a positive value to continue the search
+
+	if outcomes.iter().any(is_positive) {
+		let new_accounts = accounts
+			.into_iter()
+			.zip(outcomes.iter())
+			.filter_map(|(account, outcome)| (!is_negative(outcome)).then_some(account))
+			.collect();
+		return (Ok(true), new_accounts)
+	}
```
Kailai-Wang: Can we merge it into the `accounts.iter()` traverse above? I feel it should be possible

Kailai-Wang (commented):
Thank you, the intention is quite clear!

Have you considered [partition_point](https://doc.rust-lang.org/std/primitive.slice.html#method.partition_point)? It applies the binary search under the hood too.

We can treat the date array as "sorted"/ partitioned because `holding_time(index) == true` implies `holding_time(index + 1) == true` (but not vice versa). If that's not the case in practice, then data provider malfunctions and you won't get the right result anyway.

arieljbirnbaum:
@Kailai-Wang 

> Thank you, the intention is quite clear!

Thanks!

> Have you considered [partition_point](https://doc.rust-lang.org/std/primitive.slice.html#method.partition_point)? It applies the binary search under the hood too.

Didn't know about it; will have a look! Thanks

> We can treat the date array as "sorted"/ partitioned because `holding_time(index) == true` implies `holding_time(index + 1) == true` (but not vice versa). If that's not the case in practice, then data provider malfunctions and you won't get the right result anyway.

Exactly; that's the entire justification for the search algorithm.

tee-worker/litentry/core/assertion-build/src/holding_time.rs
arieljbirnbaum: It's intended to be a subpoint; should I format it differently to make that clear?

tee-worker/litentry/core/assertion-build/src/holding_time.rs
arieljbirnbaum: I did need it at some point (at least when I was trying my best to avoid `clone`); will check if I can ditch it now.

tee-worker/litentry/core/assertion-build/src/holding_time.rs
arieljbirnbaum: Good to know :)
Does Rust have a way of reflecting that in the type system? Otherwise it's a bit fragile to assume that there's only one pathway into this code, so maybe we want to have the redundant check here anyway.

Kailai-Wang: Hmm it's a runtime check so I doubt if you could solve it via type system - even if you had one that's practically equal to an if-check in runtime.

Yes you are free to do a double-check, just saying this shouldn't be the case if the logic functions well :D

tee-worker/litentry/core/assertion-build/src/holding_time.rs
Kailai-Wang: Oh I thought it was a major bullet point as `keep` and `discard` seem concurrent :D

Kailai-Wang:
`partition_point` is basically a binary search by predicate, applying it would make the code simpler (so better readability) and less error-prone (as it's a library func) IMO

arieljbirnbaum:
> `partition_point` is basically a binary search by predicate, applying it would make the code simpler (so better readability) and less error-prone (as it's a library func) IMO

On further look, there's a problem with error handling -- the predicate arg to `partition_point` must return `bool`, but the predicate depends on data provider queries which could fail, so the search logic needs to be able to handle that (at the very least, to bubble the error without panicking)

tee-worker/litentry/core/assertion-build/src/holding_time.rs
arieljbirnbaum: I'll add a "but" then :)

tee-worker/litentry/core/assertion-build/src/holding_time.rs
arieljbirnbaum: If I had a satoshi for every bug I've ran into that could have been avoided by a little paranoia... ;)

Kailai-Wang:
We should be able to use additional flag/storage to tackle that 🤔

arieljbirnbaum:
> We should be able to use additional flag/storage to tackle that 🤔

Care to elaborate on that? Or, do you have a link with some info? Because from the API docs (and some cursory search) I haven't found any way to work around that (without reimplementing the search algorithm) :thinking:

tee-worker/litentry/core/assertion-build/src/holding_time.rs
arieljbirnbaum: Possibly, but I'm not sure it's worth it. We only do the filtering _if_ any of the queries returned true in that iteration; if we want to merge the two loops we'd have to speculatively build the filtered array in advance, and I think that would make the logic messier to follow. In terms of runtime, I don't think the added loop to construct the filtered list makes much of a difference; what dominates here is in any case the query provider requests.

tee-worker/litentry/core/assertion-build/src/holding_time.rs
arieljbirnbaum: It's gone; thanks for the catch :)

Kailai-Wang (approved):
`partition_point` _might_ produce more readable code, but let's leave it to another PR even if that's the case

outofxxx:
Hi, @grumpygreenguy, It's good to have this try, thanks. and i have a question, how much can requests be reduced in the worst-case scenario after this optimization?

arieljbirnbaum:
@zhouhuitian 

> Hi, @grumpygreenguy, It's good to have this try, thanks. and i have a question, how much can requests be reduced in the worst-case scenario after this optimization?

The complexity goes from linear in the number of dates to logarithmic in the same; specifically, if we're checking for a single address, the number of requests in the worst case goes from 15 to 5.
If there are multiple addresses to check, the complexity is still in the worst case linear in the number of addresses as well, but depending on how quickly the search can spot and eliminate irrelevant addresses, the cost of each iteration can be drastically reduced. Would need to have a better idea of the distribution of holding times between multiple accounts of real users to make an estimate of the average runtime, tho.

tee-worker/litentry/core/assertion-build/src/holding_time.rs
```diff
@@ -239,9 +240,9 @@ fn search_holding_date(
...
 	let mut client = AchainableClient::new(data_provider_config);
 
-	let mut pred = |date| {
+	let mut pred = |date: &&str| {
 		let (outcome, new_accounts) =
-			holding_time_search_step(&mut client, q_min_balance, accounts, date);
+			holding_time_search_step(&mut client, q_min_balance, accounts.clone(), *date);
 		accounts = new_accounts;
 		outcome.map(|is_holding| !is_holding) // negated to match the partition_point API
 	};
```
arieljbirnbaum: Thanks @Kailai-Wang for fixing this issue! Still gotta wrap my head around some of the subtleties here, apparently :D

In this case the extra `clone` shouldn't make a huge difference overall, since the cost is in any case dominated by the HTTP requests; in general though it does feel a bit wasteful to have to clone the array each time, only to replace the original anyway after the call to `holding_time_search_step`. What would (in theory) have been the better approach here? (Not for this PR to be sure, just as an overall learning)

PR #2403, opened 2024-01-13
Create idgraph upon first vc request

Kailai-Wang:
### Context

As topic.

This PR:
- create an IDGraph upon the first `request_vc` call, both STF version and threaded version are adjusted
- rename `author_submitVCRequest` to `author_requestVc`, the reason is `author_submitXXRequest` is used for generic (opaque) request sending, so to not confuse with it
- adds the signature validation and origin check in the threaded vc request
- `VcIssued` event now has a `id_graph_hash` field
- `RequestVcResult` now has `pre_mutated_id_graph` and `pre_id_graph_hash` fields
- some other improvements

There's no ts-test for threaded vc requesting, so I tested it locally with CLI and it looked good.

Please pay special attention to the comment in `lc-vc-task-receiver`:

```
// To create IDGraph upon first vc request (see P-410), there're two options:
//
// 1. synchronous creation:
// we delegate the vc handling to the STF version, which only returns when the IDGraph is actually created (= InSidechainBlock state).
// The downside of this method is that the first vc_request processing time is limited to the sidechain block interval.
//
// 2. asynchronous creation (this implementation):
// we check if an IDGraph **could** be created and then process the VC request right away, meanwhile, we submit a trusted call to
// top pool. So the IDGraph will be created async: in the next sidechain block. In the `RequestVCResult` we return the pre-calculated
// `mutated_id_graph` and `id_graph_hash`.
//
// Corner case: there's a small chance that some IDGraph mutation was injectd in between. For example, a client sends `request_vc` which
// is closely followed by a `link_identity` request. In this case, the IDGrpah creation resulting from `vc_request` would fail, as
// the IDGraph would have been created already by that time. But this is OK as long as it reaches the desired state eventually.
//
// However, `RequestVCResult` might carry with outdated `mutated_id_graph` and `id_graph_hash` if it lands later than `LinkIdentityResult`.
// So we call the fields `pre_mutated_id_graph` and `pre_id_graph_hash` to show they are pre-calculated.
// The client should take proper actions against it, e.g., only use the value when the local IDGraph is empty.
//
// Please note we can't mutate the state inside vc-task-receiver via `load_for_mutation` even
// though it's lock guarded, because: a) it intereferes with the block import on another thread, which eventually
// cause state mismatch before/after applying the state diff b) it's not guaranteed to be broadcasted to other workers.
```


tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -114,18 +120,50 @@ where
...
+			Ok(v) => v,
+			Err(e) =>
+				return Ok(json!(compute_hex_encoded_return_error(
+					format!("Failed to construct AesRequest: {:?}", e).as_str()
+				))),
+		};
 
-			Ok(json!(json_value))
+		let vc_request_sender = VcRequestSender::new();
+		let (sender, receiver) = channel::<Result<Vec<u8>, String>>();
```
felixfaisal: I see that we have replaced `oneshot` channel with `mpsc` channel, Does it make sense to use an MPSC channel here?

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
```diff
@@ -110,132 +113,195 @@ where
...
-		// This internally fetches nonce from a Mutex and then updates it thereby ensuring ordering
+
+		// submit TrustedCall::maybe_create_id_graph
+		let enclave_signer: AccountId = context
+			.enclave_signer
+			.get_enclave_account()
+			.map_err(|_| "Failed to get enclave signer".to_string())?;
+		let c = TrustedCall::maybe_create_id_graph(enclave_signer.into(), who);
+		context
+			.submit_trusted_call(&request.shard, None, &c)
```
felixfaisal: Have you tried multiple requests that try to create idgraph? This when run in parallel in threadpools will likely fail due to nonce. The nonce calculation takes the existing nonce and the pending trusted calls, so it ends up having the same nonce when run in parallel. Maybe we could have a unit test or `ts-test` for this case.

tee-worker/app-libs/stf/src/trusted_call_result.rs
```diff
@@ -98,4 +98,8 @@ pub struct RequestVCResult {
 	pub vc_index: H256,
 	pub vc_hash: H256,
 	pub vc_payload: AesOutput,
+	// see comments in `lc-vc-task-receiver` why it's prefixed with `pre...`
+	// they should be referenced/used only when the client's local IDGraph is empty
+	pub pre_mutated_id_graph: AesOutput,
```
felixfaisal, line 101: Based on the comments, Shouldn;t we use `Option` here?

felixfaisal (commented):
Thanks @Kailai-Wang  
It looks really good, except for few comments from my side.

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
Kailai-Wang: > I see that we have replaced `oneshot` channel with `mpsc` channel, Does it make sense to use an MPSC channel here?

It is an mpsc channel :)

tee-worker/app-libs/stf/src/trusted_call_result.rs
Kailai-Wang: It's actually a taste of semantics - currently if nothing is mutated then the `mutated_id_graph` will be empty vector, which is fine as well IMO. `Option` will do the job too - if it's more beneficial we can introduce another PR (as all need other `xxxResult` need to be changed too)

Tagging @jonalvarezz here

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
Kailai-Wang: This is a good point.

I think either adding mutex or redirecting them to single-thread execution (e.g. using `mpsc`) to tackle this, wdyt?

felixfaisal: Yeah either of it works, I think  I've tackled it in `stf-thread` to redirect for single-threaded execution - https://github.com/litentry/litentry-parachain/pull/2159

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
felixfaisal: So what I'm trying to say is we are using `mpsc` channel here over a `oneshot` channel even though we do not have multiple producers or require more than one message, In the above RPC function we have a single producer and a single consumer and a single message, all of which are attributes of a oneshot channel. Let me know what you think.

Kailai-Wang: Yes I believe either way would work, maybe even `Arc<Mutex<DequeVec<T>> ` works too.

I just wanted to keep it simple (and consistent) so that I didn't have to introduce extra crates 😂 - do you see any downsides using mpsc? Like more resource intensive?

tee-worker/app-libs/stf/src/trusted_call_result.rs
jonalvarezz: > if nothing is mutated then the mutated_id_graph will be empty vector, which is fine as well IMO

It is perfect for clients as they can skip building the `Option` type

Kailai-Wang: The trade-off is you always have to perform the decryption :)
A `None` is detected at the first sight

Kailai-Wang: Anyways if our client representative Jonathan doesn't have objection I'll leave it as it is :D

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
felixfaisal: Yeah, that is precisely the difference while the `mpsc` internally makes use of something like `Arc<Mutex<DequeVec<T>>` for synchronisation of multiple producers, a `oneshot` channel makes use of something like `Arc<Mutex<Option<T>>`, it's much more simpler and applicable for the above usecase. (It is an oversimplification of what they are using underneath). Since we are creating a channel everytime the user requests, it is definitely more resource exhaustive to use `mpsc` channel over `oneshot` channel. We don't need all the synchronisation mechanism if we only have single producer and single message.

Kailai-Wang: That's a good arg - I'll change it to oneshot channel

tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs
Kailai-Wang: Fixed that - asking for a re-review

jonalvarezz:
Thanks, Kai!

I believed that `id_graph` was created on `vc_request` already because of https://github.com/litentry/litentry-parachain/pull/2347, thanks for fixing it.

I am a bit hesitant on this, though:

> RequestVcResult now has `pre_mutated_id_graph` and `pre_id_graph_hash fields`
> 
> ```
> // However, `RequestVCResult` might carry with outdated `mutated_id_graph` and `id_graph_hash` if it lands later than `LinkIdentityResult`.
> // So we call the fields `pre_mutated_id_graph` and `pre_id_graph_hash` to show they are pre-calculated.
> // The client should take proper actions against it, e.g., only use the value when the local IDGraph is empty.
> ```

1. Do I follow correctly that those properties are only present if the id_graph was created during `request_vc`? I.e., if an id_graph already existed, those properties won't be part of the response.
2. If so, does the `pre_` really add any value to the client? Given the conditions that local id_graph is empty and request_vc returns the mutated id graph, a client has no option but to trust that result. If a client allows scenarios where `link_identity` and `request_vc` can happen simultaneously, then it should also handle that case. Or maybe a more dramatic name could help? like the whole `pre_calculated_` prefix? Including the sidechain block in the response at which it was calculated may help too for clients to make smarter decisions.

Kailai-Wang:
> Do I follow correctly that those properties are only present if the id_graph was created during request_vc? I.e., if an id_graph already existed, those properties won't be part of the response.

No, the fields will be always there regardless. In case of an existing IDGraph, the `pre_id_graph_hash` will contain the hash of current IDGraph. The client should compare the returned hash with local hash first, if unequal, then refer to `pre_mutated_id_graph` to update the IDGraph.

We could have omitted these fields and make an offline agreement with client that whenever you request vc for the first time (= local IDGraph is empty), you should create one locally and you know same will happen on the sidechain. This will work, but I just wanted to make it more explicit as this call **could** trigger an IDGraph update.

Simultaneous calls of `request_vc` and `link_identity` from the same client aren't impossible but should be very rare - and in case the `RequestVcResult` comes later than `LinkIdentityResult` the values can be safely ignored. I mean that's why it's prefixed with `pre` while other `xxResult` doesn't have this prefix - it's supposed to be used as reference value

BillyWooo (approved):
Very nice. LGTM

pallets/vc-management/src/lib.rs
```diff
@@ -129,10 +129,12 @@ pub mod pallet {
 		},
 		// event that should be triggered by TEECallOrigin
 		// a VC is just issued
+		// we have `id_graph_hash` field since vc request could create the IDGraph
 		VCIssued {
 			identity: Identity,
 			assertion: Assertion,
 			index: VCIndex,
+			id_graph_hash: H256,
```
kziemianek, line 117: Why do we add this value here ? Just to log that idgraph was created or because we assume client will use it ? If the latter I think client will need to get idgraph anyway.

kziemianek (approved):

pallets/vc-management/src/lib.rs
jonalvarezz: it brings consistency to other operations that mutate the id_graph too. I think it is nice to have. get_id_graph from client sounds tempting until one remembers that the user has to sign that request every time.

jonalvarezz:
Thanks for elaborating, @Kailai-Wang. 

I guess I am also concerned that we need to remember what the `pre_` means and when it is useful. We better map all this knowledge out into our public documentation soon, as not everyone will look into the source code.

pallets/vc-management/src/lib.rs
Kailai-Wang: Just for logging purpose - we don't expect the client to use this event

felixfaisal (approved):
LGTM

PR #2404, opened 2024-01-14
Use whole unsigned VC as VC proof signature payload

kziemianek:
* Changes the input for VC proof signature calculation from enclave's mrenclave to VC json.
* Removes all logic/calls from `VCManagement` related to `VCRegistry`. 

After this change newly issued VC content can be verified just by verifying proof's signature, there would be no need to reach out to onchain's `VCRegistry`.


tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -244,64 +244,71 @@ export async function assertIdGraphMutationResult(
...
     const { proof, ...vcWithoutProof } = vcPayloadJson;
-    assert.equal(vcIndex, vcPayloadJson.id, 'Check VcIndex error: VcIndex should be equal to vcPayload id');
 
     // step 5
+    // prepare teerex enclave registry data for further checks
     const enclaveCount = await context.api.query.teerex.enclaveCount();
+    const parachainBlockHash = await context.api.query.system.blockHash(vcPayloadJson.parachainBlockNumber);
+    const apiAtVcIssuedBlock = await context.api.at(parachainBlockHash);
+    const lastRegisteredEnclave = (await apiAtVcIssuedBlock.query.teerex.enclaveRegistry(enclaveCount))
+        .value as TeerexPrimitivesEnclave;
```
Kailai-Wang, line 284: I think checking `scheduledEnclave` should be more accurate - enclave registration needs to comply with it too. Right now I can't think of any case where `scheduledEnclave` and `enclaveRegistry` don't match though

Btw even if we want to check `enclaveRegistry` we should `enclaveCount` from the `parachainBlockHash` height

Kailai-Wang (commented):
Nice and clean, not much to complain, thanks!

About the registry - this PR removes it completely. I'm thinking in the future, we might want to add "publish vc" feature where the user can link the selected VCs to their DID to make them public, so that anyone can see them by tracking down the DID. For that we'll need (again) a registry 🤔

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
kziemianek: But Teerex's `ScheduledEnclave` is empty so there is no value to compare against :(

kziemianek:
> Nice and clean, not much to complain, thanks!
> 
> About the registry - this PR removes it completely. I'm thinking in the future, we might want to add "publish vc" feature where the user can link the selected VCs to their DID to make them public, so that anyone can see them by tracking down the DID. For that we'll need (again) a registry 🤔

`VCRegistry` remains untouched - previously generated VCs can still be validated using it. It just can't be altered anymore because logic/calls are gone. I'm still waiting for decision on what we are going to do with old VCs.

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
Kailai-Wang: Ah in `dev` it's not populated, but in production it shouldn't be empty

Kailai-Wang: Maybe we should add the scheduled enclave even in `dev` - like here: https://github.com/litentry/litentry-parachain/pull/2382/files#diff-f941434353cbdda40f528800d59a89b42e5e51396b03c092bf41ef8e294d8fd5

kziemianek: Maybe we can always populate it with worker's mrenclave for first validateer in dev see: https://github.com/litentry/litentry-parachain/pull/2382#issuecomment-1892237305

kziemianek: @Kailai-Wang, @felixfaisal  what do you think?

jonalvarezz: fyi: the vc-sdk uses `scheduledEnclave`'s value

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -244,64 +244,72 @@ export async function assertIdGraphMutationResult(
...
+    // extrac proof and vc without proof json
     const vcPayloadJson = JSON.parse(decryptVcPayload);
     const { proof, ...vcWithoutProof } = vcPayloadJson;
-    assert.equal(vcIndex, vcPayloadJson.id, 'Check VcIndex error: VcIndex should be equal to vcPayload id');
 
     // step 5
-    const enclaveCount = await context.api.query.teerex.enclaveCount();
+    // prepare teerex enclave registry data for further checks
+    const parachainBlockHash = await context.api.query.system.blockHash(vcPayloadJson.parachainBlockNumber);
+    const apiAtVcIssuedBlock = await context.api.at(parachainBlockHash);
```
jonalvarezz, line 280: `parachainBlockNumber` is the block used to calculate the VC, and it won't match the VC's Issuance block. Is it likely that the registered enclave changes between those two instances? better said, should we account for such a case?

jonalvarezz (approved):
lgtm

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
BillyWooo: @Kailai-Wang @kziemianek I think we can always populate it for first validateer.

Kailai-Wang:
> > Nice and clean, not much to complain, thanks!
> > About the registry - this PR removes it completely. I'm thinking in the future, we might want to add "publish vc" feature where the user can link the selected VCs to their DID to make them public, so that anyone can see them by tracking down the DID. For that we'll need (again) a registry 🤔
> 
> `VCRegistry` remains untouched - previously generated VCs can still be validated using it. It just can't be altered anymore because logic/calls are gone. I'm still waiting for decision on what we are going to do with old VCs.

Yea I'm not talking about old VCs, but rather in the future - but OK we can always add it back whenever needed.

kziemianek:
I will delete `VCRegistry` storage from `VCManagement` pallet.

Kailai-Wang (approved):

BillyWooo (approved):

kziemianek:
Requested review again as there are new changes.

pallets/vc-management/src/schema.rs
```diff
@@ -17,12 +17,21 @@
...
 
-use crate::{vc_context::Status, Config};
+use crate::Config;
 use codec::{Decode, Encode, MaxEncodedLen};
 use core_primitives::{SchemaContentString, SchemaIdString};
 use scale_info::TypeInfo;
 use sp_std::vec::Vec;
 
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
+pub enum Status {
```
jonalvarezz, line 27: just wondering why we need to keep this one

jonalvarezz (approved):
set it on fire 🔥

pallets/vc-management/src/schema.rs
kziemianek: It's used by `VCSchema` struct defined below.

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
kziemianek: I think the time of VC's issuance is not important here - it doesn't matter what happens next after VC was generated as long as the VC was generated by legit enclave, but I would love to hear other opinions too.

pallets/vc-management/src/schema.rs
jonalvarezz: Right, but after removing the vc-registry, do we still need those statuses?

kziemianek: There were no plans of removing schema/schema's status so this enum cannot be deleted. You think we should remove it ?

jonalvarezz: Ah, I guess I'm missing something. If it is not part of the vc-registry then, sure, it is out of the scope 👍

kziemianek:
I'll update weights and merge it.

PR #2406, opened 2024-01-15
feat(vc): include jsonschema field

jonalvarezz:
Implements [P-385](https://linear.app/litentry/issue/P-385/include-jsonschema-field-on-assertions-to-link-to-their-json-schema)

Include a new `jsonschema` field on generated JSON VCs that point to their respective json-schema on https://github.com/litentry/vc-jsonschema

I took the approach of having everything in a single file, rather than adding the property on each assertion file, so that the `match` statement yells at us if a new assertion isn't handled in the jsonschema fn. Feedback to get better results is appreciated.

The tests for validating that the generated json complies with the schema are out of the scope of this pull request and will be tackled via [P-416](https://linear.app/litentry/issue/P-416/ensure-that-assertions-follow-the-json-schemas)

## TODO
* [x] Finish mapping assertions to their respective json-schema
* [x] Add the missing json-schemas to the vc-jsonschema repo: BAB Holder, CryptoSummary


tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
```diff
@@ -82,7 +83,8 @@ where
 			Assertion::A7(min_balance) =>
 				build_holding_time(&self.req, AmountHoldingTimeType::DOT, min_balance),
 
-			// no need to pass `networks` again because it's the same as the `get_supported_web3networks`
+			// no need to pass `networks` again because it's the same as the
+			// `get_supported_web3networks`
```
BillyWooo: do you have special config with your 'cargo fmt'? Looks like line length changed.

jonalvarezz: No I don't. I am using a VSCode extension that uses fmt to format on save; supposedly, it shouldn't be opinionated and read from our rustfmt.toml files. And yeah I noticed it breaks long lines.

Interestingly, if I put very very long one-line comment, and then run `cargo fmt` from `/tee-worker`, It doesn't break it. It should right?

jonalvarezz:
@BillyWooo @Kailai-Wang – this is ready to review now

tee-worker/litentry/core/credentials/src/jsonschema.rs
```diff
@@ -0,0 +1,123 @@
...
+			AchainableParams::AmountToken(_) =>
+				format!("{BASE_URL}/19-token-holding-amount/1-0-0.json"),
+
+			// Generates both contract-creator and token-holder credentials.
+			AchainableParams::Amount(_) => {
+				if credential_subject_type == String::from("Contract Creator") {
+					return format!("{BASE_URL}/12-contract-creator/1-0-0.json")
+				} else if credential_subject_type == String::from("Token Holder") {
+					return format!("{BASE_URL}/11-token-holder/1-0-0.json")
+				}
```
Kailai-Wang, line 65: Can we eliminate the string comparison? Like have **one** JSON file for `AchainableParams::Amount` and use subfields for categorization?

tee-worker/litentry/core/credentials/src/jsonschema.rs
```diff
@@ -0,0 +1,123 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::achainable;
+use litentry_primitives::{AchainableParams, Assertion};
+use std::string::String;
+
+const BASE_URL: &str = "https://raw.githubusercontent.com/litentry/vc-jsonschema/main/dist/schemas";
+const NOT_IMPLEMENTED: &str =
```
Kailai-Wang, line 22: I'm thinking maybe something like `FALLBACK_SCHEMA` would be better? 🤔

tee-worker/litentry/core/credentials/src/jsonschema.rs
```diff
@@ -0,0 +1,123 @@
...
+	"https://raw.githubusercontent.com/litentry/vc-jsonschema/main/dist/schemas/0-base.json";
+
+/// Returns the respective JSON Schema for the given assertion and its credential.
+/// JSON Schemas can be found on https://github.com/litentry/vc-jsonschema
+pub fn get_json_schema_url(assertion: Assertion, credential_subject_type: String) -> String {
+	match assertion {
+		Assertion::A1 => format!("{BASE_URL}/1-basic-identity-verification/1-0-0.json"),
+
+		Assertion::A2(_) => format!("{BASE_URL}/2-litentry-discord-member/1-0-0.json"),
```
Kailai-Wang, line 31: I'm generally not a fan of hard mapping :(

Can we use JSON-serialized `Assertion` enum or `IntoStaticStr` (See `Web3Network` definition)?

Kailai-Wang (commented):
Thank you! I left some comments

Kailai-Wang:
Btw do you think we need some follow-up tasks?

[1] in ts-test, verify if the generated VC matches the schema. I think we don't necessarily enforce this check in client-sdk: even if the schema doesn't match, it's still a valid VC and should throw out a warning at its maximum (it probably indicates that the code inside enclave needs to be updated)


[2] (more difficult), use the real schema as template to generate VC within enclave - if we implement this, [1] is guaranteed.

jonalvarezz:
> Btw do you think we need some follow-up tasks?
> 
> [1] in ts-test, verify if the generated VC matches the schema. I think we don't necessarily enforce this check in client-sdk: even if the schema doesn't match, it's still a valid VC and should throw out a warning at its maximum (it probably indicates that the code inside enclave needs to be updated)
> 
> [2] (more difficult), use the real schema as template to generate VC within enclave - if we implement this, [1] is guaranteed.

Totally, I wouldn't call JSON schemas completed until we make sure that whatever we generate actually follows the schema. I agree that clients don't need to validate it. You can find all the sub-tasks on the parent issue: https://linear.app/litentry/issue/P-142/enable-use-of-jsonschema-in-vc

tee-worker/litentry/core/credentials/src/jsonschema.rs
jonalvarezz: Thanks for the reference! will rework the code based on it

tee-worker/litentry/core/credentials/src/jsonschema.rs
jonalvarezz: I'm ok with that. I wanted to signal something that needs to be done and shouldn't be optional. 

Do you agree that any new VC should have a defined schema? Like, I don't want to leave the door open to situations where we put off the schema definition for the sake of delivering and later on forgetting about it or coming up with a complex schema as it wasn't thought before.

tee-worker/litentry/core/credentials/src/jsonschema.rs
jonalvarezz: Sure, do you have a reference I could use?

jonalvarezz:
I'm marking this pull request as closed to follow the suggestions above to avoid having it lingering around until I complete it. will re-open once done. 

Thanks!

Kailai-Wang:
Alternative you could turn it into draft : )

BillyWooo:
> I'm marking this pull request as closed to follow the suggestions above to avoid having it lingering around until I complete it. will re-open once done.
> 
> Thanks!

You can also convert it into "Draft PR". 😄

jonalvarezz:
elmo

tee-worker/litentry/core/credentials/src/jsonschema.rs
jonalvarezz: Wait, by one JSON you mean one schema? It will look a little off on the json-schema side due to the naming, but it should make this mapping easier.

jonalvarezz: Done in https://github.com/litentry/litentry-parachain/pull/2473

PR #2407, opened 2024-01-16
Fix build error while attesteer feature enabled

outofxxx:
build worker with command:

```shell
WORKER_FEATURES=attesteer make
```

Got error:
```shell
error[E0433]: failed to resolve: use of undeclared crate or module `prometheus_metrics`
   --> service/src/main_impl.rs:940:15
    |
940 |     let events = prometheus_metrics::fetch_marblerun_events(marblerun_base_url)
    |                  ^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `prometheus_metrics`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `litentry-worker` due to previous error
```


kziemianek (approved):

tee-worker/service/src/main_impl.rs
```diff
@@ -65,6 +65,9 @@ use substrate_api_client::{
 };
 use teerex_primitives::{Enclave as TeerexEnclave, ShardIdentifier};
 
+#[cfg(feature = "attesteer")]
+use crate::prometheus_metrics;
+
```
Kailai-Wang: IMO we should directly change
```
	let events = prometheus_metrics::fetch_marblerun_events(marblerun_base_url)
```

to 

```
	let events = crate::prometheus_metrics::fetch_marblerun_events(marblerun_base_url)
```

`mod prometheus_metrics` isn't guarded by feature, we shouldn't do that either in the derived file

Kailai-Wang (changes requested):
Although it's a simple change, I'm still being pedantic here

BillyWooo (approved):

Kailai-Wang (approved):

PR #2408, opened 2024-01-17
Register enclave with DCAP feature in SW mode

outofxxx:
build command

```shell
SGX_MODE=SW WORKER_FEATURES=dcap make
```


BillyWooo (approved):
LGTM

PR #2417, opened 2024-01-23
refactor of environment variables

BillyWooo:
as tile, a minor PR.
fixed an error within that rust file.


Kailai-Wang (approved):

tee-worker/litentry/core/data-providers/src/lib.rs
```diff
@@ -211,18 +211,18 @@ impl DataProviderConfig {
 			discord_auth_token: "".to_string(),
 			achainable_url: "https://label-production.graph.tdf-labs.io/".to_string(),
 			achainable_auth_key: "".to_string(),
-			credential_endpoint: "wss://tee-staging.litentry.io".to_string(),
+			credential_endpoint: "wss://rpc.rococo-parachain.litentry.io".to_string(),
```
kziemianek, line 214: This value will vary across environments but we cannot let it to be overriden by environment variable. I think we will need a compilation toggle for that.

Let's address it in another PR.

kziemianek (approved):

tee-worker/litentry/core/data-providers/src/lib.rs
BillyWooo: Yes. That would be a better approach.

PR #2418, opened 2024-01-23
P 419 deploy automation

BillyWooo:
> The major task is to provide a deployment script for production working mode. 
> - The whole deployment doesn't depend on source repo anymore. 
> - The script will 
>     - download all required release packages, 
>     - prepare the working env and 
>     - start the worker service. 
> 
> It requires some configs/keys files to be prepared and kept locally on the working server. 
> Try to explore the help message for more details.

Old description is deprecated. 
Auto deploy script moved into private repo. 
Here is only some minor change/fix.


tee-worker/scripts/litentry/release/prod_deploy.sh
```diff
@@ -0,0 +1,664 @@
...
+LATEST_FINALIZED_BLOCK=
+RELEASE_PKGDIR=
+TEMPLATE_DIR=
+TS_TESTS_DIR=
+TS_UTILS_DIR=
+
+SGX_SDK=/opt/intel/sgxsdk
+SGX_ENCLAVE_SIGNER=$SGX_SDK/bin/x64/sgx_sign
+
+VERSION_PATTERN="^p[0-9]+\.[0-9]+\.[0-9]+-[0-9]+-w[0-9]+\.[0-9]+\.[0-9]+-[0-9]+$"
```
Traf333: it seems that this pattern supports only versions with patch version like `p0.9.17-9172-w0.0.1-102
`, should we also support something like `p0.9.18-w0.0.2` in case we have major release?

BillyWooo: I would say we will have `p0.9.18-9180-w0.0.2-100`.

tee-worker/scripts/litentry/release/dev_deploy.sh
```diff
@@ -136,6 +125,31 @@ function main {
...
+  fi
+
+
+  # 1/ check if $USER has sudo
+  if sudo -l -U $USER 2>/dev/null | grep -q 'may run the following'; then
+    source "$SGX_SDK/environment"
+  else
+    echo "$USER doesn't have sudo permission"
+    exit 1
+  fi
```
outofxxx: We also can use this command to determine whether the current user is sudo
```shell
if [ $(id -u) -eq 0 ]; then
    echo "sudo"
else
    echo "Doesn't have sudo permission"
fi
```

tee-worker/scripts/litentry/release/dev_deploy.sh
```diff
@@ -136,6 +125,31 @@ function main {
...
+
+  # 2/ create folders if missing
+  sudo mkdir -p "$BASEDIR"
+  sudo chown -R $USER:$GROUPS "$BASEDIR"
+  for d in "$LOG_BACKUP_BASEDIR" "$WORKER_BACKUP_BASEDIR" "$RELAYCHAIN_ALICE_BASEDIR" "$RELAYCHAIN_BOB_BASEDIR" \
+    "$PARACHAIN_ALICE_BASEDIR" "$WORKER_BASEDIR"; do
+    mkdir -p "$d"
+  done
```
outofxxx: Another version:

```shell
folders=("folder1" "folder2" "folder3" "folder4")
for folder in "${folders[@]}"
do
    if [ ! -d "$folder" ]; then
        mkdir "$folder"
    else
        echo "exists"
    fi
done
```

tee-worker/scripts/litentry/release/dev_deploy.sh
```diff
@@ -353,11 +363,20 @@ function build {
...
+    # build worker
+    build_worker
   fi
 }
 
 function restart_services {
   sudo systemctl daemon-reload
+  restart_parachain_services
+  restart_worker_services
+  echo "Done"
```
outofxxx: echo "The service has been successfully restarted."

tee-worker/scripts/litentry/release/dev_deploy.sh
```diff
@@ -435,34 +455,38 @@ function register_parachain {
...
-    for ((i = 0; i < WORKER_COUNT; i++)); do
-      worker_dir="$WORKER_BASEDIR/w$i"
-      mkdir -p "$worker_dir"
-      for f in 'key.txt' 'spid.txt' 'enclave.signed.so' 'litentry-worker'; do
-        [ -f "$f" ] && cp -f "$f" "$worker_dir"
-      done
+  for ((i = 0; i < $WORKER_COUNT; i++)); do
+    worker_dir="$WORKER_BASEDIR/w$i"
+    mkdir -p "$worker_dir"
+    for f in 'key.txt' 'spid.txt' 'enclave.signed.so' 'litentry-worker'; do
```
outofxxx: Same as above, if there are more than two, put them into an array for iteration.

outofxxx (commented):
We have a `shellcheck` command, but I don't know if it has been executed before.

tee-worker/scripts/litentry/release/dev_deploy.sh
Kailai-Wang: That won't work though

root has uid 0, but non-zero uid doesn't mean it can't sudo. In fact, the code will always land in `else` branch unless executed with root.

tee-worker/scripts/litentry/release/prod_deploy.sh
```diff
@@ -0,0 +1,664 @@
...
+  echo "Restarting parachain services ..."
+
+  cp "$RELEASE_PKGDIR/polkadot" "$PARACHAIN_BASEDIR"
+  chmod a+x "$PARACHAIN_BASEDIR/polkadot"
+  cp "$RELEASE_PKGDIR/litentry-collator" "$PARACHAIN_BASEDIR"
+  chmod a+x "$PARACHAIN_BASEDIR/litentry-collator"
+
+  cd "$PARACHAIN_BASEDIR" || exit
+  ./polkadot build-spec --chain rococo-local --disable-default-bootnode --raw > rococo-local-chain-spec.json
+  ./litentry-collator export-genesis-state --chain $CHAIN-dev > genesis-state
```
felixfaisal: I had this question earlier, if this is prod server where we are starting a parachain from scratch, should this be `$CHAIN-dev` ? 
CC: @Kailai-Wang

BillyWooo: Thanks. I am refactor it, and try to break into pieces, and maybe move to another repo. It's better not give out our deployment process public.

tee-worker/local-setup/.env.dev
```diff
@@ -30,7 +30,7 @@ DISCORD_AUTH_TOKEN=
 ACHAINABLE_AUTH_KEY=
 ONEBLOCK_NOTION_KEY=
 NODEREAL_API_KEY=
-GENIIDATA_API_KEY=142cf1b0-1ca7-11ee-bb5e-9d74c2e854ac
+GENIIDATA_API_KEY=
```
Kailai-Wang, line 33: Will it affect any of our tests?
I remember @kziemianek mentioned this is the public key

tee-worker/local-setup/launch.py
```diff
@@ -156,10 +157,16 @@ def offset_port(offset):
 
 
 def setup_environment(offset, config, parachain_dir):
+    if not os.path.isfile("./local-setup/.env"):
+        shutil.copy("./local-setup/.env.dev", "./local-setup/.env")
+
```
Kailai-Wang, line 162: Is this needed?
I see the following line is 
```
load_dotenv("./local-setup/.env.dev")
```

Kailai-Wang (approved):
Looks fine! Some small comments

tee-worker/local-setup/.env.dev
BillyWooo: it won't affect any tests. The key is always overwritten.

tee-worker/local-setup/launch.py
BillyWooo: thanks. fixed.

PR #2313, opened 2023-12-06
Build match worker image

0xverin:
### Context

resolves p-330

Test steps:
- [x] Build the worker match parachain version's image and push it to Docker Hub.
- [x] Docker pull parachain and worker image and ensure it runs successfully.
- [x] Use the **release-ts-api-package** gha to generate the client API types.

PR #2423, opened 2024-01-24
Worker Docker release build without using cache

Traf333:
this PR is a follow up for #2313. At regular process using cache for worker build helps us to speed up daily workflow, but in case of preparing releases which happen not so often, we decided to avoid using cache for building docker images


.github/workflows/create-release-draft.yml
```diff
@@ -173,13 +173,9 @@ jobs:
...
-        with:
-          ref: ${{ env.RELEASE_TAG }}
-          fetch-depth: 0
-      - name: Set env
-        run: |
-          WORKER_DOCKER_TAG=$(echo ${{ env.RELEASE_TAG }} | sed 's/.*\(w.*\)/\1/;s/w/v/')
-          echo "WORKER_DOCKER_TAG=$WORKER_DOCKER_TAG" >> $GITHUB_ENV
+        # with:
+        # ref: ${{ env.RELEASE_TAG }}
+        # fetch-depth: 0
```
Traf333: I'll return it back once build release works fine

Traf333: https://github.com/litentry/litentry-parachain/actions/runs/7669315590
looks like it works fine

tee-worker/build.Dockerfile
```diff
@@ -47,17 +47,25 @@ ENV WORKER_MODE=$WORKER_MODE_ARG
...
+
 ARG FINGERPRINT=none
 
 WORKDIR $HOME/tee-worker
 COPY . $HOME
 
 RUN \
+  if [ "$IMAGE_FOR_RELEASE" = "true" ]; then \
+  echo "Omit cache for release image"; \
+  make; \
```
Kailai-Wang: Ideally with proper indention :)

.github/workflows/create-release-draft.yml
Kailai-Wang: Hmm shall we revert the comment?

.github/workflows/create-release-draft.yml
```diff
@@ -207,6 +208,7 @@ jobs:
           build-args: |
             WORKER_MODE_ARG=sidechain
             ADDITIONAL_FEATURES_ARG=
+            IMAGE_FOR_RELEASE=true
```
Kailai-Wang, line 211: Acutally we don't even need to build `local-builder:latest` here

Kailai-Wang (commented):
Thank you!

I personally prefer a separate Dockerfile where we don't need everything in local-builder (e.g. sccache) , but I'm fine with this version too :D

.github/workflows/create-release-draft.yml
Traf333: I was thinking that we need it for the next step when we build `worker` and `cli` separately

Traf333:
> I personally prefer a separate Dockerfile where we don't need everything in local-builder (e.g. sccache) , but I'm fine with this version too

 @Kailai-Wang , that's valid point and I was thinking of supporting two version of docker file over one with just minor "if" statement, but than I realised that later we may introduce new changes like including `extract_identity` binary or `ts-tests` folder to register worker to the docker image, which would require further modification. This thoughts lean me towards having single file for now

.github/workflows/create-release-draft.yml
Kailai-Wang: Oh yes I was thinking of using another Dockerfile, sorry

.github/workflows/create-release-draft.yml
```diff
@@ -174,12 +174,8 @@ jobs:
...
         uses: actions/checkout@v4
         with:
-          ref: ${{ env.RELEASE_TAG }}
-          fetch-depth: 0
-      - name: Set env
-        run: |
-          WORKER_DOCKER_TAG=$(echo ${{ env.RELEASE_TAG }} | sed 's/.*\(w.*\)/\1/;s/w/v/')
-          echo "WORKER_DOCKER_TAG=$WORKER_DOCKER_TAG" >> $GITHUB_ENV
+        ref: ${{ env.RELEASE_TAG }}
+        fetch-depth: 0
```
Kailai-Wang: The indention still doesn't seem right

Kailai-Wang (approved):
Looks good - a small indention problem

PR #2427, opened 2024-01-25
feat: pallet_account_fix

wangminqi:
A bug has recently been reported by the community, indicating a potential account reset issue under specific circumstances arising from an inconsistency between Polkadot's pallet balance and Litentry's staking mechanism.

In instances where a user has participated in staking while vesting / engaging in a crowdloan, and subsequently reduces their account balance to less than the equivalent of 0.1 LIT, the account will experience a reset.

It is important to note that triggering this bug does not result in permanent damage, but it does render the user unable to receive tokens through unstaking or staking less (a successful unstake event with unreserve 0 event)

The PR is for providing a temporary pallet with two methods for quick fixing.

One method upgrade_accounts for increasing user account with consumer + 1. (So account can not be reaped)
The other method set_balance for adding user balance (reserve and free) since pallet_balances depreciate the old method. (For recovering user data)

The pallet added is only temporary and should be removed once available.


Kailai-Wang (commented):
As mentioned, please:
- bump the litentry runtime version
- adjust the hardcoded weights

Kailai-Wang (approved):
Thanks!

PR #2430, opened 2024-01-26
Fix assertion identifiers missing the `$` prefix

outofxxx:
as topic.

**Notice that such fix needs a sync release with the client's vc-sdk  package to continuing supporting both the previous and the new claim key.**


Kailai-Wang (approved):
Thank you!

cc @jonalvarezz

jonalvarezz (approved):
> Notice that such fix needs a sync release with the client's vc-sdk package to continuing supporting both the previous and the new claim key.

Could you please create the follow-up ticket and put it on the backlog?

BillyWooo:
@jonalvarezz do we need any update from sdk point of view?

jonalvarezz:
yes totally, as well as duplicate JSON Schemas so that both versions are valid. The price of the rush 😄

kziemianek (commented):
Is it mandatory for assertion item to be prefixed with `$` ? 
If yes can we add such check and fail if condition is not met ?

It's always good to express such rules in code.

jonalvarezz:
> Is it mandatory for assertion item to be prefixed with `$` ? If yes can we add such check and fail if condition is not met ?
> 
> It's always good to express such rules in code.

I second the lint checks to enforce consitency 👍

PR #2431, opened 2024-01-26
Data providers tests for vip3

0xverin:
### Context

resolves p-366

Usage example:

- If you want to test a specific VC(`vip3-membership-card-gold`), you can simply input its ID on the command line, like: `pnpm --filter integration-tests run test-data-provider:local --id=vip3-membership-card-gold`.(you can find the ID [in vc jsonfile](https://github.com/litentry/litentry-parachain/pull/2431/files#diff-586793e6df0eb439bd5953ede84ddf21b4e6968d65a3b1d70b2bd04985e90294R3)

- Testing parameters(i.e. mockAddress or expectedCredentialValue) can be directly modified in the *.json files.
I feel this approach is relatively maintainable. If it seems reasonable, I'll add the others as well. This PR currently provides four.

If you have a better solution, please feel free to share it at any time.


tee-worker/ts-tests/integration-tests/data-providers.test.ts
```diff
@@ -0,0 +1,111 @@
...
+                console.log(`Error: ${error.stderr}`);
+                throw error;
+            }
+
+            const events = (await eventsPromise).map(({ event }) => event);
+            assert.equal(events.length, 1);
+        });
+    } else {
+        Object.keys(credentialDefinitionMap).forEach((id) => {
+            step(`linking identity-${credentialDefinitionMap[id].mockAddress} via cli`, async function () {
```
kziemianek: Can you extract and reuse this function ? I think it's exactly the same to one from line 43.

kziemianek (commented):
Could you explain how and why we are testing data provieders through typescript tests ?
Or maybe we are just testing request vc cli commands ? 🙃 
I think there should be no thing like dataproviders in ts code, it introduces unnecessary coupling.

0xverin:
>Could you explain how and why we are testing data provieders through typescript tests ?

We indeed need real accounts and a real endpoint to test the value of VCs. To be precise, we are assisting IdHub in identifying issues with VCs—for example, cases where an real account should meet the conditions for generating a credential, but IdHub receives a false value. This has occurred before.

How: We might need the CI to use the data-provider's key (expect VIP3).

>Or maybe we are just testing request vc cli commands ? 

CLI help creating an ID graph with a real account. It ensures that Alice meets the conditions for being an ETH holder, and when requesting a VC from the data-provider, it should return a value of true.

cc @kziemianek

Kailai-Wang (commented):
I haven't read thoroughly, but I feel it's a bit overcomplicated in general.

Can't we have some `testdata.json` which includes:
```
{
  [
     "idgraph": [
       "did1",
       "did2",
     ],
     "assertion": "a1",
     "expectedValue": true
   ],
   [
   ],
   ...
}
     
```

The script just iterates through the testdata, construct idgraph and request vc to see if we get the expected result.

Isn't that more clear and simple?
Just my naive thought - I could be wrong

0xverin:
> I haven't read thoroughly, but I feel it's a bit overcomplicated in general.
> 
> Can't we have some `testdata.json` which includes:
> 
> ```
> {
>   [
>      "idgraph": [
>        "did1",
>        "did2",
>      ],
>      "assertion": "a1",
>      "expectedValue": true
>    ],
>    [
>    ],
>    ...
> }
>      
> ```
> 
> The script just iterates through the testdata, construct idgraph and request vc to see if we get the expected result.
> 
> Isn't that more clear and simple? Just my naive thought - I could be wrong

I can modify it into a JSON file; the effect will be the same.

tee-worker/ts-tests/integration-tests/package.json
```diff
@@ -18,7 +18,9 @@
...
         "test-ii-vc:staging": "pnpm exec eslint . && pnpm exec cross-env NODE_ENV=staging mocha --exit --sort -r ts-node/register --loader=ts-node/esm 'ii_vc.test.ts'",
         "test-ii-batch:local": "pnpm exec eslint . && pnpm exec cross-env NODE_ENV=local mocha --exit --sort -r ts-node/register --loader=ts-node/esm 'ii_batch.test.ts'",
-        "test-ii-batch:staging": "pnpm exec eslint . && pnpm exec cross-env NODE_ENV=staging mocha --exit --sort -r ts-node/register --loader=ts-node/esm 'ii_batch.test.ts'"
+        "test-ii-batch:staging": "pnpm exec eslint . && pnpm exec cross-env NODE_ENV=staging mocha --exit --sort -r ts-node/register --loader=ts-node/esm 'ii_batch.test.ts'",
+        "test-data-providers:local": "pnpm exec eslint . && pnpm exec cross-env NODE_ENV=local mocha --exit --sort -r ts-node/register --loader=ts-node/esm 'data-providers.test.ts'",
+        "test-data-providers:staging": "pnpm exec eslint . && pnpm exec cross-env NODE_ENV=staging mocha --exit --sort -r ts-node/register --loader=ts-node/esm 'data-providers.test.ts'"
```
Traf333, line 23: it seems like our package will grow constantly with new functionality. I'm thinking that there is a space for refactoring. I'd like to see something like below and in addition the ENV variables will be resolved out of the package.json file. What do you think? Probably might be resolved as one of next tasks

```
"scripts": {
  "test": "mocha run with some arguments"
  "pretest": "pnpm exec eslint"
}
```

tee-worker/ts-tests/integration-tests/data-providers.test.ts
```diff
@@ -0,0 +1,140 @@
...
+            console.log(`Exit code: ${error.exitCode}`);
+            console.log(`Error: ${error.stderr}`);
+            throw error;
+        }
+
+        const events = (await eventsPromise).map(({ event }) => event);
+        assert.equal(events.length, 1);
+    }
+
+    async function requestVc(id: string, index: number) {
```
Traf333, line 80: is it possible to utilise this function for other scenarios, like NFT holder or BNB domain?

tee-worker/ts-tests/integration-tests/data-providers.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    const argv = process.argv.indexOf('--id');
+    argvId = process.argv[argv + 1];
+    const {
+        protocol: workerProtocal,
+        hostname: workerHostname,
+        port: workerPort,
+    } = new URL(process.env.WORKER_ENDPOINT!);
+    const { protocol: nodeProtocal, hostname: nodeHostname, port: nodePort } = new URL(process.env.NODE_ENDPOINT!);
+
+    async function linkIdentityViaCli(id: string) {
```
Traf333, line 50: is it possible to utilise this function for other VC scenarios, like NFT holder or BNB domain?

tee-worker/ts-tests/integration-tests/data-providers.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        const credentialDefinitions = vip3CredentialJson.find((item) => item.id === id) as CredentialDefinition;
+
+        const keyringPair = randomSubstrateWallet();
+        keyringPairs.push(keyringPair);
+        const formatAddress = u8aToHex(keyringPair.publicKey);
+
+        const substrateIdentity = (await buildIdentityFromKeypair(
+            new PolkadotSigner(keyringPair),
+            context
+        )) as CorePrimitivesIdentity;
```
Traf333: it seems that `buildIdentityFromKeypair` also says that it returns `CorePrimitivesIdentity` where function is declared. Does it mean, types don't work the regular way?

tee-worker/ts-tests/integration-tests/data-providers.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        const decryptVcPayload = decryptWithAes(aesKey, vcResults.vc_payload, 'utf-8').replace('0x', '');
+        const vcPayloadJson = JSON.parse(decryptVcPayload);
+
+        assert.equal(vcPayloadJson.credentialSubject.values[0], credentialDefinitions.expectedCredentialValue);
+    }
+
+    if (argvId) {
+        const credentialDefinitions = vip3CredentialJson.find((item) => item.id === argvId) as CredentialDefinition;
+        if (credentialDefinitions) {
+            const credentialDefinitions = vip3CredentialJson.find((item) => item.id === argvId) as CredentialDefinition;
```
Traf333: I've read this lines couple of times and it seems there are some either duplication or wrong naming. We should not be able to redefine const variable 🤔 @Verin1005 can you check it

Traf333 (commented):
Checked only source code and later will try to run everything locally. Put some comments.
@Verin1005 let me know what do you think about it?

tee-worker/ts-tests/integration-tests/package.json
0xverin: We can create an issue for optimization; I think it's necessary as the number of scripts is increasing.

tee-worker/ts-tests/integration-tests/data-providers.test.ts
0xverin: Thanks. Removed!

tee-worker/ts-tests/integration-tests/data-providers.test.ts
0xverin: Thanks. Removed!

tee-worker/ts-tests/integration-tests/data-providers.test.ts
0xverin: You're referring to other cases, right? Currently, this pr is for VIP3, but I am testing [the other cases](https://github.com/litentry/litentry-parachain/pull/2434). To avoid ambiguity, I am redefining credentials array [here](https://github.com/litentry/litentry-parachain/pull/2431/files#diff-29be9a24deec6c20fbd8de209e73ae83c14eaa95a86baeb38dfabb9161c8f3b2R30).

Verin1005: You're referring to other cases, right? Currently, this one is for VIP3, but I am testing [the other cases](https://github.com/litentry/litentry-parachain/pull/2434).

Traf333 (approved):
Looks good to me. Thanks @Verin1005 for initiating this big topic as it will definitely save us time for a long run

tee-worker/ts-tests/integration-tests/package.json
Traf333: Created a task for that https://linear.app/litentry/issue/P-463/refactor-integration-tests-script

PR #2433, opened 2024-01-26
remove parachain dependency on worker `itp-utils`

kziemianek:
Removes usage of `itp-utils` tee-worker crate from parachain code.


Kailai-Wang (approved):
I thought of this change too, just had the concern that we wasn't able to catch the upstream change in hex module, but there don't seem to be that many changes :D

PR #2435, opened 2024-01-29
Bump download-artifact version

Traf333:
### Context

we got several warnings within CI processes which related to Node16 version within:
- `actions/download-artifact@v3`
- `actions/upload-artifact@v3`
- `dorny/paths-filter@v2`

As a solution we just bump the version

### DIFF
--- a/.github/workflows/build-docker-with-args.yml
+++ b/.github/workflows/build-docker-with-args.yml
@@ -5,24 +5,24 @@ on:
     inputs:
       ref:
         description: The commit SHA or tag for checking out code
-        default: ''
+        default: ""
         required: false
       cargo_profile:
         type: choice
         description: which profile to use in cargo
         options:
-        - release
-        - production
+          - release
+          - production
       docker_tag:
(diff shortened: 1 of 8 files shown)


kziemianek:
Can we bump version [here](https://github.com/litentry/litentry-parachain/blob/dev/.nvmrc) too ?
So we will use same node version in development environment.
I remember there were problems with building `client-api` using node 20, but it looks like they are gone.

Kailai-Wang (approved):
Thank you - any good yml formatter? If so we can add it to CI/hooks to make sure everyone is following the same format

Traf333:
The full CI job is successful https://github.com/litentry/litentry-parachain/actions/runs/7697026419
observed that we also need to bump `actions/cache@v4` as well

Traf333:
> Thank you - any good yml formatter? If so we can add it to CI/hooks to make sure everyone is following the same format

Not really seen any styleguides for yml. I thought it's just 2 spaces indent and double quotes over single quotes

PR #2436, opened 2024-01-29
fix: bug in request-vc-direct

felixfaisal:
I think the issue description, describes the issue well and I've implemented the solution suggested by @kziemianek


Kailai-Wang (approved):
Looks fine at first glance -  can you confirm it? @kziemianek

kziemianek:
Looks fine, thanks.
I've tested A1 using `request-vc-direct` and `request-vc` commands. Both resulted in VC generation.

kziemianek (approved):
Looks fine, thanks! I've tested vc generation using `request-vc-direct` and `request-vc` commands. Both resulted in VC generation.

PR #2438, opened 2024-01-29
include only one variant in compiled code

kziemianek:
After this change only one expression will be present in compiled code, also there will be no if statement anymore. 

Previous version was failing in this case:
```
#[cfg(feature = "production")]
fn print_prod() {
	info!("Prod")
}

#[cfg(not(feature = "production"))]
fn print_dev() {
	info!("Dev")
}

if_production_or!(
	print_prod(),
	print_dev()
);
```

```
105 |         print_dev()
    |         ^^^^^^^^^ not found in this scope
```


Kailai-Wang (approved):
Looks good!

PR #2439, opened 2024-01-29
Fix `production` feature propagation

Kailai-Wang:
### Context

It's not a flawless PR due to time limit. I only make sure critical part is covered.

To solve it perfectly we need to add the feature propagation in each intermediate crate (including the **node/runtime**), and/or have a smarter way to deal with the conditional compilation.

### DIFF
--- a/primitives/core/Cargo.toml
+++ b/primitives/core/Cargo.toml
@@ -38,3 +38,6 @@ std = [
     "ring/std",
     "pallet-evm/std",
 ]
+production = [
+    "litentry-macros/production",
+]
--- a/primitives/core/macros/Cargo.toml
+++ b/primitives/core/macros/Cargo.toml
@@ -4,3 +4,6 @@ description = 'Proc-macros used by Litentry crates.'
 name = "litentry-macros"
 version = "0.9.12"
 edition = "2021"
+
+[features]
+production = []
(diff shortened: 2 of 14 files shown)


kziemianek:
Compilation doesn't fail if I put this 

```
match assertion {
			// the signer will be checked inside A13, as we don't seem to have access to ocall_api here
			Assertion::A13(_) => (),
			_ => if_production_or!(
				compile_error!("This macro should fail"),
				ensure!(
					ensure_self(&signer, &who) || ensure_alice(&signer_account),
					"Unauthorized signer",
				)
			),
		}
```

in `litentry-parachain/tee-worker/litentry/core/vc-issuance/lc-vc-task-receiver/src/lib.rs` - which indicates that production feature was not applied.
Compilation fails if I add this `compile_error` macro invocation to `getter.rs`

Command:
```
make SGX_MODE=SW
```

Do you know why ?

Kailai-Wang:
You didn't compile with `SGX_PRODUCTION=1`, so no production feature, thus no compile error - guess that's expected?

kziemianek (approved):
Looks ok, thanks.
Checked if it works using `compile_error` macro.

kziemianek:
> You didn't compile with `SGX_PRODUCTION=1`, so no production feature, thus no compile error - guess that's expected?

Yes, I was experimenting with modified `Makefile` and added `production` feature to `WORKER_FEATURES` . I forgot about this compilation option.

BillyWooo (approved):
👍

PR #2445, opened 2024-01-29
run `stf-task-handler` thread only for `sidechain`

kziemianek:


Kailai-Wang (approved):
TBH `run_stf_task_handler` does need a sidechain while `run_vc_issuance` doesn't. It works without sidechain. But I've heard there was some init order problem so we put it into sidechain initialization, which is fine.

One thing to notice is that in bitacross worker (where there's no sidechain) we'll need to have a similar task thread :D

felixfaisal (approved):
LGTM

PR #2447, opened 2024-01-30
remove unused log target

BillyWooo:
as title. a minor PR.


Kailai-Wang:
Hmm could you please explain the background of this PR a bit?
I saw such log which seemed to be fine:
```
^[[0m^[[38;5;8m[^[[0m2024-01-30T12:20:06Z ^[[0m^[[32mINFO ^[[0m aura^[[0m^[[38;5;8m]^[[0m Claiming slot (284436201)
```

kziemianek (approved):
It looks fine.

In production build we allow only certain targets to be logged.
Unfortunately, custom targets are not on this list - the list is built based on crates names.
Locally with nonprod build the entries may appear @Kailai-Wang .

see: https://github.com/litentry/litentry-parachain/blob/d275492da2622b394826a8117a3bd77bb724f2a3/tee-worker/enclave-runtime/src/lib.rs#L143-L146

Kailai-Wang:
I see, thanks!

PR #2450, opened 2024-01-30
Adds pallet-account-fix to rococo

Kailai-Wang:
### Context

As topic. This PR:
- adds `pallet-account-fix` to rococo
- adjusts call filter
- bumps versions


wangminqi (approved):
LGTM

PR #2454, opened 2024-01-31
Ethereum and Bitcoin key repository

kziemianek:
Adds `secp256k1` key repositories for managing ethereum and bitcoin keys. 
The key is sealed using protected filesystem api and stored on disk so it can be reused on next startup.


BillyWooo (approved):
LGTM

Kailai-Wang (approved):
Looks good! So we are using the same seed for btc and eth keys, aren't we (just to confirm

kziemianek:
> Looks good! So we are using the same seed for btc and eth keys, aren't we (just to confirm

The seed is initialized twice (one per `create_secp256k1_repository` call) using two different RNG instances.
This results in different keys for ethereum and bitcoin repository.

PR #2456, opened 2024-01-31
bitacross remove teeracle feature

kziemianek:
Removes `teeracle` feature from `bitacross-worker`. 

@Kailai-Wang, @felixfaisal, @wangminqi  if it somehow creates conflicts with your current work, this PR can wait.


bitacross-worker/.github/workflows/build_and_test.yml
Kailai-Wang, line 1: We can delete the whole `./github` folder, GHA is always executed from top level

Kailai-Wang (approved):
Looks good from my side

felixfaisal (approved):
I don't think I'm touching any of these files, Looks good from my side

PR #2459, opened 2024-02-01
Refactor teerex/sidechain pallet into teebag

Kailai-Wang:
### Context

This PR uses a new pallet `teebag` to replace teerex + sidechain pallet.

It uses the name `teebag` because enclave is like a teabag and the pallet is used to manage the teabag.

Additionally:
- it removes the feature `skip-ias-verify` and changes it to runtime control
- it merges the original teerex and sidechain pallets (version before MS-RSL licence) into one and rewrite/update it - it still has a mixed licence header tho, especially for the parts that I hardly made any changes
- it adds `worker_type` to allow multiple worker types
- it removes heartbeat for now - we'll need to think about how to deal with it, plus we have the enclave removal extrinsic
- it revamps the `Enclave` structure, the map uses the T::AccountId as the key, because that's (the only identifier?) guaranteed to be different for each worker and has a native parachain connection
- it moves `sgx-verify` from crate to mod
- it unifies the tests with and without ias-verify

Some follow-up TODOs - I'll create issues:
- let workers connect to pallet-teebag (while keeping teerex for state reference)
- benchmarking / weights
- remove teerex/sidechain pallet usage in runtime and worker
- runtime upgrade to include the changes, including state migration to clear the teerex state

But before I continue I'd like to see reviews about this refactoring


felixfaisal (approved):
Thanks @Kailai-Wang ,In terms of the refactor itself, I don't see any concerns, I don't know much about the attestation process, So from my side it looks good.

pallets/teebag/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+	// keep track of a list of scheduled/allowed enchalves, mainly used for enclave updates,
+	// can only be modified by AdminOrigin
+	// (worker_type, sidechain_block_number) -> expected MrEnclave
+	//
+	// about the first time enclave registration:
+	// prior to `register_enclave` this map needs to be populated with ((worker_type, 0),
+	// expected-mrenclave), otherwise the registration will fail
+	//
+	// For NON-sidechain worker_type, we still use this storage to whitelist mrenclave, in this case
+	// the `SidechainBlockNumber` is ignored - you could always set it to 0.
```
kziemianek, line 218: Maybe we could use parachain block number so we can schedule it also for non sidechain worker types ?

pallets/teebag/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+			Ok(Pays::No.into())
+		}
+
+		#[pallet::call_index(5)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal, Pays::No))]
+		pub fn force_remove_enclave_by_mrenclave(
+			origin: OriginFor<T>,
+			mrenclave: MrEnclave,
+		) -> DispatchResultWithPostInfo {
+			Self::ensure_admin_or_root(origin)?;
```
kziemianek, line 352: I assume they can register again, is it ok ?

pallets/teebag/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+		#[pallet::call_index(7)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal, Pays::No))]
+		pub fn set_scheduled_enclave(
+			origin: OriginFor<T>,
+			worker_type: WorkerType,
+			sidechain_block_number: SidechainBlockNumber,
+			mrenclave: MrEnclave,
+		) -> DispatchResultWithPostInfo {
+			Self::ensure_admin_or_root(origin)?;
+			ensure!(worker_type.is_sidechain(), Error::<T>::ScheduledEnclaveNotApplicable,);
```
kziemianek: I thought this should allow to set mrenclave for other worker types too.

	// For NON-sidechain worker_type, we still use this storage to whitelist mrenclave, in this case
	// the `SidechainBlockNumber` is ignored - you could always set it to 0.

pallets/teebag/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+	//
+	// For NON-sidechain worker_type, we still use this storage to whitelist mrenclave, in this case
+	// the `SidechainBlockNumber` is ignored - you could always set it to 0.
+	//
+	// Theorectically we could always push the enclave in `register_enclave`, but we want to
+	// limit the mrenclave that can be registered, as the parachain is supposed to process enclaves
+	// with specific mrenclaves.
+	#[pallet::storage]
+	#[pallet::getter(fn scheduled_enclave)]
+	pub type ScheduledEnclave<T: Config> =
```
kziemianek, line 225: I think it doesn't scale well.
We cannot have bitacross and identity worker both running in sidechain mode, can we ?
Each worker should have an identifier (`identity`, `bitacross`) and (WorkerIdentifier, SidechainBlockNumber) should be used as a key here. what do you think?

pallets/teebag/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+					enclave.sgx_build_mode = report.build_mode;
+				},
+			};
+
+			match Self::mode() {
+				OperationalMode::Production | OperationalMode::Maintenance => {
+					if !Self::allow_sgx_debug_mode() &&
+						enclave.sgx_build_mode == SgxBuildMode::Debug
+					{
+						return Err(Error::<T>::InvalidSgxMode.into())
```
kziemianek, line 487: Maybe it should always return error if `OperationalMode::Production && enclave.sgx_build_mode == SgxBuildMode::Debug` ?

pallets/teebag/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+					}
+					// TODO: shall we take the sidechain number into consideration?
+					ensure!(
```
kziemianek, line 489: I would keep it as it is if we really need this check.

pallets/teebag/src/tests.rs
```diff
@@ -0,0 +1,414 @@
...
+//
+// Litentry is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+// TODO: add `sidechain_block_imported` tests
```
kziemianek, line 17: Seems like it requires follow up issue too.

pallets/teebag/src/tests.rs
```diff
@@ -0,0 +1,414 @@
...
+
+#![allow(dead_code, unused_imports)]
+use crate::{
```
kziemianek, line 20: We should increase test coverage of teebag pallet, some calls are not covered, example: `force_remove_enclave_by_mrenclave`

pallets/teebag/src/lib.rs
Kailai-Wang: With "scale well" do you mean we should have more storages to facilitate the lookup?

> Each worker should have an identifier (identity, bitacross)

It's there as a field in `Enclave` struct

> (WorkerIdentifier, SidechainBlockNumber)

"WorkerIdentifier" is `T::AccountId` here, what exactly does `SidechainBlockNumber` do?

> We cannot have bitacross and identity worker both running in sidechain mode, can we ?

Not now, but maybe later (just in case they want to manage some common state).
Either way we need features of a) mrenclave whitelists b) for sidechain workers, schedule mrenclave for upgrade

pallets/teebag/src/lib.rs
Kailai-Wang: They can - do you want to blacklist them?

If so we could adjust the `ScheduledEnclave` entry and/or the `MaxEnclaveCount`

pallets/teebag/src/lib.rs
Kailai-Wang: Oh yes - it is, sorry.

pallets/teebag/src/lib.rs
Kailai-Wang: This is a tricky situation.

This mode exists when you want to test everything including RA report, but using an Intel dev endpoint - it's still a legit RA report, so you could validate all your verification logic. 

We ourselves have such situation on staging where the worker is built with `SGX_PRODUCTION`, parachain without `tee-dev`, so we expect a valid RA report. But, the report comes from a dev endpoint of Intel, which implies `SgxBuildMode == Debug`.

Upstream has this storage `AllowSgxDebugMode` too, I assume they had similar scenario.

pallets/teebag/src/lib.rs
kziemianek: By scalability I mean we cannot run both bitacross and identity worker in sidechain mode with this implementation. Just pointing this out :) but maybe that's fine for the first version.


> "WorkerIdentifier" is T::AccountId here, what exactly does SidechainBlockNumber do?

`SidechainBlockNumber` indicates sidechain block from which new mrenclave is required. I just replaced `WorkerType` with `WorkerIdentifier` in the tuple. `WorkerIdentifier` could be `identity` | `bitacross`

pallets/teebag/src/lib.rs
kziemianek: No, just thinking about the purpose of these calls. 
If the purpose is just sweeping the dust, then the mess can easily come back again - even with blacklisting I think.

pallets/teebag/src/lib.rs
Kailai-Wang: This is a good thought. I'm thinking:
1. traditionally it's used to control the sidechain block authoring, so it's natural to activate/deactivate based on sidechain block number. It also means you could still synchronize parachain (so to synchronize any updates).
2. for non-sidechain workers it's merely used as whitelist, so which chain number doesn't really matter.

It's not supposed to be used as the "only" guard against non-authorized mrenclave, we use MRSIGNER for it (for now). In other words, if anyone cracks MRSIGNER, he could change the enclave logic freely, so any check within enclave makes little sense.

pallets/teebag/src/tests.rs
Kailai-Wang: That's right - I wanted to implement/test the most important features to set the ball rolling

pallets/teebag/src/lib.rs
Kailai-Wang: Updated!

pallets/teebag/src/lib.rs
Kailai-Wang: Yea I think it will have similar effect.

If the sidechain block number doesn't match, it won't produce blocks.
It's just if we should allow it to register at all

I removed the comment

kziemianek: I think it's still not removed 😂

Kailai-Wang: Yes I didn't push it - now it's there :)

kziemianek (approved):

PR #2461, opened 2024-02-02
fix: add generating config.json

felixfaisal:
So we need to generate `config json` because it is then used by the `ts-tests` to register the parachain and needs proper path to `genesis-wasm` etc. This is need to run the local-setup of `bitacross` worker.

### DIFF
--- a/bitacross-worker/local-setup/launch.py
+++ b/bitacross-worker/local-setup/launch.py
@@ -102,6 +102,26 @@ def check_all_ports_and_reallocate():
             reallocate_ports(x, os.environ.get(x))
 
     print("All preliminary port checks completed")
+    
+# Generate `config.local.json` used by parachain ts utils
+def generate_config_local_json(parachain_dir):
+    data = {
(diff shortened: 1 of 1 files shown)


kziemianek:
How is that file used in bitacross `ts-tests` ? I tried to run bitacross locally with this command `./local-setup/launch.py --config ./local-setup/development-worker.json --parachain local-binary-standalone` and it seems like it works fine.

felixfaisal:
Standalone doesn't have to register Parachain, so it should work fine regardless. TS-Tests also would work because of default values in config.

Kailai-Wang (approved):
Make sense to me.
So it was removed when we init'ed the bitacross folder, right?

kziemianek:
Yes, I removed this function from script because it is described as :
`# Generate `config.local.json` used by parachain ts utils` 

and there are no parachain ts utils in bitacross ts-test folder.

kziemianek (approved):

PR #2462, opened 2024-02-02
add litentry archive service mock

kziemianek:
Adds litentry archive service mock which is used in A20.
We should use this mock in CI.


Kailai-Wang (approved):
Looks good, thanks for fixing CI

PR #2463, opened 2024-02-02
local dev env config

kziemianek:
For dev purposes `tee-worker` should be integrating with mock services.
Currently if we start local dev env without any config changes ts-test won't pass.

We overlooked this change in one of recent PR.


Kailai-Wang (approved):
Indeed

cc @BillyWooo in case you see any problem

Kailai-Wang:
Btw I notice we have different file structures between bitacross and normal worker: `tee-worker/local-setup/.env.dev` vs `bitacross-worker/.env.dev`

kziemianek:
Yes, I started working on bitacross before #2417 was merged so few changes may be missing.

BillyWooo:
> Indeed
> 
> cc @BillyWooo in case you see any problem

No problem. We discussed together. 😄

PR #2469, opened 2024-02-06
Adjust tee-worker counterpart of pallet-teebag

### LINKED ISSUE #1133: Review the enclave registration process

Kailai-Wang:
### Context

Only relevant extrinsics are kept - (so no `shield_funds` and `unshield_funs`), enough refactoring.

Please review (particularly) the pallet-teebag refactoring and node-api for pallet-teebag storage API access

It's a preparation of P-487 (actually, some code is already removed to make it compile).


Kailai-Wang:
cc @jonalvarezz it will be a breaking change for client (SDK) as the pallet/extrinsic interface has changed

pallets/teebag/src/lib.rs
```diff
@@ -145,14 +150,8 @@ pub mod pallet {
...
 		CollateralInvalid,
-		/// The number of `extra_topics` passed to `publish_hash` exceeds the limit.
-		TooManyTopics,
-		/// The length of the `data` passed to `publish_hash` exceeds the limit.
-		DataTooLong,
-		/// max_enclave_count overflows
-		MaxEnclaveCountOverflow,
-		/// max_enclave_count underflows (than 0 or currently registered enclave count)
-		MaxEnclaveCountUnderflow,
+		/// MaxEnclaveIdentifierO overflows
```
kziemianek: `MaxEnclaveIdentifierO` -> `MaxEnclaveIdentifier`

pallets/teebag/src/lib.rs
```diff
@@ -640,70 +613,64 @@ impl<T: Config> Pallet<T> {
...
-	pub fn add_enclave(sender: &T::AccountId, enclave: &Enclave) -> DispatchResultWithPostInfo {
+	pub fn add_enclave(sender: &T::AccountId, enclave: &Enclave) -> Result<(), Error<T>> {
 		match EnclaveRegistry::<T>::get(sender) {
 			Some(old_enclave) => {
 				if old_enclave.worker_type != enclave.worker_type {
-					// a tricky situation - we are re-registering the enclave with a different
-					// worker type
-					Self::decrement_count(old_enclave.worker_type)?;
-					Self::increment_count(enclave.worker_type)?;
+					// a tricky situation - an old enclave is registered with a different
```
kziemianek: Maybe inside the pallet everything will be fine, but in broader context, wouldn't it cause mess ? I cannot think of any case but i'm not sure. 
Have you thought about stopping here and saying: "hold on, something is wrong and we cannot register you again, start fresh instance ?"

pallets/teebag/src/types.rs
```diff
@@ -137,6 +123,16 @@ impl Enclave {
 		self
 	}
 
+	pub fn with_shielding_pubkey(mut self, shielding_pubkey: Option<Vec<u8>>) -> Self {
```
kziemianek, line 126: Are those fields mandatory for enclave ?
If yes it would be worth to add appropriate checks before struct is created.

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -276,35 +279,42 @@ export async function assertVc(context: IntegrationTestContext, subject: CorePri
...
+    const vcPubkey = Buffer.from(hexToU8a(vcPubkeyBytes.toHex()));
     const signatureStatus = await ed.verify(signature, message, vcPubkey);
 
     assert.isTrue(signatureStatus, 'Check Vc signature error: signature should be valid');
 
     // step 7
     // check VC mrenclave with enclave's mrenclave from registry
+
+    // this is awful, but I have no idea why base58.encode(registryMrenclave.mrenclave) doesn't work - it's a U8aFixed
+    const registryMrenclave = context.api.createType('[u8; 32]', lastRegisteredEnclave.mrenclave);
```
Kailai-Wang: If you have an idea why `base58.encode(registryMrenclave.mrenclave)` doesn't work - please let me know

@Traf333 @jonalvarezz @Verin1005 @grumpygreenguy

tee-worker/service/src/main_impl.rs
```diff
@@ -824,7 +789,7 @@ where
 }
 
 /// Start polling loop to wait until we have a worker for a shard registered on
-/// the parentchain (TEEREX WorkerForShard). This is the pre-requisite to be
+/// the parentchain (TEEBAG WorkerForShard). This is the pre-requisite to be
```
kziemianek: I think there is no more `WorkerForShard` in `teebag` pallet.

tee-worker/core-primitives/node-api/api-client-extensions/src/pallet_teebag.rs
```diff
@@ -0,0 +1,128 @@
...
+		shard: &ShardIdentifier,
+		at_block: Option<Self::Hash>,
+	) -> ApiResult<Option<AccountId>> {
+		let identifiers: Vec<AccountId> = self
+			.get_storage_map(TEEBAG, "EnclaveIdentifier", worker_type, at_block)?
+			.unwrap_or_default();
+		let maybe_account =
+			identifiers.iter().find(|account| match self.enclave(account, at_block) {
+				Ok(Some(e)) => e.mrenclave == shard.as_ref(),
+				_ => false,
```
kziemianek: In case of error we might miss the "real" first enclave, is this correct ?

tee-worker/core-primitives/node-api/api-client-extensions/src/pallet_teebag.rs
```diff
@@ -0,0 +1,128 @@
...
+		worker_type: WorkerType,
+		shard: &ShardIdentifier,
+		at_block: Option<Self::Hash>,
+	) -> ApiResult<Option<AccountId>> {
+		let identifiers: Vec<AccountId> = self
+			.get_storage_map(TEEBAG, "EnclaveIdentifier", worker_type, at_block)?
+			.unwrap_or_default();
+		let maybe_account =
+			identifiers.iter().find(|account| match self.enclave(account, at_block) {
+				Ok(Some(e)) => e.mrenclave == shard.as_ref(),
```
kziemianek: Is this a valid assumption that `mrenclave` will be equal to shard ?
I think `shard` can be provided and then it wouldn't match mrenclave (default).
Previously it was comparing shard identifiers, not shard with mrenclave

tee-worker/enclave-runtime/src/attestation.rs
```diff
@@ -142,7 +142,8 @@ pub unsafe extern "C" fn generate_ias_ra_extrinsic(
 	}
 	let mut url_slice = slice::from_raw_parts(w_url, w_url_size as usize);
 	let url = match String::decode(&mut url_slice) {
```
kziemianek, line 144: Wasn't it previously decoded (`String`) , not encoded (`Vec<u8>`) ? 😕 
Your comment says like there was no decoding performed.

tee-worker/service/src/main_impl.rs
```diff
@@ -521,86 +521,51 @@ fn start_worker<E, T, D, InitializationHandler, WorkerModeProvider>(
...
+		.enclave(&tee_accountid, None)
+		.unwrap()
+		.expect("our enclave should be registered at this point");
+	trace!("verified that our enclave is registered: {:?}", my_enclave);
 
-	if let Some(register_enclave_xt_header) = register_enclave_xt_header.clone() {
-		we_are_primary_validateer =
-			we_are_primary_worker(&litentry_rpc_api, &register_enclave_xt_header).unwrap();
-	}
+	let is_primary_enclave = match litentry_rpc_api
```
kziemianek, line 557: It's just a little optimization but maybe the `Teebag` `EnclaveAdded` event could contain information whether it was the primary enclave registered or not.

Then we could look up for this event in [`ExtrinsicReport`](https://github.com/scs/substrate-api-client/blob/master/src/api/mod.rs#L39) and this additional api call(s) could be omitted.

tee-worker/service/src/main_impl.rs
```diff
@@ -840,7 +805,11 @@ fn spawn_worker_for_shard_polling<InitializationHandler>(
 
 		loop {
 			info!("Polling for worker for shard ({} seconds interval)", POLL_INTERVAL_SECS);
-			if let Ok(Some(_enclave)) = node_api.worker_for_shard(&shard_for_initialized, None) {
+			if let Ok(Some(_account)) = node_api.primary_enclave_identifier_for_shard(
```
kziemianek, line 807: The `is_primary_enclave` value is calculated before so either we are the primary enclave for shard or there is already other. It sound like we don't need to perform this additional call, what do you think?

tee-worker/app-libs/parentchain-interface/src/integritee/mod.rs
```diff
@@ -143,11 +143,7 @@ where
 			"[ShieldFundsAndInvokeFilter] attempting to execute indirect call with index {:?}",
 			index
 		);
-		if index == metadata.shield_funds_call_indexes().ok()? {
```
felixfaisal, line 146: I have a question here, So we are only skipping parsing `shield_funds`, is this sufficient?

Kailai-Wang:
fixes P-10
fixes #1133

pallets/teebag/src/lib.rs
Kailai-Wang: I think the core question is shall we allow registering a different worker type using the same parachain account - as if we are re-using the same machine with previously sealed parachain account to re-reigster itself (I can only imagine this in test).

Either way is fine for me - can you think of any cases where we heavily prefer A over B?

pallets/teebag/src/types.rs
Kailai-Wang: They are semi-mandatory in our business logic - without it you can't perform any writing.
It's not strictly reqruied in reading, as well as a broader enclave sense

So I'd still keep it optional

tee-worker/core-primitives/node-api/api-client-extensions/src/pallet_teebag.rs
Kailai-Wang: When you say error do you mean network error or whatever causes `self.enclave` to abort abnormally? If so then we might miss, yes, but that's expected. If you can't do rpc query, you are expected to not get the right result

tee-worker/app-libs/parentchain-interface/src/integritee/mod.rs
Kailai-Wang: Since we don't have such extrinsic in pallet-teebag, we shouldn't land in this arm at all.

tee-worker/service/src/main_impl.rs
Kailai-Wang: if `is_primary_enclave` is true, it's all good. If it's false, we don't know if some primary enclave (other than us) exists or no enclave is registered at all.

This line checks if "some" enclave has registered.

This is my understanding purely from the code logic - it might seem redundant because we don't properly take `shard` into consideration, I had comment about it somehwere 😂 

that said I'll keep it for now

tee-worker/enclave-runtime/src/attestation.rs
Kailai-Wang: The url is read from config and SCALE encoded before passing into the enclave (this is upstream code), while teebag accepts an opaque `Vec<u8>`, the UI can display utf8 encoded string well (but not string encoded Vec) - that's why I did the conversion here

pallets/teebag/src/lib.rs
kziemianek: I think I would prefer to have a clear rule 1 parachain account - 1 worker type.
It's relatively cheap to create new account compared to how it can potentially make things complex (now or future).

tee-worker/core-primitives/node-api/api-client-extensions/src/pallet_teebag.rs
Kailai-Wang: It's valid until we support multi shard - and by then we'll probably need a bigger revamp (for code outside of this PR). It seems upstream sometimes takes the shard into consideration, but not always consistent.

pallets/teebag/src/lib.rs
Kailai-Wang: Sure I'll change it

tee-worker/core-primitives/node-api/api-client-extensions/src/pallet_teebag.rs
kziemianek: I mean whatever causes `self.enclave` call to return `Result::Error`

Kailai-Wang: This is basically about the error handling inside a boolean predicate.

I could use something like this:
```
		let mut maybe_account: Option<AccountId> = None;
		for account in identifiers {
			match self.enclave(&account, at_block)? {
				Some(e) => if e.mrenclave == shard.as_ref() {
					maybe_account = Some(account.clone())
					break;
				},
				None => continue,
			}
		}
		maybe_account
```

but I think the readability is awful 😂 

wdyt?

tee-worker/service/src/main_impl.rs
Kailai-Wang: I think that's actually more complicated 😂 You need to verify it's the first `EnclaveAdded` event you've got?

tee-worker/core-primitives/node-api/api-client-extensions/src/pallet_teebag.rs
kziemianek: I was thinking more about returning the error rather than silently skipping the enclave and fetchin following one.

Kailai-Wang: Yea then that's the code snippet above

Kailai-Wang:
The lit-resume-worker test fails at the moment, I created an issue P-501

tee-worker/core-primitives/node-api/api-client-extensions/src/pallet_teebag.rs
Kailai-Wang: code updated

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
jonalvarezz: can't tell exactly, I'd need to toy around it, but the `.toHuman() as PalletTeebagEnclave` of L291 might be encoding it in a different format? the `toHuman` output isn't exactly `PalletTeebagEnclave` – `api.createType` should be more accurate.

jonalvarezz:
> cc @jonalvarezz it will be a breaking change for client (SDK) as the pallet/extrinsic interface has changed

Yes, quite a breaking change. We'd need to put IDH under maintenance for a few minutes before releasing it.

Btw, will this be released on the original plan of releasing every 3 weeks? So I can plan the client-sdk update accordingly.

Kailai-Wang:
> > cc @jonalvarezz it will be a breaking change for client (SDK) as the pallet/extrinsic interface has changed
> 
> Yes, quite a breaking change. We'd need to put IDH under maintenance for a few minutes before releasing it.
> 
> Btw, will this be released on the original plan of releasing every 3 weeks? So I can plan the client-sdk update accordingly.

I'd prefer an ad-hoc release if the next regular release is distant. However, the majority of IDHub colleagues is on vacation next week so maybe we can't do it in a hurry anyway :(

Kailai-Wang:
Well mainly the change from `teerex` to `teebag` - and the extrinsic/event names have changed.
It affects the registration process, the enclave registry on parachain as well as the callback events (e.g. the famous `teerex.ProcessedParentchainBlock` event :D )

jonalvarezz:
> I'd prefer an ad-hoc release if the next regular release is distant. However, the majority of IDHub colleagues is on vacation next week so maybe we can't do it in a hurry anyway :(

We can handle that. It is mostly infrastructure so we can definitely do it on behalf of them.

jonalvarezz:
> e.g. the famous `teerex.ProcessedParentchainBlock` event :D )

Is that one changing names as well? The monitoring service we have relies on it. It'd need update too then

Kailai-Wang:
Yes it's now `teebag.ParentchainBlockProcessed` - we'll add more lifecycle events later

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
Kailai-Wang: Thanks @jonalvarezz - that seems to be the cause!

Although I haven't understood it comprehensively, but the `teebag.enclaveRegistry` query returns an `Option<_>`, so maybe `toHuman()` then `as PalletTeebagEnclave` will pull in some tricky decoding problem which silently gets ignored.

Either way, I just updated it - it seems to be working now :)
https://github.com/litentry/litentry-parachain/pull/2469/commits/a00123ba0f3d7e5e2f8698cb81400a58c1253a11

Lesson learned is be cautious with using `as unknown as` and `toHuman`

jonalvarezz: Yeah, `toHuman` and `toJson` should ideally be used for displaying or logging purposes only. We've found ourselves in situations where we had to use it when an `api`/ `registry` instance was not worth creating.

felixfaisal (approved):
LGTM, Tested locally

kziemianek (approved):

PR #2470, opened 2024-02-07
p-496 Add vip3 service mock

0xverin:
### Context

resolves p-496

Additionally, I also found that this API:
`https://dappapi.vip3.io/api/v1/sbt/info?addr%E2%80%8B=0x651614cA9097C5ba189Ef85e7851Ef9cff592B2c`
 sometimes returns an error with `temporarily unavaliable`


tee-worker/litentry/core/mock-server/src/vip3.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2020-2023 Trust Computing GmbH.
```
kziemianek: Should be 2024 :)

kziemianek (approved):

tee-worker/litentry/core/mock-server/src/vip3.rs
Kailai-Wang: Since @Verin1005 is on vacation I'll just update and merge it 😂

PR #2472, opened 2024-02-08
P 491 tee prod env suffers sgx error enclave crashed

BillyWooo:
- add error handling when parsing url. (within build_client)
- propagate error
- remove compiling warnings.


kziemianek (commented):
Hey @BillyWooo , have you considered using `Url` in `DataProviderConfig` so url can be parsed as soon as config is initialized ? The best option is to fail as soon as possible in case wrong config is provided. What do you think ?

Kailai-Wang:
> Hey @BillyWooo , have you considered using `Url` in `DataProviderConfig` so url can be parsed as soon as config is initialized ? The best option is to fail as soon as possible in case wrong config is provided. What do you think ?

This is a good suggestion, but AFAIK the parsing error doesn't always happen. So it _looks_ like a runtime error rather than statically-parsing error when reading the config 🤔  (correct me if I'm wrong) - are we sending requests to different URLs when building VCs?

kziemianek:
But we still can do this whenever we expect valid url in config (not template like in case of nodereal).

jonalvarezz:
> are we sending requests to different URLs when building VCs?

The Achainable data prov implementation is, for instance, very open, and it accepts anything from clients. Things probably go wrong when building URLs based on it

Kailai-Wang:
> The Achainable data prov implementation is, for instance, very open, and it accepts anything from clients

Even for endpoints? 😲 
I mean bad actors can easily provide his own website and log down all identities via requests.

And do we know which URL went wrong?
It should be easy to find out why it errored out, the `url` crate isn't a blackbox :D

BillyWooo:
> > The Achainable data prov implementation is, for instance, very open, and it accepts anything from clients
> 
> Even for endpoints? 😲 I mean bad actors can easily provide his own website and log down all identities via requests.
> 
> And do we know which URL went wrong? It should be easy to find out why it errored out, the `url` crate isn't a blackbox :D

The error is known: InvalidPort when parsing url. 

@jonalvarezz can you point me to which ones for Achainable could lead to that problem?

Hi @kziemianek, as Kai pointed out, the error coming from run-time. Checking within DataProviderConfig can't solve that. Besides, our url for production is hardcoded.

Kailai-Wang:
> The error is known: InvalidPort when parsing url.

I mean the concrete URL value when the error happens.

If the URL is hardcoded, I'm also wondering why it happens _sometimes_.

jonalvarezz:
> can you point me to which ones for Achainable could lead to that problem?

I'm not sure if it is exactly related or if we are doing strict checks somewhere else, but it may be worth checking since we are at it:

Take `AchainableParams:Amount` as example. The `name` param accepts any string. but it will only return the desired response if client sends exactly `Created over {amount} contracts`. A client could send other things from P-js for example.

BillyWooo:
> > can you point me to which ones for Achainable could lead to that problem?
> 
> I'm not sure if it is exactly related or if we are doing strict checks somewhere else, but it may be worth checking since we are at it:
> 
> Take `AchainableParams:Amount` as example. The `name` param accepts any string. but it will only return the desired response if client sends exactly `Created over {amount} contracts`. A client could send other things from P-js for example.

The error happens at building client with base url. These input parameter has not yet come into picture.

BillyWooo:
> > The error is known: InvalidPort when parsing url.
> 
> I mean the concrete URL value when the error happens.
> 
> If the URL is hardcoded, I'm also wondering why it happens _sometimes_.

Unfortunately we didn't have that debug info before. I just added error message to include url which cause crash within this PR.

Kailai-Wang (commented):
Thanks!

- Yes I believe checking error in each `..Client::new` is too verbose, if it expects a `Url`, we should pass in a `Url` and check it beforehand. The URL parsing could even happen in reading config, so that `DataProviderConfig` accepts `Url` type instead of `String`
- I feel we should all use `build_client_with_cert`? Any reason we need to use `build_client`? Maybe for mock server only?
- If it still takes time to update the PR, we should fix the special quote character problem in another PR and push the change to prod ASAP, otherwise it will be crashing from time to time (whenever users do discord-related stuff)
- the PR might need a better title :)

BillyWooo:
Will create a new PR to:
- check URL at the very beginning
- remove `build_client`. (previously it's decided to do it separately)
- Production env is running on a hotfix for the moment. At least it won't crash because of this reason.

PR #2474, opened 2024-02-09
BitAcross direct calls

kziemianek:
* Adds two `DirectCall` variants to perform `ecdsa` and `schnorr` signing.
* Replaces `libsecp256k1` crate with `k256` crate, so both keypair generation and signing support comes from single dependency.
* `DirectCall` processing result is `aes` encrypted using key provided by the client.
* Adds CLI commands to perform signing.


bitacross-worker/litentry/core/direct-call/src/lib.rs
```diff
@@ -0,0 +1,75 @@
+// Copyright 2020-2023 Trust Computing GmbH.
```
Kailai-Wang: 2024 :)

bitacross-worker/litentry/core/direct-call/Cargo.toml
Kailai-Wang, line 1: Good that we start to have a crate for such calls

Not going to change this here but we'll probably have another name for such a "direct" call which doesn't have to go through the sidechain.

I remember I proposed something like `NativeCall` in one of @felixfaisal 's PRs

bitacross-worker/litentry/core/direct-call/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false }
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false }
+sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false }
+
+# internal dependencies
+itp-stf-primitives = { path = "../../../core-primitives/stf-primitives", default-features = false }
+litentry-primitives = { path = "../../primitives", default-features = false }
+parentchain-primitives = { package = "core-primitives", path = "../../../../primitives/core", default-features = false }
+teerex-primitives = { path = "../../../../primitives/teerex", default-features = false }
```
Kailai-Wang: Shall we ditch `teerex-*` stuff now? Like use `itp_types` or `litentry_primitives` instead for `ShardIdentifier`

I mean there will be a clean-up, but we can try to not add new references

bitacross-worker/bitacross/core/bc-task-receiver/src/lib.rs
```diff
@@ -50,44 +58,81 @@ pub enum Error {
...
+		}
 	}
 }
 
-pub fn run_bit_across_handler_runner<ShieldingKeyRepository, S, H, O>(
-	_context: Arc<BitAcrossTaskContext<ShieldingKeyRepository, S, H, O>>,
+pub fn run_bit_across_handler_runner<
+	ShieldingKeyRepository,
+	EthereumKeyRepository,
+	BitcoinKeyRepository,
```
Kailai-Wang: Do you fancy using a shorter name to increase the readability?
Like SR, ER, BR ..

Kailai-Wang (approved):
Thanks, looks great! I haven't verified the crypto impl though

bitacross-worker/litentry/core/direct-call/Cargo.toml
kziemianek: Yes, I've removed this dep.

bitacross-worker/bitacross/core/bc-task-receiver/src/lib.rs
kziemianek: Renamed 👍🏼

bitacross-worker/bitacross/core/bc-task-receiver/src/lib.rs
```diff
@@ -111,8 +140,48 @@ pub fn run_bit_across_handler_runner<ShieldingKeyRepository, S, H, O>(
...
+	let mrenclave = match context.ocall_api.get_mrenclave_of_self() {
+		Ok(m) => m.m,
+		Err(_) => return Err("Failed to get mrenclave".to_string()),
+	};
+	ensure!(dc.verify_signature(&mrenclave, &request.shard), "Failed to verify sig".to_string());
+	match dc.call {
+		DirectCall::SignBitcoin(_, aes_key, payload) => {
+			if_production_or!(unimplemented!(), {
+				let key = context.bitcoin_key_repository.retrieve_key().unwrap();
+				let signature = key.sign(&payload).unwrap();
```
felixfaisal, line 172: Is there a way we can write a unit test to verify the signature?

felixfaisal (approved):
It looks good, ,It would be great to have a simple unit test to check if the signature signing works as expected, but we could do it in another PR.

Kailai-Wang:
I believe the CI failed because of the introduction of teebag usage in ts-test, which is common code between identity and bc worker.
So either we force-merge it, or wait until P-486 is done

Kailai-Wang:
Maybe I misunderstood, the failed CI is tee-worker, not bc-worker. Then it can be related to the too old tee-worker docker image 🤔  let's see if this run works

PR #2476, opened 2024-02-09
Decouple woker_type and worker_mode

Kailai-Wang:
### Context

As topic


kziemianek (approved):
That's a good improvement.

PR #2478, opened 2024-02-11
fix ecdsa and schnorr tests

kziemianek:
* runs ecdsa and schnorr crypto tests alongside other sgx crypto tests
* adds basic signature verification unit tests


felixfaisal (approved):
LGTM

Kailai-Wang (approved):

PR #2483, opened 2024-02-12
authorize signing tasks

kziemianek:
* adds logic for signing tasks authorization


Kailai-Wang (approved):
Looks good, thanks!

We might be able to further improve the readability by defining a trait for DirectCall to verify the signer's identity (something like `verifyCaller`) but it's just nice-to-have

PR #2486, opened 2024-02-13
More cleanup around the workers

Kailai-Wang:
### Context

This PR:
- removes heartbeat-timeout related test which is unused
- removes pallet-drop3 integration (the pallet itself is still kept)
- removes unused itp-types definitions


bitacross-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -208,7 +210,8 @@ impl TrustedCallVerification for TrustedCallSigned {
 		payload.append(&mut mrenclave.encode());
 		payload.append(&mut shard.encode());
 
-		self.signature.verify(payload.as_slice(), self.call.sender_identity())
+		self.signature.verify(&blake2_256(&payload), self.call.sender_identity())
```
felixfaisal, line 213: I would say we keep only one way of signing message, that is hashing it before we sign it. I always assumed the sign and verify methods did the hashing internally 😅

bitacross-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -155,7 +156,8 @@ impl TrustedCallSigning<TrustedCallSigned> for TrustedCall {
 		payload.append(&mut mrenclave.encode());
 		payload.append(&mut shard.encode());
 
-		TrustedCallSigned { call: self.clone(), nonce, signature: pair.sign(payload.as_slice()) }
+		// use blake2_256 hash to shorten the payload - see `verify_signature` below
+		TrustedCallSigned { call: self.clone(), nonce, signature: pair.sign(&blake2_256(&payload)) }
```
felixfaisal, line 160: The size of the signature remains same, I think we are replacing the computation of signing a large message with the computation of producing the hash for a large message. I do prefer hashing a message before signing it for consistency.

felixfaisal (approved):
LGTM, Just had a few comments regarding the payload signing.

bitacross-worker/app-libs/stf/src/trusted_call.rs
felixfaisal: Please ignore this, It was rhetorical

bitacross-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: Yea this is changed in tee-worker already, just not in bc-worker.

kziemianek (approved):

PR #2490, opened 2024-02-14
refactor: remove redundant assertion handling code

felixfaisal:
I think I'm doing the opposite than the issue description, Since there is a lot we don't need from the assertion handler code, I've just extracted a common function that can be used by `vc-issuance` and `stf-task`. This way direct invocation and direct rpc methods of requesting vc can co-exist atleast for a while.


jonalvarezz (approved):
before it was easy to forget update one of the files, it's better now! thanks!

PR #2492, opened 2024-02-15
feat: P-212 Club3 NFT VC

wli-pro:
### Context

1 added new data provider morails (other existing data provider does not support the required evm network)
2 added arbitrum and polygon network
3 added general nft holder VC using this new data provider
4 added Club3 VC
5 migrated existing WeirdoGhostGang VC under it

### Labels

### Testing Evidences

```
{
   "@context":[
      "https://www.w3.org/2018/credentials/v1",
      "https://w3id.org/security/suites/ed25519-2020/v1"
   ],
   "id":"0xb7d828814aeb7bc9c8fa674370b45a3f38eea1b249f400bb48c71e305a46a4af",
   "type":[
      "VerifiableCredential"
   ],
   "credentialSubject":{
      "id":"did:litentry:substrate:0xd4e35b16ec6b417386b948e7eaf5cc642a243096cecf366e6313689b90969f42",
      "description":"You are a holder of a certain kind of NFT",
      "type":"NFT Holder",
      "assertionText":"NftHolder(Club3Sbt)",
      "assertions":[
         {
            "and":[
               {
                  "src":"$nft",
                  "op":"==",
                  "dst":"Club3 SBT"
               },
               {
                  "or":[
                     {
                        "and":[
                           {
                              "src":"$network",
                              "op":"==",
                              "dst":"bsc"
                           },
                           {
                              "src":"$address",
                              "op":"==",
                              "dst":"0x9f488C0dafb1B3bFeeD3e886e0E6E5f3f4517925"
                           }
                        ]
                     },
                     {
                        "and":[
                           {
                              "src":"$network",
                              "op":"==",
                              "dst":"polygon"
                           },
                           {
                              "src":"$address",
                              "op":"==",
                              "dst":"0xAc2e4e67cffa5E82bfA1e169e5F9aa405114C982"
                           }
                        ]
                     },
                     {
                        "and":[
                           {
                              "src":"$network",
                              "op":"==",
                              "dst":"arbitrum"
                           },
                           {
                              "src":"$address",
                              "op":"==",
                              "dst":"0xcccFF19FB8a4a2A206d07842b8F8c8c0A11904C2"
                           }
                        ]
                     }
                  ]
               }
            ]
         }
      ],
      "values":[
         true
      ],
      "endpoint":"wss://rpc.rococo-parachain.litentry.io"
   },
   "issuer":{
      "id":"did:litentry:substrate:0x304209a0581f23786118458a63f7be3b46554fa59be4a12a194c9f3c6304f191",
      "name":"Litentry TEE Worker",
      "mrenclave":"5tJCvdgWoScE27uQPC5YafWePiR7Qn2V9apJPR44ZfMT"
   },
   "issuanceDate":"2024-02-15T12:00:37.078191645+00:00",
   "parachainBlockNumber":50,
   "sidechainBlockNumber":92,
   "proof":{
      "created":"2024-02-15T12:00:37.079184850+00:00",
      "type":"Ed25519Signature2020",
      "proofPurpose":"assertionMethod",
      "proofValue":"4c2469c052dbb131f5a1bc26cefe72856cb332417beec8b9507dd4c9e714a6937eb362b020018b78de87c8c96b9fc85ff7b75d89d0fe14d4c7d556b9d1814103",
      "verificationMethod":"0x304209a0581f23786118458a63f7be3b46554fa59be4a12a194c9f3c6304f191"
   }
}
```


wli-pro:
added new evm network support for polygon and arbitrum, pls see how client side wanted to extend the supported network. @YoshiyukiSakura

felixfaisal (commented):
Hey @higherordertech 
Not related to the PR directly, But i see these crates `assertion-build-v2` and `credential-v2`, 
1. I may have missed out why we are creating separate crates for different versions? 
2. Do we plan on continuing to iter this way, wherein we have `v3` and then `v4`? 
3. If not are we planning on merging `assertion-build-v2` and `assertion-build`, So what we waiting for in order to merge? 

We have existing feature gating and version controlling features  in Rust without having to explicitly define `v2` crates and so on.

tee-worker/litentry/core/data-providers/src/moralis.rs
```diff
@@ -0,0 +1,268 @@
...
+			match action(self) {
+				Ok(response) => return Ok(response),
+				Err(err) => {
+					let req_err: Error = Error::RequestError(format!("moralis api error: {}", err));
+					match err {
+						HttpError::HttpError(code, _) =>
+							if code == 429 {
+								// Too Many Requests
+								// exponential back off
+								thread::sleep(base_delay * 2u32.pow(retries as u32));
```
felixfaisal: The exponential backing off here, will make the thread sleep for x seconds, during which the thread is not available for any other computation. Exponential backoff should not be implemented single threaded, Because it essentially is blocking this single thread for x seconds and affecting throughput.

wli-pro:
> Hey @higherordertech Not related to the PR directly, But i see these crates `assertion-build-v2` and `credential-v2`,
> 
> 1. I may have missed out why we are creating separate crates for different versions?
> 2. Do we plan on continuing to iter this way, wherein we have `v3` and then `v4`?
> 3. If not are we planning on merging `assertion-build-v2` and `assertion-build`, So what we waiting for in order to merge?
> 
> We have existing feature gating and version controlling features in Rust without having to explicitly define `v2` crates and so on.

The plan is to migrate all v1 credentials under v2 and then v1 can be deprecated and v2 can be renamed to be the only version. Currently they co-exists because the migration work is not prioritised yet (while we are still catching new VC requirement), while at same time some of the new VC can not really be implemented via the v1 structure.

tee-worker/litentry/core/data-providers/src/moralis.rs
wli-pro: For now, in this case should we simply stop the whole VC logic and throw error and buble up the error and pass this rate limiting error to client side maybe?

Just this is at data-provider layer,  for one VC, the caller could call this method in a for loop too, and if one fail then even we might have some results from other calls, we will have to fail the whole flow.

I feel it's a trade off between throughput and trying server-side's best to make the logic work without requiring client-side retry. 

For an overal enhancement, for all VC logic, some rough ideas how this can be strategically dealt, though more insight of how exactly this is feasible will be greatly appreciated):

1 Ideally, if we can break one VC generation into many independent sub tasks that can be run in different threads (or async wait), and allow them to execute and retry without affecting the main thread here, that would be the ultimate solution?

2 Or for retryable exception, we can throw the request back into the task pool, which is what @kziemianek suggested previously.

tee-worker/local-setup/.env.dev
```diff
@@ -31,6 +31,8 @@ ACHAINABLE_AUTH_KEY=
 ONEBLOCK_NOTION_KEY=
 NODEREAL_API_KEY=
 GENIIDATA_API_KEY=
+GENIIDATA_API_KEY=142cf1b0-1ca7-11ee-bb5e-9d74c2e854ac
```
Kailai-Wang: We have this key already - see the line above

tee-worker/litentry/core/data-providers/src/moralis.rs
felixfaisal: See, here's the scenario, in isolated request-vc we have about 4 threads, So if we have 4 threads occupied with an exponential backoff sleep, We potentially have a DOS implemented by ourselves in the code. The ultimate solution is definitely 1 or 2, but until we have that, let's not explicitly sleep(x to the power of n) the entire thread such that no other VCs of other data providers can use the threads?

tee-worker/litentry/core/data-providers/src/moralis.rs
```diff
@@ -0,0 +1,268 @@
...
+	api_retry_times: u16,
+	client: RestClient<HttpClient<DefaultSend>>,
+}
+
+impl MoralisClient {
+	pub fn new(data_provider_config: &DataProviderConfig) -> Self {
+		let api_key = data_provider_config.moralis_api_key.clone();
+		let api_retry_delay = data_provider_config.moralis_api_retry_delay;
+		let api_retry_times = data_provider_config.moralis_api_retry_times;
+		let api_url = data_provider_config.moralis_api_url.clone();
```
Kailai-Wang, line 56: It's out of this PR's scope, but sooner or later we'll need a refactoring to have a trait-based impl. The background is I see a lot of duplicate code, we could have something like `build_client_with auth` or `build_client_with retry` - or as fn parameters. 🤔

Kailai-Wang (commented):
Not much to complain from my side :)
I've been wondering the difference between lc-credential and lc-crediential-v2

tee-worker/litentry/core/data-providers/src/moralis.rs
arieljbirnbaum: Given the thread limitation, I'm not sure if there's a one-size-fits-all solution.
On one hand, if a VC only needs a single data provider request, it's probably best to free up the worker and let the client retry.
On the other hand, if a lot of provider requests are needed (like the holding time assertion, even [after optimizing the logic](github.com/litentry/litentry-parachain/pull/2401), it could be less harmful to retry a couple times rather than forcing the client to restart from scratch.

felixfaisal: There is never a one size fits solution 😅 . So basically assertion logic that retry are going to take up more of our computation time which increases the response time for other more simpler assertion. Why can't we have temporary storage of already received results and only retry the ones that didn't succeed instead of blocking the thread for **exponential time**.

wli-pro: added fail fast boolean option in service layer, and trait in data provider layer now

tee-worker/litentry/core/data-providers/src/moralis.rs
wli-pro: added fail fast boolean option in service layer, and trait in data provider layer now

wli-pro:
With minimum changes, also added RetryableError for client to differentiate retryable error or non retryable error: 

```
>>>command: PlatformUser(
    KaratDaoUser,
)
Trusted call 0x0b19000591014461746150726f76696465724572726f7228426f756e64656456 is TopExecuted([11, 25, 0, 5, 145, 1, 68, 97, 116, 97, 80, 114, 111, 118, 105, 100, 101, 114, 69, 114, 114, 111, 114, 40, 66, 111, 117, 110, 100, 101, 100, 86, 101, 99, 40, 91, 56, 50, 44, 32, 49, 48, 49, 44, 32, 49, 49, 54, 44, 32, 49, 49, 52, 44, 32, 49, 50, 49, 44, 32, 57, 55, 44, 32, 57, 56, 44, 32, 49, 48, 56, 44, 32, 49, 48, 49, 44, 32, 54, 57, 44, 32, 49, 49, 52, 44, 32, 49, 49, 52, 44, 32, 49, 49, 49, 44, 32, 49, 49, 52, 44, 32, 52, 48, 44, 32], false)
Default { msg: "[Error] Error occurred while executing trusted call: RequestVCFailed(PlatformUser(KaratDaoUser), DataProviderError(BoundedVec([68, 97, 116, 97, 80, 114, 111, 118, 105, 100, 101, 114, 69, 114, 114, 111, 114, 40, 66, 111, 117, 110, 100, 101, 100, 86, 101, 99, 40, 91, 56, 50, 44, 32, 49, 48, 49, 44, 32, 49, 49, 54, 44, 32, 49, 49, 52, 44, 32, 49, 50, 49, 44, 32, 57, 55, 44, 32, 57, 56, 44, 32, 49, 48, 56, 44, 32, 49, 48, 49, 44, 32, 54, 57, 44, 32, 49, 49, 52, 44, 32, 49, 49, 52, 44, 32, 49, 49, 49, 44, 32, 49, 49, 52, 44, 32, 52, 48, 44, 32], 100)))" }

RetryableError  BoundedVec converted to string is:
RetryableError("Fail to call rest api within 0 retries")
```

felixfaisal (approved):
LGTM

Kailai-Wang:
Asking for reviews from @BillyWooo @zhouhuitian

wli-pro:
> Not much to complain from my side :) I've been wondering the difference between lc-credential and lc-crediential-v2

@Kailai-Wang let me take this chance to explain the differences:

lc-credentials-v2  does not depend on lc-data-providers, and doesn't depend on any data provider from the credential level, nor limit how many data provider can be used. e.g. there won't be so called achainable credential or noderal credentail, instead a credential is just a credential decoupled from any (one or many) data providers it would require to use.

while lc-credentials depend on lc-data-providers and has always hard coded the data provider dependency in credential level, and also constraint one credential can only use max one data provider

primitives/core/src/network.rs
```diff
@@ -90,6 +90,12 @@ pub enum Web3Network {
 	BitcoinP2wpkh,
 	#[codec(index = 13)]
 	BitcoinP2wsh,
+
+	// evm
```
BillyWooo, line 94: My review comes a bit late, better to put these two options above with `Ethereum/Bsc`.

PR #2493, opened 2024-02-15
publish custodian wallets

kziemianek:
* publish enclave's bitcoin and ethereum public keys in parachain bitacross pallet.

Keys are always published during the startup - even if they are no freshly generated but I assume it's enough for first version.


pallets/bitacross/src/custodial_wallet.rs
```diff
@@ -15,14 +15,15 @@
 // along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
 
 use codec::{Decode, Encode};
-use core_primitives::{Address20, Address33};
 use scale_info::TypeInfo;
 
+pub type PubKey = [u8; 33];
```
Kailai-Wang, line 20: Did you try it?

Are `ethereum_key.public_bytes` and `bitcoin_key.public_bytes` guaranteed to be convertible to this type?

kziemianek: Yes, it's 
```
pub type CompressedPoint = GenericArray<u8, U33>;
```

Kailai-Wang: Ah I see, so it's the pubkey, not address

Kailai-Wang (approved):

felixfaisal (approved):
It looks good, yeah not sure if the bitacross colleagues need pubkey or address

PR #2495, opened 2024-02-15
Use ed25519::Public as vc_pubkey

Kailai-Wang:
### Context

As topic, prefer explicit type over opaque type.


Kailai-Wang:
Noticeable for the client-sdk but it's not really a breaking change cc @jonalvarezz

jonalvarezz:
> Noticeable for the client-sdk but it's not really a breaking change cc @jonalvarezz

Thanks, I added a client-sdk follow-up task to ensure integration: [p-521-client-sdk-ensure-compabitibility-after-vc_pubkeys](https://linear.app/litentry/issue/P-521/client-sdk-ensure-compabitibility-after-vc-pubkeys-ed25519public)

Kailai-Wang:
It should be compatible though - you could check the generated metadata/type, it's still `Option<Bytes>`.  The passing CI proves it too

kziemianek (approved):

PR #2496, opened 2024-02-16
Check all subcommands / option flags of litentry-worker and litentry-cli

Traf333:
### Context
This PR is tend to adjust missing or misleading descriptions. In addition it was decided to remove indirect linking identity command due to having direct cli command


tee-worker/service/src/cli.yml
```diff
@@ -155,6 +155,22 @@ subcommands:
...
+                required: false
+                long: teeracle-interval
+                short: i
+                help: Set the teeracle exchange rate update interval. Example of accepted syntax <5 seconds 15 minutes 2 hours 1 days> or short <5s15m2h1d>
+                takes_value: true
+            - reregister-teeracle-interval:
+                required: false
+                long: reregister
+                help: Set the teeracle reregistration interval. Example of accepted syntax <5 seconds 15 minutes 2 hours 1 days> or short <5s15m2h1d>
+                takes_value: true
```
Kailai-Wang: I think this should be removed actually, see https://github.com/litentry/litentry-parachain/pull/2488/files#diff-6acbe3078fcde879295ce6a6b23acd22ba163f837b3fde7700b268a03ec3930b

tee-worker/cli/src/base_cli/mod.rs
```diff
@@ -103,7 +97,6 @@ impl BaseCommand {
 			BaseCommand::RegisterTcbInfo(cmd) => cmd.run(cli),
 			// Litentry's commands below
 			BaseCommand::PrintSgxMetadataRaw => print_sgx_metadata_raw(cli),
-			BaseCommand::LinkIdentity(cmd) => cmd.run(cli),
```
Kailai-Wang, line 106: What about `PrintSgxMetadataRaw` (just the line above)?  Do we have help message for it?

Traf333: Yes, it has help message, though it's a bit verbose 🤔 
image

tee-worker/service/src/cli.yml
Traf333: good catch, it seems I've resolved conflict in a wrong way. Thanks

tee-worker/service/src/cli.yml
Kailai-Wang, line 1: Is it possible to keep the format? It's been changing back and forth .. I know it's probably some automatic formatting from IDE, but still..

Kailai-Wang (approved):
Thanks! I invited @BillyWooo and @zhouhuitian for review too

tee-worker/service/src/cli.yml
Traf333: egh, again. Will be more cautious with yml files and will setup proper formating in IDE. Thanks for pointing

BillyWooo (approved):
The changes look good for me. 
Just want to confirm, have all subcommands been checked?

PR #2497, opened 2024-02-16
Remove sidechain feature and sidechain related code from bitacross

kziemianek:
Nothing built here, just removal of unused sidechain code from bitacross worker.


Kailai-Wang:
Thanks!

I'm also wondering how hard it would be to add sidechain back (in case we need it someday) 😂

kziemianek:
> Thanks!
> 
> I'm also wondering how hard it would be to add sidechain back (in case we need it someday) 😂

That's a good point. 

As long as bitacross and identity workers are based on the same upstream commit it shouldn't be hard, taking code from upstream shouldn't be hard too.

Things might complicate once identity and bitacross workers are based on different upstream commits and we want to 'copy' identity worker sidechain into bitacross.

In general it might potentially require work from us to bring this code back, but from the other hand if any further development is planned it might also save time because we don't need to maintain unused code.

Kailai-Wang (approved):
Feel free to merge it once CI passes

PR #2498, opened 2024-02-16
purge relayer registry file

kziemianek:
As topic


bitacross-worker/service/src/setup.rs
```diff
@@ -17,9 +17,10 @@
 */
 
 use crate::error::{Error, ServiceResult};
+use frame_support::storage::child::root;
```
Kailai-Wang: Is it redundant?

Kailai-Wang (approved):
Thanks!

felixfaisal (approved):
LGTM

bitacross-worker/service/src/setup.rs
kziemianek: Yes, it was added by IDE autoimport ;)

Kailai-Wang:
We have to resolve the conflict and we are good to go

PR #2500, opened 2024-02-19
fix: merge request-vc commands

felixfaisal:
So I'm using a flag `--isolated` to differentiate between direct invocation and isolated request-vc 
[Update] 
So we want `--isolated` to be default, therefore we are using the flag `--stf` to request-vc via sidechain, it's basically the other way round.


tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
```diff
@@ -505,17 +508,33 @@ impl RequestVcCommand {
...
+			match perform_trusted_operation::<RequestVCResult>(cli, trusted_cli, &top) {
+				Ok(mut vc) => {
+					let decrypted = aes_decrypt(&key, &mut vc.vc_payload).unwrap();
+					let credential_str = String::from_utf8(decrypted).expect("Found invalid UTF-8");
+					println!("----Generated VC-----");
+					println!("{}", credential_str);
+				},
+				Err(e) => {
+					println!("{:?}", e);
+				},
```
Kailai-Wang: It seems we can avoid the repetition by something like:
```
let maybe_vc = if isolated { .. } else { .. }
match maybe_vc ...
```

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
```diff
@@ -75,6 +75,9 @@ pub struct RequestVcCommand {
 	/// subcommand to define the vc type requested
 	#[clap(subcommand)]
 	command: Command,
+	/// mode for the request-vc
+	#[clap(short = 'i', long, default_value_t = false)]
+	isolated: bool,
```
Kailai-Wang: Shall we consider using `isolated` by default? Because this should become the standard way of requesting VCs

Maybe rename it to something opposite

Kailai-Wang (commented):
Thanks!

jonalvarezz (approved):
I'm ok with the direction, would only emphasize to make it clearer that isolated is the preferred way of requesting vc, as Kai pointed out.

BillyWooo (approved):
Second to Kai's opinion.

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
felixfaisal: Renamed it to `stf` ✅

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
felixfaisal: Updated ✅

BillyWooo:
Should we also remove `request_vc_direct.rs` ?

Kailai-Wang:
Can this be merged?

And yes it seems that we don't need `request_vc_direct.rs`

PR #2502, opened 2024-02-19
feat(vc tests): use the schema validator package

jonalvarezz:
### Context

[P-142 : Enable use of JSONSchema in VC](https://linear.app/litentry/issue/P-142/enable-use-of-jsonschema-in-vc) introduced stricter JSON Schemas for the generated VC. The schemas can be found at [vc-jsonschema repo](https://github.com/litentry/vc-jsonschema/).

This change, uses the schema validator package distributed from the same repo above to assert the generated VCs' schemas.

### Challenge

What's another good place we could add this schema check in our codebase? Ideally, VC's template should from the schema, but that's out of the scope of this task. 

I found that this schema check faces the same main blocker of "we can't generate all VCs in development due to the lack of right accounts/ data provider setup"


tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -38,7 +37,7 @@ export async function assertFailedEvent(
     type EventLike = Parameters<typeof isFailed>[0];
     const ievents: EventLike[] = events.map(({ event }) => event);
     const failedEvent = ievents.filter(isFailed);
-    /* 
+    /*
```
Kailai-Wang: Maybe beyond the reach of this PR, but this way of event assertion is a bit awful - would something like `context.api.events.identityManagement.LinkIdentityFailed.is(event)` and then checking `event.detail` work?

cc @Verin1005

jonalvarezz: Yes, it'd be great, but first we need to fix typescript to pick the right types. Same issue as: https://linear.app/litentry/issue/P-508/avoid-as-unknown-as-when-possible

Otherwise, the type guard on `.is(event)` won't work.

It took us some time to get that right on IDH 😄

0xverin: We're using is(event) approach in [di evm tests](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/integration-tests/di_evm_identity.test.ts#L166). So I feel like `as unknown as` might be being overused. I'll double-check when I start this task.

tee-worker/ts-tests/integration-tests/common/utils/vc-helper.ts
```diff
@@ -166,85 +166,3 @@ export const unconfiguredAssertions = [
         },
     },
 ];
-export const jsonSchema = {
```
0xverin, line 169: I tested it locally, and it works. In fact, this crude jsonSchema should have been removed and replaced long ago. It was created during the validation of the first VC request.

0xverin (commented):
Looks really good; I like this lib. >3

tee-worker/ts-tests/integration-tests/common/utils/vc-helper.ts
jonalvarezz: Thanks @Verin1005 , if there isn't anything against, can I get the approval check to merge it? 😄

0xverin (approved):

PR #2452, opened 2024-01-31
feat: bitacross worker<>parachain external communication

felixfaisal:
So this PR is supposed to be the pathway for communication between Worker and Parachain and External Communication 
The things I've implemented are: 
- Placeholder `bit_across` Pallet 
- Placeholder `bit_across` extrinsic 
- Indirect Call filter for single extrinsic of the placeholder pallet 
- Separate threads with threadpools for handling JSONRPC requests 

I'm not sure if this PR is sufficient because I don't know the underlying or next implementation details, I will refactor the PR based off any specific requirements needed by the bitacross team.

PR #2505, opened 2024-02-20
Always init GLOBAL_SCHEDULED_ENCLAVE

kziemianek:
* initializes `GLOBAL_SCHEDULED_ENCLAVE` for any worker type, previously it was initialized only in case of sidechain worker.

### DIFF
--- a/bitacross-worker/README.md
+++ b/bitacross-worker/README.md
@@ -186,7 +186,7 @@ They take single argument representing raw payload bytes to sign.
 #### Example usage
 
 ```bash
-./bitacross-cli trusted -m 7ppBUcnjGir4szRHCG59p2dTnbtRwKRbLZPpR32ACjbK request-direct-call-sign-bitcoin 00
+./bitacross-cli trusted request-direct-call-sign-bitcoin 00
 ```
 
 ### Obtaining data from parachain's teebag pallet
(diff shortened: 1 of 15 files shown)


BillyWooo (approved):
LGTM

felixfaisal (approved):
LGTM

Kailai-Wang (approved):
While I have no objection to the PR itself, I'm wondering if we actually need/use `GLOBAL_SCHEDULED_ENCLAVE` in bc-worker, except for `state_getMrenclave`.

We shouldn't care about the scheduled_enclave in bc-worker, the validity of mrenclave will be checked upon enclave registration 🤔 

If `state_getMrenclave` is the only reference, maybe we should use other ways to retrieve mrenclave in the rpc handler

kziemianek:
> While I have no objection to the PR itself, I'm wondering if we actually need/use `GLOBAL_SCHEDULED_ENCLAVE` in bc-worker, except for `state_getMrenclave`.
> 
> We shouldn't care about the scheduled_enclave in bc-worker, the validity of mrenclave will be checked upon enclave registration 🤔
> 
> If `state_getMrenclave` is the only reference, maybe we should use other ways to retrieve mrenclave in the rpc handler

You are right, it was [raised](https://github.com/litentry/litentry-parachain/pull/2452#discussion_r1475670688) before. 
I'll remove all related code to scheduled enclave to avoid ambiguity.

PR #2506, opened 2024-02-20
bitacross remove evm feature

kziemianek:
* removes `evm` feature from bitacross worker

This is the last big PR related to unused features removal from bitacross worker.


BillyWooo (approved):
LGTM

PR #2507, opened 2024-02-20
fix state_getMrenclave

kziemianek:
* return mrenclave from `MY_MRENCLAVE` cache


Kailai-Wang (approved):
Nice improvement, thanks for the unittest too

PR #2508, opened 2024-02-20
P-328 Make unconfiguredAssertions pass the test

0xverin:
### Context

resolves p-328


kziemianek (approved):

Traf333 (commented):
Overall looks good, but seems mock server doesn't work as expected since tests are failing. Once they are green :shipit:

0xverin:
> Overall looks good, but seems mock server doesn't work as expected since tests are failing. Once they are green :shipit:

~~Unfortunately ci still fails but I can confirm that everything is fine locally. any ideas?~~
I think I got the reason.

Kailai-Wang:
The error says:
```
NoderealError("Nodereal response error: Server returned non-success status: 400, details: Bad Reques
```

0xverin:
> The error says:
> 
> ```
> NoderealError("Nodereal response error: Server returned non-success status: 400, details: Bad Reques
> ```

I think I know what the reason is, apikey use defalut value [here](https://github.com/litentry/litentry-parachain/pull/2508/files#diff-8f944cd5bdc2ce343a7b25fb2e5e2d96afa006c82b6dce2d6482c4a5ffa1c433L132)

PR #2510, opened 2024-02-20
Use `sign_prehash_recoverable` for ETH signing

Kailai-Wang:
### Context

Using the test vector from colleagues.

The DirectCall interface is adapted to enforce `[u8; 32]` type too.


felixfaisal (approved):
It looks good, So we are using `sign_prehash_recoverable` and we are sending a pre-hashed message for consistency.

BillyWooo (approved):
LGTM

PR #2512, opened 2024-02-21
Bump docker/login-action to v3

Traf333:
### Context

We observed node16 within GHA actions again https://github.com/litentry/litentry-parachain/actions/runs/7967753339 and this PR is tend to fix it


Kailai-Wang (approved):
Thank you!

PR #2513, opened 2024-02-21
P-508 Avoid using as sometype

0xverin:
### Context

- Remove `as sometype` for `api.createType`.
- Checked the usage of `toHuman()`, currently `toHuman()` is mostly used for for final assertions of content, there is no usage of `toHuman() as sometype`.

The clien-api has been properly configured for TypeScript types. Whether it's `parachain-ap`i or `sidechain-api`, as long as the import is correct, api.createType can correctly create types without the need for `as sometype.`

also resolved https://github.com/litentry/litentry-parachain/pull/2502#discussion_r1495017714, this is an unnecessary method, just remove it; we already have assertions in place [here](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/integration-tests/di_substrate_identity.test.ts#L316).


Kailai-Wang (approved):
Looks like a nice improvement for me!

Waiting for reviews from others :)

jonalvarezz (approved):
fantastic!

PR #2515, opened 2024-02-21
Fix CI worker integration tests

Traf333:
this PR fixes CI as during ts-tests refactoring, the configuration was wrong


0xverin (approved):
Thanks for the quick fix, LGTM. Just a minor detail, in the [readme](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/README.md), it's labeled as 'run test xx.ts,' so it might be better to keep it consistent here?

PR #2516, opened 2024-02-21
fix: add metrics for isolated request-vc

felixfaisal:
Adding metrics for the following 
1. Succesfull VC Request Count 
2. Failed VC Request Count 
3. Time taken for a VC Build


tee-worker/core-primitives/enclave-metrics/src/lib.rs
```diff
@@ -44,4 +44,7 @@ pub enum EnclaveMetric {
 	SidechainSlotStfExecutionTime(Duration),
 	SidechainSlotBlockCompositionTime(Duration),
 	SidechainBlockBroadcastingTime(Duration),
+	VcBuildTime(String, Duration),
+	SuccessfullVCIssuance,
```
kziemianek, line 48: Can you make it consistent and use `Vc` instead of `VC` ?

tee-worker/service/src/prometheus_metrics.rs
```diff
@@ -88,6 +88,15 @@ lazy_static! {
 	static ref ENCLAVE_SIDECHAIN_BLOCK_BROADCASTING_TIME: Histogram =
 		register_histogram!("litentry_worker_enclave_sidechain_block_broadcasting_time", "Time taken to broadcast sidechain block")
 			.unwrap();
+	static ref VC_BUILD_TIME: HistogramVec =
+		register_histogram_vec!("litentry_worker_vc_build_time", "Time taken to build a vc", &["variant"])
+			.unwrap();
+	static ref VC_SUCCESSFULL_RPC_CALLS: Counter =
```
kziemianek: I'm not entirely convinced that we should associate this name with RPC.
What do you think about `SUCCESSFULL_VC_ISSUANCE_TASKS`?

kziemianek (approved):

tee-worker/core-primitives/enclave-metrics/src/lib.rs
felixfaisal: Updated ✅

tee-worker/service/src/prometheus_metrics.rs
felixfaisal: Updated ✅

PR #2517, opened 2024-02-22
fix: change nft holder credentials type name to 'Generic NFT Holder'

wli-pro:
### Context

There is existing VC type being NFT Holder, although parachain doesn't have such unique constraints, FE does assume so, hence changing the VC type of the newly published one.

### Labels

### Testing Evidences


BillyWooo (approved):
It should be a breaking change for FE.

wli-pro:
> It should be a breaking change for FE.

FE is having issue without this PR, so it should be fine.

PR #2518, opened 2024-02-22
Fix typo add -> aad

Kailai-Wang:
### Context

A trivial PR thanks to bc colleagues.

Somehow we haven't got any runtime error even though it's correctly written in the definitions:
https://github.com/litentry/litentry-parachain/blob/78c79c949206acf1360b7dea7ea894c27decacbb/tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts#L17

Maybe it worked because of default values.


jonalvarezz (approved):
lol, I think it works because most likely createType defaults non-specified properties to undefined anyway.

PR #2520, opened 2024-02-23
Move achainable types to a separate file to make it as clear as other types

outofxxx:
as topic.


Kailai-Wang (approved):
It makes sense.

But now that we are talking about refactoring, could we pack everything assertion-related to `pritives/core/src/assertion` mod, or vc mod?

It's just a bit strange that e.g. `vip3.rs` is on the same level as `identity` and `vc` - and we can only have more and more such data provider primtiives..

PR #2523, opened 2024-02-26
Refactor ClassOfYear name type to enum

outofxxx:
This PR only deals with the `ClassOfYear` Achainable type alone, if there are no issues, the subsequent PR will change the type of the Name field from `string` to `enum` for the remaining Achainable types.

**Also, this is a breaking change.**


jonalvarezz:
Hi @zhouhuitian , can you elaborate on how clients are impacted so we can follow up and update client-sdk correspondingly?

Kailai-Wang:
> Hi @zhouhuitian , can you elaborate on how clients are impacted so we can follow up and update client-sdk correspondingly?

I guess it's the type change for `name` - it uses enum instead of string now.
It will affect how the client constructs the vc request.

outofxxx:
> Hi @zhouhuitian , can you elaborate on how clients are impacted so we can follow up and update client-sdk correspondingly?


Yes, after this modification, when constructing the request, the "name" field no longer needs to be entered as characters; you can only select from the provided enum options.

0xverin (commented):
overall looks good, I only checked type-definitions

jonalvarezz (approved):

jonalvarezz:
fyi: I created this sub-ticket to track the client-sdk change https://linear.app/litentry/issue/P-649/client-sdk-update-achainable-class-of-year-request-payload

BillyWooo:
Maybe let's hold on this topic for a moment.

Kailai-Wang:
Shall we continue this PR?

kziemianek:
> Shall we continue this PR?

This PR is open since February, what's the decision?

BillyWooo:
Close it for the moment:
1. We will take over Achainable service.
2. We will move to Solidity solution.

PR #2524, opened 2024-02-26
Eliminate warnings in tests mod

outofxxx:
as topic.


Kailai-Wang (commented):
Thanks - does it mean the tests are not run in CI?

BillyWooo:
> Thanks - does it mean the tests are not run in CI?

I think this is mainly building warnings. Because of not used return value.

BillyWooo (approved):

PR #2525, opened 2024-02-26
Should use non-compact sbn in scheduled_enclave

Kailai-Wang:
### Context

As topic, a bug found during the upgrade.

### DIFF
--- a/bitacross-worker/app-libs/parentchain-interface/src/indirect_calls/litentry/scheduled_enclave.rs
+++ b/bitacross-worker/app-libs/parentchain-interface/src/indirect_calls/litentry/scheduled_enclave.rs
@@ -39,7 +39,7 @@ impl<Executor: IndirectExecutor<TrustedCallSigned, Error>>
 	fn dispatch(&self, _executor: &Executor, _args: Self::Args) -> Result<()> {
 		debug!("execute indirect call: SetScheduledEnclave, worker_type: {:?}, sidechain_block_number: {:?}, mrenclave: {:?}", self.worker_type, self.sbn, self.mrenclave);
 		if self.worker_type == WorkerType::BitAcross {
-			GLOBAL_SCHEDULED_ENCLAVE.update(self.sbn.into(), self.mrenclave)?;
+			GLOBAL_SCHEDULED_ENCLAVE.update(self.sbn, self.mrenclave)?;
 		} else {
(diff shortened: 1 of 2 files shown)


bitacross-worker/app-libs/parentchain-interface/src/indirect_calls/litentry/scheduled_enclave.rs
```diff
@@ -28,7 +28,7 @@ use log::*;
 #[derive(Debug, Clone, Encode, Decode, Eq, PartialEq)]
 pub struct SetScheduledEnclaveArgs {
 	worker_type: WorkerType,
-	sbn: codec::Compact<SidechainBlockNumber>,
```
felixfaisal, line 31: Hey @Kailai-Wang I'm just curious on what the bug was and how it was discovered, does it affect other places where we are using `compact` encoding.

Kailai-Wang: Hi @felixfaisal - it's caused by parameter type mismatch: int `teebag.set_scheduled_enclave`, the `sbn` isn't compact encoded (so just plain `u64`), so you will get a decode error here without this fix.

It's easily reproducible when you try to set or remove scheduled enclave via indirect call.

felixfaisal: I think in `teerex` we are/were using `compact` - https://github.com/litentry/litentry-parachain/blob/3548f09b7ff80c2f8f9579d0c100cdff0846220e/pallets/teerex/src/lib.rs#L501 

It goes back to this issue - https://github.com/litentry/litentry-parachain/issues/1433 because of which we made the extrinsic parameter compact, and now we are kind of going back. Should we instead update `teebag` to use compact, or do you think there is no significant advantage in using `compact` encoding for our usecase currently?

Kailai-Wang: The latter :) compact encoding a u64 won't save us a lot but it's quite error-prone. We had the case where we forgot this compact encoding and had problems. It caused some trouble in ts-test.

That said maybe we should simply use plain encoding

felixfaisal (approved):
LGTM

PR #2529, opened 2024-02-27
Remove teerex/sidechain runtime integration

Kailai-Wang:
### Context

This PR removes teerex/sidechain from runtime completely.
I didn't bother writing runtime state migration (to clean them up) as these pallets only existed in test parachain so far.

I also moved `test-util` pallet to `teebag`.

The `tee-dev` and `skip-ias-verify` features are removed now as they are not relevant anymore.

The `fast-runtime` feature is not built by default (but still included in CI) - cc @Traf333


pallets/vc-management/Cargo.toml
```diff
@@ -26,25 +26,25 @@ pallet-teebag = { path = "../teebag", default-features = false }
...
-test-utils = { path = "../test-utils" }
+pallet-group = { path = "../../pallets/group" }
+pallet-teebag = { path = "../teebag", features = ["test-util"] }
+
 
 [features]
 default = ["std"]
 runtime-benchmarks = [
-    "frame-benchmarking/runtime-benchmarks",
+    "frame-benchmarking?/runtime-benchmarks",
```
Traf333: is it typo or specific syntax?

Kailai-Wang: Thanks Igor, it's not a typo - but it's wrong :)

We should have enabled this by default - I'll fix it

BillyWooo (approved):
LGTM

PR #2530, opened 2024-02-27
Rename vc-issuance crate

Kailai-Wang:
### Context

As topic, analog to stf-task sender and receiver.

`lc` means `litentry-core`, it exists in the folder path already, no need to start the subfolder with `lc-*`. Let me know if that makes sense.


felixfaisal (approved):
LGTM

kziemianek (approved):

PR #2532, opened 2024-02-27
P-422 add a batched request vc method

BillyWooo:
This PR provides a new functionality of handling `batch_request_vc`, along with existed `request_vc`. 
```
request_vc(Identity, Identity, Assertion, Option<RequestAesKey>, H256),
batch_request_vc(Identity, Identity, Vec<Assertion>, Option<RequestAesKey>, H256),
```
This new feature should not break existed functionality. But provide a new possibility that new `batch_request_vc` can support a vector of assertions as input. 

Both `request_vc` and `batch_request_vc` go with the same rpc method `author_requestVc` as before. 
- `request_vc` will expect: 
    - one regular trustedcall submitted response (similar like link_identity)
    - one result response. (error regarding processing assertion is a kind of result)
- `batch_request_vc` will expect: 
    - one regular trustedcall submitted response (similar like link_identity)
    - the same number of input assertion vector length result response. (error regarding processing assertion is a kind of result)
 
`result response` is defined like this:
```
#[derive(Debug, Encode, Decode, Clone)]
pub struct RequestVcResultOrError {
	pub payload: Vec<u8>,
	pub is_error: bool,
	pub idx: u8,
	pub len: u8,
}

pub struct RequestVCResult {
	pub vc_payload: AesOutput,
	pub pre_mutated_id_graph: AesOutput,
	pub pre_id_graph_hash: H256,
}
```
And the existed `struct RequestVCResult` is encoded into `payload` of `struct RequestVcResultOrError`.

There are two kinds of result scenarios:
1. Normal, `is_error==FALSE && len > 0` -> a valid assertion response; use `idx` and `len` to indicate which assertion it is. Expect total response number == assertion vector length (request_vc: len is 1).
2. Assertion Error, `is_error==TRUE && len > 0` -> error from processing assertion; use `idx` and `len` to indicate which assertion it is. 
3. Pre-Assertion Error, `is_error==TRUE && len==0` -> error happens before starting to process any assertion, e.g. 
    - Wrong trusted call (not request_vc, nor request_batch_vc)
    - decoding error (can't decode out trusted call, no idea what's the input)
    - can't verify signature (illegal input)
    - failed to retrieve shielding key
    - failed to get mrenclave
   
The first two will expect total response number == assertion vector length (request_vc: len is 1). 
For the third one, there will be only one response.


Kailai-Wang (commented):
A naive question before going into the review - how did you test it? CLI? ts-test?

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -154,30 +162,122 @@ where
...
+			node_metadata_repo,
+			tc_sender,
+			tcs.call.clone(),
+			1,
+		);
+		send_vc_response(sender, context, response);
+	} else if let TrustedCall::batch_request_vc(signer, who, assertions, maybe_key, req_ext_hash) =
+		tcs.call
+	{
+		let pool = ThreadPool::new(4);
```
felixfaisal: Should we limit the batch-vc-requests to 4, to avoid blocking time here for the pool to be free when the number of assertions are greater than 4, wdyt? @BillyWooo

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -154,30 +162,122 @@ where
...
+			);
+			let context_pool = context.clone();
+			let extrinsic_factory_pool = extrinsic_factory.clone();
+			let node_metadata_repo_pool = node_metadata_repo.clone();
+			let tc_sender_pool = tc_sender.clone();
+			let shard = request.shard;
+			let sender_clone = sender.clone();
+
+			pool.execute(move || {
+				let response = process_single_request(
```
felixfaisal: So every single request we are submitting a callback extrinsic upon success, Can we batch them as well for the batch request-vc? wdyt? Could be useful when we implement callback extrinsics for failure as well.

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -333,27 +340,64 @@ where
...
-			log::error!("Received error in jsonresponse: {:?} ", e);
-			Err(e)
-		},
-		Err(_) => {
-			// This case will only happen if the sender has been dropped
-			Err("The sender has been dropped".to_string())
-		},
+	let mut req_cnt = 0u8;
+	let mut req_num = 0u8;
+	loop {
```
felixfaisal: We can instead use `while receiver.recv()`

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -333,27 +340,64 @@ where
...
+		req_cnt += 1;
+		match receiver.recv() {
+			Ok(Ok(response)) =>
+				if let Ok(vc_res) = VCResponse::decode(&mut response.as_slice()) {
+					req_num = vc_res.vc_req_num;
+					author.send_rpc_response(hash, vc_res.vc_res_payload, req_cnt < req_num);
+					if !req_cnt < req_num {
```
felixfaisal: I don't think we have to manually count this, If all the senders in the channel are dropped, it signals it. https://docs.rs/futures/latest/futures/channel/mpsc/index.html#disconnection

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -119,6 +119,8 @@ pub enum TrustedCall {
 	#[cfg(not(feature = "production"))]
 	#[codec(index = 5)]
 	remove_identity(Identity, Identity, Vec<Identity>),
+	#[codec(index = 6)]
+	batch_request_vc(Identity, Identity, Vec<Assertion>, Option<RequestAesKey>, H256),
```
felixfaisal: I think we should use a `BoundedVec` to limit the number of assertions.

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
BillyWooo: Thanks for the doc. But the count here is more for the connection token, which manages (do_watch) socket to send out result. Wdyt?

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
BillyWooo: Thought about it before. But there the use case would easily result in a "select all" option from FE; then a big vector of assertions (e.g. 30) come inside. If we combine all result together, that would too big as one result. That's the reason to split it and send back result separately.
Otherwise, it would much simpler only to extend the old arch to process vector of assertions, gather result together, and return. Only need to touch one file is enough.

felixfaisal: The FE is not using callback extrinsics for anything, It's only meant for logging. We shouldn't allow a big vector of assertion, which is my other comment.

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
felixfaisal: Isn't the count being used to exit the loop, when all requests are received? Also if we do end up using this count could you use something like `req_cnt >= req_num`

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
```diff
@@ -37,6 +37,7 @@ export default {
                 deactivate_identity: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, Option<RequestAesKey>, H256)",
                 activate_identity: "(LitentryIdentity, LitentryIdentity, LitentryIdentity, Option<RequestAesKey>, H256)",
                 request_vc: "(LitentryIdentity, LitentryIdentity, Assertion, Option<RequestAesKey>, H256)",
+                batch_request_vc: "(LitentryIdentity, LitentryIdentity, Vec<Assertion>, Option<RequestAesKey>, H256)",
```
jonalvarezz: Order is important here. It should go after remove_identity. Same as in `tee-worker/app-libs/stf/src/trusted_call.rs`

jonalvarezz (commented):
Thanks, Yang,


I left one comment that may be the culprit of ts-tests failing, as no new tests have been added or modified. 

I look forward to seeing the dedup and index for the responses. It'd be great if we have the corresponding ts-tests as well.

tee-worker/app-libs/stf/src/trusted_call_result.rs
```diff
@@ -100,4 +100,6 @@ pub struct RequestVCResult {
 	// they should be referenced/used only when the client's local IDGraph is empty
 	pub pre_mutated_id_graph: AesOutput,
 	pub pre_id_graph_hash: H256,
+	// can be ignored when it's for one vc request
+	pub batch_vc_idx: u8,
```
jonalvarezz: should we move this to a higher layer like `WorkerRpcReturnValue`? I'm thinking of responses that are not `RequestVCResult` like errors or further batch methods we want to enable

tee-worker/ts-tests/integration-tests/dr_vc.test.ts
```diff
@@ -147,24 +148,39 @@ describe('Test Vc (direct request)', function () {
...
+            }
 
             const isVcDirect = true;
-            // const res = await sendRequestFromTrustedCall(context, teeShieldingKey, requestVcCall, isVcDirect);
-
             // Instead of waiting for final response we will listen all responses from the call
             const onMessageReceived = async (res: WorkerRpcReturnValue) => {
                 // if response is a A1 or A2, etc....
-                await assertVc(context, aliceSubstrateIdentity, res.value);
+                const vcresponse = context.api.createType('VCResponse', res.value);
```
jonalvarezz: I'm a bit iffy on the `VCResponse` naming here:

`WorkerRpcReturnValue` contains `VCResponse` which contains either `RequestVCResult` if it's `len` prop is > 0, or String with the error msg, otherwise.

How about something more idiomatic like `RequestVcResultOrError`.

Also do I understand correctly that adding the `idx` into `WorkerRpcReturnValue` is not possible and thus we had to introduce a new layer called `VCResponse`?

BillyWooo: Hi @jonalvarezz `idx` is also include within this VCResponse. Same layer as `len`. Is it OK?

BillyWooo: Yes. I can renaming as you suggested.

jonalvarezz: > idx is also include within this VCResponse. Same layer as len. Is it OK?

That was my second question. VCResult didn't exist before so we are purposely adding it. Do we need to because adding idx to `WorkerRpcReturnValue` was not possible?

tee-worker/ts-tests/integration-tests/dr_vc.test.ts
```diff
@@ -147,30 +148,58 @@ describe('Test Vc (direct request)', function () {
...
-                context.mrEnclave,
-                context.api.createType('Index', nonce),
-                new PolkadotSigner(context.substrateWallet.alice),
-                aliceSubstrateIdentity,
-                context.api.createType('Assertion', assertion).toHex(),
-                context.api.createType('Option<RequestAesKey>', aesKey).toHex(),
-                requestIdentifier
-            );
+            let requestVcCall;
+            if (Array.isArray(assertion)) {
```
0xverin, line 152: The di and dr vc test codes are basically the same, so you may need to change the code of di to be the same as dr's, otherwise I think ci should not pass.  cc @BillyWooo 

Since they only have a difference between isVcDirect true and false, do we need to consider merging them? Otherwise, in the future changing the same code will look too bloated. cc @Traf333

0xverin (commented):
Thanks! Please copy the code of `dr_vc.test.ts` to di_vc.tests.ts to let ci pass. As I mentioned in comment, hand it over to ts-tests side.  :p

tee-worker/ts-tests/integration-tests/dr_vc.test.ts
BillyWooo: Let's visit this topic together with **Whether we should remove `request_vc`, and only use `request_batch_vc` in the future?**. Then we maybe do some test refactor together.

tee-worker/core-primitives/top-pool/src/watcher.rs
```diff
@@ -139,6 +139,15 @@ where
 		}
 	}
 
+	pub fn send_rpc_response(&mut self, encoded_value: Vec<u8>, do_watch: bool) {
```
Kailai-Wang: Is this function actually used?

tee-worker/sidechain/rpc-handler/Cargo.toml
```diff
@@ -8,6 +8,9 @@ edition = "2021"
 # sgx dependencies
 sgx_tstd = { branch = "master", git = "https://github.com/apache/teaclave-sgx-sdk.git", optional = true }
 
+# substrate deps
+sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
```
Kailai-Wang, line 12: it's a good practice to always add `sp-runtime/std` to the `std` feature list

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -333,27 +340,59 @@ where
...
-		Ok(Err(e)) => {
-			log::error!("Received error in jsonresponse: {:?} ", e);
-			Err(e)
-		},
-		Err(_) => {
-			// This case will only happen if the sender has been dropped
-			Err("The sender has been dropped".to_string())
-		},
+	let mut req_cnt = 0u8;
+	while let Ok(res) = receiver.recv() {
```
Kailai-Wang: Something to improve here is to properly handle the error - so far the loop will be ended silently. It will cause the client to wait indefinitely

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -333,27 +340,59 @@ where
...
+
+	// Use this hash to trigger the storage of connection_registry
+	let hash = request.using_encoded(|x| BlakeTwo256::hash(x));
+
+	thread::spawn(move || request_vc_inner1(author, request, hash));
+
+	Ok(hash)
+}
 
+fn request_vc_inner1<R, TCS, G>(author: Arc<R>, request: AesRequest, hash: H256)
```
Kailai-Wang: it seems that `hash` is redundant (as it can be computed from `request`)

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -154,30 +171,145 @@ where
...
+	{
+		// Filter out duplicate assertions
+		let mut seen: HashSet<H256> = HashSet::new();
+		let mut unique_assertions = Vec::new();
+		for assertion in assertions.into_iter() {
+			let hash = H256::from(blake2_256(&assertion.encode()));
+			if seen.insert(hash) {
+				unique_assertions.push(assertion);
+			}
+		}
```
Kailai-Wang: I think HashSet itself is enough to dedup, something like `let set: HashSet<_> = vec.drain(..).collect();`

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -333,27 +340,59 @@ where
...
-		},
+	let mut req_cnt = 0u8;
+	while let Ok(res) = receiver.recv() {
+		match res {
+			Ok(response) => {
+				req_cnt += 1;
+				if let Ok(vc_res) = RequestVcResultOrError::decode(&mut response.as_slice()) {
+					author.send_rpc_response(hash, response, req_cnt < vc_res.len);
+					if req_cnt >= vc_res.len {
+						break
```
Kailai-Wang: Does it actually close the ws connection?

If not, could we end up exhausting the connection limit quite soon?

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -333,27 +340,59 @@ where
...
+{
 	let payload = get_request_payload(params)?;
-	let request = AesRequest::from_hex(&payload)
-		.map_err(|e| format!("AesRequest construction error: {:?}", e))?;
+	let request = AesRequest::from_hex(&payload).map_err(|e| format!("{:?}", e))?;
+
+	// Use this hash to trigger the storage of connection_registry
+	let hash = request.using_encoded(|x| BlakeTwo256::hash(x));
+
+	thread::spawn(move || request_vc_inner1(author, request, hash));
```
Kailai-Wang: As mentioned, although unlikely - it's still better to make sure this thread only runs after the connection registry has been populated - can be followed up by another issue

tee-worker/litentry/core/vc-task/sender/src/lib.rs
```diff
@@ -37,13 +36,16 @@ use std::{
...
-	pub sender: oneshot::Sender<Result<Vec<u8>, String>>,
+	pub sender: MpscSender<Result<Vec<u8>, String>>,
 	pub request: AesRequest,
 }
 
-#[derive(Encode, Decode, Clone)]
-pub struct VCResponse {
-	pub vc_payload: Vec<u8>,
+#[derive(Debug, Encode, Decode, Clone)]
+pub struct RequestVcResultOrError {
```
Kailai-Wang: I actually prefer using enum to represent "result or error"

Kailai-Wang (commented):
The general idea looks good, thanks!

However, I'm not convinced by the two level thread-pool design, it seems to complicate the whole procedure.. why don't we deserialise the request and simply post each assertion request as task to the thread pool we had? There seems to be no demand of being grouped.

Btw we can remove `P 422` in the PR title - or rewrite it as `P-422:`

tee-worker/core-primitives/top-pool/src/watcher.rs
BillyWooo: Thanks. It should be removed.

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
BillyWooo: Yes. Once the receiver goes out of scope. The connection will be cut down automatically.

BillyWooo: It will.  Once `do_watch` is false, the wss connection will be cut down.

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
BillyWooo: add `receiver.recv()` failure checking.

BillyWooo:
> read pool we ha

Originally I had the same idea. But when analysing the use case, request will always come with vector of assertions. I don't have any profiling statistics to guarantee that all assertions will executed very fast; then the following new vector will be executed. So I made it two layers. At least requests coming from 4 different users can be processed in parallel. If request from one user has a bigger vector, it won't block other users too much.

Kailai-Wang:
> > read pool we ha
> 
> Originally I had the same idea. But when analysing the use case, request will always come with vector of assertions. I don't have any profiling statistics to guarantee that all assertions will executed very fast; then the following new vector will be executed. So I made it two layers. At least requests coming from 4 different users can be processed in parallel. If request from one user has a bigger vector, it won't block other users too much.

That's effectively the same - as long as the assertions aren't user-specific and don't have to be bound to be executed in one group, it's equivalent to simply increase the worker number IMO.

kziemianek:
Can we have a metric for pending vc tasks ?

Kailai-Wang:
Let's try to get this PR merged today so that it's not prolonged to next week :)

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -68,11 +68,13 @@ use sp_core::{
 	ed25519,
 };
 use sp_io::hashing::blake2_256;
-use sp_runtime::MultiAddress;
+use sp_runtime::{traits::ConstU32, BoundedVec, MultiAddress};
 use std::{format, prelude::v1::*, sync::Arc};
 
 pub type IMTCall = ita_sgx_runtime::IdentityManagementCall<Runtime>;
 pub type IMT = ita_sgx_runtime::pallet_imt::Pallet<Runtime>;
+pub type MaxAssertionLength = ConstU32<32>;
```
felixfaisal, line 76: This should not be greater than `4` , I don't think we even have `32` assertions.

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -333,27 +341,94 @@ where
...
+{
 	let payload = get_request_payload(params)?;
-	let request = AesRequest::from_hex(&payload)
-		.map_err(|e| format!("AesRequest construction error: {:?}", e))?;
+	let request = AesRequest::from_hex(&payload).map_err(|e| format!("{:?}", e))?;
+
+	// Use this hash to trigger the storage of connection_registry
+	let hash = request.using_encoded(|x| BlakeTwo256::hash(x));
 
+	thread::spawn(move || request_vc_inner(author, request, hash));
```
felixfaisal: This does not look right, We are creating a thread every time we receive a request?

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -154,30 +174,158 @@ where
...
+					maybe_key,
+					req_ext_hash,
+				);
+				let extrinsic_factory_pool = extrinsic_factory.clone();
+				let node_metadata_repo_pool = node_metadata_repo.clone();
+				let tc_sender_pool = tc_sender.clone();
+				let shard = request.shard;
+
+				pool.execute(move || {
+					let response = process_single_request(
```
felixfaisal: There is an edge case here, That is if the `identity-graph` is empty, We will try to create `link-identity` trusted-call for all of these requests and submit, So then each response will have `mutated_id_graph`. 
CC: @Kailai-Wang

felixfaisal (commented):
From just generally going through the PR, I can see a few edge cases missed. 
- I would have preferred a separate handle for `batch_requests` and a separate one for `single_requests`. The coupling seems to add more complications.
- As @Kailai-Wang mentioned we could just set the total number of worker threads to be 20 (worst case is `20` threads not `16` ) 
- There is a lingering `thread::spawn` that seems to be forgotten

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
BillyWooo: Yes. Because `request_vc_inner` is going to process the request and wait for the response.

Kailai-Wang: I think what Faisal means is it could easily spawn many concurrent threads - it could cause problems especially for enclave where the number of threads are limited

Kailai-Wang: I think it should directly post the task to thread pool and return

Kailai-Wang: I think it depends on how jsonrpc handles the handling of sync method internally - the while loop might need to be executed on threads in the first threadpool

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
Kailai-Wang: OK not if you want to find out the duplicated assertion and return the error for it...

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
Kailai-Wang: This is the major change we need before we can merge it

tee-worker/litentry/core/vc-task/sender/src/lib.rs
Kailai-Wang: Maybe we should create an issue and solve it in another PR - in case you don't have any capacity for this one

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
Kailai-Wang: Good catch.

I'd argue the impact is insignificant though - it can be mitigated on the client side @jonalvarezz . I don't know how the sdk currently deals with it, but in general you could igore the `pre_mutated_id_graph` if the local hash is equal to `pre_id_graph_hash`.

We could adjust the parachain side so that only the last response contain meaningful values, but again what if that vc request ends up being an error? The client would never get a response with these fields, which doesn't seem to be right either. So I'd say let's leave it as it is.

One improvement that I can see is we only need to submit `TrustedCall::maybe_create_id_graph` once per batch. In general, all batch level work should be extracted from `process_single_request`, so it's like:
```
- prepocessing
- n * vc request handling
- postprocessing
```

Please create an issue for that

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
BillyWooo: Thanks. I see the problem. @felixfaisal good catch.

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
jonalvarezz: > I'd argue the impact is insignificant though - it can be mitigated on the client side @jonalvarezz .

Furthermore, `pre_mutated_id_graph` isn't used so far. IDH rely on the `get_id_graph_hash` getter.

tee-worker/app-libs/stf/src/trusted_call_result.rs
```diff
@@ -101,3 +101,12 @@ pub struct RequestVCResult {
...
 	pub pre_id_graph_hash: H256,
 }
+
+#[derive(Debug, Encode, Decode, Clone)]
+pub struct RequestVcResultOrError {
+	pub payload: Vec<u8>,
+	pub is_error: bool,
+	// When `is_error` true, the following `idx` and `len` should be ignored.
+	pub idx: u8,
+	pub len: u8,
```
jonalvarezz: If this is the case, How should clients map the errors back to the corresponding assertion request?

Kailai-Wang: Yes I think in error cases it should still contain the expected `idx` and `len`

tee-worker/sidechain/rpc-handler/Cargo.toml
```diff
@@ -8,7 +8,11 @@ edition = "2021"
...
+
 # local dependencies
+ita-stf = { path = "../../app-libs/stf", default-features = false }
```
Kailai-Wang, line 15: Same as above - I don't find any references

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -45,14 +45,12 @@ use std::{
 	vec::Vec,
 };
 
-type Hash = sp_core::H256;
```
Kailai-Wang, line 48: Although I don't know why we change this - it should be fine too

tee-worker/sidechain/rpc-handler/Cargo.toml
```diff
@@ -8,7 +8,11 @@ edition = "2021"
 # sgx dependencies
 sgx_tstd = { branch = "master", git = "https://github.com/apache/teaclave-sgx-sdk.git", optional = true }
 
+# substrate deps
+sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
```
Kailai-Wang, line 12: Why do we need this crate here?

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -84,62 +91,300 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
+						let response = process_single_request(
+							shard_pool,
+							context_pool.clone(),
+							extrinsic_factory_pool,
+							node_metadata_repo_pool,
+							tc_sender_pool,
+							new_call,
+						);
+
+						//Totally fine to `unwrap` here. Because new item just added above.
```
Kailai-Wang, line 256: Space - comma - big/small letter - missing `is`/`was` : )

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -84,62 +91,300 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
+			},
+		};
+		if !tcs.verify_signature(&mrenclave, &request.shard) {
+			send_vc_response(
+				connection_hash,
+				context,
+				Err("Failed to verify sig".to_string()),
+				0u8,
+				0u8,
+				false,
```
Kailai-Wang, line 177: The readability could be better - we should refactor it together with `RequestVcResultOrError` - need an issue

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -84,62 +91,300 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
+				let response = process_single_request(
+					shard_pool,
+					context_pool.clone(),
+					extrinsic_factory_pool,
+					node_metadata_repo_pool,
+					tc_sender_pool,
+					tcs.call.clone(),
+				);
+
+				//Totally fine to `unwrap` here. Because new item just added above.
```
Kailai-Wang, line 203: Space - comma - big/small letter - missing `is`/`was` : )

Kailai-Wang: Actually it's not that safe to unwrap it - at least not from the code logic. You have a separate thread which runs periodic cleanup tasks, so it's possible that the task gets removed inbetween (for whatever reason)

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -84,62 +91,300 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
+			.filter(|(_, status)| {
+				status.time_stamp.elapsed() >= Duration::from_secs(wait_seconds)
+					&& status.indexes.len() as u8 == status.len
+			})
+			.map(|(key, _)| *key)
+			.collect();
+
+		for key in items_to_remove {
+			map.remove(&key);
+		}
```
Kailai-Wang, line 340: I think there's `retain` or `drain_filter`

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -84,62 +91,300 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
+							context_pool.clone(),
+							extrinsic_factory_pool,
+							node_metadata_repo_pool,
+							tc_sender_pool,
+							new_call,
+						);
+
+						//Totally fine to `unwrap` here. Because new item just added above.
+						let (size, len) =
+							req_registry_pool.update_item(connection_hash, idx as u8).unwrap();
```
Kailai-Wang, line 258: Is `len` alway equal to `assertion_len`? If so we don't need to retun `len` here

tee-worker/app-libs/stf/src/trusted_call_result.rs
BillyWooo: You are right. I removed this comment. The idx and len are still valid.

BillyWooo: Besides, when response is error, and both idx and len are 0: it means error happens before any assertion processing.

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -84,62 +91,300 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
-	request: &mut AesRequest,
+#[derive(Clone)]
+struct RequestRegistry {
+	status_map: Arc<Mutex<HashMap<H256, AssertionStatus>>>,
+}
+
+struct AssertionStatus {
+	pub time_stamp: Instant,
+	pub len: u8,
+	pub indexes: Vec<u8>,
```
Kailai-Wang, line 303: We could have better names, like `total` / `processed` etc

And it seems we don't need concrete indexes, just `processed_count`, do we?

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -84,62 +91,300 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
 
+	// use local registry to manage request reponse status
+	let req_registry = RequestRegistry::new();
+	let req_registry_cloned = req_registry.clone();
+	thread::spawn(move || loop {
+		// Doesn't harm anything to clean up slowly.
+		let wait_seconds = 200u64;
+		req_registry_cloned.clean_up(wait_seconds);
+		sleep(Duration::from_secs(wait_seconds));
+	});
```
Kailai-Wang, line 116: It sounds like something could be done via reference counting - like `Arc` or so - not in this PR though...

Kailai-Wang (commented):
I think we are close to merging it.

IIUC there's only one universal thread pool now, isn't it?

felixfaisal (approved):
It looks good to me, It's good to be merged if client side(@jonalvarezz) hasn't found anything troubling. However there are follow up issues that need to be addressed and created, Please follow up @BillyWooo

BillyWooo:
Auto-merged. 
Will have another follow up PR to address the improvement comments.

tee-worker/sidechain/rpc-handler/Cargo.toml
BillyWooo: will removed. 
It was added before the refactor. Missed out to remove.

tee-worker/sidechain/rpc-handler/src/direct_top_pool_api.rs
BillyWooo: Same reason as above. I think it's fine to keep it as it is.

tee-worker/sidechain/rpc-handler/Cargo.toml
BillyWooo: same reason as above. will remove it.

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
BillyWooo: Forget about them. Will replace.

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
BillyWooo: >  better names
Yes.
> concrete indexes
Originally planned to check idx and find any missing items. Looks like not necessary now.

PR #2533, opened 2024-02-28
fix: remove fees for TEE callback extrinsics

felixfaisal:
~This is a straightforward PR. We observed an `out of balance` for TEE-worker in one of our latest deployment. I went through three pallets namely `teebag` , `identity-management` and `vc-management`. I found two extrinsics in `teebag` that are not feeless and most likely the reason why we observed an `out of balance`. However I'm confused about `post-opaque-task` extrinsic, Are we actively using it or is there an update to it?~
[UPDATE] 
_One thing we can improve is:
prefer returning Pays::No over weight macro, as the latter will waive the fee even for failed extrinsic (which can be spammed)
see https://substrate.stackexchange.com/questions/3964/feeless-transactions-unsigned-vs-signed-extension__ 

I checked the extrinsics called by the Worker, only two in `teebag` use the weight macro

### DIFF
--- a/pallets/teebag/src/lib.rs
+++ b/pallets/teebag/src/lib.rs
@@ -528,7 +528,7 @@ pub mod pallet {
 		pub fn post_opaque_task(origin: OriginFor<T>, request: RsaRequest) -> DispatchResult {
 			let _ = ensure_signed(origin)?;
 			Self::deposit_event(Event::OpaqueTaskPosted { shard: request.shard });
-			Ok(().into())
+			Ok(())
 		}
 
 		#[pallet::call_index(21)]
--- a/pallets/teebag/src/lib.rs
+++ b/pallets/teebag/src/lib.rs
@@ -525,10 +525,13 @@ pub mod pallet {
 
 		#[pallet::call_index(20)]
 		#[pallet::weight((195_000_000, DispatchClass::Normal, Pays::Yes))]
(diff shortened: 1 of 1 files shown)


pallets/teebag/src/lib.rs
```diff
@@ -549,7 +549,7 @@ pub mod pallet {
 				block_hash,
 				task_merkle_root,
 			});
-			Ok(().into())
+			Ok(Pays::No.into())
```
Kailai-Wang, line 552: I don't think this is needed because of `#[pallet::weight((195_000_000, DispatchClass::Normal, Pays::No))]` in line 535, although this might be the preferred way.

The actual problem comes from the `pallet-identity-management` and `pallet-vc-management`, please double check it there.

See the issue description:
> Pallets to check:
> - teebag
> - identity-management
> - vc-management

Kailai-Wang (changes requested):

pallets/teebag/src/lib.rs
felixfaisal: I think i missed out on this detail, I checked both identity-management and vc-management, All the extrinsics created to be called by the tee-worker, are `Pays::No`.

Kailai-Wang (approved):
Thanks!

kziemianek (approved):

PR #2534, opened 2024-02-28
Enhance local setup/launch with python script

Traf333:
### Context

This PR introduces next changes:
* moves `local-setup` folder to the root folder
* adjusts `launch.py` 
    - to consider multi worker setup with correct offset
    - omits use of `--config` option, as hardcoded json files are too specific and some of them don't work as expected


docker/pnpm-lock.yaml
Traf333, line 1: I've got the changes for this file every time once I try local-docker mode 🤔

Kailai-Wang:
Cool attempt 👍 , can we keep both? 
For some people (like me) they know exactly what they want and are used to one-liner

Maybe have a command line option `-i` for interactive mode

Traf333:
> Cool attempt 👍 , can we keep both? For some people (like me) they know exactly what they want and are used to one-liner
> 
> Maybe have a command line option `-i` for interactive mode

@Kailai-Wang  by oneliner you want to select specific json config or do you mean to have ability to pass arguments with ports and stuff? 
Or keep default preset, like `Parachain in standalone mode + 1 worker`?

Asking because I've noticed that hardcoded json configs don't work stable and some of them already outdated. It sounds reasonable to have `-i` though 👍

Traf333:
By the way, if we are sticking to this approach, I would also like to include:
1. getting rid of the `bitacross-worker/local-setup` and other python scripts
2. removing  `*worker.json` config files as they are static and might not working well, when somebody else run infra on the server
3. updated Readme 
4. moved `local-setup` folder to the root directory

BillyWooo:
"removing *worker.json* ", what's the replacement? Besides, are we still using python files?

felixfaisal (commented):
Thanks @Traf333 , I think I'm a little bit biased, Here's what I think:
1. Majority of the changes felt like the same logic in python ported to js. I don't see the benefits of moving from python to a js script. I may have missed the discussion on this. 
2. No update in Readme, Missing steps of installation such as `yarn install` etc. 
3. I'm of the same opinion, that I generally prefer commands and flags over an interactive script, it's quicker to restart and make changes. So would be great if we can have both.

0xverin:
>I'm of the same opinion, that I generally prefer commands and flags over an interactive script, it's quicker to restart and make changes. So would be great if we can have both.

I agree with this opinion. :p

docker/pnpm-lock.yaml
0xverin: Version issue?

local-setup/.env
```diff
@@ -0,0 +1,53 @@
...
+
+# The following key/token are MANDATORY to give when running worker.
+# Otherwise request-vc might suffering from data provider error.
+TWITTER_AUTH_TOKEN_V2=
+DISCORD_AUTH_TOKEN=
+ACHAINABLE_AUTH_KEY=
+ONEBLOCK_NOTION_KEY=
+NODEREAL_API_KEY=
+GENIIDATA_API_KEY=
+
```
wli-pro: Might have missed this newly introduced config: https://github.com/litentry/litentry-parachain/blob/92d705fe5ef44d5f53aca31dd380830d78cd9fe8/local-setup/.env.dev#L34

Might worth check others too.

local-setup/utils/port.js
wli-pro, line 1: thought we are gonna stick to python only for now, or we still need use under certain cases?

local-setup/.env
Traf333: egh, this one should be gitignored. Thanks for the raising it 👍

local-setup/utils/port.js
Traf333: it's a leftover. Removing it, thanks

local-setup/.env
Traf333: .env file shouldn't be here at all, though for .env.dev probably we will add new variables

felixfaisal (approved):
Thanks @Traf333 For revamping the entire PR
I tested the following: 
1. local-binary 
2. local-binary-standalone 
3. local-binary with two workers (WOW!) 

It looks good from my side 🙌

0xverin (approved):
Thanks. I tested:

- local binary

- local-binary-standalone

LGTM!

BillyWooo:
Is the "remote" working mode still valid?

Traf333:
> Is the "remote" working mode still valid?

@BillyWooo yes it works as well, but there are couple of moments we should be aware of.
1. parachain should be run on the same machine
2. offset for remote should also be considered. if you run command with offset 44 then your parachain port should be 9988
3. without offset everything will work as before

if the remote mode would gain more popularity, in that case we can introduce extra argument for pointing to remote url

Traf333:
> > Is the "remote" working mode still valid?
> 
> @BillyWooo yes it works as well, but there are couple of moments we should be aware of.
> 
> 1. parachain should be run on the same machine
> 2. offset for remote should also be considered. if you run command with offset 44 then your parachain port should be 9988
> 3. without offset everything will work as before
> 
> if the remote mode would gain more popularity, in that case we can introduce extra argument for pointing to remote url

it seems I was wrong. We still shifting ports if 9944 is busy, :( so for now remote option is unavaliable. I'll prepare fix with adding extra argument for remote url

PR #2535, opened 2024-02-28
Use workspace dependencies for parachain

Kailai-Wang:
### Context

As topic.

It serves as a preface for the incoming polkadot dep update.


BillyWooo (approved):
Nice. LGTM.
Small question: when should we use `std` feature like this `litmus-parachain-runtime = { workspace = true, features = ["std"] }` ? How could we know it? Is there any other feature we need to take care?

felixfaisal (approved):

Kailai-Wang:
> Nice. LGTM. Small question: when should we use `std` feature like this `litmus-parachain-runtime = { workspace = true, features = ["std"] }` ? How could we know it? Is there any other feature we need to take care?

You only need that for the node compilation, we know that becuase there's no `default-features = false` in the original toml.
I believe all necessary features are already taken care of :)

PR #2538, opened 2024-02-29
fix(nft-holder): schema url

jonalvarezz:
As topic.

vc-jsonschema counterpart: https://github.com/litentry/vc-jsonschema/pull/22/


tee-worker/litentry/core/credentials/src/credential_schema.rs
```diff
@@ -104,6 +104,6 @@ pub fn get_schema_url(assertion: &Assertion) -> String {
 
 		Assertion::PlatformUser(_) => format!("{BASE_URL}/24-platform-user/1-0-0.json"),
 
-		Assertion::NftHolder(_) => format!("{BASE_URL}/26-nft-holder/1-0-0.json"),
+		Assertion::NftHolder(_) => format!("{BASE_URL}/25-nft-holder/1-0-0.json"),
```
Kailai-Wang, line 107: The `PlatformUser` and `NftHolder` are indexed 25 and 26 in SCALE-codec, respectively:
```
	#[codec(index = 25)]
	PlatformUser(PlatformUserType),

	#[codec(index = 26)]
	NftHolder(Web3NftType),
```

Although it's not obligatory to have the same numbers, could it cause any confusion? (also to any new entries)

jonalvarezz: Good point. 

Those indexes are overlooked by the client and not used at all. Not even Assertion type construction is needed explicitly (Polkadot-js does it under the hood). 

From that perspective, I don't think it will cause any confusion.

Looking at past schemas, most of them don't match. It wasn't intended from the beginning. But if you agree, we can skip schema 25 so from 26 onwards schema id and assertion index match. What do you think?

jonalvarezz: After a second thought. 

This change is available on production already and users already issued VCs where the url points to 26-nft-holder.

So, to keep it simple, we definitely need to match that on the vc-jsonschemas 😄

This can be closed and should be kept as is.

Thanks Kai!

PR #2540, opened 2024-03-01
feat: link identity in cli

felixfaisal:
So I've implemented `link-identity` in `cli` for web3 identities, with validation data payloads that can be used in production as well. For now this supports only via indirect-invocation, we can add this support for direct-invocation. 

You can try it by using the below command: 
```rust
./litentry-cli link-identity //Bob did:litentry:substrate:0xae5e2acb3f950b04e7fc31a1c531ed6d462224c98564af08009c0a790454c91b  litentry
```


wli-pro (approved):
LGTM

tee-worker/cli/src/base_cli/commands/litentry/mod.rs
```diff
@@ -15,3 +15,7 @@
 // along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
 
 pub mod id_graph_hash;
+pub mod link_identity;
+
+// TODO: maybe move it to use itp_node_api::api_client
```
kziemianek: Are you moving it or leaving here as it is ?

felixfaisal: This is an old comment, I'm leaving it here as it is 😅

BillyWooo (approved):

PR #2542, opened 2024-03-03
Pin the parachain-launch commit hash

Kailai-Wang:
### Context

As topic, the reason is I updated the `use-raw-binary` branch for the next polkadot upgrade - and using branch tip is probably not a good idea anyway.


Traf333 (approved):
Checked also what has changed there https://github.com/open-web3-stack/parachain-launch/commit/e5ccde9cbd94a43c3212b437c5aebcf9e8aa0474 🚀 

:shipit:

arieljbirnbaum (approved):

PR #2543, opened 2024-03-04
feat: test the failed parachain extrinsic handling in worker

felixfaisal:
So the idea of this PR was: 
1. Call `link-identity` for `Bob`  
2. Wait for it to be succesful and capture the id-graph hash 
3. Call `link-identity` but use a delegate signer, the extrinsic will fail since the delegate is not authorized 
4. We wait, and capture the id-graph hash, no change in id-graph-hash would signify that the failed extrinsic did not execute.


tee-worker/cli/src/base_cli/commands/litentry/link_identity.rs
```diff
@@ -55,7 +55,8 @@ impl LinkIdentityCommand {
 		let nonce = direct_api.get_next_nonce(&shard, &self.get_primary_account_id()).unwrap();
 
 		let signer = self.get_signer();
-		chain_api.set_signer(signer.clone().into());
+		let who: sr25519_core::Pair = get_pair_from_str(&self.account).into();
```
jonalvarezz, line 58: can you educate me how this line brings more value compared to before?

jonalvarezz (commented):
Naïve question: why not a ts-test? i.e. when should we prefer cli test over ts-test?

felixfaisal:
> Naïve question: why not a ts-test? i.e. when should we prefer cli test over ts-test?

A good rule of thumb would be if it is assigned to rust dev, then cli would be preferred xD 
However, we did have a few CI tests that make use of CLI, and have been removed, so it's not exactly reinventing the wheel since the support always has been there. This is basically my rationale.

tee-worker/cli/src/base_cli/commands/litentry/link_identity.rs
felixfaisal: The single line itself doesn't bring more value, but the changes in the file collectively alter the logic. So earlier we were using `let signer = self.get_signer()` and then using it to set the signer of the extrinsic and also in the extrinsic parameter. However, when using a delegate signer, we are setting the delegate signer as the parameter in the extrinsic. So to avoid this, I'm creating a separate `who` for the primary account, which will go to the `extrinsic` parameter and `signer` will be the signer of the extrinsic depending on what arg/flag has been provided.

jonalvarezz: got it, it makes sense now, thanks!

Kailai-Wang:
> Naïve question: why not a ts-test? i.e. when should we prefer cli test over ts-test?

This is actually a good one.
I'd say either will do its job, but rust CLI generally provides more flexibility and is simpler (from code) as it's closer to call existing rust functions. So achieving the same functionality requires more code in ts :D 

Upstream doesn't have ts-test btw, we need that because we want some e2e test and our F/E is ofc in ts

tee-worker/cli/lit_test_failed_parentchain_extrinsic.sh
```diff
@@ -0,0 +1,71 @@
...
+echo "Finished Linking identity to Bob"
+sleep 30
+
+echo "Capturing IDGraph Hash of Bob" 
+INITIAL_ID_GRAPH_HASH=$(${CLIENT} id-graph-hash did:litentry:substrate:0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48) || { echo "Failed to get ID Graph hash"; exit 1; }
+echo "Initial ID Graph Hash of Bob: ${INITIAL_ID_GRAPH_HASH}"
+
+SECOND_NEW_ACCOUNT=$(${CLIENT} new-account)
+echo "New Account created: ${SECOND_NEW_ACCOUNT}" 
+# Step 5: Link new account to Bob with Alice as delegate signer
```
Kailai-Wang: What's this step 5? :)

Kailai-Wang (commented):
Thanks! 

While I understand what the code tries to do, I'm not sure if it meets the test goal. We want to make sure that failed extrinsics **don't even get executed** - imagine you transfer 1m tokens to someone and it fails as you have insufficient balance, the indirect call executor should detect that the extrinsic fails and never try to parse and execute it.

This PR tests however, when an error is thrown out during the execution of the II dispatcher, it won't write any wrong state - which is something good to test, but not the same as what we wanted to test IMO

felixfaisal:
> This PR tests however, when an error is thrown out during the execution of the II dispatcher, it won't write any wrong state - which is something good to test, but not the same as what we wanted to test IMO

I don't think so it is testing the error thrown out during the execution of the `II` dispatch. It is doing the former. 
- So we first `link-identity` to `Bob` via the `II` providing the correct inputs, meaning it will not fail in `II` dispatch, (https://github.com/litentry/litentry-parachain/blob/94e09b2e4fccdb01989f78484fcfc25554c56dc7/tee-worker/cli/lit_test_failed_parentchain_extrinsic.sh#L47) We check the `id-graph-hash` to conclude that it executed succesfully. 
-  We create a second `link-identity` however we use `Eve` as delegate signer - https://github.com/litentry/litentry-parachain/blob/94e09b2e4fccdb01989f78484fcfc25554c56dc7/tee-worker/cli/lit_test_failed_parentchain_extrinsic.sh#L59 
- This extrinsic will fail on parentchain since `Eve` has not been set as the delegate signer for the account  - https://github.com/litentry/litentry-parachain/blob/94e09b2e4fccdb01989f78484fcfc25554c56dc7/pallets/identity-management/src/lib.rs#L211
- We check the `id-graph-hash` again, no state change would mean that the above failed extrinsic did not get executed
- From the logs, it parses but it does not execute it - https://github.com/litentry/litentry-parachain/blob/94e09b2e4fccdb01989f78484fcfc25554c56dc7/tee-worker/core/parentchain/indirect-calls-executor/src/executor.rs#L190

The failed extrinsic gets parsed but is not executed, Let me know if I'm missing something here.

kziemianek:
A more naive question: Have you considered writing unit test for it ? 😂
There is already `indirect_call_can_be_added_to_pool_successfully`

Kailai-Wang:
I see, thanks. I missed the delegator on-chain logic part.

Kasper's comment makes sense too - I'm happy to have both : )

Kailai-Wang (approved):

PR #2550, opened 2024-03-04
Cargo update

BillyWooo:
as title
merge other dependabot PRs together.


Kailai-Wang:
Sometimes dependabot doesn't update the worker Cargo.lock - can you please verify if that's the case?

Kailai-Wang (approved):

PR #2551, opened 2024-03-05
bitacross client sample golang code

kziemianek:
Sample golang code for requesting `SignEthereum` direct call from bitacross worker.


felixfaisal (approved):
LGTM
I was wondering if we should also verify the signature received with the public key, to make sure there are no shenanigans in the decoding

kziemianek:
> LGTM I was wondering if we should also verify the signature received with the public key, to make sure there are no shenanigans in the decoding

This can be done later once bitacross worker exposes rpc method for retrieving public keys. In general I think it's not necessary to have such check in this example.

Kailai-Wang:
A naive question: have you tried it on the actual bc-worker host (or connect to it)?

kziemianek:
> A naive question: have you tried it on the actual bc-worker host (or connect to it)?

No I've only tested it against bc-worker run on my host.

PR #2552, opened 2024-03-05
P-564 Sudo remove migration

wangminqi:
This closes Linear issue P-564 for sudo remove of litentry parachain.


runtime/litentry/src/migration/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+		log::info!("Post check Sudo");
+		let res: KillStorageResult =
+			frame_support::storage::unhashed::clear_prefix(&Twox128::hash(b"Sudo"), Some(0), None)
+				.into();
+
+		match res {
+			KillStorageResult::AllRemoved(0) | KillStorageResult::SomeRemaining(0) => {},
+			KillStorageResult::AllRemoved(n) | KillStorageResult::SomeRemaining(n) => {
+				log::error!("Remaining entries: {:?}", n);
```
kziemianek, line 63: Are you sure the error should be returned in case of `AllRemoved` ?

wangminqi: Yes. Post_upgrade is the checking hook after killing the storage. It will not be truly executed in production since it is under feature try-runtime. It will try re-killing storage in test mode and should expect killing nothing/left nothing.

Kailai-Wang (commented):
The clean-up migration looks good - I'm just wondering have you gone through the origin control? Especially those demand the root permission - e.g. what happens with `pallet-account-fix` when we need it later?

wangminqi:
> The clean-up migration looks good - I'm just wondering have you gone through the origin control? Especially those demand the root permission - e.g. what happens with `pallet-account-fix` when we need it later?

Except committee set member and account-fix, there is no other pallets/origin using simple root origin.
What's your opinion regarding about account-fix? If we do not change the setup, we can still do account_fix but will require a democracy process to do so.

Anyway, I will put another PR if this concerns us.

Kailai-Wang:
> > The clean-up migration looks good - I'm just wondering have you gone through the origin control? Especially those demand the root permission - e.g. what happens with `pallet-account-fix` when we need it later?
> 
> Except committee set member and account-fix, there is no other pallets/origin using simple root origin. What's your opinion regarding about account-fix? If we do not change the setup, we can still do account_fix but will require a democracy process to do so.
> 
> Anyway, I will put another PR if this concerns us.

As it direclty modifies the balance, it's not appropriate to use centralised origin on it - let's keep the root check so that it needs a referendum to do so.

Kailai-Wang (approved):
Thanks!

PR #2553, opened 2024-03-05
fix: p-460 bitcore message signature issue

0xverin:
### Context

resolves:p-460

Root cause: bitcore uses the ECDSA signature algorithm, which uses a nonce to sign messages. This causes the signature (base64) to not be consistent. When decoding, we might need to manually pad it with a 0 byte.

For example, under normal the signature is:
`IJyax3CRyGOCceIyoBDpH9a03dwi3SoRJ4N1RZh36FxARkDCpEmUgrHqIhE5H8THYF8TcE5HuH/9BkovbYmMSbk=`
But sometimes:
`H6FuCBQt/fjshZIym0BBgGjbykkqBWO6CisWA9tBDDMD/5iODJehfyUXX3MrM21l5dHSuG46KqXiFEA7X94apw==`
This is a base64 encoding issue.

~~I tested it and adding a `0` can make it work normally:~~
```
await signer.sign(payloadStr)------ Uint8Array(65) [
    0,  32, 157, 209, 236, 186,  54,  14, 139,  90, 150,
   47, 194, 137,  67,  86,  82, 112, 120, 102,  63, 223,
  156, 127, 240, 243, 129, 151,  19,  58, 202,  25, 246,
  121,  56, 254, 247,  39, 203,  68, 213, 168, 236,  53,
  163,  69, 207, 182,   3, 156, 234, 203, 145, 236,  34,
  150,  18, 212, 121, 148,  33, 130, 253, 117, 127
]
Got response: {"value":"0x010400","do_watch":false,"status":"Ok"}
```
Updates:
The solution of adding 0 was wrong, will change the content of sig, causing it to be invalid when used. I updated it with a new way.

After 2000 cycles of testing, there were no issues with the do-while sign method:
```
Got response: {"value":"0x010400","do_watch":false,"status":"Ok"}
    ✔ check idGraph from sidechain storage before linking (2034579ms)
```


tee-worker/ts-tests/integration-tests/common/utils/crypto.ts
```diff
@@ -145,7 +145,11 @@ export class BitcoinSigner implements Signer {
         return new Promise((resolve, reject) => {
             if (isString(message)) {
                 const sig = new bitcore.Message(message).sign(this.keypair);
-                resolve(bufferToU8a(Buffer.from(sig, 'base64')));
+                let buf = Buffer.from(sig, 'base64');
+                if (buf.length === 64) {
```
Kailai-Wang: Is it possible to have more idiomatic code by using e.g. a fixed length buffer? Something like `[u8; 65]` in rust to make sure it's always 65 bytes 🤔

Kailai-Wang (commented):
Thanks for finding the root cause! 👍 

Not sure if it's related to nonce tho (I haven't checked the lib impl and am not sure if the `k` is random or not) - maybe it's a sole consequence of having random btc key pairs each time.

BillyWooo (approved):
Thank you for finding the root cause.

tee-worker/ts-tests/integration-tests/common/utils/crypto.ts
0xverin: >I haven't checked the lib impl and am not sure if the k is random or not
I checked:
image

0xverin:
>I haven't checked the lib impl and am not sure if the k is random or not

I checked:
image

image

0xverin:
~~Please do not merge first, there is a small issue that needs to be confirmed.~~

tee-worker/ts-tests/integration-tests/common/utils/crypto.ts
0xverin: After testing, it is not just the error case where `buffer.length === 64`. In fact, there will also be cases where it equals 63 (not excluding other cases, 63 is the result of 1000 loops). Therefore, change to using do while to ensure getting u8aArray 65 sig.

```
1) Test Identity (bitcoin direct invocation)
       check idGraph from sidechain storage before linking:
     Error: createType(LitentryMultiSignature):: Enum(bitcoin):: Expected input with 65 bytes (520 bits), found 63 bytes
```

The way of creating a buffer seems to be impractical, as it will change the content of the signature, such as this way:
```
let buf = Buffer.from(sig, 'base64');
Buffer.alloc(65)
buf.copy(resultBuf, 65 - buf.length);
```
Will get an invalid error: 
`Got response: {"value":"TGetterIsNotAuthorized","do_watch":false,"status":"Error"}`

So my initial approach was also incorrect. The reason why the test passed is that when I [console.log here](https://github.com/litentry/litentry-parachain/pull/2553/commits/d87def6fd3655f4f555c64fc50db3091e7690c9c#diff-cfd728a08e2960d755d2a6976c5b9a2863031978a63a2390673ec7d9305da3c4L126-L127), actually called the sign method. Due to the uncertainty of the signature result, after two console calls in front, the third call [here](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/integration-tests/common/di-utils.ts#L126) happened to match the sig we used, so the test passed strangely.

tee-worker/ts-tests/integration-tests/common/utils/crypto.ts
```diff
@@ -144,8 +145,17 @@ export class BitcoinSigner implements Signer {
...
         return new Promise((resolve, reject) => {
             if (isString(message)) {
-                const sig = new bitcore.Message(message).sign(this.keypair);
-                resolve(bufferToU8a(Buffer.from(sig, 'base64')));
+                let buffer: Uint8Array;
+
+                // Here we must use the format [u8,65],
+                // so we can only accept a signature that has been decoded from base64 and has a length of 65.
+                // Therefore, using a do -while loop to resign unnecessary signatures.We should not encounter an infinite loop situation, so no condition is set to break out of the loop.
+                do {
```
Kailai-Wang: I somehow feel this isn't the right way, still.

A btc message signature should be 65 bytes - the encoded base64 array should be 88 bytes. This should be firm. If we get less than 65 bytes in decoding, I can only think of padding?

Let it run again and hope to get 65 bytes isn't the good practice IMO, as in reality you should be able to verify each btc signature the user generated, with arbitrary `k`

cc @kziemianek  - maybe you know it as you've done some btc sig related impl

Kailai-Wang (changes requested):
Btw what about `bitcore.Message` in `identity-helper.ts`?

tee-worker/ts-tests/integration-tests/common/utils/crypto.ts
kziemianek: In my opinion it should always produce 65 bytes signature, even `bitcore-lib` developers assumes that. 
If you try to use `bitcode-lib` api to recreate signature from `Buffer` it will fail:

```
const buffer = Buffer.from(sig, 'base64')
Signature.fromCompact(buffer)

Invalid Argument: Error: s must be 32 bytes
Error
    at new NodeError (/home/kziemianek/projects/kziemianek/sig-test/node_modules/bitcore-lib/lib/errors/index.js:20:41)
    at Object.checkArgument (/home/kziemianek/projects/kziemianek/sig-test/node_modules/bitcore-lib/lib/util/preconditions.js:14:13)
    at Signature.fromCompact (/home/kziemianek/projects/kziemianek/sig-test/node_modules/bitcore-lib/lib/crypto/signature.js:53:5)
```

I debugged it for a while and it turns out the `r` value is sometimes 31 bytes long, which results in 64 (1+31+32) bytes signature (instead of 1+32+32). Above error says `s must be 32 bytes` but the root cause is still `r` being 31 bytes long.

64-byte-sig

It happens ocassionally so it's really low probability of happeining twice in the row.

kziemianek: So to sum up, I see two options:

* switch to other js bitcoin lib (https://www.npmjs.com/package/bitcoinjs-lib maybe ?)
* implement a workaround (try to generate new signature untill it has 65 bytes) and wait for the potential fix in `bitcore-lib` - but before that happening it needs to be reported, confirmed and released.

0xverin: Thanks for your summary @kziemianek , I will switch to bitcoinjs for implementation. Btw, I have looked at the [SDK of unisat-walle](https://github.com/unisat-wallet/wallet-sdk/blob/master/package.json#L40)t and switched to the version they use, but there is still the same problem.

Kailai-Wang: cc @jonalvarezz - in case you have some experience when implementing SDK..

0xverin: The new bitcoinjs-lib was tested twice in 2000 loops with no signature decoding issues. I also tested identity-helper several times, and there was no problem:
```
Got response: {"value":"0x010400","do_watch":false,"status":"Ok"}
    ✔ check idGraph from sidechain storage before linking (2012011ms)
```

cc @kziemianek @Kailai-Wang

tee-worker/ts-tests/integration-tests/common/utils/vc-helper.ts
```diff
@@ -190,11 +190,13 @@ export const mockAssertions = [
...
     },
+
     //karat_dao
-    {
-        description: 'You are a user of a certain platform',
-        assertion: {
-            PlatformUser: 'KaratDaoUser',
-        },
-    },
+    // @fixme: Not sure why it fails in CI but wokrs locally, temporarily comment out and let the PR(2553) pass, solve it in another issue.
```
kziemianek: It pass on dev https://github.com/litentry/litentry-parachain/actions?query=branch%3Adev

Does this PR make it fail ?
If yes, you should resolve it before merge.

0xverin: Should not, my PR did not break this part, if it is broken, all VC requests should fail. Let me try again, and I have never reproduced it locally.

0xverin: I think I know why, the ci uses a real endpoint and been unavailable that time, do you mind if I add a mock service for this later with a new pr? I've created an [issue](https://linear.app/litentry/issue/P-587/request-platformuser-vc-failed-in-ci) this afternoon to document this.

```
 lc_vc_task_receiver] should_create_id_graph: false
[2024-03-07T06:10:55Z DEBUG lc_data_providers] base_url: https://api.karatdao.com/
[2024-03-07T06:10:55Z DEBUG lc_data_providers::karat_dao] user_verification, params: KaraDaoRequest { path: "user/verification", query: Some([("address", "0xff93b45308fd417df303d6515ab04d9e89a750ca")]) }
[2024-03-07T06:10:55Z DEBUG itc_rest_client::http_client] no body to send
[2024-03-07T06:10:55Z DEBUG lc_data_providers::karat_dao] user_verification, error: RequestError("call rest api error: Server returned non-success status: 503, details: Service Unavailable")
```

kziemianek: Would be great if we do not comment out any test cases in this PR.
We already have mock for it https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/litentry/core/mock-server/src/karat_dao.rs.

kziemianek: CI passed 🙃

tee-worker/ts-tests/integration-tests/common/common-types.ts
```diff
@@ -16,7 +15,7 @@ interface SubstrateWalletItem {
     [key: string]: KeyringPair;
 }
 interface BitcoinWalletItem {
-    [key: string]: bitcore.PrivateKey;
+    [key: string]: ECPairInterface;
```
kziemianek, line 18: I'm just wondering if we can get rid of `EthersWalletItem`, `BitcoinWalletItem` and `SubstrateWalletItem`. 

We already have `Signer` interface so we can create wallet map:

```
{
    "evm": {
        "alice": Signer,
        "bob": Signer
    },
    "substrate": {
        "alice": Signer,
        "bob": Signer
    },
    "bitcoin": {
        "alice": Signer,
        "bob": Signer
    }
}
```

What do you think @0xverin ? Could be done in separate PR. Probably we should have Solana signer as well...

tee-worker/ts-tests/integration-tests/di_bitcoin_identity.test.ts
```diff
@@ -66,13 +64,13 @@ describe('Test Identity (bitcoin direct invocation)', function () {
         );
         teeShieldingKey = await getTeeShieldingKey(context);
         aliceBitcoinIdentity = await buildIdentityHelper(
-            u8aToHex(bufferToU8a(context.bitcoinWallet.alice.toPublicKey().toBuffer())),
+            '0x' + context.bitcoinWallet.alice.publicKey.toString('hex'),
```
kziemianek: Can it be simply `u8aToHex(context.bitcoinWallet.alice.publicKey)` ?

tee-worker/ts-tests/integration-tests/common/helpers.ts
```diff
@@ -6,7 +6,9 @@ import type { HexString } from '@polkadot/util/types';
 import './config';
 import { IntegrationTestContext, JsonRpcRequest } from './common-types';
 import { randomBytes } from 'crypto';
-
+import { ECPairFactory, ECPairInterface } from 'ecpair';
+import * as ecc from 'tiny-secp256k1';
+import * as bitcoinjs from 'bitcoinjs-lib';
```
kziemianek: This dependency seems to be unused, can we remove it from project ?

tee-worker/ts-tests/integration-tests/common/utils/crypto.ts
```diff
@@ -131,21 +132,24 @@ export class EthersSigner implements Signer {
...
-        return bufferToU8a(this.keypair.toPublicKey().toBuffer());
+        return bufferToU8a(this.keypair.publicKey);
     }
 
     sign(message: HexString | string | Uint8Array): Promise<Uint8Array> {
         return new Promise((resolve, reject) => {
             if (isString(message)) {
-                const sig = new bitcore.Message(message).sign(this.keypair);
-                resolve(bufferToU8a(Buffer.from(sig, 'base64')));
+                // produce non-deterministic signatures
```
kziemianek: I would say it's preferred to have deterministic outputs in tests, can we get rid of `extraEntropy` ?

tee-worker/ts-tests/integration-tests/common/utils/crypto.ts
jonalvarezz: jumping in late, but I hope it helps. 

Here is what we do on the SDK for Bitcoin in a nutshell:
1. The hex challenge code should be signed without the `0x` prefix. Only relevant if signing in hex/string format. We were using raw uint signing before, but I remember we had to change this because of bitcoin/ unisat.

https://github.com/litentry/identity-hub/blob/develop/libs/enclave/src/lib/requests/link-identity.request.ts#L63

2. The produced signature is base64 decoded and the `0x` prefix put back. Otherwise, `LitentryValidationData` fails at encoding.

https://github.com/litentry/identity-hub/blob/develop/libs/enclave/src/lib/type-creators/validation-data.ts#L145-L153

tee-worker/ts-tests/integration-tests/common/common-types.ts
0xverin: >Could be done in separate PR. Probably we should have Solana signer as well

will create a new pr and then I think  need to check if there are similar redundant codes in other places.

tee-worker/ts-tests/integration-tests/di_bitcoin_identity.test.ts
0xverin: Modified

tee-worker/ts-tests/integration-tests/common/utils/crypto.ts
0xverin: thanks @jonalvarezz  I actually looked at the code on idehub, and in the Bitcoin signature part, mainly use unisat-wallet. This wallet uses their own integrated [unisat-sdk](https://github.com/unisat-wallet/extension/blob/master/package.json#L54), and you will find that they also use [bitcore-lib and bitcoinjs-lib](https://github.com/unisat-wallet/wallet-sdk/blob/master/package.json#L39-L43).So, essentially there is no difference,  I guess.:p

Kailai-Wang (approved):
For me it's ready to be merged

PR #2554, opened 2024-03-05
remove skip-ias-check leftovers

kziemianek:
This feature is not present in any package.


Kailai-Wang:
Good catch, it should have been removed in my last PR, sorry! I only checked `tee-dev` feature at that time...

Kailai-Wang (approved):
Happy to merge it once CI passes

PR #2555, opened 2024-03-06
fix: use futures executor ThreadPool instead of standalone ThreadPool

felixfaisal:
This PR replaces `Threadpool` with `Threadpool` from `futures::executors` that let's us spawn tasks which are run concurrently.


tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -82,8 +82,7 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
 	N::MetadataType: NodeMetadataTrait,
 {
 	let vc_task_receiver = init_vc_task_sender_storage();
-	let n_workers = 4;
-	let pool = ThreadPool::new(n_workers);
+	let pool = ThreadPool::new().unwrap();
```
Kailai-Wang, line 85: It uses the default builder config - did you check what the pool size it? (It's bound to number of cpus but I don't know the number actually)

Kailai-Wang (commented):
Could you explain a bit what is the advantage of it?

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
felixfaisal: Yeah I'm using the default, It is equal to the number of cpus available on the machine. I think i mentioned it in the issue. 
https://github.com/mesalock-linux/futures-rs-sgx/blob/d54882f24ddf7d61327a067b2f608d6940a36444/futures-executor/src/thread_pool.rs#L194

Kailai-Wang: I just want to make sure it's not some value like 1 :D 

because enclave is quite special, I won't be surprised if the sdk (Hw or SW) is doing some schenanigans

Kailai-Wang (approved):
Thanks - please double check the default configuration within enclave though (both SW and HW) - especially the number of workers (or you pass it in as parameter explicitly)

I approve the PR for now :)

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
BillyWooo: I don't think to give default is a good idea. Quote "It is equal to the number of cpus available on the machine.", "available" means at that moment of executing or in general the CPU physically has? what `num_cpus::get()` gives out?

BillyWooo: And I think it will also change the behaviour of batch_VC PR. Please hold this PR for a moment.

BillyWooo (changes requested):
And I think it will also change the behaviour of batch_VC PR. Please hold this PR for a moment.

BillyWooo (approved):

Kailai-Wang:
> Thanks - please double check the default configuration within enclave though (both SW and HW) - especially the number of workers (or you pass it in as parameter explicitly)

@felixfaisal do you have answers for it? : )

PR #2559, opened 2024-03-08
Use fixed private key in bitacross golang example

kziemianek:
* use fixed private key (so it can be registered as relayer)
* use bitacross dev instance 
* `go mod tidy`


bitacross-worker/example/client/example.go
```diff
@@ -76,11 +75,13 @@ func main() {
 	getStateMrEnclaveResult, _ := decodeRpcReturnValue(res.Result)
 	//at this point we got all stuff from worker - shielding key, mrenclave and shard (shard == mrenclave)
 
-	//** generate private key and ethereum address
-	curve := secp256k1.S256()
-	key, _ := ecdsa.GenerateKey(curve, rand.Reader)
+	//** WARNING: use this key only for environment without real value
```
Kailai-Wang, line 78: Does it make sense to log pubkey in comment too? Just for reference

Kailai-Wang (approved):
Thanks!

bitacross-worker/example/client/example.go
kziemianek: Added alongside with eth address

PR #2560, opened 2024-03-08
remove identity_test.sh from bitacross

kziemianek:
Small leftover - this script is not related to bitacross worker as there are no identity related trusted calls.


kziemianek:
Looking at the script itself, maybe we should remove it from `tee-worker` also. What do you think ?

Kailai-Wang (approved):

Kailai-Wang:
> Looking at the script itself, maybe we should remove it from `tee-worker` also. What do you think ?

I was too fast in the merge, sorry. Yes it seems obsolete.

PR #2561, opened 2024-03-08
Add an option to start bitacross worker using `launch.py`

kziemianek:
We need script to run both `identity` and `bitacross` worker

* adds an option to specify which worker should be launched through `--worker` arg, example: `--worker identity`
* renames short version of argument used to specify number of workers to launch from `-w` to `-wn`


kziemianek:
I've checked locally following configurations:

| w | p | result |
|----|---|---------|
| bitacross | local-binary | 🟢 
| identity | local-binary | 🟢 
| - | local-binary | 🟢 
| - | local-binary-standalone | 🟢 
| bitacross | local-binary-standalone | 🟢 
| identity | local-binary-standalone | 🟢

BillyWooo (approved):

local-setup/launch.py
```diff
@@ -215,24 +216,38 @@ def get_subcommand_flags(index):
...
+        worker_dir = "bitacross-worker"
+    else:
+        sys.exit("Unsupported worker")
+
+    if worker == "identity":
+        worker_bin = "litentry-worker"
+    elif worker == "bitacross":
+        worker_bin = "bitacross-worker"
+    else:
+        sys.exit("Unsupported worker")
```
Traf333: shall we just smush it together?

```suggestion
    if worker == "identity":
        worker_dir = "tee-worker"
        worker_bin = "litentry-worker"
    elif worker == "bitacross":
        worker_dir = "bitacross-worker"
        worker_bin = "bitacross-worker"
    else:
        sys.exit("Unsupported worker")
```

Traf333 (commented):
Thanks for the improvement ❤️ Observed one bit to DRY and apart from that looks good

PR #2562, opened 2024-03-11
Add RPC inteface for getting wallet info

Traf333:
### Context
this PR introduces new RPC call to provide info about signer, bitcoin and ethereum public keys

### Testing Evidences
checked only via `websocat` commands

```
echo '{"jsonrpc":"2.0","method":"bitacross_getPublicKeys","id":1}' | websocat -n1 -k -B 99999999 wss://localhost:2000

// => {"jsonrpc":"2.0","result": {"bitcoin_key":"0x023ab07017cef21e0042e123e83cebde2e838ab2bd903fb31428b45f3bd1761dfc","ethereum_key":"0x037d54c9ebee18a8b54bf3ac560015a87a16af36edc1bd8d698463cb7afb9038ab","signer":"0x953edfcc064cfa66c564739e9988233e358604baa3f0aa34c13df367661e4ffb"},"id":1}
```


Traf333:
@kziemianek do we plan to have tests for RPC calls or it's not relevant at this moment?

Kailai-Wang (approved):
Looks good to me, I'm fine either one method or three. Thank you for the manual tests too.

One thing to notice is that the returned `result` is actually the SCALE-encoded `RpcReturnValue`, so extra decoding work is needed to extract the key.

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -145,6 +150,66 @@ where
 		}
 	});
 
+	io.add_method("bitacross_getParachainKey", move |_: Params| {
+		debug!("worker_api_direct rpc was called: bitacross_getParachainKey");
+
+		async move {
+			let json_value = match GLOBAL_SIGNING_KEY_REPOSITORY_COMPONENT.get() {
```
kziemianek: Could you pass components to function instead of coupling to globals ?
Some of them are already passed like `top_pool_author`.

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -145,6 +150,66 @@ where
...
+				Ok(key_repository) => match key_repository.retrieve_key() {
+					Ok(key) =>
+						RpcReturnValue::new(key.public().0.encode(), false, DirectRequestStatus::Ok)
+							.to_hex(),
+					Err(_) => compute_hex_encoded_return_error("Some error happened"),
```
kziemianek: Please log the error

kziemianek (commented):
We don't need unit tests at this point for read operations.
I left two small comments.

Traf333:
> One thing to notice is that the returned result is actually the SCALE-encoded RpcReturnValue, so extra decoding work is needed to extract the key.

Do you mean, it's totally fine to return just public key without any encoding, right?

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
Traf333: is it ok to have functions with 7 args passing through?

kziemianek: It's fine for me.

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -74,6 +80,9 @@ pub fn public_api_rpc_handler<Author, GetterExecutor, AccessShieldingKey, OcallA
 	getter_executor: Arc<GetterExecutor>,
 	shielding_key: Arc<AccessShieldingKey>,
 	ocall_api: Arc<OcallApi>,
+	signer: ed25519::Pair,
```
kziemianek: I suggested passing `KeyRepository`, not keys directly. 
It's possible programaticaly to change the keys during the runtime, see:
`impl<KeyType, SealedIo> MutateKey<KeyType> for KeyRepository<KeyType, SealedIo>`

Traf333:
@Kailai-Wang @kziemianek , I've modified functionality according to your comments + turned into 1 RPC method. Updated description with testing evidence

kziemianek (approved):

PR #2568, opened 2024-03-12
feat: write async data provider code using and add async feature

felixfaisal:
So I've created a feature `async` under which I'm writing all the equivalent async functions. In this PR, I'm mostly adding this feature to two crates `data providers` and `rest client`. It will not tamper any ongoing vc development since it is feature gated. This PR will allow me to move on to the next part of async assertions.


BillyWooo (commented):
Over all, the changes look good to me. I am just wondering whether we still need the sync code? Should we simply remove the sync code, only use async?

felixfaisal:
> Over all, the changes look good to me. I am just wondering whether we still need the sync code? Should we simply remove the sync code, only use async?

I think we can have this discussion once we have the entire async feature working in `isolated` vc request, In the case of this PR, I'm adding async code only to the `data providers` and I'm feature to not block the ongoing development of VCs.

wli-pro:
> > Over all, the changes look good to me. I am just wondering whether we still need the sync code? Should we simply remove the sync code, only use async?
> 
> I think we can have this discussion once we have the entire async feature working in `isolated` vc request, In the case of this PR, I'm adding async code only to the `data providers` and I'm feature to not block the ongoing development of VCs.

Yeah, we can/should remove the sync one only once all moved to async one.

wli-pro (approved):
LGTM

tee-worker/core/rest-client/src/lib.rs
```diff
@@ -119,6 +139,46 @@ pub trait RestPost {
 		K: serde::de::DeserializeOwned;
 }
 
+#[cfg(feature = "async")]
+#[async_trait]
+pub trait RestPost {
+	/// Plain POST request.
+	async fn post<U, T>(&mut self, params: U, data: &T) -> Result<(), Error>
+	where
+		T: serde::Serialize + RestPath<U> + std::marker::Send + std::marker::Sync,
```
Kailai-Wang: Do you consider writing `Send + Sync` with `use`? Will look less verbose

Kailai-Wang (commented):
Thanks!

I'm not sure about two things:
- is it better to keep both versions at least? Like have `sync_get_with` and `async_get_with` - I'm generally not a fan of feature-guarded implementation of same/similar functionalities, especially when they are mutually exclusive - unless absolutely required
- is it worth having it when we move to evm-based impl later?

felixfaisal:
> Thanks!
> 
> I'm not sure about two things:
> 
> * is it better to keep both versions at least? Like have `sync_get_with` and `async_get_with` - I'm generally not a fan of feature-guarded implementation of same/similar functionalities, especially when they are mutually exclusive - unless absolutely required
> * is it worth having it when we move to evm-based impl later?

1. So both caller and callee have to be async, and we will have to create separate traits for the functions, It would be a much bigger refactor. The rationale for me is by feature-gating async implementations,we allow for incremental updates without disrupting existing synchronous code paths. This approach lets us refine the async implementation over time. I would say we keep both implementation and do a refactor after we have everything working. 

2.  I think this depends on the implementation of evm, if the evm doesn't make use of threadpool (async executor) then it makes little sense, but it also has implications on the performance. It is worth having it for supporting as many VCs in batch_vc.

tee-worker/core/rest-client/src/rest_client.rs
```diff
@@ -182,6 +185,36 @@ where
...
+	/// Make a GET request.
+	async fn get<U, T>(&mut self, params: U) -> Result<T, Error>
+	where
+		T: serde::de::DeserializeOwned + RestPath<U> + std::marker::Send,
+		U: std::marker::Send,
+	{
+		let body = self.make_request::<U, T>(Method::GET, params, None, None)?;
+
+		serde_json::from_str(body.as_str()).map_err(|err| Error::DeserializeParseError(err, body))
+	}
```
jonalvarezz, line 203: If response is != 200, will it throw a `DeserializeParseError` error? could that be misleading? 

furthermore, should we keep this generic and not assume the body's content is json string?

PR #2570, opened 2024-03-12
Follow-up-of-PR-2532

BillyWooo:
As title.
Update based on comments from PR #2532.
1. remove separate thread to clean up fully processed registry item; Relay on assertion-build return value / error / time out.
2. update according to PR comments from PR-2532

### DIFF
--- a/tee-worker/Cargo.lock
+++ b/tee-worker/Cargo.lock
@@ -5619,7 +5619,6 @@ version = "0.9.0"
 dependencies = [
  "futures 0.3.28",
  "futures 0.3.8",
- "ita-stf",
  "itp-rpc",
  "itp-stf-primitives",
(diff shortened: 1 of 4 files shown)


Kailai-Wang (approved):
Looks better, thanks!

We still need two follow-up issues tho IIRC, one is about code readability improvement (low prior), one is about `RequestVcResultOrError` struct refactoring - we should avoid using `&str` to represent errors (medium - high prior) :)

BillyWooo:
[[P-609 : refactoring struct RequestVcResultOrError](https://linear.app/litentry/issue/P-609/refactoring-struct-requestvcresultorerror)](https://linear.app/litentry/issue/P-609/refactoring-struct-requestvcresultorerror)

PR #2571, opened 2024-03-12
Update identity worker upstream codebase to `9a3b032`

Kailai-Wang:
### Context

As topic.

The most important change is the decoupling of parachain runtime against the service/CLI deps. Hence you'll notice a (potentially significantly) faster compilation time.

Other notes:
- pallet-parentchain has a per-instance config. We don't necessarily use it but I keep it to minimise the future maintainability cost
- I add back `tee-worker/.github/workflows/build_and_test.yml` for reference purpose
- ~~The `StfError` enum index has changed, depending on how client-sdk handles it, it could be a breaking or non-breaking change. cc @jonalvarezz @0xverin~~ It's a breaking change for the client
- I skip the timestamp indirect calls as we don't use it for now (see `BlockMetadata`)
- Some shard related settings are not yet used / propagated (a legacy problem - we should have a complete review later for multi-worker setup)
- I've noticed some changes regarding state provisioning but it should be clearer to have a separate PR - to be followed.


tee-worker/app-libs/parentchain-interface/src/event_subscriber.rs
```diff
@@ -0,0 +1,60 @@
...
+	See the License for the specific language governing permissions and
+	limitations under the License.
+
+*/
+
+use itp_api_client_types::ParentchainApi;
+use itp_types::parentchain::{BalanceTransfer, ParentchainId};
+use substrate_api_client::SubscribeEvents;
+
+pub fn subscribe_to_parentchain_events(api: &ParentchainApi, parentchain_id: ParentchainId) {
```
jonalvarezz, line 22: what's this one for? if I follow correctly, it subscribes forever and log event's info only

tee-worker/core-primitives/stf-primitives/src/error.rs
jonalvarezz, line 1: since decoding changes, we should mark it as breaking change. I've created this follow-up ticket https://linear.app/litentry/issue/P-620/client-sdk-update-stferror-definition

tee-worker/app-libs/parentchain-interface/src/event_subscriber.rs
Kailai-Wang: It's been like this since the beginning 😂  for now it's used in tests and CLI.
My understanding is it prefaces the upcoming change where parentchain events should be used to trigger indirect calls instead of extrinsics => even if upstream doesn't do it we might want it

tee-worker/core-primitives/stf-primitives/src/error.rs
Kailai-Wang: I see, thanks!

tee-worker/app-libs/stf/src/stf_sgx.rs
```diff
@@ -254,27 +266,174 @@ where
...
+		state: &mut State,
+		header: ParentchainHeader,
+	) -> Result<(), Self::Error> {
+		trace!("updating litentry parentchain block : {:?}", header);
+		state.execute_with(|| {
+			pallet_parentchain::Call::<Runtime, ParentchainInstanceLitentry>::set_block { header }
+				.dispatch_bypass_filter(Runtime::RuntimeOrigin::root())
+				.map_err(|e| {
+					Self::Error::Dispatch(format!(
+						"Update parentchain integritee block error: {:?}",
```
kziemianek: Maybe "Update parentchain litentry block error" - you probably updated funcion name 🤔

tee-worker/core-primitives/stf-primitives/src/error.rs
```diff
@@ -24,54 +24,64 @@ pub type StfResult<T> = Result<T, StfError>;
...
-	InvalidNonce(Nonce, Nonce),
-	#[codec(index = 5)]
-	StorageHashMismatch,
-	#[codec(index = 6)]
-	InvalidStorageDiff,
-	#[codec(index = 7)]
-	InvalidMetadata,
 	// litentry
-	#[codec(index = 8)]
+	#[codec(index = 0)]
```
kziemianek, line 28: Why don't we keep old indices?

tee-worker/core/parentchain/light-client/src/justification.rs
```diff
@@ -151,10 +151,10 @@ impl<Block: BlockT> GrandpaJustification<Block> {
 				set_id,
 				&mut buf,
 			) {
-				debug!("Bad signature on message from {:?}", &signed.id);
-				return Err(ClientError::BadJustification(
-					"invalid signature for precommit in grandpa justification".to_string(),
-				))
+				warn!("Bad signature on message from {:?}", &signed.id);
+				//				return Err(ClientError::BadJustification(
```
kziemianek, line 155: This is a workaround from upstream. 
Let's keep an eye on the fix.

tee-worker/enclave-runtime/src/shard_config.rs
```diff
@@ -0,0 +1,24 @@
...
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+use crate::error::Result as EnclaveResult;
+use itp_types::ShardIdentifier;
+use log::*;
+
+pub(crate) fn init_shard_config(_shard: ShardIdentifier) -> EnclaveResult<()> {
+	warn!("TODO(Litentry): init_shard_config");
```
kziemianek: Is this going to be resolved in follow up PR?

tee-worker/service/src/account_funding.rs
```diff
@@ -54,66 +60,64 @@ impl EnclaveAccountInfoProvider {
...
-	if missing_funds > 0 {
-		info!("Transfer {:?} from Alice to {}", missing_funds, accountid);
-		bootstrap_funds_from_alice(api, accountid, missing_funds)?;
-	}
-	Ok(())
+	let transfer_fee = estimate_transfer_fee(api)?;
+	info!("[{:?}] a single transfer costs {:?}", parentchain_id, transfer_fee);
+	min_required_funds += 1000 * transfer_fee;
+
+	// TODO(Litentry): shall we charge RA fee?
```
kziemianek: please create follow up issue for discussion

tee-worker/service/src/sync_state.rs
```diff
@@ -42,13 +44,10 @@ pub(crate) fn sync_state<
 	enclave_api: &E,
 	skip_ra: bool,
 ) {
-	// FIXME: we now assume that keys are equal for all shards.
 	let provider_url = match WorkerModeProvider::worker_mode() {
-		WorkerMode::Sidechain =>
+		WorkerMode::Sidechain | WorkerMode::OffChainWorker =>
 			executor::block_on(get_author_url_of_last_finalized_sidechain_block(node_api, shard))
```
kziemianek: How is it possible to get author of last finalized sidechain block in offchain mode ?

kziemianek (approved):

tee-worker/core-primitives/stf-primitives/src/error.rs
Kailai-Wang: Because they keep adding new entries which will break the litentry entries anyway (you could append them only, but it's a bit awful in terms of readability)

So I made the same changes as what we did to `TrustedCall` enum - put litentry's entries at front so that future upstream changes won't change the codec index (for us)

tee-worker/core/parentchain/light-client/src/justification.rs
Kailai-Wang: Yes 👍  We are not using grandpa tho so we are less affected :D

tee-worker/enclave-runtime/src/shard_config.rs
Kailai-Wang: Unlikely - it's to update some upstream parachain pallet storage which we don't use.
But let me create an issue for it, with low priority

tee-worker/service/src/account_funding.rs
Kailai-Wang: Will create an issue!

tee-worker/service/src/sync_state.rs
Kailai-Wang: Good catch - the function isn't well named, I'll change that.

Edit: maybe we should implement `get_enclave_url_of_last_active` - created an issue for it

tee-worker/enclave-runtime/src/shard_config.rs
Kailai-Wang: P-627

Kailai-Wang: Added the issue number to comment too

PR #2573, opened 2024-03-13
Bump versions for next release

Kailai-Wang:
### Context

As topic. Worker version is bumped too as we have some updates in non-enclave part


Kailai-Wang:
wrong base - will reopen it

PR #2575, opened 2024-03-13
fix error of request vc for cli

BillyWooo:
Minor fix. Just ignore the first response message. Not a perfect solution. 
Need to refactor together with `Struct RequestVcResultOrError`.

May remove `request_vc` support. So didn't spend too much time here for now.


tee-worker/cli/src/trusted_operation.rs
```diff
@@ -469,6 +393,10 @@ pub(crate) fn send_direct_batch_vc_request(
 				let response: RpcResponse = serde_json::from_str(&response).unwrap();
 				if let Ok(return_value) = RpcReturnValue::from_hex(&response.result) {
 					debug!("successfully decoded rpc response: {:?}", return_value);
+					if ignore_first {
```
felixfaisal, line 396: I could not understand this, Are we deliberately ignoring the first response, if so why?

tee-worker/cli/src/trusted_base_cli/commands/litentry/request_vc.rs
```diff
@@ -509,34 +507,69 @@ impl RequestVcCommand {
...
-			Err(e) => {
-				println!("{:?}", e);
-			},
+		let (sender, receiver) = channel::<Result<RequestVcResultOrError, String>>();
+
+		send_direct_batch_vc_request(cli, trusted_cli, &top, key, sender);
+
+		let mut len = 0u8;
+		let mut cnt = 0u8;
+		loop {
```
felixfaisal, line 537: Can we do this inside `send_direct_batch_vc` which should collect responses, and give a vector of responses here and we can just print them. wdyt?

tee-worker/cli/src/trusted_operation.rs
```diff
@@ -461,6 +384,7 @@ pub(crate) fn send_direct_batch_vc_request(
 
 	debug!("waiting for rpc response");
 	let mut req_cnt = 0u8;
+	let mut ignore_first = true;
 	std::thread::spawn(move || loop {
```
felixfaisal, line 388: > Can we do this inside send_direct_batch_vc which should collect responses, and give a vector of responses here and we can just print them. wdyt?

If we want to do this as mentioned, then we should probably not move to different thread.

felixfaisal (commented):
Thanks @BillyWooo I couldn't test it myself, I will do it in some time. I have few comments for you.

BillyWooo:
replaced by #2595

PR #2576, opened 2024-03-14
Prefer author_getNextNonce for nonce retrieval

silva-fj:
- Updated `getSidechainNonce` in `di-utils.ts` to retrieve the nonce using `author_getNextNonce` RPC call and updated the corresponding tests.
- Refactored `author_getNextNonce` to expect `Identity` to support other type of accounts.


silva-fj:
The failing tests are caused by some change in this [PR](https://github.com/litentry/litentry-parachain/pull/2571). c.c @Kailai-Wang

Kailai-Wang (approved):
Looks good to me!

I removed `P610` from the PR title :)

tee-worker/ts-tests/integration-tests/common/di-utils.ts
```diff
@@ -297,15 +297,21 @@ export async function createSignedTrustedGetterIdGraph(
...
+        'author_getNextNonce',
+        [base58Encode(hexToU8a(context.mrEnclave)), primeIdentity.toHex()],
+        nextRequestId(context)
+    );
+    const res = await sendRequest(context.tee, request, context.api);
+    const nonceHex = res.value.toHex();
+    let nonce = 0;
+
+    if (nonceHex) {
+        nonce = context.api.createType('Index', '0x' + nonceHex.slice(2)?.match(/../g)?.reverse().join('')).toNumber();
```
jonalvarezz, line 312: hwo should we read this line: 

```
nonceHex.slice(2)?.match(/../g)?.reverse().join('')
```

silva-fj: I borrowed that logic from [here](https://github.com/litentry/litentry-parachain/blob/6ad83aa755cdc5118fbecb3240881c170e1f2cb7/tee-worker/ts-tests/integration-tests/ii_identity.test.ts#L47). What this line does is to reverse the order of two character groups in the hex string while keeping the '0x' prefix. (also known as byte swapping or endian swapping).

PR #2577, opened 2024-03-14
Check condition before creating BRC20 vc

BillyWooo:
Minor fix.
When response is empty, no VC should be created.


Kailai-Wang (approved):
Looks good to me

Kailai-Wang:
Btw - accidentally found that this PR had two authors again :)

PR #2578, opened 2024-03-15
Re-register enclave after state provisioning

Kailai-Wang:
### Context

As topic.

In #2571 the state provisioning happens after the worker is started, thus we need to re-register the enclave, otherwise the registry will contain the old data.

Please also note: following upstream `request-state` arg is removed, the subcommand is still there but marked as "DEPRECATED" too.

### DIFF
--- a/tee-worker/docker/multiworker-docker-compose.yml
+++ b/tee-worker/docker/multiworker-docker-compose.yml
@@ -208,7 +208,7 @@ services:
     entrypoint:
       "/usr/local/bin/litentry-worker --clean-reset --ws-external -M litentry-worker-2 -T wss://litentry-worker-2
       -u ws://litentry-node -U ws://litentry-worker-2 -P 2011 -w 2101 -p 9912 -h 4645 --enable-mock-server
-      run --dev --skip-ra --request-state"
+      run --dev --skip-ra"
     restart: "no"
   litentry-worker-3:
     image: litentry/litentry-worker:latest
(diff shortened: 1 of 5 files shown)


kziemianek (commented):
Does it still make sense to perform state synce if `request-state` is provided ? Can we just accept argument but do not perform it?

I checked and we have the same logic as upstream, but it seems like it causes problems, am I correct ?

Kailai-Wang:
> Does it still make sense to perform state synce if `request-state` is provided ? Can we just accept argument but do not perform it?
> 
> I checked and we have the same logic as upstream, but it seems like it causes problems, am I correct ?

I don't fully understand the question.
`request-state` as arg is removed.
`request-state` as subcommand still exists but is marked as deprecated - so theoretically it should never be required.

And what problem do you mean?

kziemianek:
Ok, I see it now. It's my misunderstanding.

1. I meant the problem described in the issue
2. I confused subcommand with arg.
3. We somehow decided to keep arg even if it was deleted in upstream https://github.com/integritee-network/worker/commit/8a3ea6e40ae54da377da47945d672da38d27426d#diff-a4bded0511a00449aaf3e6f8b33720ffb6ceb277ede2419ad3454668c1a50d63

I agree that there is no point in explicitly requesting state if workers are capable of autodetecting the need. (covered by upstream update)

kziemianek (approved):
It would be nice if we could send only one register extrinsic in both cases.

Kailai-Wang:
> It would be nice if we could send only one register extrinsic in both cases.

Yea had this thought too - can be done as an improvement. I'll file an issue 
For now I don't bother changing upstream too much with non-critical features.

PR #2586, opened 2024-03-18
fix: update DirectClient and link-identity

felixfaisal:
The issue describes it.


silva-fj (approved):

silva-fj:
Thank you for adding this @felixfaisal, I missed checking the usages of that method. 

[This mock](https://github.com/litentry/litentry-parachain/blob/489930ca6329c8c407e403c58086dc32b5c71657/tee-worker/core/rpc-client/src/mock.rs#L114) probably needs to be updated as well

Kailai-Wang:
There seems to be compile errors still :)

Kailai-Wang:
I added the test to watchlist too

PR #2587, opened 2024-03-18
apply codebase update to bc worker

kziemianek:
New upstream commit: `9a3b032`


Kailai-Wang (approved):
I'm good as long as CI passes

PR #2588, opened 2024-03-18
Refactoring types under primitives

silva-fj:
Moving assertion related primitive types to sub-modules.


BillyWooo (approved):
LGTM

primitives/core/src/assertion/mod.rs
```diff
@@ -17,14 +17,33 @@
...
+use generic_discord_role::GenericDiscordRoleType;
+use network::{all_web3networks, BoundedWeb3Network, Web3Network};
+use oneblock::OneBlockCourseType;
 use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
+use platform_user::PlatformUserType;
 use scale_info::TypeInfo;
 use sp_std::{vec, vec::Vec};
+use vip3::VIP3MembershipCardLevel;
+use web3_nft::Web3NftType;
+use web3_token::Web3TokenType;
```
Kailai-Wang: Is it possible to separate those imports at least under:
- crate level (`use crate::...`)
- local (sub) mod (e.g. `network::...`)
- extern crates (e.g. `sp_std`, `parity_scale_codec` etc

primitives/core/src/assertion/mod.rs
```diff
@@ -17,14 +17,33 @@
...
+pub mod contest;
+pub mod evm_amount_holding;
+pub mod generic_discord_role;
+pub mod network;
+pub mod oneblock;
+pub mod platform_user;
+pub mod soraquiz;
+pub mod vip3;
+pub mod web3_nft;
+pub mod web3_token;
```
Kailai-Wang: Any reason to not follow the idiom
```
mod xxx;
pub use xxx::yyy;
```
?

Not trying to enforce it, maybe there's a case where this one fits better

silva-fj: Not a particular reason, I'm just used to declare all modules at the top but if this is a convention we use I can change it to follow it

PR #2589, opened 2024-03-18
Handshaking between workers for local setup

Traf333:
### Context

this PR fixes couple of minor issues for multiworker setup in local environment
1. `error: Found argument '--request-state' which wasn't expected, or isn't valid in this context`
2. 
```
[Sidechain] imported block (number: 306, tcalls: 0, author: 0x26491bbf431770cc4b0b23aad6f1216a40c97c098600d981a022d9c246409cad), based on parentchain block 751
[2024-03-18T14:01:54.020389Z ERROR itc_peer_top_broadcaster] Could not connect to peer wss://127.0.0.1:2100, reason: "Could not open websocket connection: HandshakeError::Failure(Tls(Dns(InvalidDNSNameError)))"
```


BillyWooo (commented):
I don't think this is the fix to something. 
I think `--request-state` flag is still valid.

Kailai-Wang (approved):
Thanks - `request-state` was indeed removed earlier (cc @BillyWooo)

PR #2597, opened 2024-03-19
fix: use async threadpool for stf

felixfaisal:
We are using the same number of workers that we were using previously.


BillyWooo (approved):
LGTM

Kailai-Wang (approved):

PR #2599, opened 2024-03-19
Remove redundant HttpClientBuilder

Traf333:
### Context

Initially, the task was intent to make sure that all data providers have defined timeout for http client, and during sanity check I discovered that there is no use of `HttpClientBuilder` and no other changes for task is required


kziemianek (approved):
This is fine for me but personally I would prefer to use this builder in place where `HttpClient` is instantiated instead of removing it.

Traf333:
> This is fine for me but personally I would prefer to use this builder in place where HttpClient is instantiated instead of removing it.

Good point @kziemianek, just for my curiosity, does our team prefer builder pattern over initialisation function? 

I've noticed with this particular case that we have initialisation function and builder, so I decided to keep the one which is used and involved in tests at this moment

BillyWooo:
> HttpClientBuilder

How to make use of `HttpClientBuilder`? I just feel it's another wrapper. What's the benefit?

kziemianek:
> > This is fine for me but personally I would prefer to use this builder in place where HttpClient is instantiated instead of removing it.
> 
> Good point @kziemianek, just for my curiosity, does our team prefer builder pattern over initialisation function?
> 
> I've noticed with this particular case that we have initialisation function and builder, so I decided to keep the one which is used and involved in tests at this moment

I don't think so, you are free to use whatever design pattern is convenient for you.
This one comes from upstream repository.
Why somebody decided to use associate function instead of existing builder ? I don't know.
It's just handy if you need to instantiate objects in different configurations.

Traf333:
Thanks for the feedback. I'll keep in mind possible options and for this particular case since it's not used or having any tests covered I'll 🔥 it

PR #2600, opened 2024-03-19
Removing id_graph_stats getter and related tests

silva-fj:


tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
```diff
@@ -26,7 +26,6 @@ export default {
                 __Unused_evm_account_codes: "Null",
                 __Unused_evm_account_storages: "Null",
                 id_graph: "(LitentryIdentity)",
-                id_graph_stats: "(LitentryIdentity)",
```
BillyWooo, line 29: Would suggest: It's better not to simply remove it. But copy the above example `__Unused_xxx` .

BillyWooo (approved):
LGTM

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
```diff
@@ -26,7 +26,7 @@ export default {
...
+                __Unused_id_graph_stats: "(LitentryIdentity)",
```
Kailai-Wang: This is not strictly necessary though as it's the last item of the enum (thus nothing there in the first place

Kailai-Wang (approved):
Thanks!

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
kziemianek: If we assign this index to new getter it may cause problem for clients using old sdk in the future, that's the only risk a see but at this point I also opt for not having this.

kziemianek (approved):

tee-worker/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
silva-fj: I'll remove it then 👍🏼

PR #2601, opened 2024-03-19
Enforce prefix in AssertionLogic Item src

silva-fj:
I have removed the `Logic` trait as it looks it is not used anywhere. Also the test for the `eval` method breaks when enforcing the prefix only on `src`.


tee-worker/litentry/core/credentials/src/assertion_logic.rs
```diff
@@ -145,11 +130,20 @@ mod tests {
...
+	fn assertion_new_item_adds_dollar_prefix_if_missing() {
+		let item = AssertionLogic::new_item("some_src", Op::GreaterEq, "7");
+		assert_eq!(
+			item,
+			AssertionLogic::Item { src: "$some_src".into(), op: Op::GreaterEq, dst: "7".into() }
+		);
+	}
 
-		let a1 = AssertionLogic::new_or().add_item(web2_item).add_item(web3_item);
-		assert_eq!(a1.eval(), true);
```
BillyWooo, line 153: Would like to get opinion from @zhouhuitian ? Do we still need this `eval()`?

outofxxx:
LGTM.

outofxxx (approved):

PR #2602, opened 2024-03-19
P-618 explore the performance of batch vc

BillyWooo:
minor change:
1. explicitly set threadpool to size of 960; and enclave TCSNum to 1024.
2. remove `unwrap`, catch the error. (Even though it looks like the error won't  happen).


tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -195,8 +196,22 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
 
 				// Totally fine to `unwrap` here. Because new item was just added above.
-				let do_watch = req_registry_pool.update_item(connection_hash).unwrap();
-				send_vc_response(connection_hash, context_pool, response, 0u8, 1, do_watch);
+				match req_registry_pool.update_item(connection_hash) {
+					Ok(do_watch) => {
+						send_vc_response(connection_hash, context_pool, response, 0u8, 1, do_watch);
+					},
+					Err(e) => {
+						error!("1 couldn't find connection_hash: {:?}", e);
```
felixfaisal, line 204: There's a typo here or is `1` to differentiate?

felixfaisal (approved):
LGTM, However there is no evidence present to suggest we can handle 128 size batch_vc.

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -195,8 +196,22 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
...
+						send_vc_response(
+							connection_hash,
+							context_pool,
+							Err(format!("1 couldn't find connection_hash: {:?}", e)),
+							0u8,
+							1,
+							false,
+						);
+					},
```
Kailai-Wang, line 213: It seems no matter `Ok` or `Err` we are going to call `send_vc_response` anyway, only the response and do_watch are different.

Can we have code like this
```
let (response, do_watch) = match req_registry_pool.update_item(connection_hash) { .. };
send_vc_response(....)
```

I think it has a better readability

Kailai-Wang: Same to other places

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -90,7 +90,8 @@ pub fn run_vc_handler_runner<ShieldingKeyRepository, A, S, H, O, Z, N>(
 	N::MetadataType: NodeMetadataTrait,
 {
 	let vc_task_receiver = init_vc_task_sender_storage();
-	let pool = ThreadPool::new().unwrap();
+	let n_workers = 12;
```
Kailai-Wang: I suggest putting this at the beginning with `const`.

kziemianek: Can we make it parametrized instead of hardcoded ? what do you think ?

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
BillyWooo: `response` is not returned from `update_item`.

PR #2603, opened 2024-03-19
Hotfix unit-test in CI

Kailai-Wang:
### Context

CI fails even on the current `dev` branch.

The root cause is we hit the expiration date of 
https://github.com/litentry/litentry-parachain/blob/973c913b9593ceb46f5d3a39645b6adf1dcdbf3c/tee-worker/litentry/core/assertion-build/src/nodereal/bnb_domain/mod.rs#L269

I'm not sure if this is the ideal fix (or if we can manually edit the responses), as we want a stable CI in the long run ofc.

Edit: `dev` branch was updated, so this PR only updates the lock files

### DIFF
--- a/bitacross-worker/Cargo.lock
+++ b/bitacross-worker/Cargo.lock
@@ -6221,9 +6221,9 @@ dependencies = [
 
 [[package]]
 name = "scale-info"
-version = "2.10.0"
+version = "2.11.0"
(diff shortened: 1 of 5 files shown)


BillyWooo (approved):
I accidentally pushed the same thing. But to `dev` branch directly. Not get blocked. Why?

PR #2605, opened 2024-03-20
Compiling non-prod code only when explicitly specified

silva-fj:
Including dev only code when the `production` feature is not enabled makes it posible to include unsafe code to the production build in the cases where the feature has not been properly propagated. These changes switch the logic to include prod code by default and only include dev code when explicitly specified.

### DIFF
--- a/bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
+++ b/bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
@@ -50,7 +50,7 @@ use itp_types::{DirectRequestStatus, RsaRequest, ShardIdentifier, H256};
 use itp_utils::{FromHexPrefixed, ToHexPrefixed};
 use jsonrpc_core::{serde_json::json, IoHandler, Params, Value};
 use lc_scheduled_enclave::{ScheduledEnclaveUpdater, GLOBAL_SCHEDULED_ENCLAVE};
-use litentry_macros::if_not_production;
+use litentry_macros::if_development;
 use litentry_primitives::{AesRequest, DecryptableRequest};
 use log::debug;
 use sgx_crypto_helper::rsa3072::Rsa3072PubKey;
@@ -311,7 +311,7 @@ where
 		Ok(json!(json_value))
 	});
 
-	if_not_production!({
+	if_development!({
 		use itp_types::{MrEnclave, SidechainBlockNumber};
 		// state_setScheduledEnclave, params: sidechainBlockNumber, hex encoded mrenclave
 		io.add_sync_method("state_setScheduledEnclave", move |params: Params| {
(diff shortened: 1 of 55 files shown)


tee-worker/service/src/main_impl.rs
```diff
@@ -98,7 +98,7 @@ pub(crate) fn main() {
 	// stdout from the service
 	#[cfg(feature = "production")]
```
silva-fj: Should we keep the `production` feature? or should we do:

```rust
 #[cfg(not(feature = "development"))]
 ```

the latter looks a bit weird but I think it makes more sense as `production` is the default now, I think having a production feature as well makes it a bit confusing

Comment (Kailai-Wang) [reply tee-worker/service/src/main_impl.rs]: I actually prefer latter as `production` should be the default now.

Is there a better name for `development`? :D 

cc @kziemianek too

Comment (arieljbirnbaum) [reply tee-worker/service/src/main_impl.rs]: `development` makes sense to me tbh :+1:

Comment (kziemianek) [reply tee-worker/service/src/main_impl.rs]: Agree, imho we should have only `development` feature as default code refers to production.

Comment (silva-fj) [reply tee-worker/service/src/main_impl.rs]: It makes sense, thank you guys! 👍🏼

File: bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs [hunk 2]
```diff
@@ -49,8 +49,9 @@ use itp_top_pool_author::traits::AuthorApi;
 use itp_types::{DirectRequestStatus, RsaRequest, ShardIdentifier, H256};
 use itp_utils::{FromHexPrefixed, ToHexPrefixed};
 use jsonrpc_core::{serde_json::json, IoHandler, Params, Value};
+#[cfg(feature = "development")]
 use lc_scheduled_enclave::{ScheduledEnclaveUpdater, GLOBAL_SCHEDULED_ENCLAVE};
-use litentry_macros::if_not_production;
+use litentry_macros::if_development;
```
Kailai-Wang, line 54: How about moving this line
```
use lc_scheduled_enclave::{ScheduledEnclaveUpdater, GLOBAL_SCHEDULED_ENCLAVE};
```

into the `if_development!({` block below?

There might be multiple places.

Kailai-Wang (approved):
Looks good in general, nicely done!

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
silva-fj: I tried using the macro with `use` statements but it does not compile in some cases

e.g:
```
error: expected item after attributes
  --> app-libs/stf/src/trusted_call.rs:76:1
   |
76 | / if_development!({
77 | |     use crate::helpers::ensure_enclave_signer_or_alice;
78 | | });
   | |__^ in this macro invocation
   |
   = note: this error originates in the macro `if_development` (in Nightly builds, run with -Z macro-backtrace for more info)

```

PR #2609, opened 2024-03-22
feat(call verify): support prefixes for user signature

jonalvarezz:
### Context

Add prefix messages support to `link_identity`, `request_batch_vc`, and `set_identity_networks`.

Some things to notice:
1. Signature messages can come in four different ways. [P-639 : clean-up: remove raw payload signature verification](https://linear.app/litentry/issue/P-639/clean-up-remove-raw-payload-signature-verification) will reduce it to two.
2. I didn't add ts-tests cases to keep the PR short. ts-tests will be reworked anyway on P-639. client-sdk was used to ensure the functionality. See video below.
3. Calls' variations with no prefix default to "Token: ".

### Testing Evidences

Test done with IDH:

https://github.com/litentry/litentry-parachain/assets/2167222/314a5179-9818-4ef1-85c3-45bb4407835d

> The video demo a prime bitcoin identity performing link identity, set identity networks, and request batch vc operations.


tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
```diff
@@ -94,9 +94,7 @@ export default {
                 Sr25519: "Sr25519Signature",
                 Ecdsa: "EcdsaSignature",
                 Ethereum: "EthereumSignature",
-                EthereumPrettified: "EthereumSignature",
                 Bitcoin: "BitcoinSignature",
-                BitcoinPrettified: "BitcoinSignature",
             },
```
jonalvarezz, line 98: opportunistic clean-up: I don't see much value in the Prettified variations, so I removed them. My take is: either we add *Prettified to all of them or to none (Substrate and Evm don't have it, yet they support prettified messages signatures)

tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -157,26 +148,18 @@ impl LitentryMultiSignature {
...
-				let prettified_msg =
-					LITENTRY_PRETTIFIED_MESSAGE_PREFIX.to_string() + &hex_encode(msg);
-				verify_bitcoin_signature(prettified_msg.as_str(), sig, signer)
-			},
+				verify_bitcoin_signature(hex::encode(msg).as_str(), sig, signer)
+					|| match std::str::from_utf8(msg) {
+						Err(_) => false,
+						Ok(prettified) => verify_bitcoin_signature(prettified, sig, signer),
+					},
 			_ => false,
```
jonalvarezz, line 163: Is `str::from_utf8` our preferred way for `&[u8]` -> `&str`?

Maybe reworking `verify_bitcoin_signature` could avoid this dance, as it isn't needed in other `verify_` functions, but I didn't want to touch that one :D

tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -107,15 +104,9 @@ pub enum LitentryMultiSignature {
 	/// An ECDSA/keccak256 signature. An Ethereum signature. hash message with keccak256
 	#[codec(index = 3)]
 	Ethereum(EthereumSignature),
-	/// Same as above, but the payload bytes are prepended with a readable prefix and `0x`
-	#[codec(index = 4)]
-	EthereumPrettified(EthereumSignature),
 	/// Bitcoin signed message, a hex-encoded string of original &[u8] message, without `0x` prefix
-	#[codec(index = 5)]
+	#[codec(index = 4)]
```
kziemianek, line 108: Is this a breaking change ?

tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -157,26 +148,18 @@ impl LitentryMultiSignature {
...
 			Self::Bitcoin(ref sig) =>
-				verify_bitcoin_signature(hex::encode(msg).as_str(), sig, signer),
-			Self::BitcoinPrettified(ref sig) => {
-				let prettified_msg =
-					LITENTRY_PRETTIFIED_MESSAGE_PREFIX.to_string() + &hex_encode(msg);
-				verify_bitcoin_signature(prettified_msg.as_str(), sig, signer)
-			},
+				verify_bitcoin_signature(hex::encode(msg).as_str(), sig, signer)
+					|| match std::str::from_utf8(msg) {
+						Err(_) => false,
```
kziemianek, line 160: I think we should rework verify interface and return the actual reason to client.

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -326,9 +338,26 @@ impl TrustedCallVerification for TrustedCallSigned {
...
+
+		let hashed = blake2_256(&payload);
+
+		let prettified_msg_raw = self.call.signature_message_prefix() + &hex_encode(&payload);
+		let prettified_msg_raw = prettified_msg_raw.as_bytes();
+
+		let prettified_msg_hash = self.call.signature_message_prefix() + &hex_encode(&hashed);
+		let prettified_msg_hash = prettified_msg_hash.as_bytes();
+
+		self.signature.verify(&payload, self.call.sender_identity())
```
kziemianek: I'm thinking about having an enum representing which type of signature it is instead of naively verifying them one by one.

We could have:
`Raw`
`Prettified`
`CustomPrettified` (containing client defined prefix ?)

Kailai-Wang: Yes this is better than trying to verify all cases - I'd suggest doing it in a separate PR though as it would be a breaking change to client.

Additionally, we are going to remove 1 and 3 so we only need to verify twice at maximum. If we know IDHub (the only client for now) is using prettified hash, we could verify this case first to slightly improve the processing time :)

tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
Kailai-Wang: Yes - or use the way that Kasper suggested - again it's better to do it separately (together with enum change as it will be breaking)

tee-worker/litentry/primitives/src/lib.rs
```diff
@@ -157,26 +148,18 @@ impl LitentryMultiSignature {
...
+						Ok(prettified) => verify_bitcoin_signature(prettified, sig, signer),
```
Kailai-Wang, line 161: Hmm why do we verify the utf8 string? It seems to be different from what we did.

tee-worker/app-libs/stf/src/trusted_call.rs
jonalvarezz: > We could have:
> `Raw`
> `Prettified`

If there is nothing against removing 1 and 3, only Prettified version is needed. I suggest moving this discussion to https://linear.app/litentry/issue/P-639/clean-up-remove-raw-payload-signature-verification

> we could verify this case first to slightly improve the processing time

Good call – will do. And yes, IDHub is already using prettified hashes only.

tee-worker/litentry/primitives/src/lib.rs
jonalvarezz: Yes, it is, although I already managed it on IDH side. It shouldn't impact current ts-test as no prettified versions are used.

I'll add the breaking-change tag

tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
jonalvarezz: I already managed the breaking change on IDH (the only one impacted by this) so we can proceed with it 👍

tee-worker/litentry/primitives/src/lib.rs
jonalvarezz: For the prettified version. 

Maybe there is another way but this is what I found.

`TrustedCall` controls the signature prefix based on the call and creates the raw + prettified versions:

https://github.com/litentry/litentry-parachain/blob/31050aeb44e159cba5128b3d59886842f68b168c/tee-worker/app-libs/stf/src/trusted_call.rs#L357-L360

I noticed that for Bitcoin signature to work, the prettified version should be encoded as utf8. It won't work when `&str` is passed as `&u8`, so I had to decode back to utf8. I tried to mimic what we currently have:

raw check as hex:
https://github.com/litentry/litentry-parachain/blob/534c2d369eaec2a8a93c062a5a5e8a6f7cd5d7cd/tee-worker/litentry/primitives/src/lib.rs#L174

str prettified check:
https://github.com/litentry/litentry-parachain/blob/534c2d369eaec2a8a93c062a5a5e8a6f7cd5d7cd/tee-worker/litentry/primitives/src/lib.rs#L178

For `verify_bitcoin`, it'd be more convenient to expect `msg` as `&str` but I didn't want to change it because it would involve reworking any other `verify_*` function.

tee-worker/app-libs/stf/src/trusted_call.rs
jonalvarezz: Done

tee-worker/litentry/primitives/src/lib.rs
jonalvarezz: I aligned with Kai about the utf_8 conversion. By looking at the code we thought it wouldn't work, but as shown in the video from the description it works well for Bitcoin, which I have tested throughly.

We agreed the issue comes mainly from `verify_bitcoin_signature` which expects a `str` instead of raw bytes.

tee-worker/ts-tests/integration-tests/common/utils/identity-helper.ts
```diff
@@ -128,32 +128,6 @@ export async function buildValidations(
             'LitentryValidationData',
             bitcoinValidationData
         );
-        validation = encodedVerifyIdentityValidation;
-    } else if (network === 'bitcoinPrettified') {
```
jonalvarezz, line 132: So, removing this means: Prettified messages are supported for `TrustedCall` and `Getter` signatures only.

This code was testing a prettified message on `ValidationData` (the link_identity's challenge code), which clients don't prettify.

It is also a consequence of moving the prettify signature check from `LitentryMultiSignature` to `TrustedCall`/ `Getter`.

`ts-tests` doesn't really cover prettified messages or blake256 payloads. All messages are raw. I'm leaving the ts-test rework to https://linear.app/litentry/issue/P-639/clean-up-remove-raw-payload-signature-verification

cc @Kailai-Wang – this was the missing fix for the CI checks.

jonalvarezz:
@Kailai-Wang , @kziemianek , can we get an approving review on this one? anything else to tackle before merging? :)

kziemianek (approved):

tee-worker/ts-tests/integration-tests/common/utils/identity-helper.ts
Kailai-Wang: > This code was testing a prettified message on ValidationData (the link_identity's challenge code), which clients don't prettify.

This is a good point, shall we prettify it too? I assume not, but we need to let IDHub know about it

Kailai-Wang (approved):

tee-worker/ts-tests/integration-tests/common/utils/identity-helper.ts
jonalvarezz: > IDHub know about it

I think IDHub never knew challenge codes could be prettified 😂

But to be fair, it hasn't been a high-priority issue. Now that I am touching this msg prefix I'll bring up the discussion if we want prettified challenge codes too.

PR #2610, opened 2024-03-22
Register the enclave only once

silva-fj:
Here is the logic:
 
`primary_enclave_identifier_for_shard` returns `None` if there is no worker registered for the shard.

In the case it's `Some`, it means there is a primary worker registered for the shard, we are a non-primary worker if it is first run so we need to request provisioning. Then in any case we register worker.

In case it's `None`, it means there is no primary worker registered for the shard, then we are the primary worker


tee-worker/service/src/main_impl.rs
```diff
@@ -505,110 +505,103 @@ fn start_worker<E, T, D, InitializationHandler, WorkerModeProvider>(
...
+		Some(primary_account_id) => {
+			println!("Primary enclave for shard is: {:?}", primary_account_id);
+			let is_first_run = enclave
+				.get_shard_creation_info(shard)
+				.unwrap()
+				.for_parentchain(ParentchainId::Litentry)
+				.is_none();
+
+			if is_first_run {
+				// we are not primary worker
```
felixfaisal, line 540: So trying to understand the logic here, We are not checking if `primary_account_id` is equal to `tee_account_id` or not, and we are assuming `is_first_run` also means that it is a non-primary worker? If `tee_account_id` is equal to `primary_account_id` are we effectively trying to perform `sync state` with ourselves? Let me know if i'm missing something

silva-fj: In the previous logic, we first registered the enclave, then we checked if there was some enclave registered (the first one, so the primary) and if it was the first run or not to run the init shard logic. With these changes we first try to fetch the primary enclave (before registering), then:

- if `primary_enclave_id` is `Some`, it can be two things:
1. It's the first run, so we assume we are not the primary worker because there is already an enclave registered.  (We could also double check that `tee_accountid` is different than the `primary_account_id`, but I guess it's redundant at this point?). So we sync state.
2. It's not first run, so it may be the primary worker. (In this case there is no additional logic we need to run).
  
   Then we register the enclave.
  
- if `primary_enclave_id` is `None`, there is no enclave registered so we are the primary worker. We register the enclave and run the init shard logic. 

Does this makes sense?

silva-fj: I updated the code to double check the cases where we are not the primary worker to sync state

felixfaisal: Ah, I see, Totally makes sense.

tee-worker/service/src/main_impl.rs
```diff
@@ -505,110 +505,103 @@ fn start_worker<E, T, D, InitializationHandler, WorkerModeProvider>(
...
+	let (we_are_primary_validateer, re_init_parentchain_needed) = match primary_enclave_id {
+		Some(primary_account_id) => {
+			println!("Primary enclave for shard is: {:?}", primary_account_id);
+			let is_first_run = enclave
+				.get_shard_creation_info(shard)
+				.unwrap()
+				.for_parentchain(ParentchainId::Litentry)
+				.is_none();
+
+			if is_first_run && tee_accountid != primary_account_id {
```
kziemianek: Pleas compute this once and reuse in `:540` and  `:565` 

`let we_are_primary_validateer = tee_accountid == primary_account_id`

tee-worker/service/src/main_impl.rs
```diff
@@ -505,110 +505,103 @@ fn start_worker<E, T, D, InitializationHandler, WorkerModeProvider>(
...
+				get_registered_enclave_xt_header(&litentry_rpc_api, register_enclave_block_hash);
 
-	// TODO: #1451: Fix api-client type hacks
-	let register_enclave_xt_header =
-		Header::decode(&mut api_register_enclave_xt_header.encode().as_slice())
-			.expect("Can decode previously encoded header; qed");
+			(primary_account_id == tee_accountid, is_first_run)
+		},
+		None => {
+			println!("No primary enclave for shard was found");
```
kziemianek: Sometimes we use `println!` and sometimes logger. What do you think about changing `println!` to `log::info!` calls ?

tee-worker/service/src/main_impl.rs
```diff
@@ -495,120 +495,112 @@ fn start_worker<E, T, D, InitializationHandler, WorkerModeProvider>(
...
-	let api_register_enclave_xt_header =
-		litentry_rpc_api.get_header(Some(register_enclave_block_hash)).unwrap().unwrap();
+			register_enclave_xt_header =
+				get_registered_enclave_xt_header(&litentry_rpc_api, register_enclave_block_hash);
 
-	// TODO: #1451: Fix api-client type hacks
-	let register_enclave_xt_header =
-		Header::decode(&mut api_register_enclave_xt_header.encode().as_slice())
-			.expect("Can decode previously encoded header; qed");
+			(we_are_primary_validateer, is_first_run)
```
Kailai-Wang, line 564: Overall looks good to me 

One small distinction to the original impl is:
for non-primary worker and not the first run, we'll return `(false, false)` here.

In the old impl we'll never have such value, so `re_init_parentchain_needed` will always be true for non-primary workers.

Shall we change it to the same logic as before?

silva-fj: Good catch, I'll change `re_init_parentchain_needed` to follow the previous logic

Kailai-Wang (approved):

PR #2611, opened 2024-03-22
Prefer no-std url crate

Kailai-Wang:
### Context

IMO we should always prefer the no-std version of a crate over sgx-ported one.

This PR tries to replace `url-sgx` with a no-std `rust-url` crate (a forked one).

Please note in some scenarios where a `IntoClientRequest` is expected (e.g. `tungstenite.client_tls_with_config`) we are using the `&str` that has to include scheme. It's because `Url` to `SocketAddress` conversion is only possible under `std`. cc @kziemianek 

We should be able to switch to the more "standard" crate once `core::net` is supported, see https://github.com/servo/rust-url/pull/831#discussion_r1534293460

It removes some unused code/import and renames the `stf-task-sender` a bit too - I think the path `core/stf-task/sender/src/stf_task_sender.rs` makes little sense when we include `stf_task_sender` twice


kziemianek (approved):
I totally agree that we should prefer no std crates over std/sgx.
I hope we will be able to switch to `url` from crates soon :)

PR #2595, opened 2024-03-18
fix request vc cli tool

BillyWooo:
Minor fix. Just ignore the first response message. Not a perfect solution.
Need to refactor together with Struct RequestVcResultOrError.

`link-identity`:
```
user@parachain-dev-server-01:~/repo/litentry-parachain/tee-worker$ ./bin/litentry-cli trusted -d link-identity did:litentry:substrate:0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 did:litentry:evm:0x0D9bFD1f18f5f4FD08247DC54aD3528909c4b3E9 bsc,ethereum
```

`request-batch-vc` test result:
```
user@parachain-dev-server-01:~/repo/litentry-parachain/tee-worker$ ./bin/litentry-cli trusted -d request-batch-vc did:litentry:substrate:0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 a a8 litentry,litmus  a a8 polkadot
Commands: [("a8", ["litentry,litmus"]), ("a8", ["polkadot"])]
command: "a8", param: ["litentry,litmus"]
command: "a8", param: ["polkadot"]
assertions: [A8(BoundedVec([Litentry, Litmus], 128)), A8(BoundedVec([Polkadot], 128))]
>>>id: Substrate(Address32([142, 175, 4, 21, 22, 135, 115, 99, 38, 201, 254, 161, 126, 37, 252, 82, 135, 97, 54, 147, 201, 18, 144, 156, 178, 38, 170, 71, 148, 242, 106, 72]))
>>>nonce: 1
----Generated VC-----
[...]

PR #2612, opened 2024-03-24
Refactor request-vc CLI

Kailai-Wang:
### Context

A small refactoring of request-vc CLI.

The reasons are:
- duplicate code in `request-vc` and `request-batch-vc` CLI (they were probably copied around)
- IMO the string comparison for `request-batch-vc` is too verbose and insufficient hint is given when entering the wrong parameter (type/count).
- https://github.com/litentry/litentry-parachain/pull/2595#pullrequestreview-1945716984

So this PR tries to:
- provide only one CLI (single vc can be done via batching with size 1)
- replace string comparison with subcommand
- improves the interface a bit, hoping it to be clearer

Basically it looks like this now: 
```
<binary> trusted -d request-vc <did> [-s] -a "<value>" -a "<value>" ...
```

Example of batched vc a1 + a2 + a3:
```
./bin/litentry-cli trusted -d request-vc \
  did:litentry:substrate:0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 \
  -a "a1" -a "a2 gid" -a "a3 gid cid rid"
```
see the comment in the code for more examples


BillyWooo (approved):
Thanks. Very nice refactoring. 
Maybe we don't need option `stf: bool` anymore.

felixfaisal (approved):
LGTM

PR #2607, opened 2024-03-21
P-418 Add Signer check in ts tests

0xverin:
### Context
as the topic

PR #2619, opened 2024-03-26
Fix sidechain nonce retrieval

0xverin:
### Context

`invalidNonce` occurred in [ci](https://github.com/litentry/litentry-parachain/actions/runs/8426759411/job/23077676305?pr=2618#step:10:660). After checking, the sidechain nonce is missing auto-increment  in the tests step.


Traf333 (commented):
It seems like test still fails
```
  expected LinkIdentityFailed, received InvalidNonce instead
```

silva-fj:
Interesting, it fails now in the single worker setup but it passes in the multi worker one

0xverin:
> It seems like test still fails
> 
> ```
>   expected LinkIdentityFailed, received InvalidNonce instead
> ```

I refactored this part of the code(`getSidechainNonce`), chose nonce incremental way, and will finally check the value of sidechiannonce. I deleted the tests case for `link invalid identity` because the use of nonce here is confusing. I will correct this in this [PR](https://github.com/litentry/litentry-parachain/pull/2607). If you have time, please take a look at this PR as well. :p

0xverin:
> Interesting, it fails now in the single worker setup but it passes in the multi worker one

This is another [issue](https://linear.app/litentry/issue/P-588/multi-worker-vc-tests-sometimes-fail-in-ci). I guess it might be caused by the problem here..

0xverin:
In addition, getSidechainNonce was not able to keep up with the speed of the ts-tests update before, as shown below:

```
    ✔ check user sidechain storage after linking (2007ms)
Got response: {"value":"0x00000000","do_watch":false,"status":"Ok"}
context.api.createType('Index', nonce) <BN: 0>
```
 it should not be 0 after linking
I think the speed of ts-tests is accelerated after we cancel the parachain event listening, and sidechainnonce is not updated in time. 
Either we should sleep every time we get sidechain nonce, or can increase it by self. I prefer the latter, because each sleep should be at least 10 seconds, which will lengthen the test time.
This is also why I chose to rewrite this part of the code, and I did check in the end whether the final sidechain nonce is the value we expect.

tee-worker/ts-tests/integration-tests/di_evm_identity.test.ts
```diff
@@ -361,4 +355,9 @@ describe('Test Identity (evm direct invocation)', function () {
 
         await assertIdGraphHash(context, teeShieldingKey, aliceEvmIdentity, idGraph);
     });
+    step('check sidechain nonce', async function () {
+        await sleep(20);
+        const nonce = await getSidechainNonce(context, aliceEvmIdentity);
+        assert.equal(nonce.toNumber(), initialNonce);
```
silva-fj: how about to call it `currentNonce` instead of `initialNonce`?

silva-fj (approved):

tee-worker/ts-tests/integration-tests/di_evm_identity.test.ts
0xverin: Renamed.

PR #2620, opened 2024-03-26
Add more ways to read scheduled enclave

Traf333:
### Context

this PR adds ~2 ways~ a way to read scheduled enclave list
1. ~via binary:`./bitacross-worker scheduled-enclave`~
2. via RPC: `echo '{"jsonrpc":"2.0","method":"state_getScheduledEnclaveList","id":1}' | websocat -n1 -k -B 99999999 wss://localhost:2000`


bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -177,6 +176,11 @@ where
 		}))
 	});
 
+	io.add_sync_method("bitacross_getScheduledEnclaveList", move |_: Params| {
+		debug!("worker_api_direct rpc was called: bitacross_getScheduledEnclaveList");
+		Ok(json!({ "signer": format!("{:?}", &GLOBAL_SCHEDULED_ENCLAVE.current_mrenclave) }))
```
felixfaisal: This is a `RwLock`, you have to first obtain the `read` lock and then try to get the value. Reference - https://github.com/litentry/litentry-parachain/blob/d893864ee3dbded7d05e9c7921936a30c58841c5/bitacross-worker/litentry/core/scheduled-enclave/src/lib.rs#L51C44-L52C1

felixfaisal: Use this function instead `get_current_mrenclave()`, reference - https://github.com/litentry/litentry-parachain/blob/d893864ee3dbded7d05e9c7921936a30c58841c5/bitacross-worker/litentry/core/scheduled-enclave/src/io.rs#L134

bitacross-worker/service/src/main_impl.rs
```diff
@@ -195,6 +196,11 @@ pub(crate) fn main() {
 		setup::generate_signing_key_file(enclave.as_ref());
 		let tee_accountid = enclave_account(enclave.as_ref());
 		println!("Enclave signing account: {:}", &tee_accountid.to_ss58check());
+	} else if matches.is_present("scheduled-enclave") {
+		let mrenclave = enclave.get_fingerprint().unwrap();
+		println!("check mrenclave, {:?}", &mrenclave.0);
+		&GLOBAL_SCHEDULED_ENCLAVE.init(mrenclave.0).unwrap();
```
Traf333: it seems like getting result via RPC works fine, but if I want to obtain data via binary it returns empty result and sometimes fails to run. Do you know why it might happen? @kziemianek @felixfaisal 

on the screenshot it shows when triggering binary several times, some executions are failing and at the end I've got data via RPC and it works fine. Might it mean that `GLOBAL_SCHEDULED_ENCLAVE.registry` is every time different?


image

Kailai-Wang: The error implies a race condition.

`Component factory has not been set` means you try to access the global variable before it's initialized (most likely GLOBAL_SCHEDULED_ENCLAVE in this case)

Kailai-Wang: It's not necessarily your code's problem - it can be "you simply run the test binary too early"

Kailai-Wang: In general I recommend adding more lifecycle events representating the worker state (not in this PR ofc) - currently the only thing we have is `EnclaveAdded`

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -177,6 +176,11 @@ where
 		}))
 	});
 
+	io.add_sync_method("bitacross_getScheduledEnclaveList", move |_: Params| {
```
silva-fj: should this be behind the "development" feature flag ?

kziemianek: I suggest naming it `state_getScheduledEnclaveList`

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -177,6 +176,11 @@ where
...
+		debug!("worker_api_direct rpc was called: bitacross_getScheduledEnclaveList");
+		Ok(json!(format!("{:?}", &GLOBAL_SCHEDULED_ENCLAVE.registry.read())))
```
kziemianek: You should embed it with `RpcReturnValue` to be consistent with most of methods.

kziemianek (changes requested):

kziemianek:
Do you plan adding this method to `tee-worker`?

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
Kailai-Wang: It should be OK for production too as it's meant to be public information

silva-fj: I see, if this is not secret info why do we seal it?

Kailai-Wang: To prevent it from being modified (intentionally or unintentionally) by worker operators. So the public should have read access but no write access.

I mean even if you can't read it, you should be able to infer the content just by checking the parachain tx and tee-worker code : )

silva-fj: Got it 👍🏼

Traf333:
> Do you plan adding this method to `tee-worker`?
@kziemianek I'd say yes and have only one concern, how would it work with multi-worker setup. Does `scheduled_enclave_sealed.bin` synchronises for each worker once it's run?

silva-fj:
> > Do you plan adding this method to `tee-worker`?
> > @kziemianek I'd say yes and have only one concern, how would it work with multi-worker setup. Does `scheduled_enclave_sealed.bin` synchronises for each worker once it's run?

It will once we merge https://github.com/litentry/litentry-parachain/pull/2628

bitacross-worker/service/src/main_impl.rs
Traf333: It seems that this issue `Component factory has not been set` happens with other commands as well. Do we need need to create an issue for that or not worth it? as it seems really low impact

Traf333:
at the end decided to keep only one method to obtain scheduled enclave list as through the binary it seems unstable

silva-fj (approved):

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -177,6 +176,13 @@ where
 		}))
 	});
 
+	io.add_sync_method("state_getScheduledEnclaveList", move |_: Params| {
+		debug!("worker_api_direct rpc was called: state_getScheduledEnclaveList");
+		let value = format!("{:?}", &GLOBAL_SCHEDULED_ENCLAVE.registry.read());
```
kziemianek: This results in something like this:
`Ok({0: [177, 155, 22, 22, 112, 239, 100, 252, 239, 19, 10, 78, 170, 5, 181, 229, 13, 234, 219, 91, 29, 160, 23, 147, 133, 219, 42, 14, 151, 253, 19, 134]})`

1. You should return only value contained in `Result` not `Result` itself.
2. In my opinion value should be serialized directly, not formated to string then serialized.
3. Can we have an integration test for this ?

kziemianek (changes requested):

Traf333:
Included integration tests which also addresses next linear issue [P-645 : Add test case for scheduled-enclave setting](https://linear.app/litentry/issue/P-645/add-test-case-for-scheduled-enclave-setting) where tests covered:
- `state_getScheduledEnclave` 
- couple of invalid and valid scenarios for `state_setScheduledEnclave`

kziemianek (approved):
Nicely done

tee-worker/ts-tests/integration-tests/scheduled_enclave.test.ts
```diff
@@ -0,0 +1,84 @@
...
+import { initIntegrationTestContext } from './common/utils';
+import { getTeeShieldingKey } from './common/di-utils';
+import type { IntegrationTestContext, JsonRpcRequest } from './common/common-types';
+import { decodeRpcBytesAsString, sendRequest } from './common/call';
+import type { CorePrimitivesIdentity } from 'parachain-api';
+import { assert } from 'chai';
+import { createJsonRpcRequest, nextRequestId } from './common/helpers';
+
+describe('Scheduled Enclave', function () {
+    let context: IntegrationTestContext = undefined as any;
```
silva-fj, line 12: why do we define the type and then cast it to `any`?

silva-fj (approved):

tee-worker/ts-tests/integration-tests/scheduled_enclave.test.ts
Traf333: good question 🤔 this one was copy-paste from another test. I just created [chore task](https://linear.app/litentry/issue/P-686/worker-ts-tests-cleanup) to cleanup our tests, as I've also seen some fixme's were copied from one test to another

PR #2621, opened 2024-03-26
Use LinkIdentityRequested event instead of link_identity extrinsic

silva-fj:
This is the step to migrating to using events instead of extrinsics as indirect call triggers. I'll be adding one by one in separate PR's to make it easier to review

### DIFF
--- a/pallets/identity-management/src/lib.rs
+++ b/pallets/identity-management/src/lib.rs
@@ -81,6 +81,10 @@ pub mod pallet {
 		// TODO: do we need account as event parameter? This needs to be decided by F/E
 		LinkIdentityRequested {
 			shard: ShardIdentifier,
+			account: T::AccountId,
+			encrypted_identity: Vec<u8>, // TODO: check if this is the correct type
+			encrypted_validation_data: Vec<u8>,
+			encrypted_web3networks: Vec<u8>,
 		},
 		DeactivateIdentityRequested {
 			shard: ShardIdentifier,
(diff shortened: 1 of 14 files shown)


pallets/identity-management/src/lib.rs
```diff
@@ -81,6 +81,10 @@ pub mod pallet {
 		// TODO: do we need account as event parameter? This needs to be decided by F/E
 		LinkIdentityRequested {
 			shard: ShardIdentifier,
+			account: T::AccountId,
```
silva-fj, line 84: Do we need to update something on the clients/sdk for this ? @jonalvarezz

BillyWooo (commented):
Over all, it looks good to me. 
I just recall my memory.  I think we had something like this before for Indirect Invocation mode. Then we removed it. Now we add it back. I agree to add it back. Just forget what's the reason to remove before.

pallets/identity-management/src/lib.rs
jonalvarezz: Thanks @silva-fj , I checked it up, and can confirm the type isn't use on the client-sdk. Furthermore, client-sdk doesn't not really on parachain other than for fetching metadata and registered enclave data.

silva-fj:
> Over all, it looks good to me. I just recall my memory. I think we had something like this before for Indirect Invocation mode. Then we removed it. Now we add it back. I agree to add it back. Just forget what's the reason to remove before.

Maybe it's worth it to find the issue/PR that removed it to know the reason

silva-fj:
I will create a new PR with new changes

PR #2622, opened 2024-03-27
Using events to trigger deactivate and activate identity indirect calls

silva-fj:
This is a follow up of https://github.com/litentry/litentry-parachain/pull/2621

Note: currently there is no way for me to test these calls, I'll create a separate PR to add commands to the cli to activate/deactivate identities.

Once the previous PR get's merged, I'll change the base branch


silva-fj:
I will create a new PR with the changes combined

PR #2624, opened 2024-03-28
Adding Deactivate and Activate identity in the cli

silva-fj:
Context: Needed to be able to easily test indirect call invocations

### DIFF
--- a/tee-worker/cli/src/base_cli/commands/litentry/activate_identity.rs
+++ b/tee-worker/cli/src/base_cli/commands/litentry/activate_identity.rs
@@ -20,14 +20,11 @@ use crate::{
 	Cli, CliResult, CliResultOk,
 };
 use codec::{Decode, Encode};
-use ita_stf::{helpers::get_expected_raw_message, Web3Network};
 use itc_rpc_client::direct_client::DirectApi;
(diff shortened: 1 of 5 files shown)


tee-worker/cli/src/base_cli/commands/litentry/activate_identity.rs
```diff
@@ -0,0 +1,74 @@
...
+		let signer = self.get_signer();
+
+		chain_api.set_signer(signer.into());
+
+		let (_, encrypted_identity) = self.encrypt_identity(cli);
+
+		let xt = compose_extrinsic!(chain_api, IMP, "activate_identity", shard, encrypted_identity);
+
+		let tx_hash = chain_api.submit_and_watch_extrinsic_until(xt, XtStatus::Broadcast).unwrap();
+		println!("[+] ActivateIdentityCommand got finalized. Hash: {:?}", tx_hash);
```
Kailai-Wang: Should it be `XtStatus::Finalized` to match the log below?

Kailai-Wang (approved):
Looks good from my side, thanks!

Please also be aware that identity-related indirect calls are not actively maintained, so it is more crucial to make sure other II calls (e.g. set_scheduled_enclave) work well - but I can understand the need of having this PR

PR #2625, opened 2024-03-28
Using events instead of extrinsics as indirect call triggers

silva-fj:
[P-642 Use events instead of extrinsics as indirect call triggers](https://linear.app/litentry/issue/P-642/use-events-instead-of-extrinsics-as-indirect-call-triggers)


pallets/identity-management/src/benchmarking.rs
```diff
@@ -62,7 +62,13 @@ benchmarks! {
 		let encrypted_web3networks = vec![1u8; 2048];
 	}: _(RawOrigin::Signed(caller.clone()), shard, caller.clone(), encrypted_did, encrypted_validation_data, encrypted_web3networks)
 	verify {
-		assert_last_event::<T>(Event::LinkIdentityRequested{ shard }.into());
+		assert_last_event::<T>(Event::LinkIdentityRequested{
+			shard,
+			account: caller,
+			encrypted_identity: vec![1u8; 2048],
```
kziemianek: There are `encrypted_did`, `encrypted_validation_data`, `encrypted_web3networks` variables declared before with exact same values, why don't you use them ?

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -53,6 +63,134 @@ impl ParentchainEventHandler {
...
+		let trusted_operation =
+			TrustedOperation::<TrustedCallSigned, Getter>::indirect_call(signed_trusted_call);
+
+		let encrypted_trusted_call = executor.encrypt(&trusted_operation.encode())?;
+		executor.submit_trusted_call(shard, encrypted_trusted_call);
+
+		Ok(())
+	}
+
+	fn call_identity_action<Executor: IndirectExecutor<TrustedCallSigned, Error>>(
```
kziemianek: I doubt is worth expressing it this way. 
From my perspective it's an attempt to extract shared code for processing events that consist only `encrypted_identity` - this might be misleading. 

`link_identity` can also be treated as "identity_action" but it's not handled here.

Given that `shard` and `enclave_account_id` retrieval, identity decoding are also performed in other parachains event handling functions I would suggest replacing this functiion with `activate_identity` and `deactivate_identity`

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -53,6 +63,134 @@ impl ParentchainEventHandler {
...
+		Ok(())
+	}
+
+	fn set_scheduled_enclave(
+		worker_type: WorkerType,
+		sbn: SidechainBlockNumber,
+		mrenclave: MrEnclave,
+	) -> Result<(), Error> {
+		if worker_type != WorkerType::Identity {
+			warn!("Ignore RemoveScheduledEnclave due to wrong worker_type");
```
kziemianek: Could you improve log ? This handling is not related to removing scheduled enclave.

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -82,6 +219,98 @@ where
 				})
 				.map_err(|_| ParentchainError::ShieldFundsFailure)?;
 		}
+
+		if let Ok(events) = events.get_events::<LinkIdentityRequested>() {
+			debug!("Handling link_identity events");
+			events
+				.iter()
+				.try_for_each(|event| {
+					info!("found link_identity_event: {}", event);
```
kziemianek: have you considered changing the log level to `debug` ?

tee-worker/core-primitives/types/src/parentchain/events.rs
```diff
@@ -0,0 +1,225 @@
...
+pub struct ScheduledEnclaveSet {
+	pub worker_type: WorkerType,
+	pub sidechain_block_number: SidechainBlockNumber,
+	pub mrenclave: MrEnclave,
+}
+
+impl core::fmt::Display for ScheduledEnclaveSet {
+	fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+		let message = format!(
+			"ScheduledEnclaveSet :: worker_type: {:?}, sidechain_block_number: {}, mrenclave: {:?}",
```
kziemianek: Have you considered referring to `StaticEvent` trait you implemented for this struct? 
You can get event name and pallet name:

```
	fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
		let message = format!(
			"[{}:{}] :: worker_type: {:?}, sidechain_block_number: {}, mrenclave: {:?}",
			ScheduledEnclaveSet::PALLET, ScheduledEnclaveSet::EVENT, self.worker_type, self.sidechain_block_number, self.mrenclave
		);
		write!(f, "{}", message)
	}
```

tee-worker/core-primitives/types/src/parentchain/mod.rs
```diff
@@ -180,13 +118,29 @@ where
 pub enum ParentchainError {
```
kziemianek: What do you think about removing this enum and replacing it with just `String` ? 

If you decide to keep it, maybe rename it to `ParentchainEventProcessingError` ?

kziemianek (commented):
Overall it looks fine, but I doubt it's finished. For example is `parse_batch_all` going to be removed ?

tee-worker/core-primitives/types/src/parentchain/mod.rs
silva-fj: I'll rename it for now

silva-fj:
> Overall it looks fine, but I doubt it's finished. For example is `parse_batch_all` going to be removed ?

Good question, I'm not sure what to do in this case. As far as I understand `batchAll` will deposit `BatchCompleted` event which has no info about the data sent in the extrinsics (please correct me if I'm wrong). So, I guess we need to keep it. c.c: @Kailai-Wang

kziemianek:
> > Overall it looks fine, but I doubt it's finished. For example is `parse_batch_all` going to be removed ?
> 
> Good question, I'm not sure what to do in this case. As far as I understand `batchAll` will deposit `BatchCompleted` event which has no info about the data sent in the extrinsics (please correct me if I'm wrong). So, I guess we need to keep it. c.c: @Kailai-Wang

Batched calls should still emit events upon execution.

silva-fj:
> > > Overall it looks fine, but I doubt it's finished. For example is `parse_batch_all` going to be removed ?
> > 
> > 
> > Good question, I'm not sure what to do in this case. As far as I understand `batchAll` will deposit `BatchCompleted` event which has no info about the data sent in the extrinsics (please correct me if I'm wrong). So, I guess we need to keep it. c.c: @Kailai-Wang
> 
> Batched calls should still emit events upon execution.

Oh, I see. 👍🏼

silva-fj:
Do we also want to migrate calls between parachain and worker? like `post_opaque_task` ?

kziemianek:
> Do we also want to migrate calls between parachain and worker? like `post_opaque_task` ?

I don't see any reason for such exception, do you?
Can it emit event so we can handle it on worker's side ?

Then can we remove handling of `batch_all` and `post_opaque_task` calls from worker completely and rely purely on events?

silva-fj:
> post_opaque_task

Yes we can, we just need to include the `RsaRequest` in the event

tee-worker/core/parentchain/indirect-calls-executor/src/executor.rs
```diff
@@ -153,66 +145,33 @@ impl<
...
+		ParentchainEventHandler::handle_events(self, events)?;
+
+		debug!("successfully processed {} indirect invocations", events_count);
+
+		// TODO: what to pass to create_processed_parentchain_block_call??
+		// if self.parentchain_id == ParentchainId::Litentry {
+		// Include a processed parentchain block confirmation for each block.
+		// Ok(Some(self.create_processed_parentchain_block_call::<ParentchainBlock>(
+		// 	block_hash,
+		// 	executed_calls,
```
silva-fj, line 167: Should we return the calls somehow from the event handler to pass them here?

kziemianek (approved):

PR #2627, opened 2024-03-29
p-657 Upgrade web3.js version

0xverin:
### Context

as the topic

Rewrote the `executeTransaction` method of precompile-contract: changed to use `ethers` instead of `web3.js`, because `web3.js` has a strange [issue](https://github.com/litentry/litentry-parachain/actions/runs/8510235570/job/23307659315#step:6:1040) that is difficult to debug.


ts-tests/integration-tests/precompile-contract.test.ts
```diff
@@ -132,6 +134,7 @@ describeLitentry('Test Parachain Precompile Contract', ``, (context) => {
                     to: precompileStakingContractAddress,
                     data: delegateWithAutoCompound.encodeABI(),
                     gas: await delegateWithAutoCompound.estimateGas(),
+                    gasPrice: await web3.eth.getGasPrice(),
```
BillyWooo: missing `nonce` ?

ts-tests/integration-tests/evm-contract.test.ts
```diff
@@ -147,7 +148,7 @@ describeLitentry('Test EVM Module Contract', ``, (context) => {
             return data;
         };
 
-        const message = await sayMessage(deployed.contractAddress!);
+        const message = (await sayMessage(deployed.contractAddress!)) as unknown as string;
```
silva-fj: If we know that message should always be `string`, how about making the return type of `sayMessage` explicit:
 `const sayMessage = async (contractAddress: string): Promise<string> => {...}`

arieljbirnbaum (approved):

PR #2628, opened 2024-04-01
Provisioning scheduled_enclave_sealed.bin file

silva-fj:


tee-worker/litentry/core/scheduled-enclave/src/lib.rs
```diff
@@ -39,6 +41,8 @@ use std::sync::SgxRwLock as RwLock;
 use lazy_static::lazy_static;
 use std::sync::Arc;
 
+pub mod mock;
+
 lazy_static! {
 	/// Global instance of a ScheduledEnclave
 	pub static ref GLOBAL_SCHEDULED_ENCLAVE: Arc<ScheduledEnclave> = Default::default();
```
kziemianek, line 48: Have you checked if this runtime state is populated after provisioning ?

silva-fj: Yes, it gets initialized after the state have been synced, so during initialization it finds the file and gets the data from it.

Here:
https://github.com/litentry/litentry-parachain/blob/bc9896b5f8651b547e789691bb0a6cbecb3cc898/tee-worker/litentry/core/scheduled-enclave/src/io.rs#L105

tee-worker/enclave-runtime/src/tls_ra/mocks.rs
```diff
@@ -29,6 +29,7 @@ pub struct SealHandlerMock {
 	pub state_key: Arc<RwLock<Vec<u8>>>,
 	pub state: Arc<RwLock<Vec<u8>>>,
 	pub light_client_state: Arc<RwLock<Vec<u8>>>,
+	pub schedule_enclave_state: Arc<RwLock<Vec<u8>>>,
```
kziemianek: schedule -> scheduled ?

tee-worker/litentry/core/scheduled-enclave/src/lib.rs
kziemianek: Ok I see, so there is no need to pass `mrenclave` to `GLOBAL_SCHEDULED_ENCLAVE.init` in `init_enclave_sidechain_components` for workers that requested state provisioning because the file is already sealed with provisioned state.

kziemianek (approved):

tee-worker/litentry/core/scheduled-enclave/src/lib.rs
silva-fj: Well, it is used to init the current mrenclave with `set_current_mrenclave`, which is used in other places

kziemianek: But in case of running non primary worker it's not used ?
If yes, can it be optional if worker is non primary?

silva-fj: it seems that the value needs to be initialized regardless of the type of worker. it is used [here](https://github.com/litentry/litentry-parachain/blob/7c0cec586e785069f3545cfafad063263c9bc7c6/tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs#L331)

Kailai-Wang (approved):
Looks good!

tee-worker/litentry/core/scheduled-enclave/src/lib.rs
kziemianek: Ok, name is a bit misleading. 
It's worker's mrenclave, not current active mrenclave from scheduled mrencalve set.

PR #2635, opened 2024-04-02
fix: P-654 added solana signature check

wli-pro:
### Context

as titles

### Labels

### Testing Evidences

Tested via unit test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -685,6 +685,7 @@ jobs:
           - test_name: lit-di-substrate-identity-test
           - test_name: lit-di-evm-identity-test
           - test_name: lit-di-bitcoin-identity-test
+          - test_name: lit-di-solana-identity-test
           - test_name: lit-di-vc-test
           - test_name: lit-dr-vc-test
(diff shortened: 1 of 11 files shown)


silva-fj (commented):
Do we have a place where we put shared/common code between workers? if not, what's the reason?

BillyWooo:
> Do we have a place where we put shared/common code between workers? if not, what's the reason?

This is under discussion. Check issue P-600 and related ones.

BillyWooo:
The changes looks good to me. Can you take a look at the CI test failure of one worker?

Kailai-Wang (commented):
Shall we have ts-test for it (at least happy path)?

wli-pro:
> Shall we have ts-test for it (at least happy path)?

Yes integration tests have been added too.

Kailai-Wang:
> > Shall we have ts-test for it (at least happy path)?
> 
> Yes integration tests have been added too.

I'm not sure if the test is run in CI though

Kailai-Wang (approved):
Thanks!

PR #2636, opened 2024-04-02
Removing shield/unshield funds related code

silva-fj:


BillyWooo:
LGTM. Maybe @kziemianek and @felixfaisal  please also take a look. Thank you.

tee-worker/cli/demo_shielding_unshielding.sh
```diff
@@ -181,19 +181,6 @@ echo "* Create a new incognito account for Bob"
 echo "  Bob's incognito account = ${ICGACCOUNTBOB}"
```
silva-fj: Should we remove all `*shielding_unshielding*.sh` files ? I kept them because they do other calls as well that may be good to test (I'm not sure if we are testing those somewhere else), but the file names for these tests are misleading

kziemianek: If we want to keep it we should have a test for a nonce increment by failed trusted call - that is actually run in CI.
From what I know, this test is not run by CI so my suggestion is to remove it and create issue for adding a test case for scenario described above.

kziemianek (approved):

PR #2638, opened 2024-04-03
Link sgx_epid in DCAP feature

outofxxx:
Fix P-446

using command below:
```
WORKER_FEATURES=dcap make
running worker without `skip-ra` flag
```

Partial P-117 solved. The error under `HW` mode has been resolved, but the `200e` issue still remains under SW mode.
 If possible, We can merge this PR first to allow MAA's PR to continue. I will address the issues under SW mode later.


BillyWooo (approved):
LGTM

PR #2641, opened 2024-04-05
Adding support for non public twitter verification (OAuth2)

silva-fj:
This PR introduces a new option to verify twitter accounts, using twitter's OAuth2.0. We will support multiple ways to verify accounts (later we will explore ZK proofs).

This new option works in the following way:

1. The client requests the authorize url from a new RPC method `identity_getTwitterAuthorizeUrl`, sending the `redirect_uri`  and the `identity` of the user wanting to link the twitter account. This generates the code verifier for that `account_id`, this code is saved for later verification.
2. The client presents this link to the user, so the user can authorize the app. Once the authorizes the app, it will be redirected to the client's app with the oauth `code`.
3. The client submits the `link_identity` trusted call, sending the oauth `code` from twitter and the `redirect_url` sent previously as validation data.

> [!IMPORTANT]
> - There are 2 new environment variables needed for the twitter API as the OAuth2.0 endpoints use a different mechanism for authentication: `TWITTER_CLIENT_ID` and `TWITTER_CLIENT_SECRET`. 
> - The "Twitter App" we use (the developer account we use to get the API KEY) needs to be configured properly with name, description, etc. As the users will see that info on the authorize page.


tee-worker/litentry/core/identity-verification/src/web2/twitter/code_verifier_store.rs
```diff
@@ -0,0 +1,44 @@
...
+
+lazy_static! {
+	static ref TWITTER_CODE_VERIFIER_STORE: Arc<RwLock<HashMap<String, String>>> =
+		Arc::new(RwLock::new(HashMap::new()));
+}
+
+pub struct CodeVerifierStore;
+
+impl CodeVerifierStore {
+	pub fn save_code(account_id: AccountId, code: String) -> Result<(), String> {
```
kziemianek: Can we remove codes after some time? Otherwise this will grow and grow.

silva-fj: The code is removed when consumed. I guess there could be some cases where the code was created but never consumed, for example if a user started the link identity process on the client but did not continue with the verification, but I'm not sure if this would be a problem. We would need to implement some sort of expire mechanism

kziemianek: I suggest adding an expiration logic and a limit of N codes at time stored untill we find more memory safety storage.

silva-fj: I changed the data structure to use LRU cache, when the data is consumed it is removed from the store and if the store hits the capacity limit the new values will replace the least recently used values.

.github/workflows/ci.yml
```diff
@@ -808,6 +809,7 @@ jobs:
           - test_name: lit-dr-vc-multiworker-test
           - test_name: lit-resume-worker
           - test_name: lit-scheduled-enclave-multiworker-test
+          - test_name: lit-twitter-identity-multiworker-test
```
Kailai-Wang: In fact, I don't think multi-worker test is much needed here

tee-worker/core/rest-client/src/http_client.rs
```diff
@@ -51,6 +51,7 @@ pub trait SendHttpRequest {
 		params: U,
 		query: Option<&Query<'_>>,
 		maybe_body: Option<String>,
+		headers: Option<Headers>,
```
Kailai-Wang: Hmm there's a `HttpClient::set_header` method already, does it meet the requirement?

tee-worker/enclave-runtime/src/initialization/mod.rs
```diff
@@ -200,12 +200,20 @@ pub(crate) fn init_enclave(
...
+	}
+
+	let data_provider_config = GLOBAL_DATA_PROVIDER_CONFIG.get()?;
 	let getter_executor = Arc::new(EnclaveGetterExecutor::new(state_observer));
 	let io_handler = public_api_rpc_handler(
 		top_pool_author,
 		getter_executor,
 		shielding_key_repository,
 		Some(state_handler),
+		data_provider_config,
```
Kailai-Wang, line 216: It's good practice to replace the global variable with dependency injection, however, I was thinking to have a dedicated PR for it (so not address it in this PR)

Otherwise it might be confusing why we mix them up - sometimes we use GLOBAL, sometimes parameter

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -472,6 +475,52 @@ where
 		Ok(Value::String(format!("hello, {}", parsed)))
 	});
 
+	io.add_sync_method("identity_getTwitterAuthorizeUrl", move |params: Params| {
+		debug!("worker_api_direct rpc was called: identity_getTwitterAuthorizeUrl");
+
+		match params.parse::<(String, String)>() {
+			Ok((encoded_identity, redirect_url)) => {
```
Kailai-Wang: How about using the did string form here? Maybe it makes the interface simple for the client

tee-worker/litentry/core/identity-verification/src/web2/twitter/oauth_store.rs
```diff
@@ -0,0 +1,42 @@
+#[cfg(feature = "std")]
+use std::sync::RwLock;
+#[cfg(feature = "sgx")]
+use std::sync::SgxRwLock as RwLock;
+
+#[cfg(feature = "std")]
+use hex;
+#[cfg(feature = "sgx")]
+use hex_sgx as hex;
```
Kailai-Wang: `hex` has no_std version

tee-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -472,6 +475,52 @@ where
...
+
+				let json_value = RpcReturnValue::new(
+					authorize_data.authorize_url.encode(),
+					false,
+					DirectRequestStatus::Ok,
+				);
+
+				Ok(json!(json_value.to_hex()))
+			},
+
```
Kailai-Wang: I think the error and return value handling could be improved to have better readability

tee-worker/litentry/core/identity-verification/src/web2/mod.rs
```diff
@@ -56,32 +54,103 @@ pub fn verify(
...
+							Error::LinkIdentityFailed(ErrorDetail::StfError(
+								ErrorString::truncate_from(
+									std::format!(
+										"no oauth data found for {}",
+										account_id_to_string(&account_id)
+									)
+									.into(),
+								),
+							))
+						})?,
```
Kailai-Wang: Is there any `into` for it?

tee-worker/litentry/core/data-providers/src/twitter_official.rs
```diff
@@ -99,6 +101,22 @@ pub struct TargetTwitterUser {
 	pub followed_by: bool,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct CreateTwitterUserAccessToken {
```
Kailai-Wang: It's a bit uncommon that a struct name starts with `Create...`

tee-worker/litentry/core/identity-verification/src/web2/twitter/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[derive(Debug)]
+pub struct AuthorizeData {
+	pub authorize_url: String,
+	pub code_verifier: String,
+	pub state: String,
+}
+
+const BASE_URL: &str = "https://twitter.com/i/oauth2/authorize";
+const SCOPES: &str = "tweet.read%20users.read";
```
Kailai-Wang, line 18: Do we need tweet.read?
Would users.read suffice for verification purpose?

Kailai-Wang (commented):
The logic looks good, thanks!

I think in the future we should allow the twitter_client_id to be registered on the parachain (as part of the enclave metadata as verification service)

.github/workflows/ci.yml
Kailai-Wang: IMO it's OK to stick to one worker throughout the verification process.

But maybe we want to turn the `unwrap` into an error message here: https://github.com/litentry/litentry-parachain/actions/runs/8819687857/job/24212432931?pr=2641

local-setup/.env.dev
```diff
@@ -26,6 +26,8 @@ NODE_ENV=local
 # The following key/token are MANDATORY to give when running worker.
 # Otherwise request-vc might suffering from data provider error.
 TWITTER_AUTH_TOKEN_V2=
+TWITTER_CLIENT_ID=twitter_client_id
+TWITTER_CLIENT_SECRET=twitter_client_secret
```
BillyWooo, line 30: Should these two parameters be put somewhere else? I think they are not keys we need to protect?

tee-worker/litentry/core/identity-verification/src/web2/twitter/mod.rs
silva-fj: We need both to [request](https://developer.twitter.com/en/docs/twitter-api/users/lookup/api-reference/get-users-me) the user

tee-worker/core/rest-client/src/http_client.rs
silva-fj: I've adjusted the code to expose it to the rest client

.github/workflows/ci.yml
silva-fj: 👍🏼

tee-worker/enclave-runtime/src/initialization/mod.rs
silva-fj: Yeah, it makes sense. For the case of `GLOBAL_DATA_PROVIDER_CONFIG` we don't use it as a global variable, we use dependency injection in all places

Kailai-Wang: I see - `GLOBAL_DATA_PROVIDER_CONFIG` is only used when initialising the vc-task and stf-task thread

Kailai-Wang (approved):
Thanks!

PR #2644, opened 2024-04-09
P-681 Merge EVM and non-EVM code

wangminqi:
This is for p-681.


node/src/rpc.rs
```diff
@@ -50,11 +48,9 @@ use crate::tracing;
 /// A type representing all RPC extensions.
 pub type RpcExtension = jsonrpsee::RpcModule<()>;
 
-#[reuse(evm, no_evm)]
-pub mod __ {
+pub mod evm {
```
Kailai-Wang: Shall we simply remove this mod scope?

Kailai-Wang (commented):
Looks good in general! One small thing only

node/src/rpc.rs
wangminqi: let's save it and remove in next pr? since if we want to revert, I hate to handle the whitespace error by clippy.

Kailai-Wang (approved):

PR #2645, opened 2024-04-09
revert info! back to println!

BillyWooo:
very minor change, but `info!` can't be catched within shell script.


BillyWooo:
doesn't need it anymore.

PR #2646, opened 2024-04-09
Do not monitor writable events for ws server

Kailai-Wang:
### Context

As topic, otherwise we are trapped in an endless loop of writable readiness, which causes 100% cpu.

So IMO we shouldn't monitor the writable event as there's no case that server proactively writes something after the handshake (and even if there is we don't want to handle it)

This PR adds the `deregister` call too.

This PR **might** fix the multi-worker CI failure - or at least a remedy for it.


felixfaisal (approved):
Thanks @Kailai-Wang , It looks good to me.

kziemianek (approved):

PR #2648, opened 2024-04-10
Removing init_shard_config

silva-fj:
This is used upstream to update a shard registry (on-chain) storage with the configs of shards on the parentchain. We don't have this in our parachain.


tee-worker/enclave-runtime/Cargo.toml
```diff
@@ -91,7 +91,6 @@ webpki = { git = "https://github.com/mesalock-linux/webpki", branch = "mesalock_
 base58 = { rev = "sgx_1.1.3", package = "rust-base58", default-features = false, features = ["mesalock_sgx"], git = "https://github.com/mesalock-linux/rust-base58-sgx" }
 
 cid = { default-features = false, git = "https://github.com/whalelephant/rust-cid", branch = "nstd" }
-enclave-bridge-primitives = { default-features = false, git = "https://github.com/integritee-network/pallets.git", branch = "sdk-v0.12.11-polkadot-v0.9.42" }
```
Kailai-Wang, line 94: Thanks, good catch :)

Kailai-Wang (approved):
Looks good from my side

PR #2657, opened 2024-04-11
Fix evm build

kziemianek:
The build for `evm` feature is broken again. There is a check for that in `scripts/pre-commit.sh` script but CI does not ensure code correctness.

This PR:
* fixes `evm` build
* adds `evm` feature check to CI


Kailai-Wang (approved):
Thanks!

PR #2658, opened 2024-04-12
chore: add missing tests in ci

wli-pro:
### Context

as titles

added all ts tests for multi-worker tests too
also renamed bit to make the unique test name to show in ci:

Screenshot 2024-04-12 at 12 14 36 pm

Screenshot 2024-04-12 at 1 36 10 pm

### Labels

### Testing Evidences

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -776,6 +776,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - test_name: lit-di-bitcoin-identity-multiworker-test
           - test_name: lit-di-evm-identity-multiworker-test
           - test_name: lit-di-solana-identity-multiworker-test
           - test_name: lit-di-substrate-identity-multiworker-test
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -686,14 +686,19 @@ jobs:
         include:
           # Litentry
           - test_name: lit-di-substrate-identity-test
+            overridden_test_name: '[substrate] tee-single-worker-test(lit-di-identity)'
```
Traf333: what do you think if we will call it **label** instead? Otherwise we will get a question, why it's overridden

Traf333 (approved):
overall looks good. Just small nitpicking to the wording

.github/workflows/ci.yml
BillyWooo: what's the benefit of "overridden" ?

Traf333: it helps to see which tests failed. Otherwise it's hidden or difficult to read
Screenshot 2024-04-12 at 17 40 19

wli-pro: per Igor's suggest, I renamed overridden to label now, used in the above graph to show important text first without being hidden

PR #2659, opened 2024-04-12
Run `identity-worker` CI tests without `development` feature

kziemianek:
CI should run tests for `identity-worker` build without `development` feature to ensure correctness of production code

This PR:

- adds `env-data-providers-config`  feature, if enabled enclave will read data providers config from env variables
- adds `mock-server` feature, if enabled enclave will start mock server if other conditions are met 
- adds `WORKER_ENV_DATA_PROVIDERS_CONFIG` argument to `Makefile`, if `WORKER_ENV_DATA_PROVIDERS_CONFIG=1` is passed the worker will be build with `env-data-providers-config` feature, by default worker is built without it 
- adds `WORKER_MOCK_SERVER` argument to `Makefile`, if `WORKER_MOCK_SERVER=1` is passed the worker will be build with `mock-server` feature, by default worker is built without it 
- fixes build without `development` feature
- adds clippy checks without `development` feature
- fixes clippy warnings
- makes CI tests to be run against `identity-worker` built without `development` feature

> [!IMPORTANT]
> All deployment script that expect worker to read data providers config from environment and start mock server must be adjusted.


tee-worker/litentry/pallets/identity-management/src/lib.rs
```diff
@@ -221,7 +221,6 @@ pub mod pallet {
 			})
 		}
 
-		#[cfg(feature = "development")]
```
kziemianek: I assume this removal is safe - runtime is reachable only through trusted calls.

tee-worker/ts-tests/integration-tests/scheduled_enclave.test.ts
```diff
@@ -38,47 +38,4 @@ describe('Scheduled Enclave', function () {
         assert.equal(blockNumber, 0);
         assert.equal(mrEnclave, context.mrEnclave);
     });
-
```
kziemianek: Sadly I had to remove this step. This is `development` feature so it's not possible to test it with worker built without it.

Traf333: sounds reasonable 👍  It's me who implemented this test in a wrong way. I had to use indirect invocation instead

Traf333: for this case I'll implement the proper testing. Thanks for raising this up

Traf333: I've prepared PR to use II for setting enclave instead of using dev RPC method
https://github.com/litentry/litentry-parachain/pull/2664

tee-worker/service/src/main_impl.rs
```diff
@@ -193,18 +193,16 @@ pub(crate) fn main() {
...
-						.try_parse_mock_server_port()
-						.expect("mock server port to be a valid port number");
-					thread::spawn(move || {
-						info!("*** Starting mock server");
-						let _ = lc_mock_server::run(mock_server_port);
-					});
-				},
-				warn!("Mock server not started. Node is running in production mode.")
-			)
+			#[cfg(any(feature = "mock-server", feature = "development"))]
```
Kailai-Wang, line 196: Since it's `any` here, why don't we use `development` feature only to control it? Do we have a case where we want to compile/launch mock-server for production build?

Additionally, it could be a bit misleading if we have `WORKER_MOCK_SERVER=0` and `WORKER_DEV=1` => the mock-server will be launched (even if we set it explicitly to 0)

kziemianek: We cannot use only `development` feature in this case because we want to run CI checks with mock server while having build without `development` feature.


Two conditions needs to be met in order to start mock server:
1. worker needs to be built with one of mentioned features
2. `config.enable_mock_server` must be set to true

I agree that it can be misleading that worker built without `mock-server` feature will launch it if config allows for that - but I assumed it's still fine since there is `development` feature. I can change it so it will be included only by `mock-server` feature, what do you think?

felixfaisal (approved):
Thanks @kziemianek , LGTM

jonalvarezz (approved):

PR #2660, opened 2024-04-12
Remove sidechain functionality from bc-worker

Traf333:
### Context

this PR tends to remove redundant code for bc-worker since we don't use sidechain there. In addition it removes a `changelog` folder which uses ruby language and not used at our project


bitacross-worker/enclave-runtime/src/lib.rs
```diff
@@ -486,203 +476,6 @@ pub unsafe extern "C" fn migrate_shard(
...
-	) {
-		error!("Error synching parentchain: {:?}", e);
-		return sgx_status_t::SGX_ERROR_UNEXPECTED
-	}
-
-	sgx_status_t::SGX_SUCCESS
-}
-
-#[allow(clippy::too_many_arguments)]
-unsafe fn sync_parentchain_internal(
```
Traf333: it seems that with removal of this code omits the event `teebag.ParentchainBlockProcessed` which is triggered every parachain block. I've removed this logic since it was referenced to `bitacross-worker/core-primitives/enclave-api/src/sidechain.rs` and I'm not sure  do we want to initiate this event or not? 
@kziemianek @felixfaisal , do you have any suggestions?

Kailai-Wang: IMO we'll need this event - it signals the processing of parentchain block syncing

Kailai-Wang:
cc @kziemianek @felixfaisal could you please take a look at this PR? Thanks!

bitacross-worker/enclave-runtime/src/lib.rs
```diff
@@ -486,203 +476,6 @@ pub unsafe extern "C" fn migrate_shard(
...
-	let _ = va.execute_mut_on_validator(|v| v.set_ignore_validation_until(*until));
-
-	sgx_status_t::SGX_SUCCESS
-}
-
-/// Dispatch the parentchain blocks for import.
-/// Depending on the worker mode, a different dispatcher is used:
-///
-/// * An immediate dispatcher will immediately import any parentchain blocks and execute
-///   the corresponding extrinsics (offchain-worker executor).
```
kziemianek: The doc mentions `offchain-worker` - this mode is currently used by bitacross worker. Are you sure it's safe to delete this ?

Traf333: it seems I've deleted some useful functionality. I was pretty sure that `sidechain.rs` file is referencing to only code which is irrelevant for bc-worker, but seems it just changes the context. Rolling it back.

what do you think, would it make sense to rename file to `offchain.rs`?

kziemianek: I prefer to keep the original file name - less problems while applying upstream changes. Please add comment why file is kept.

Traf333:
returned back `sidechain.rs` and related code + extended comment on why do we use it. 

@kziemianek , @Kailai-Wang any tips what else might be addressed within this task?

felixfaisal (commented):
Looks good, I ran it locally, able to sync, I'm not very familiar with what all is absolutely  required for sidechain functionality and what is not. So I might be missing or overlooking something.

bitacross-worker/service/src/prometheus_metrics.rs
```diff
@@ -191,12 +173,6 @@ pub struct EnclaveMetricsReceiver;
...
 	fn receive_enclave_metric(&self, metric: EnclaveMetric) -> ServiceResult<()> {
 		match metric {
-			EnclaveMetric::SetSidechainBlockHeight(h) => {
-				ENCLAVE_SIDECHAIN_BLOCK_HEIGHT.set(h as i64);
-			},
-			EnclaveMetric::TopPoolSizeSet(pool_size) => {
-				ENCLAVE_SIDECHAIN_TOP_POOL_SIZE.set(pool_size as i64);
-			},
 			EnclaveMetric::TopPoolSizeIncrement => {
 				ENCLAVE_SIDECHAIN_TOP_POOL_SIZE.inc();
```
kziemianek: Can this metric be removed together with related code ?

Traf333: it used at two functions `process_top` and `remove_top`, so it seems they used at code we utilise.

```rs
// Update metric
  if let Err(e) = self.ocall_api.update_metric(EnclaveMetric::TopPoolSizeIncrement) {
	warn!("Failed to update metric for top pool size: {:?}", e);
  }
```

do you think this metrics irrelevant anymore?

kziemianek: top pool is a sidechain functionality. If we aim for getting rid of sidechain related code from bc-worker it should be removed too.

Traf333: great then I'll review author functionality and will try to remove as much as possible. Thanks

Kailai-Wang:
Want to get this PR done @Traf333 ? :)

Traf333:
> Want to get this PR done @Traf333 ? :)

I have to remove a trusted operations functionality. Should be pretty quick, though I was focused on higher priority tasks

bitacross-worker/service/src/prometheus_metrics.rs
Traf333: Hey @kziemianek, I've managed to remove some functionality with one of [latest commits](https://github.com/litentry/litentry-parachain/pull/2660/commits/fa90de77162c185077079b283417350d5fb7aca6) but it seems whole Author's logic used across the bitacross project and not sure is it relevant or not. :thinking: 

Probably I'd finish with only this chunk of work and merge it. What do you think?

Traf333:
the changes are not relevant anymore

Mr Snail Meme

PR #2661, opened 2024-04-12
Fix bc-worker build

Traf333:
### Context
this PR fixes missing usage of type

```rs
   --> src/rpc/worker_api_direct.rs:346:46
    |
346 | ...                   return match GLOBAL_SCHEDULED_ENCLAVE.update(bn, enclave_to_set) {
    |                                                             ^^^^^^ method not found in `GLOBAL_SCHEDULED_ENCLAVE`
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
18  | use lc_scheduled_enclave::ScheduledEnclaveUpdater;
 ```

was broken by [this PR](https://github.com/litentry/litentry-parachain/pull/2620/files#diff-be10719c49f6f4fc36bc4b54ffda6f5a023254abf927033cd2b93396b8a027dcR52)

PR #2662, opened 2024-04-12
check `development` feature for bitacross in ci

kziemianek:
Adds check for `development` feature to bitacross GH jobs.


kziemianek:
This will fail untill #2661 is merged.

Traf333 (approved):
thanks for addressing this one ✔️

PR #2663, opened 2024-04-15
update dockerfile

m1iktea:
### Context

Modify the Dockerfile to adapt for deploying tee-worker and parachain using Docker.

### Labels

- `C1-noteworthy`

### How (Optional)

- The Docker build for tee-worker is based on the worker's [base image](https://github.com/litentry/devops/blob/main/worker_deploy/tee-worker-base-docker/Dockerfile). For a new build, only the binary needs to be updated.

- The base image for parachain has been changed to ubuntu:22.04.

### Testing Evidences

Now it's running on tee-staging.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,7 +21,7 @@ RUN cargo build --locked --profile $PROFILE $BUILD_ARGS
 # ==========================
 # stage 2: packaging
 # ==========================
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 LABEL maintainer="Trust Computing GmbH <info@litentry.com>"
 
 ARG PROFILE
--- a/tee-worker/build.Dockerfile
+++ b/tee-worker/build.Dockerfile
@@ -102,28 +102,12 @@ ENTRYPOINT ["/usr/local/bin/litentry-cli"]
 
 ### Deployed worker service
 ##################################################
-FROM runner AS deployed-worker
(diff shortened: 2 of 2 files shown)


tee-worker/build.Dockerfile
Kailai-Wang, line 1: Hmm this will remove the possibilty of caching in CI I asusme? (As the updated content in local-builder won't be synced into `litentry/litentry-worker:base`

m1iktea: In my understanding, litentry/litentry-worker:base only provides the runtime environment for the worker and adapts the entrypoint script. When building a new image, it's only necessary to update the binary files within it.

Kailai-Wang: It's more than that => for CI builds (dev-build) it's responsible to update the cache to accelerate the build too.

In production (release) build I agree we should do a clean build without caching

docker/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN cargo build --locked --profile $PROFILE $BUILD_ARGS
 # ==========================
 # stage 2: packaging
 # ==========================
-FROM ubuntu:20.04
+FROM ubuntu:22.04
```
Kailai-Wang, line 24: We can update this only when our production parachain nodes are all deployed with docker.

Kailai-Wang:
I'll close it for now - it's superseded by https://github.com/litentry/litentry-parachain/pull/2733

PR #2664, opened 2024-04-15
Integration test: Set scheduled enclave

Traf333:
### Context

this PR implements integration test for II setting scheduled enclave


silva-fj (approved):
LGTM!

PR #2665, opened 2024-04-15
feat: add unit tests for vc-task

felixfaisal:
So this PR aims to mock some parts of the `vc-task` thread and test. I've created `mock.rs` to create all the mock instances: ;
A few problems I want to address in different issues: 
- Currently, I'm able to run these tests properly only under `--test-threads 1`, if not it fails with `SendError`. I want to debug this in another issue, for now I'm running these tests on a single thread. Not related to `tee-dev` 
- There is no way to test more than one assertion `A1` as all others require network calls, So I'm only testing the `A1` assertion, and secondly I'm not asserting if the credential str produced is correct or not.


tee-worker/build.Dockerfile
```diff
@@ -68,6 +68,7 @@ RUN \
   fi
 
 RUN cargo test --release --features development
+RUN cargo test -p lc-vc-task-receiver -- --ignored --test-threads 1
```
kziemianek, line 71: If there is no rush i suggest debugging `--test-threads` before this PR gets merged.

tee-worker/core-primitives/top-pool-author/src/mocks.rs
```diff
@@ -264,7 +266,12 @@ where
 
 	fn update_connection_state(&self, _updates: Vec<(H256, (Vec<u8>, bool))>) {}
 
-	fn send_rpc_response(&self, _hash: H256, _encoded_value: Vec<u8>, _do_watch: bool) {}
+	fn send_rpc_response(&self, _hash: H256, encoded_value: Vec<u8>, _do_watch: bool) {
+		let sender_guard = GLOBAL_MOCK_RPC_API.lock().unwrap();
+		let sender = &*sender_guard;
+		sender.as_ref().expect("Not yet initialized").send(encoded_value).unwrap();
+		// Box::pin(ready(Ok([0u8; 32].into())))
```
kziemianek, line 273: Please remove the comment.

tee-worker/litentry/core/vc-task/receiver/src/mock.rs
```diff
@@ -0,0 +1,229 @@
...
+
+pub fn execute_with_vc_task<F>(process_fn: F)
+where
+	F: FnOnce(&Receiver<Vec<u8>>, &Receiver<Vec<u8>>),
+{
+	init_vc_task();
+	let top_calls_receiver = init_global_mock_author_api().unwrap();
+	let rpc_calls_receiver = init_global_rpc_api().unwrap();
+
+	thread::sleep(Duration::from_secs(2));
```
kziemianek, line 94: What's the purpose of this `sleep` ?

tee-worker/litentry/core/vc-task/receiver/src/mock.rs
```diff
@@ -0,0 +1,229 @@
...
+use sp_core::{blake2_256, sr25519, Pair};
+use std::{sync::mpsc::Receiver, vec::Vec};
+
+pub const COMMON_SEED: &[u8] =
+	b"crouch whisper apple ladder skull blouse ridge oven despair cloth pony";
+lazy_static! {
+	pub static ref SHIELDING_KEY: ShieldingCryptoMock = ShieldingCryptoMock::default();
+}
+
+pub fn init_global_mock_author_api() -> Result<Receiver<Vec<u8>>> {
```
kziemianek, line 33: What do you think about keeping only mock definitions here and removing test util functions and functions that initializes tests context (move everything to `test.rs` and remove this file) ?

tee-worker/litentry/core/vc-task/receiver/src/test.rs
```diff
@@ -0,0 +1,105 @@
+use core::time::Duration;
+
+use lc_vc_task_sender::{VCRequest, VcRequestSender};
+use litentry_primitives::AesRequest;
+
+use super::*;
+use crate::mock::*;
+
+#[test]
+#[ignore]
```
kziemianek, line 10: Why is this test case ignored ?

kziemianek, line 10: Why is this test case ignored ?

tee-worker/litentry/core/vc-task/receiver/src/mock.rs
felixfaisal: Yeah I can do that, I followed the pattern of unit tests in pallets.

tee-worker/litentry/core/vc-task/receiver/src/test.rs
felixfaisal: Yeah this is so that it is not run with all the tests, and only run with the `test-threads 1`

felixfaisal: Same reason as above

tee-worker/litentry/core/vc-task/receiver/src/mock.rs
felixfaisal: This is a sanity check to make sure things are initialised.

tee-worker/build.Dockerfile
felixfaisal: The reason why I'm choosing to debug this in another issue, is because the most likely reason why it fails is because of the two globals we are using i.e. `GLOBAL_MOCK_RPC_API` and `GLOBAL_MOCK_AUTHOR_API`, it is something being used for only the unit testing, So I prefer doing it another PR. I would have to use a completely different mechanism to capture the values, and I think I should take a dig at it in a different/follow-up PR.

tee-worker/litentry/core/vc-task/receiver/src/test.rs
kziemianek: Don't know why this duplicated comment was added :)

kziemianek: This means if we add more `ignored` tests to this package they will be run alongside  ones added in this PR, is this fine ?

felixfaisal: Yeah that should be fine imo, they will be run with `cargo test -p lc-vc-task-receiver -- --ignored --test-threads 1 `

felixfaisal:
I'll re-open this soon, it has a few conflicts, and I will add it with Dynamic Assertion.

PR #2671, opened 2024-04-17
Batch vc send error test

BillyWooo:
try to catch the panic/error from batch vc task.


tee-worker/enclave-runtime/src/initialization/mod.rs
```diff
@@ -295,7 +295,17 @@ fn run_vc_issuance() -> Result<(), Error> {
 	);
 	let extrinsic_factory = get_extrinsic_factory_from_integritee_solo_or_parachain()?;
 	let node_metadata_repo = get_node_metadata_repository_from_integritee_solo_or_parachain()?;
-	run_vc_handler_runner(Arc::new(stf_task_context), extrinsic_factory, node_metadata_repo);
+
+	let vc_handler = std::thread::spawn(move || {
```
silva-fj, line 299: why do we need to spawn a new thread? `run_vc_issuance` is called within an already spawned thread

felixfaisal (approved):
Thanks @BillyWooo , I think for the temporary need to figure out the `SendError` we need this, However for a long term, it is redundant to have these two threads and we need to do this more ergonomically.

BillyWooo:
> Thanks @BillyWooo , I think for the temporary need to figure out the `SendError` we need this, However for a long term, it is redundant to have these two threads and we need to do this more ergonomically.

Yes. Once we detect/fix it, we need to remove this wrapper layer.

tee-worker/enclave-runtime/src/initialization/mod.rs
BillyWooo: Because it's hard to reproduce. This is only for test purpose. Will remove it later.

PR #2680, opened 2024-04-22
Hot fix send error

BillyWooo:
### Labels

### Testing Evidences


tee-worker/enclave-runtime/src/initialization/mod.rs
```diff
@@ -296,15 +296,7 @@ fn run_vc_issuance() -> Result<(), Error> {
...
-	let vc_handler = std::thread::spawn(move || {
-		error!("*** testing vc task start");
-		run_vc_handler_runner(Arc::new(stf_task_context), extrinsic_factory, node_metadata_repo);
-		error!("*** testing vc task finish");
-	});
-	match vc_handler.join() {
-		Ok(v) => error!("Joined vc_handler OK: {:?}", v),
-		Err(e) => error!("Joined vc_handler Err: {:?}", e),
-	}
+	run_vc_handler_runner(Arc::new(stf_task_context), extrinsic_factory, node_metadata_repo);
```
Kailai-Wang, line 299: You might want to use its return value too - so without `;` or `Ok(())` - similar to `run_stf_task_receiver`

PR #2687, opened 2024-04-24
remove unwrap from `VcRequestSender`

kziemianek:
Prevents from panics if sender is not yet initialized and vc requests are consumed.


BillyWooo:
The fix looks fine to me. But what do you think to revisit the structure/working flow of threads/initialization steps? I think it's better to eliminate it by design.

kziemianek:
1. We must eliminate unwraps - otherwise worker is prone to panics
2. We may optionally improve design

The issue has Urgent priority and S estimation so I believe we aim here only `panic` as a root cause of a worker failure. Same fix was applied to `StfRequestSender` before.

I think we can try to improve the design and avoid using globals - this should be addressed separately.

BillyWooo (approved):

PR #2688, opened 2024-04-24
Fix pnpm not found

outofxxx:
As topic.

### DIFF
--- a/scripts/launch-standalone.sh
+++ b/scripts/launch-standalone.sh
@@ -45,7 +45,7 @@ if [[ -z "${NODE_ENV}" ]]; then
 else
     echo "NODE_ENV=${NODE_ENV}" > .env
 fi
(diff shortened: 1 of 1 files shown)


scripts/launch-standalone.sh
```diff
@@ -45,7 +45,7 @@ if [[ -z "${NODE_ENV}" ]]; then
 else
     echo "NODE_ENV=${NODE_ENV}" > .env
 fi
-corepack pnpm install
+corepack enable pnpm
```
jonalvarezz, line 48: don't we need to run `corepack pnpm install` afterward?

There are more scripts that may need this update, like `scripts/local-binary.sh`. It makes wonder if fixing it per file is the right fix or if we can tackle it more hollistically in the CI environment . I also left a comment in the issue on Linear

PR #2691, opened 2024-04-25
Fixes multi-worker VC test

Kailai-Wang:
### Context

fixes P-588

See the comment in that issue.
The error comes from a race condition between swapping connection and broadcasting the state (with force_wait = true to retain the connection to client):

```
There are basically two ways to fix it:

1. ensure the order of 1 and 2 (by syncing / serializing)
2. delay the removal of connection hash from registry until we are sure that it's done
```

The fix implements the second option.

I ran 9 * 3 = 27 times vc test with the fix and they all went through:

https://github.com/litentry/litentry-parachain/actions/runs/8828860816
https://github.com/litentry/litentry-parachain/actions/runs/8828318163
https://github.com/litentry/litentry-parachain/actions/runs/8824440994

However, I'm still not so happy with the whole flow, we might need bigger refactoring, see the comment in the issue.


tee-worker/core/direct-rpc-server/src/rpc_responder.rs
```diff
@@ -226,9 +239,51 @@ where
...
+						old_hash,
+						connection_token,
+						new_response,
+						force_wait,
+					);
+				}
+			},
+			_ => {
+				self.connection_registry.store(
+					old_hash,
```
silva-fj, line 279: why do we need to put back the `old_hash` for the other cases?

Kailai-Wang: Here other case means status other than `TrustedOperationStatus`, so it's either `Ok` or `Invalid`. In either case we don't want to touch it (at least not within swap_hash)

tee-worker/core-primitives/types/src/lib.rs
```diff
@@ -132,6 +132,9 @@ pub enum TrustedOperationStatus {
 	/// TrustedOperation has been executed.
 	#[codec(index = 11)]
 	TopExecuted(Vec<u8>, bool),
+	/// TrustedOperation whose successor top has been executed, but the top itself not
+	#[codec(index = 12)]
+	SuccessorExecuted,
```
kziemianek, line 137: I was very unconvinced about adding `TopExecuted` status to this enum because it is leaked to the client, this one looks even worse as it exposes unnecessary details but i assume there is no other easy way to fix it.

kziemianek (approved):
In general it looks fine. It looks like a very challenging task to debug and solve due to race conditions. 

In addition to that I think we should try to get rid of those trusted call callback and hash swapping because I think it's the cause of current and potentially future problems. I've put my suggestion in the issue.

tee-worker/core-primitives/types/src/lib.rs
Kailai-Wang: Exactly - because of the fact that 3 related events could show up in arbitrary order(see comment in the issue), so we need a status to mark a connection hash as "can be removed later" in case the swap_hash operation shows up as the first event. 

Later when the desired status is reached the hash will get removed.  Otherwise the connection will be kept forever

felixfaisal (approved):
Thanks @Kailai-Wang for documenting the issue and fixing it with one of the options. So based on my understanding, much of the multi-worker workflow is a happy path and not taking into account network failures or websockets closing abruptly(just one example). I suggest we run the multi-worker setup in one of our 3 server environments. This again depends on when we want to run the multi-worker setup in prod.

Kailai-Wang:
> Thanks @Kailai-Wang for documenting the issue and fixing it with one of the options. So based on my understanding, much of the multi-worker workflow is a happy path and not taking into account network failures or websockets closing abruptly(just one example). I suggest we run the multi-worker setup in one of our 3 server environments. This again depends on when we want to run the multi-worker setup in prod.

Yes they mostly come from the uncertainty of network latency (not failure, just how long it takes to transmit/broadcast the event) , plus we have multiple thread in each node. These combined can create race conditions.

New problems might occur if we have 3 workers located on 3 different instances so that it undergoes non-local network transmission

PR #2695, opened 2024-04-26
Dynamic EVM assertions

kziemianek:
### General

This PR adds foundation for creating user defined runtime assertions called _Dynamic Assertions_

_Dynamic Assertion_ is a assertion logic represented by code/binary that can be added to parachain's storage by any account and executed by dynamic assertion runtime inside sidechain's sgx enclave. 

The main purpose of dynamic assertions is to provide the infrastructure for the community to freely define and create assertions. Secondary goal is to remove assertions written in Rust that are baked into identity-worker's code.

Dynamic assertion has an access to:
1. user id graph
2. assertion secrets 

Assertion secret is a string encrypted with worker's shielding key that is added during dynamic assertion creation and is passed to assertion execution. It can be an API-KEY used for authorizing request sent during execution.

Included assertions:

- A1 - uses only sidechain's state
- A20 - reaches out to HTTP server without authentication
- A6 - reaches out to X api with authentication

This evm assertion runtime implementation should allow for fulfilling any of the scenarios above. It might need further extension in order to support other cases.

Assertions are propagated to sidechain through parachain's event sync mechanism and are kept in memory + sealed to file. 
Storage implementation is far from ideal and we should aim in the future for it's improvement.

###  Writing evm dynamic assertions
It should be possible to use any language that compiles to valid evm byte code (I tested only `Solidity`). Some of the features are still available but it doesn't make sense to use them in assertions because each assertion is executed inside clean evm runtime, for example writing to evm storage doesn't make sense because it's purged after the execution. 

#### Prerequisites for evm dynamic assertions:
* smart contract must have function with signature `function execute(Identity[] memory identities, string[] memory secrets) public`
* should be compiled with `solc 0.8.11` (the most recent version tested)

Abstract `DynamicAssertion` contract can be used for easier development. It contains `virtual` `execute` function matching signature above. It has also various util functions to work with precompiles and `IDGraph` data structures available during the execution.

#### Reaching out to HTTP servers from assertion
EVM assertion runtime contains various precompiles providing ability for making HTTP calls and extracting data using [JSONPointer](https://datatracker.ietf.org/doc/html/rfc6901) .

Example of getting followers count from X api:

```
string memory url = concatenateStrings("http://https://api.twitter.com/2/users/by/username/", string(identities[i].value));
string memory full_url = concatenateStrings(url, "?user.fields=public_metrics");
HttpHeader[] memory headers = new HttpHeader[](1);
// we expect first secret to be twitter api key
headers[0] = HttpHeader("authorization", secrets[0]);
int64 followers_count = GetI64(full_url, "/data/public_metrics/followers_count", headers);
```

List of precompiles:

- `0x03E8` - `GetI64` - for getting I64 value from HTTP server
- `0x03E9` - `GetBool` - for getting bool value from HTTP server
- `0x03EA` - `GetString`  - for getting string value from HTTP server
- `0x041B` - `toHex` - for turning bytes into hex string

Assertion's Solidity source code can be found under `litentry-parachain/tee-worker/litentry/core/assertion-build/src/dynamic/contracts`.


tee-worker/litentry/core/evm-dynamic-assertions/src/repository.rs
```diff
@@ -0,0 +1,129 @@
...
+
+		Ok(EvmAssertionRepository { state: state.into(), path: path.to_string() })
+	}
+}
+
+impl AssertionLogicRepository for EvmAssertionRepository {
+	type Id = AssertionId;
+	type Item = (SmartContractByteCode, Vec<String>);
+
+	fn get(&self, id: &Self::Id) -> Result<Option<Self::Item>, String> {
```
silva-fj, line 66: How would this work in multi worker setup? I think the sealed files will get synced during initialization only (please correct me if I'm wrong)

tee-worker/litentry/core/evm-dynamic-assertions/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+			caller,
+			U256::zero(),
+			smart_contract_byte_code,
+			u64::MAX,
+			Vec::new(),
+		);
+
+		// call assertion smart contract
+		let call_result =
+			executor.transact_call(caller, address, U256::zero(), input, u64::MAX, Vec::new());
```
silva-fj, line 116: I'm not sure if this is a real concern, but will the `gas limit` be enough to prevent the smart contracts (untrusted code) from doing something malicious?

silva-fj (commented):
It looks pretty good overall, the idea of dynamic assertions is quite nice.

I have a few observations. Since we are storing the assertions/smart contracts in the parachain, why don't we use that state instead of storing them also in a sealed file? or (if my assumption that there could by a sync issue in a multi worker setup is correct), how about storing this data on the sidechain storage?

tee-worker/litentry/core/evm-dynamic-assertions/src/repository.rs
kziemianek: This is the initial sync, after that assertions are synced with parachain state in `event_handler.rs`

tee-worker/litentry/core/evm-dynamic-assertions/src/lib.rs
kziemianek: This only prevents from infinite execution.

kziemianek:
> It looks pretty good overall, the idea of dynamic assertions is quite nice.If i'
> 
> I have a few observations. Since we are storing the assertions/smart contracts in the parachain, why don't we use that state instead of storing them also in a sealed file? or (if my assumption that there could by a sync issue in a multi worker setup is correct), how about storing this data on the sidechain storage?

Thanks @silva-fj for review.

Getting state directly from parachain's store was my first choice but it requires parachain block header in order to verify date see `get_storage_verified`. That would lead us to stoing parachain's block headers by workers. 

There should be no sync issue because after initial sync workers will continously process parachain's events and apply changes to local state. Additionaly for data that is not used/produced by STF I see no point for using sidechain's storage.

tee-worker/litentry/core/evm-dynamic-assertions/src/repository.rs
silva-fj: True, I forgot about that 👍🏼

silva-fj (approved):

pallets/evm-assertions/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+			secrets: Vec<Vec<u8>>,
+		) -> DispatchResultWithPostInfo {
+			ensure_signed(origin)?;
+			ensure!(!Assertions::<T>::contains_key(id), Error::<T>::AssertionExists);
+			Assertions::<T>::insert(
+				id,
+				Assertion { byte_code: byte_code.clone(), secrets: secrets.clone() },
+			);
+			Self::deposit_event(Event::AssertionCreated { id, byte_code, secrets });
+			Ok(Pays::No.into())
```
Kailai-Wang, line 96: We'll need to charge it for sure - OK to keep it for now

tee-worker/app-libs/parentchain-interface/src/integritee/event_filter.rs
```diff
@@ -34,6 +36,23 @@ use std::vec::Vec;
 #[derive(Clone)]
 pub struct FilterableEvents(pub Events<H256>);
 
+impl FilterableEvents {
+	fn filter<T: StaticEvent, E>(&self) -> Result<Vec<T>, E> {
```
Kailai-Wang, line 40: Thank you for the refactoring

tee-worker/litentry/core/evm-dynamic-assertions/src/precompiles/to_hex.rs
```diff
@@ -0,0 +1,60 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::precompiles::{macros::prepare_custom_failure, PrecompileResult};
+use std::{format, vec::Vec};
+
+#[cfg(all(not(feature = "std"), feature = "sgx"))]
+extern crate hex_sgx as hex;
```
Kailai-Wang: We should be able to use no_std hex

primitives/core/src/assertion/mod.rs
```diff
@@ -135,6 +136,9 @@ pub enum Assertion {
 
 	#[codec(index = 26)]
 	NftHolder(Web3NftType),
+
+	#[codec(index = 27)]
+	Dynamic(H160) // smart contract code identifier
```
Kailai-Wang, line 141: Shall we have a list of pre-deployed contracts somewhere as comment?

pallets/evm-assertions/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+		AssertionExists,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight((T::DbWeight::get().write, DispatchClass::Normal, Pays::No))]
+		pub fn create_assertion(
+			origin: OriginFor<T>,
+			id: T::AssertionId,
```
Kailai-Wang, line 85: Hmm I suggest we allocate AssertionId internally, and reserve some sections for Litentry precompiles

Kailai-Wang (approved):
Not much I can complain, good job 👍 

We might need a mechanism to "forcibly" sync the parachain assertion registry at once in case we lose the track of the events, but that's a general question between parachain <> worker sync

primitives/core/src/assertion/mod.rs
kziemianek: There are no pre-deployed contracts. 
We should decouple it from parachain and let product apps prepare envs for use (deploy assertions if needed)

pallets/evm-assertions/src/lib.rs
kziemianek: What do you mean by Litentry precompiles ?
`AssertionId` is not a smart contract address, it's just (bytecode,secrets) identifier.

pallets/evm-assertions/src/lib.rs
kziemianek: P-722

tee-worker/litentry/core/evm-dynamic-assertions/src/precompiles/to_hex.rs
kziemianek: applied

pallets/evm-assertions/src/lib.rs
Kailai-Wang: But it's using H160 right - that could be misleading and might cause (semantic) abuse 😂 

> Litentry precompiles

I mean those contracts wirtten by Litentry ourselves (e.g. A1 - A20)

kziemianek: From the client perspective you dont need to be aware about implementation details. All you need to do is to generate unique `H160` identifier and use it during `create` and `request_vc`. I think it's natural to use `H160` because that's how smart contracts are identified - but we can change to another type depends if it's going to be generated internally or by client. 

I suggest to treat all assertions equally so there is no need for identifier reservation for assertions currently used by IdHub.

Given that, do you still think it's worth it ?

Kailai-Wang: Technically there's so such requirement.

I'm thinking of how it looks like in the future - like when we introduce "reputation system" to find and slash bad contracts, it's easier to identify internal and external contracts just by looking at their identifiers. It's less error-prone (for clients to avoid scams) and easier to document.

Internal contracts don't carry along any particular feature, but they are marked as "written by litentry" so people know the trustness of the logic should be at least as high as Litentry the project team.

I see similar implementation to multi asset management, where most projects do draw a line between "internally defined assets" (assets that are registered by the chain team itself) and "externally defined assets" that can be registered arbitrarily by anyone.

So I'm ok to have H160 as identifier but I think it makes sense (at least I don't see any harm) to have a reserved section for Litentry contracts (both primitive precompiles and concrete logic contracts) - like 65536 should be enough? Wdyt?

kziemianek: Having a reputation system for assertions is a nice idea. I would like to draw another way of meeting these requirements.

We could extend parachain's storage with `bool` value, let's name it `verified`. Ideally, assertions could be marked as `verified` upon collective decision. The `verified` mark could be also removed in the same manner, for example after detecting that assertion generates false-positive vcs.

All assertions written by the Litentry team could get `verified` mark from the start. It would also open the possibility to mark other assertions as `verified` after auditing their code, for example by the Litentry team.

We can also extend storage with `owner` address. That would allow for detecting which assertions are added by the Litentry team.

PR #2696, opened 2024-04-27
BitAcross musig2 ceremony

kziemianek:
Adds a logic for running musig2 ceremony between `bc-workers`. 

Authorized relayers can start a musig2 ceremony. The ceremony must be started on all registered enclaves in order to finish. Once started it will be driven by the `runner` to the end. Communication between the workers takes place through trusted rpc endpoints. Once the ceremony is finished, relayer receives response containing schnorr signature.

This is still a PoC phase. All additional information regarding next steps is wrote down in the issue.


Kailai-Wang (approved):
I just roughly browsed this grand feature implementation - if bc colleagues are happy with it, I'm happy to merge it

PR #2697, opened 2024-04-29
hot fix unwrap withink cli id-graph-hash

BillyWooo:
as title. a small fix.

### DIFF
--- a/tee-worker/cli/src/base_cli/commands/litentry/id_graph_hash.rs
+++ b/tee-worker/cli/src/base_cli/commands/litentry/id_graph_hash.rs
@@ -36,8 +36,14 @@ impl IDGraphHashCommand {
 		let mrenclave = direct_api.get_state_mrenclave().unwrap();
 		let shard = ShardIdentifier::decode(&mut &mrenclave[..]).unwrap();
 		let identity = Identity::from_did(self.did.as_str()).unwrap();
-		let id_graph_hash = direct_api.get_id_graph_hash(&shard, &identity).unwrap();
-		println!("{:?}", id_graph_hash);
+		match direct_api.get_id_graph_hash(&shard, &identity) {
+			Ok(id_graph_hash) => {
+				println!("{:?}", id_graph_hash);
+			},
+			Err(e) => {
+				println!("{:?}", e);
+			},
+		}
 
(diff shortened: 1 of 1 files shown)


Kailai-Wang:
I'm fine with it but if we want to do this, we probably need to do it for all CLI commands.
Normally CLI isn't that critical as it's not part of runtime - so unwrap = CLI program exit (which is fine)

tee-worker/cli/src/base_cli/commands/litentry/id_graph_hash.rs
```diff
@@ -36,8 +36,14 @@ impl IDGraphHashCommand {
 		let mrenclave = direct_api.get_state_mrenclave().unwrap();
 		let shard = ShardIdentifier::decode(&mut &mrenclave[..]).unwrap();
 		let identity = Identity::from_did(self.did.as_str()).unwrap();
-		let id_graph_hash = direct_api.get_id_graph_hash(&shard, &identity).unwrap();
-		println!("{:?}", id_graph_hash);
+		match direct_api.get_id_graph_hash(&shard, &identity) {
```
kziemianek, line 39: Why not `expect` ?
It can panic with more friendly message, additionaly `Ok` is not returned (there was a failure).

BillyWooo: What `Ok` is not returned?

BillyWooo:
The reason I want to add it is: I need to do a group of `id-graph-hash` ops. If it simply crashes, it's hard to be managed within scripts. 
I will put this on hold and discuss later whether to check all cli commands.

tee-worker/cli/src/base_cli/commands/litentry/id_graph_hash.rs
kziemianek: `unwrap` previously prevented from returning `Ok` from function.
After you fix, even if there is an error the function will return `Ok`

BillyWooo:
close for the moment. 
[P-755 : take care of crash/panic from cli tools/commands](https://linear.app/litentry/issue/P-755/take-care-of-crashpanic-from-cli-toolscommands)

PR #2700, opened 2024-05-03
Fix multi worker scheduled enclave test 

Traf333:
### Context

test failed because workers have to sync state and it might take a bit of time, where test was designed to check immediately a the block for scheduled enclave. This PR fixes the the test by increasing awaiting time

### Testing Evidences

Before https://github.com/litentry/litentry-parachain/actions/runs/8936500614
After https://github.com/litentry/litentry-parachain/actions/runs/8936891499

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -800,37 +800,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


BillyWooo (approved):
LGTM

PR #2701, opened 2024-05-03
Fix crate unit test

Kailai-Wang:
### Context

This PR adds the crate unit tests under `tee-worker/litentry/` into CI and fixes them:

```
        working-directory: ./tee-worker
        run: |
          cargo test --release -p pallet-* --lib
          cargo test --release -p lc-* --lib
          cargo test --release -p litentry-primitives --lib
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,6 +260,8 @@ jobs:
         working-directory: ./tee-worker
         run: |
           cargo test --release -p pallet-* --lib
+          cargo test --release -p lc-* --lib
+          cargo test --release -p litentry-primitives --lib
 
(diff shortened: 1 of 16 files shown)


kziemianek:
I think we got more cases like this, another one
`cargo test --release -p lc-stf-task-receiver --lib`

Should we check it in CI ?

Kailai-Wang:
> cargo test --release -p lc-stf-task-receiver --lib

Thanks 

Yes actually - I prefer to have it, at least for our own crates (lc-*) - let me try to add and fix them

Basically the crate unit tests should pass under both `std` and `sgx`

kziemianek (approved):

PR #2707, opened 2024-05-07
feat: P-298 add abort strategy for rest VC api call, add vc ts tests

wli-pro:
### Context

This PR added abort strategy to rest of VCs as well as the ts tests.

### Labels

### Testing Evidences


tee-worker/litentry/core/assertion-build/src/a14.rs
```diff
@@ -115,24 +116,30 @@ pub fn build(
...
+				params: A14DataParams { address: address.clone() },
+				include_metadata: false,
+				include_widgets: false,
+			};
+			let response = client.send_request(&data)?;
+
+			let result = response
+				.data
+				.get("result")
+				.and_then(|r| r.as_bool())
```
kziemianek: Do I understand correctly that:

- old version stops if `RequestVCFailed` occours
- new version doesn't stop if `RequestVCFailed` occours 

I checked `test_continue_until_end_success` and that's my understanding. Am I right ?

wli-pro: Yes, well spot.
It's changed or fixed because the VC result should be true if any address has result true return from api, and it shouldn't fail if only one of them failed and it should continue until get one true or exhaust all address.

kziemianek: So if I have 10 addresses and for one of them the result would be true, but due to the network problems request failed I will get VC with `value: [false]`. 

Maybe there should be an additional information saying that not all addresses were checked ? But such information would reveal idgraph details (there is more than one identity linked related to this VC).

Honestly, I'm not sure which version is better (previous or current) but it seems like the previous was more accurate.

tee-worker/litentry/core/assertion-build/src/a14.rs
```diff
@@ -113,26 +50,28 @@ pub fn build(
...
+				.map_err(|e| Error::RequestVCFailed(Assertion::A14, e.into_error_detail()))?;
+			if result {
+				value = result;
+				Ok(LoopControls::Break)
+			} else {
+				Ok(LoopControls::Continue)
+			}
+		},
+		AbortStrategy::ContinueUntilEnd::<fn(&_) -> bool>,
+	)?;
```
wli-pro: Hey @kziemianek see below hypothetic scenario here:

(1) response 1 gets a network error
code goes to L56 in below code, and because we set AbortStrategy::ContinueUntilEnd at L73 above, i.e. if not the last element then it won't break the loop, so loop continue:
Screenshot 2024-05-10 at 7 47 01 pm

(2) response 2 gets a false value
code goes to else block (L69 above), i.e. continue

(3) response 3 gets a true value
code goes to if branch (L66 above), i.e. break

So any network error won't break the loop (same as false response either). Does this make more sense?

higherordertech: Also
Screenshot 2024-05-10 at 7 47 01 pm

Once AbortStrategy::ContinueUntilEnd is used in above L73, if not the last element then it won't break the loop.

kziemianek: It makes same sense as before 🙃. Honestly, I'm not sure which version is better (previous or current) but it seems like the previous was more accurate (assuming previously request vc failed).


(1) network error (potentailly true)
(2) false
(3) false

result: false

tee-worker/litentry/core/assertion-build/src/a14.rs
kziemianek: Just wanted to confirm that the behaviour has slightly changed.

BillyWooo (approved):
LGTM.

tee-worker/litentry/core/assertion-build/src/a14.rs
```diff
@@ -114,26 +51,28 @@ pub fn build(
...
+				.query_label("a719e99c-1f9b-432e-8f1d-cb3de0f14dde", &data)
+				.map_err(|e| Error::RequestVCFailed(Assertion::A14, e.into_error_detail()))?;
+			if result {
+				value = result;
+				Ok(LoopControls::Break)
+			} else {
+				Ok(LoopControls::Continue)
+			}
+		},
+		AbortStrategy::ContinueUntilEnd::<fn(&_) -> bool>,
```
felixfaisal, line 72: So i tracked the logic of this, If we are continuing until end, we are only propagating the last error we received in the loop, Does it make sense to accumulate all the errors and propagate that?  Or maybe if not for the client, but for some kind of observability?

tee-worker/litentry/core/assertion-build/src/a14.rs
```diff
@@ -114,26 +51,28 @@ pub fn build(
...
+				params: LabelQueryReqParams { address: address.clone() },
+				include_metadata: false,
+				include_widgets: false,
+			};
+			let result = client
+				.query_label("a719e99c-1f9b-432e-8f1d-cb3de0f14dde", &data)
+				.map_err(|e| Error::RequestVCFailed(Assertion::A14, e.into_error_detail()))?;
+			if result {
+				value = result;
+				Ok(LoopControls::Break)
```
felixfaisal, line 67: Seems like `LoopControls` will only ever have two variants, In that case can we use `Option<>` ? I think that's more idiomatic, wdyt?

wli-pro: IMO LoopControls is better than using Option<>
1 LoopControls is more readable
2 LoopControls is more extensible, e.g. we can add a new value: LoopControls.RETURN

tee-worker/litentry/core/assertion-build/src/a14.rs
wli-pro: Hey @felixfaisal all errors are now being accumulated now in this: https://github.com/litentry/litentry-parachain/pull/2707/commits/bf386dcec959135abb60ac8f0ce0d8217426bdcb

felixfaisal (approved):
Thanks @higherordertech . LGTM from my side. If nothing breaking for the client side then we should go ahead and merge imo.

PR #2709, opened 2024-05-07
P-145 Add {para,side}chain runtime versions to VC

arieljbirnbaum:
### Context

Implementing P-145
Following VC schema changes added in [vc-jsonschema#24](https://github.com/litentry/vc-jsonschema/pull/24)

### How (Optional)

- Sidechain version is taken from the build time constant.
- Parachain version is taken from the `node_metadata_repo` argument at VC construction time.


Traf333 (commented):
@grumpygreenguy should we exclude bitacross-worker since there are no sidechain and VC?

Kailai-Wang:
> @grumpygreenguy should we exclude bitacross-worker since there are no sidechain and VC?

IMO yes - there should be no stf/sidechain for bc-worker (there's a PR to remove it)

tee-worker/core-primitives/stf-interface/src/lib.rs
```diff
@@ -20,6 +20,12 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+#[cfg(all(feature = "std", feature = "sgx"))]
+compile_error!("feature \"std\" and feature \"sgx\" cannot be enabled at the same time");
+
+#[cfg(all(not(feature = "std"), feature = "sgx"))]
+extern crate sgx_tstd as std;
```
silva-fj: Is this really needed?

tee-worker/core-primitives/node-api/metadata/src/lib.rs
```diff
@@ -56,20 +56,21 @@ pub trait NodeMetadataTrait:
...
-			+ VCMPCallIndexes
-			+ SystemSs58Prefix
-			+ UtilityCallIndexes
-			+ ProxyCallIndexes
-			+ BalancesCallIndexes
-			+ TimestampCallIndexes,
-	> NodeMetadataTrait for T
-{
-}
+// impl<
```
silva-fj: Can this be removed?

silva-fj (commented):
it looks like you had a lot of fun adding just two values 😆

tee-worker/core-primitives/node-api/metadata/src/lib.rs
arieljbirnbaum: Possibly. @Kailai-Wang suggested having the method in the `SystemSs58Prefix` trait tho (which should then probably be renamed to sth like `SystemPalletTrait`) so I'm having a look at that; if that works then I can keep the old impl

tee-worker/core-primitives/stf-interface/src/lib.rs
arieljbirnbaum: It _was_ when I was passing the version numbers as `String` (as opposed to `u32`) :D
Will try yeeting it; thanks for the catch :raised_hands:

tee-worker/core-primitives/node-api/metadata/src/lib.rs
arieljbirnbaum: Fixed :D

tee-worker/core-primitives/stf-interface/src/lib.rs
arieljbirnbaum: Yeeted :D

silva-fj (approved):
LGTM!

PR #2710, opened 2024-05-08
Adding support for discord oauth verification

silva-fj:
This PR introduces a new option to verify discord accounts using OAuth2.0

The discord oauth 2.0 mechanism works a little differently than Twitter's, there is no need to generate a code challenge from our side so clients only need one interaction with the worker to link identities.

Just like with twitter, we need to register the redirect uri clients will use in the discord developer account.

As Kai suggested here https://github.com/litentry/litentry-parachain/pull/2641#pullrequestreview-2035558395, in the future, we could register these oauth clients in the parachain.

> [!IMPORTANT]
> - There are 2 new environment variables needed for the Discord API needed for OAuth2.0: `DISCORD_CLIENT_ID` and `DISCORD_CLIENT_SECRET`. 
> - The clients must use the following template link to present to their users `https://discord.com/oauth2/authorize?client_id={DISCORD_CLIENT_ID}&response_type=code&redirect_uri={CLIENTS-REDIRECT-URI}&scope=identify`.


tee-worker/client-api/parachain-api/prepare-build/interfaces/identity/definitions.ts
```diff
@@ -70,22 +70,32 @@ export default {
...
             state: "Vec<u8>",
             redirect_uri: "Vec<u8>",
         },
         DiscordValidationData: {
+            _enum: {
+                PublicMessage: "PublicMessage",
+                OAuth2: "DiscordOAuth2",
+            },
+        },
+        PublicMessage: {
```
kziemianek, line 90: `DiscordPublicMessage` or `DiscordChannelMessage` (I find latter more accurate) ? Just to avoid future rename like it was with `OAuth2` -> `TwitterOAuth2`

tee-worker/ts-tests/integration-tests/common/utils/identity-helper.ts
```diff
@@ -49,15 +49,33 @@ export function parseIdGraph(
...
 }
 
-type TwitterValidationConfig =
+type Web2ValidationConfig =
     | {
+          identityType: 'Discord';
+          context: IntegrationTestContext;
+          signerIdentitity: CorePrimitivesIdentity;
+          linkIdentity: CorePrimitivesIdentity;
+          verificationType: 'PublicMessage';
```
kziemianek: Would it work `verificationType: 'PublicMessage' | 'OAuth2';` ?

kziemianek (approved):
I left small comments but overall it looks great 👍🏼

BillyWooo (approved):
LGTM. 
Is there anything setup from our Discord server?

tee-worker/ts-tests/integration-tests/common/utils/identity-helper.ts
silva-fj: yep 👍

silva-fj:
> LGTM. Is there anything setup from our Discord server?

We need to setup OAuth2 on the app we use or create a new app in the account and set it up to get the client id and client secret.

tee-worker/litentry/primitives/src/validation_data.rs
```diff
@@ -34,10 +34,16 @@ pub enum TwitterValidationData {
 
 #[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct DiscordValidationData {
-	pub channel_id: ValidationString,
-	pub message_id: ValidationString,
-	pub guild_id: ValidationString,
+pub enum DiscordValidationData {
+	DiscordPublicMessage {
```
Kailai-Wang: Shall we have a consitent naming with the Twitter?
I mean `DiscordPublicMessage` vs `PublicTweet`

tee-worker/litentry/core/identity-verification/src/web2/mod.rs
```diff
@@ -20,13 +20,14 @@ extern crate sgx_tstd as std;
 #[cfg(all(feature = "std", feature = "sgx"))]
 compile_error!("feature \"std\" and feature \"sgx\" cannot be enabled at the same time");
 
+mod discord;
```
Kailai-Wang, line 23: hmm discord is private mod while twitter is public mod, why?

Kailai-Wang (commented):
Thanks, only have some small naming questions, otherwise it looks good

tee-worker/litentry/core/identity-verification/src/web2/mod.rs
BillyWooo: @Kailai-Wang  does it make sense to clean up a bit the "pub" key words in our code in general ? I think there are quite some places where ”pub" is not necessary.

silva-fj: > hmm discord is private mod while twitter is public mod, why?

Discord's oauth mechanism is a bit different, there is no need to generate and store a code challenges. This is why there are no exports for discord

tee-worker/litentry/primitives/src/validation_data.rs
silva-fj: on a second thought, I think `Discord*` is redundant as it's already in `DiscordValidationData`. I'll change it back to just `PublicMessage`

tee-worker/litentry/core/identity-verification/src/web2/mod.rs
Kailai-Wang: I see, thanks - maybe we could have `mod twitter;` and only export the symbols that we need, but it's a minor problem (maybe there're too many symbols to export, which can be very verbose)

PR #2730, opened 2024-05-15
Adding prometheus metrics for processed parentchain events

silva-fj:
As part of the preparation for the identity worker monitoring, we are including the parentchain events. All the other metrics mentioned in the issue are already included.


felixfaisal (commented):
Thanks @silva-fj  
I think we shouldn't be monitoring the parentchain events, via the indirect executor and filter for the sake of monitoring. We can use the `indexer` if we want to monitor the events of the parachain. Doesn't the sidechain block height suffice for monitoring? wdyt?

silva-fj:
> Thanks @silva-fj I think we shouldn't be monitoring the parentchain events, via the indirect executor and filter for the sake of monitoring. We can use the `indexer` if we want to monitor the events of the parachain. Doesn't the sidechain block height suffice for monitoring? wdyt?

@felixfaisal Monitoring events from an indexer makes a lot of sense, I didn’t know we have one. However, following the discussion in the issue if we set these metrics from the indexer, the worker operators would need to include the indexer as target (or build their own events exporter) in their prometheus setup (assuming they are interested in these metrics and will use the same dashboard).

According to the issue the idea of including the (teebag) events is to show it's still parachain. Maybe @Kailai-Wang could give us more clarity on the purpose

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -378,10 +380,19 @@ where
...
 
+		if let Ok(events) = events.get_parentchain_block_proccessed_events() {
+			debug!("Handling ParentchainBlockProcessed events");
+			events.iter().for_each(|event| {
+				debug!("found ParentchainBlockProcessed event: {:?}", event);
+				// This is for monitoring purposes
+				handled_events.push(hash_of(ParentchainBlockProcessed::EVENT));
+			});
+		}
+
```
BillyWooo, line 391: Maybe a dumb question: why not keep the same structure as the previous above block of `if let Ok(events)` ?

silva-fj: In this case the function applied to each event is not fallible, that's why we use `for_each` and we don't need ownership of the event as don't use it (apart from the log), so we just use `iter()`. 

Is this what you mean by  "the same structure"?

Kailai-Wang:
> Doesn't the sidechain block height suffice for monitoring? wdyt?

This is not enough, unfortunately. They represent different aspects.
Sidechain height is arguably the most important metric => it indicates if sidechain itself is alive
Teebag event can only come from worker => We use it as an indicator to show worker is still connected to parachain => any better metric for such purpose?

It reminds of another question:
theoretically it's possible that the vc-handler thread is dead while sidechain is still producing blocks (as they are different threads) - do we have a way to detect such? I guess not yet.

Indexer introduces too much external dependency to me:
a) you need to implement it (we have one but it's maintained by IDH - not sure if that will last forever)
b) we rely on the availability of indexer too, which might cause false alarm
c) it complicates the input source setting for grafana and whole system

BillyWooo:
>do we have a way to detect such? I guess not yet.

No, we don't have yet.
Besides, we may also miss the mechanism to detect once worker is disconnected with parachain, while it's still running by itself.

Kailai-Wang:
Could we get this PR over the line? Mainly asking @silva-fj and @felixfaisal 
We are going to set up the monitor in grafana.

felixfaisal (approved):
LGTM

PR #2735, opened 2024-05-16
Bitacross worker build and publish workflow

kziemianek:
Adds github actions workflow for building`bitacross-worker` in `HW` mode and pushing it to dockerhub.


.github/workflows/build_bitacross.yml
```diff
@@ -0,0 +1,52 @@
+name: Build bitacross-worker
+
+on:
+  workflow_dispatch:
```
Kailai-Wang, line 4: Do you want to have a tag/branch/commit hash as input parameter?

.github/workflows/build_bitacross.yml
```diff
@@ -0,0 +1,52 @@
...
+            SGX_MODE=HW
+            ADDITIONAL_FEATURES_ARG=      
+            IMAGE_FOR_RELEASE=true
+
+      - name: Build worker
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: bitacross-worker/build.Dockerfile
+          tags: litentry/bitacross-worker:latest
```
Kailai-Wang: I suggest that the docker tag is parameterized too

Kailai-Wang (approved):
Looks good - the comment is more of a flexibility improvement

.github/workflows/build_bitacross.yml
kziemianek: It'll use `input.ref` for both checking out the code and tagging image

Kailai-Wang: tags should be `${{ github.event.inputs.ref || github.ref }}` instead of latest?

kziemianek: I would like to test this workflow now but seems I cannot start it from GitHub.

.github/workflows/build_bitacross.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Build local builder
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: bitacross-worker/build.Dockerfile
+          tags: local-builder:${{ github.event.inputs.ref || github.ref }}
+          target: builder
+          build-args: |
+            WORKER_MODE_ARG=offchain-worker
+            SGX_MODE=HW
```
Traf333, line 34: does GH actions run on machine with HW support? I thought with GHA we can only use software mode :thinking:

kziemianek:
Closing for now as procedure for building with `SGX_PRODUCTION=1` is not determined yet.

PR #2736, opened 2024-05-16
Web3networks: Remove from IDgraphs and make optional for requesting VCs

Traf333:
### Context

this PR is the first step towards removing `web3networks` from id graphs since we decided to utilize it only on requesting VC. Also if no networks specified for requesting VC, by default we set all relevant

### Labels

### Note

Due to migration issues, we can not drop completely the field from idgraph as well as dropping SubstrateTestnet, since it prevents getting idgraph's data

### TODO

- [ ] Add state migration


tee-worker/litentry/pallets/identity-management/src/migrations.rs
```diff
@@ -61,3 +63,16 @@ pub fn migrate_to_v2<T: crate::Config, P: GetStorageVersion + PalletInfoAccess>(
 	}
 	Weight::zero()
 }
+
+pub fn drop_web3networks_from_id_graph<T: crate::Config>() -> Result<Weight, Error<T>> {
```
kziemianek: Is this function called only in tests? How is migration performed ?

Traf333: yeah, I'm on the way to figuring out where to put it to trigger on runtime upgrade. And according to our guide https://www.notion.so/web3builders/How-to-upgrade-enclave-one-worker-edfaf5871b4441579f9471074032ed1e

the best place is `tee-worker/app-libs/sgx-runtime/src/migration.rs`

local-setup/launch.py
```diff
@@ -196,7 +196,7 @@ def get_flags(index, worker):
 
     return list(filter(None, [
         "--clean-reset",
-        "-T", "wss://localhost",
+        "-T", "ws://localhost",
```
jonalvarezz: could this impact production deployments?

jonalvarezz (approved):
I'm fine overall with the direction. I created a client-sdk issue to follow-up on integrating this changes [P-765](https://linear.app/litentry/issue/P-710/include-optional-network-parameter-to-vc-request)

local-setup/launch.py
Traf333: nope, it related to not more than just local run. This particular change allows to run ` ./litentry-cli print-sgx-metadata-raw`, which we need for running integration tests locally. Otherwise the execution will hang and won't go further

kziemianek: It hangs because it tries to use default value

```
	#[clap(short = 'U', long, default_value_t = String::from("ws://127.0.0.1"))]
	worker_url: String,
```

You could explicitly pass `-U wss://localhost` to CLI but having it working out of the box is more handy.

Traf333: Thanks for the tip, modified package.json instead of python lanucher `./bin/litentry-cli -U wss://localhost print-sgx-metadata-raw`

Kailai-Wang:
Thank you! 

We'll need to exercise this on our staging first to test it thoroughly - also it will change the IDGraph hash, will it have any impact on the IDHub (in case they store and use the hash in some way)?

jonalvarezz:
> also it will change the IDGraph hash, will it have any impact on the IDHub (in case they store and use the hash in some way)?

That's a great point. IDH stores the hash to tell if the local is out of date. I guess all users will get an "id_graph out of sync" notice which is ok-ish, but we may want to check with them.

wrt to timing, do we agree on postponing this change until the next IDH campaign has finished?

Traf333:
> wrt to timing, do we agree on postponing this change until the next IDH campaign has finished?

Definitely, it would make sense to thoroughly test it and merge only after the release. ⏳

tee-worker/client-api/sidechain-api/package.json
```diff
@@ -6,7 +6,7 @@
     "version": "0.0.1-1",
     "scripts": {
         "clean": "rm -rf dist build node_modules",
-        "update-metadata": "../../bin/litentry-cli print-sgx-metadata-raw > prepare-build/litentry-sidechain-metadata.json",
+        "update-metadata": "../../bin/litentry-cli -U wss://localhost print-sgx-metadata-raw > prepare-build/litentry-sidechain-metadata.json",
```
0xverin, line 11: Can you also change the [WORKER_ENDPOINT](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/integration-tests/.env.local.example#L2) in `integration-tests/env.local.example`? I think it needs to be modified to `wss://localhost`. Thanks!

tee-worker/litentry/pallets/identity-management/src/identity_context.rs
```diff
@@ -38,18 +38,14 @@ pub struct IdentityContext<T: Config> {
 	// the sidechain block number at which the identity is linked
 	pub link_block: BlockNumberOf<T>,
 	// a list of web3 networks on which the identity should be used
-	pub web3networks: Vec<Web3Network>,
+	pub web3networks: Option<Vec<Web3Network>>,
```
jonalvarezz: Why not to remove it?

BillyWooo: I think we need to remove it. We don't need to store it anymore. 
Just passed in as an option parameter when requesting VC. That would be enough.

Traf333: From what I understood, we should do the migration first and then remove it. Otherwise the migration will fail here https://github.com/litentry/litentry-parachain/pull/2736/files#diff-c45ca5a4e78440095e592f8784465f19278a7c2c23a0bedb9ed2a82d6817e233R67

Traf333: returning it back, since otherwise it can not read idgraph :(

jonalvarezz:
Hi folks, anything holding us before merging this in?

Kailai-Wang:
> Hi folks, anything holding us before merging this in?

Yes - I prefer that we have the state migration code within this PR too

Traf333:
> Hi folks, anything holding us before merging this in?

It would be nice to have opened, since we still need to remove web3networks from idgraph with proper migration. And for testing proper migration we have to link identity with some network which is unavailable within current changes

felixfaisal:
I'm closing this PR, Due to inactivity, I will not delete the branch if someone wants to take the work further. Feel free to re-open.

PR #2739, opened 2024-05-17
fix: P-441 remove NoEligibleIdentity

wli-pro:
return false value VC if has no eligible identity

### Labels

### Testing Evidences

### DIFF
--- a/tee-worker/app-libs/stf/src/trusted_call_litentry.rs
+++ b/tee-worker/app-libs/stf/src/trusted_call_litentry.rs
@@ -197,7 +197,7 @@ impl TrustedCallSigned {
 		let identities = get_eligible_identities(id_graph.as_ref(), assertion_networks);
 
 		if identities.is_empty() {
-			warn!("[RequestVc] : No eligible identities");
+			debug!("[RequestVc]: No eligible identities");
 		}
 
 		let parachain_block_number = ParentchainLitentry::block_number();
--- a/tee-worker/app-libs/stf/src/trusted_call_litentry.rs
+++ b/tee-worker/app-libs/stf/src/trusted_call_litentry.rs
@@ -196,10 +196,9 @@ impl TrustedCallSigned {
 		let assertion_networks = assertion.get_supported_web3networks();
 		let identities = get_eligible_identities(id_graph.as_ref(), assertion_networks);
 
-		ensure!(
-			!identities.is_empty(),
-			StfError::RequestVCFailed(assertion, ErrorDetail::NoEligibleIdentity)
-		);
+		if identities.is_empty() {
+			warn!("[RequestVc] : No eligible identities");
+		}
 
 		let parachain_block_number = ParentchainLitentry::block_number();
 		let sidechain_block_number = System::block_number();
--- a/tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
+++ b/tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
@@ -86,7 +86,7 @@ export default {
                 UnexpectedMessage: "Null",
                 __Unused_WrongSignatureType: "Null",
                 VerifyWeb3SignatureFailed: "Null",
-                NoEligibleIdentity: "Null",
+                __Unused_NoEligibleIdentity: "Null",
             },
         },
         ShardIdentifier: "H256",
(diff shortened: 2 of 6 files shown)


tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
```diff
@@ -86,7 +86,7 @@ export default {
                 UnexpectedMessage: "Null",
                 __Unused_WrongSignatureType: "Null",
                 VerifyWeb3SignatureFailed: "Null",
-                NoEligibleIdentity: "Null",
+                __Unused_NoEligibleIdentity: "Null",
```
Kailai-Wang: I suggest removing it completely.

Imagine we have a new entry later, we might erroneously append them like:
```
// in rust
#[codec(index = 11)]
NewEntry,

// in definitions.ts
__Unused_NoEligibleIdentity: "Null",
NewEntry: "Null",
```

Which is wrong, so it's very error-prone IMO.

jonalvarezz (approved):
thanks!

tee-worker/app-libs/stf/src/trusted_call_litentry.rs
```diff
@@ -196,10 +196,9 @@ impl TrustedCallSigned {
 		let assertion_networks = assertion.get_supported_web3networks();
 		let identities = get_eligible_identities(id_graph.as_ref(), assertion_networks);
 
-		ensure!(
-			!identities.is_empty(),
-			StfError::RequestVCFailed(assertion, ErrorDetail::NoEligibleIdentity)
-		);
+		if identities.is_empty() {
+			warn!("[RequestVc] : No eligible identities");
```
BillyWooo: if the purpose is to not leak user info, should here still print out such text?

tee-worker/litentry/core/assertion-build/src/brc20/amount_holder.rs
```diff
@@ -46,7 +46,12 @@ pub fn build(
 	})?;
 
 	if response.is_empty() {
-		Err(Error::RequestVCFailed(Assertion::BRC20AmountHolder, ErrorDetail::NoEligibleIdentity))
+		Err(Error::RequestVCFailed(
+			Assertion::BRC20AmountHolder,
+			ErrorDetail::StfError(ErrorString::truncate_from(
+				"BRC20 token not found for linked identity".as_bytes().to_vec(),
+			)),
```
BillyWooo: same here. 
Do I understand it right: even there is no eligible identity, an empty/false VC will still be created, right? Here will just return error message. No empty/dummy VC.

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
```diff
@@ -529,7 +529,10 @@ where
 		let id_graph_hash = H256::from(blake2_256(&id_graph.encode()));
 		let assertion_networks = assertion.get_supported_web3networks();
 		let identities = get_eligible_identities(id_graph.as_ref(), assertion_networks);
-		ensure!(!identities.is_empty(), "No eligible identity".to_string());
+
+		if identities.is_empty() {
+			warn!("[RequestVc] : No eligible identities");
```
BillyWooo: same here.

BillyWooo (commented):
In general, if we are going to create dummy/false value VC, we need to apply this change to all VCs. 
Or as we are moving to solidity, should we continue on this topic?

tee-worker/litentry/core/assertion-build/src/brc20/amount_holder.rs
wli-pro: Fixed in this commit now: https://github.com/litentry/litentry-parachain/pull/2739/commits/9bf130b8b44100863a9ff89d3c5502bd2ceef238

wli-pro:
> In general, if we are going to create dummy/false value VC, we need to apply this change to all VCs. Or as we are moving to solidity, should we continue on this topic?

All VCs are being applied with this same handling.

tee-worker/litentry/core/vc-task/receiver/src/lib.rs
wli-pro: The previous logic returns an error if the response is empty, but there is already code to process the empty response in `tee-worker/litentry/core/credentials/src/brc20/amount_holder.rs#line124~131`, So this VC logic is wrong, should return an empty/false VC here.

image

Kailai-Wang:
Hi @higherordertech - just to confirm: this will **not** be a breaking change for the sdk or idh, right?

If so I guess it's ready to be merged

tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
```diff
@@ -86,7 +86,6 @@ export default {
                 UnexpectedMessage: "Null",
                 __Unused_WrongSignatureType: "Null",
                 VerifyWeb3SignatureFailed: "Null",
-                NoEligibleIdentity: "Null",
```
wli-pro, line 89: Should the corresponding line to be removed in client-sdk?

PR #2740, opened 2024-05-17
`shield-text` CLI command

kziemianek:
Adds CLI command `shield-text` for encrypting arbitrary string with worker's shielding key.

Assertion's secrets are stored in parachain storage in such form.


jonalvarezz (approved):
looking good to me. 

nitpick: the naming. `shield-string`  sounds more straightforward to me. Or simply `sign-message`, `encrypt`. naming is hard. so no strong opinion about it 😄

kziemianek:
> looking good to me.
> 
> nitpick: the naming. `shield-string` sounds more straightforward to me. Or simply `sign-message`, `encrypt`. naming is hard. so no strong opinion about it 😄

Thanks @jonalvarezz , I've renamed the command to `shield-text`.  Naming is important, especially for public interface.

PR #2741, opened 2024-05-20
Add SPDX license in Dynamic contracts

outofxxx:
as topic.


tee-worker/litentry/core/assertion-build/src/dynamic/contracts/A1.sol
```diff
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
 
+// SPDX-License-Identifier: AGPL-3.0-or-later
```
kziemianek: Why not https://spdx.org/licenses/GPL-3.0-or-later.html ?

outofxxx: Yes, i should be `GPL-3.0-or-later`

kziemianek (approved):

PR #2742, opened 2024-05-20
Remove `A1` assertion from worker's code

kziemianek:
Removes `A1` assertion from worker's code.

It's not needed since we converted it to Dynamic Assertion.


Kailai-Wang:
Would you keep it for a while? At least as long as IDH hasn't switched to DA in the production release 🤔

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
```diff
@@ -276,11 +269,17 @@ where
...
 
-		Assertion::Dynamic(smart_contract_id) => lc_assertion_build::dynamic::build(
-			req,
-			smart_contract_id,
-			context.assertion_repository.clone(),
-		),
+		Assertion::Dynamic(smart_contract_id) => {
+			#[cfg(test)]
+			{
+				std::thread::sleep(core::time::Duration::from_secs(5));
```
Traf333, line 275: for what reason do we need the sleep? Is it awaits something to be created before?

Traf333 (commented):
Looks good to me :+1:

Traf333 (approved):

BillyWooo (changes requested):
Thanks. 
Nice movement to trigger it out. But let me confirm with IDHub, whether they are ready. Then we merge this PR. I will put this PR on hold.

jonalvarezz:
> Would you keep it for a while? At least as long as IDH hasn't switched to DA in the production release 🤔

I'd suggest the same.

is DA ready to use on production? we migrated A1 but it is not used in IDH on production.

tee-worker/litentry/core/stf-task/receiver/src/handler/assertion.rs
kziemianek: It's used by test to detect if two assertions are processed in parralel.

kziemianek:
i'll reopen once we decide we need it

PR #2747, opened 2024-05-21
Fix warnings and import installed openzeppelin

outofxxx:
I'm using https://remix.ethereum.org/ to test.

1. fix unused varibales warning
2. Error: not found DynamicAssertion.sol
3. Use the installed code as openzepplin suggested.

### DIFF
--- a/tee-worker/litentry/core/assertion-build/src/dynamic/contracts/A6.sol
+++ b/tee-worker/litentry/core/assertion-build/src/dynamic/contracts/A6.sol
@@ -19,7 +19,7 @@
 pragma solidity ^0.8.8;
 
 import {DynamicAssertion, Identity, HttpHeader} from "./DynamicAssertion.sol";
-import "https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.0/contracts/utils/Strings.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
 
 contract A6 is DynamicAssertion {
     function execute(Identity[] memory identities, string[] memory secrets)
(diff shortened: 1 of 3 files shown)


tee-worker/litentry/core/assertion-build/src/dynamic/contracts/A6.sol
```diff
@@ -19,7 +19,7 @@
 pragma solidity ^0.8.8;
 
 import {DynamicAssertion, Identity, HttpHeader} from "./DynamicAssertion.sol";
-import "https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.0/contracts/utils/Strings.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
```
kziemianek, line 22: What version is being used here ?

outofxxx: I don't think the version of this library Strings is important. In a real contract development environment, developers can choose the appropriate version to install and import this Strings from the installed version.

kziemianek: As you can see sometime it's important, good example is the fix you try to apply in another PR #2748.
The compilation error is caused by the dependency version change.
If we don't lock on exact version, the code may stop compiling one day.

outofxxx: Given that these codes are currently working properly, I will not make any modifications and will close these two PRs.

PR #2748, opened 2024-05-21
Fix Invalid implicit conversion

outofxxx:
as topic.


kziemianek:
How do you get error with code from `dev` branch ?

In my case there are nor errors/warnings.

I only get error if I use code from #2747

outofxxx:
Ok, I have found the reason. Regarding Strings, in version `4.9.0`, there is a function `toString (int256 value) internal pure returns (string memory)`, so there is no problem with the code on dev and you can directly use toStrings. However, I am using Strings from `npm install @ openzeppelin/contracts`, version `5.0.2` In this version, the signature of the method has changed to function `toStringSigned (int256 value) internal pure returns (string memory)`, resulting in this error

outofxxx:
So, if the three you tested were fine at the time and you prefer to directly import `Strings` from GitHub, then I will keep them in their current state without making any changes.

Related [PR](https://github.com/litentry/litentry-parachain/pull/2747)

PR #2749, opened 2024-05-21
Do not pass this PR

wangminqi:
chore: add wrong script on purpose
To make sure that token bridge **Resource** storage is not used.


wangminqi:
Good.. Now we are certain that this storage is faking. And I have test on rococo with manipulated resource storage of wrong method. Not infecting the success of procedure at all. 

So code of several repo's CI and logic should be discarded/modified. We will raise issue accrodingly.

PR #2750, opened 2024-05-21
basically revert changes from PR 2610.

BillyWooo:
PR 2610 breaks enclave upgrade process. "state.get_block_number()" will return None. 
https://github.com/litentry/litentry-parachain/blob/686c00f21642e9c35f1203e72bd43ed945578060/tee-worker/sidechain/consensus/slots/src/lib.rs#L355


Traf333 (approved):
Not sure about code, but tested locally full cycle of running and upgrading worker - works fine :rocket: 

Thanks @BillyWooo

PR #2755, opened 2024-05-24
RPC getter method with AES encryption

arieljbirnbaum:
### Context

Implement P-183

### How (Optional)

Added as a new method `state_executeAesGetter` to avoid breaking existing clients.
Ideally this ad-hoc solution should be replaced with a "proper" end-to-end encrypted channel (ie "TLS at home").

### Testing Note

I didn't replace all usages of `sendRsaRequestFromGetter` in ts-tests yet since this method is still in use by clients, therefore it should still be exercised.


Kailai-Wang (commented):
The logic looks good, thanks! We'll still need to get CI passed ofc

Kailai-Wang:
This is a breaking change, isn't

arieljbirnbaum:
@Kailai-Wang 

> This is a breaking change, isn't

It shouldn't be --- the encrypted getter response is added as a new method, so existing calls to `state_executeGetter` should not be affected.

0xverin (commented):
Thanks! Ts part looks good, and ci failures seems to be caused by other issues(your error case and schema issue), just fix it.

Kailai-Wang (approved):
Cool, so finally CI passed?

PR #2759, opened 2024-05-26
Build and publish a ready-to-use docker image

Kailai-Wang:
### Context

As topic - it uses the same tag for all docker images.

This is trial run of building identity worker and bitacross-worker: https://github.com/litentry/litentry-parachain/releases/tag/untagged-2f18187a98096d4ad600

It seems that the real SGX support is only required during runtime, not compiling time. So everything is compiled using gh runners. Yang did a smoke test showing that the compiled binary and enclave can be directly run on a HW host.

For the "packaging" part I pretty much copied what we have in devops repo. It's intentionally kept compatible to the current structure (/origin, /data, /user/local/bin) although it's not ideal.


0xverin (commented):
Will we publish all new docker images after merging this PR?

Kailai-Wang:
> Will we publish all new docker images after merging this PR?

Only when you run this release build

Traf333:
Is there any way to protect pushing to docker image versions only once? Or lock the version once it was published to dockerhub. I'm thinking the way to prevent pushing other versions of parachain or workers to the same docker image tag

Makefile
```diff
@@ -245,6 +245,12 @@ shellcheck:
 	@find tee-worker/scripts/litentry/ -name "*.sh" | xargs shellcheck -S warning
 	@echo "Ok"
 
+# generate release tag
+
+.PHONY: generate-tag ## generate full release tag
+generate-tag:
+	@./scripts/generate-release-tag.sh
```
Traf333: did you forget to `git add` some file? Or it should be `generate-release-notes`?

Kailai-Wang: Good catch - there was a script to generate that composited tag `pxxx-ixxx-bxxx` 😂  - but not valid anyone after we decided to use a general tag

I'll remove it

Kailai-Wang:
> Is there any way to protect pushing to docker image versions only once? Or lock the version once it was published to dockerhub. I'm thinking the way to prevent pushing other versions of parachain or workers to the same docker image tag

There is a way (like check-before-push), but I could imagine there might be cases we do want a force-push or replacement.

So I basically say, both cases could happen, so it's the GHA actor's responsibility to take care of this

BillyWooo (approved):
LGTM
@m1iktea please also take a look. Thank you.

m1iktea:
Why is the worker image still being built using `target: deployed-worker` in the `ci.yaml file`?

Kailai-Wang:
> deployed-worker

This is our normal CI - which requires an image built with SW and WORKER_DEV

Our **release** image is built with HW mode.

.github/workflows/create-release-draft.yml
```diff
@@ -145,17 +138,17 @@ jobs:
 
       - name: Push docker image
         run: |
-          docker push litentry/litentry-parachain:$PARACHAIN_DOCKER_TAG
+          docker push litentry/litentry-parachain:${{ env.RELEASE_TAG }}
```
kziemianek: I would suggest pushing at the end of the workflow - to make sure we publish only once all builds/tests/checks pass.

Otherwise some images might be published some not.

Traf333 (approved):
Looks good. Thanks 👍

PR #2766, opened 2024-05-28
use secp256k1 for producing ecdsa signature

kziemianek:
* Switches ecdsa signing from `k256` to `secp256k1` due to known problems under SGX HW mode
* removes schnorr `k256` signing due to known problems under SGX HW mode


Kailai-Wang (approved):
Thanks!

PR #2771, opened 2024-05-29
feat: support bean token holding on combo mainnet

felixfaisal:
This adds new web3network `combo` and supports `holding amount` calculation for bean. 
I couldn't perform e2e test.

### DIFF
--- a/tee-worker/litentry/core/common/src/lib.rs
+++ b/tee-worker/litentry/core/common/src/lib.rs
@@ -49,5 +49,6 @@ pub fn web3_network_to_chain(network: &Web3Network) -> &'static str {
 		Web3Network::Polygon => "polygon",
 		Web3Network::Arbitrum => "arbitrum",
 		Web3Network::Solana => "solana",
+		Web3Network::Combo => "combo-mainnet",
 	}
 }
--- a/tee-worker/litentry/core/common/src/lib.rs
+++ b/tee-worker/litentry/core/common/src/lib.rs
@@ -49,6 +49,6 @@ pub fn web3_network_to_chain(network: &Web3Network) -> &'static str {
 		Web3Network::Polygon => "polygon",
 		Web3Network::Arbitrum => "arbitrum",
 		Web3Network::Solana => "solana",
-		Web3Network::Combo => "combo-mainnet",
+		Web3Network::Combo => "combo",
 	}
 }
--- a/tee-worker/litentry/core/data-providers/src/achainable.rs
+++ b/tee-worker/litentry/core/data-providers/src/achainable.rs
@@ -188,6 +188,7 @@ pub fn web3_network_to_chain(network: &Web3Network) -> String {
 		Web3Network::Polygon => "polygon".into(),
 		Web3Network::Arbitrum => "arbitrum".into(),
 		Web3Network::Solana => "solana".into(),
+		Web3Network::Combo => "combo-mainnet".into(),
 	}
 }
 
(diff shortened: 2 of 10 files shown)


tee-worker/litentry/core/common/src/lib.rs
```diff
@@ -49,5 +49,6 @@ pub fn web3_network_to_chain(network: &Web3Network) -> &'static str {
 		Web3Network::Polygon => "polygon",
 		Web3Network::Arbitrum => "arbitrum",
 		Web3Network::Solana => "solana",
+		Web3Network::Combo => "combo-mainnet",
```
Kailai-Wang: Maybe we should use `combo` here - similar to `bsc`

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -188,6 +188,7 @@ pub fn web3_network_to_chain(network: &Web3Network) -> String {
 		Web3Network::Polygon => "polygon".into(),
 		Web3Network::Arbitrum => "arbitrum".into(),
 		Web3Network::Solana => "solana".into(),
+		Web3Network::Combo => "combo-mainnet".into(),
```
Kailai-Wang: same here - although we might not use it for achainable

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
```diff
@@ -91,6 +95,7 @@ impl NoderealChain {
 			NoderealChain::Aptos => "aptos",
 			NoderealChain::Opt => "opt",
 			NoderealChain::Polygon => "polygon",
+			NoderealChain::Combo => "combo-mainnet",
```
Kailai-Wang: It has to be `combo` here - the `mainnet` is appended by default into the url

Kailai-Wang (changes requested):

Kailai-Wang:
Another thing is: we are adding new item to `Web3Network`, I assume this won't require any state migration, right? (meaning the stored user IDGraph could still be retrieved after this upgrade)

Kailai-Wang:
What's still missing is we need to **add** token amount on Bsc and Combo to get the final balance, similar to e.g. LIT - so you need to modify here

```
	match token_type {
		Web3TokenType::Bnb => bnb_balance::get_balance(addresses, data_provider_config),
		Web3TokenType::Eth => eth_balance::get_balance(addresses, data_provider_config),
		Web3TokenType::Lit => lit_balance::get_balance(addresses, data_provider_config),
		Web3TokenType::Sol => sol_balance::get_balance(addresses, data_provider_config),
		Web3TokenType::Btc => btc_balance::get_balance(addresses, data_provider_config),
		_ => common::get_balance(addresses, token_type, data_provider_config),
```

tee-worker/litentry/core/common/src/lib.rs
felixfaisal: Updated

tee-worker/litentry/core/data-providers/src/achainable.rs
felixfaisal: Updated

tee-worker/litentry/core/data-providers/src/nodereal_jsonrpc.rs
felixfaisal: Updated

Kailai-Wang (approved):

PR #2779, opened 2024-06-01
Take web2 identities into account when filtering identities

Kailai-Wang:
### Context

Current `get_eligible_identities` will retain web2 identities even if `desired_web3networks` is non-empty, this was done intentioanlly (see `get_eligible_identities_works` unit test).

However, this will pass web2 identities through the filter and eventually be fed into the data provider requests. Sometimes the data provider (as well as our logic) can't handle unexpected request params properly - see related issue.

This PR filters web2 identities out when `desired_web3networks` are non-empty, except for `A1`.


primitives/core/src/assertion/mod.rs
```diff
@@ -189,4 +189,19 @@ impl Assertion {
...
+	// However, there's one exception: A1, which requires at least 1 web2 identity and 1 web3
+	// identity.
+	//
+	// This fn is used in `get_eligible_identities` to decide if we should keep web2 identities
+	// regardless of the value of `get_supported_web3networks`
+	#[allow(clippy::match_like_matches_macro)]
+	pub fn force_retain_web2_identity(&self) -> bool {
+		match self {
+			Self::A1 => true,
+			_ => false,
```
kziemianek, line 204: Can we always keep web2 identities and eventually filter them out at assertion logic level ?

Kailai-Wang: Yes thought of those too - but then we'll need to handle that in all 30+ assertions 😂 
It's duplicate code (even if it's just fn call) and easy to forget.

And when I think of relations assertion <> identity. I think most (currently 99%) assertions only need one type of identiies

Are you asking this on behalf of dynamic assertions? There we'll need to handle them all manually anyway I think

silva-fj (approved):

tee-worker/litentry/pallets/identity-management/src/identity_context.rs
```diff
@@ -101,10 +103,26 @@ pub fn get_eligible_identities<T: Config>(
...
+					} else {
+						None
+					}
 				} else {
-					Some((item.0.clone(), networks))
+					if item.0.is_web3() && !networks.is_empty() {
+						Some((item.0.clone(), networks))
+					} else {
+						None
+					}
```
wli-pro, line 125: Maybe use below code to simplify the nesting levels of if-else:
```
if (force_retain_web2_identity || desired_web3networks.is_empty()) && item.0.is_web2() {
	Some((item.0.clone(), networks))
} else if item.0.is_web3() && !networks.is_empty() {
	Some((item.0.clone(), networks))
} else {
	None
}
```

Kailai-Wang: Thanks!

I intentionally keep the nested structure to (hopefully) retain the readability. While it can be logiacally further simplified, down goes the readability too (IMO).

So I'll keep it, it's probably a style problem.

tee-worker/litentry/pallets/identity-management/src/identity_context.rs
```diff
@@ -101,10 +103,26 @@ pub fn get_eligible_identities<T: Config>(
 				//   care about web2 identities, this step will empty `IdentityContext.web3networks`
 				// - it helps to reduce the request size a bit
 				networks.retain(|n| desired_web3networks.contains(n));
-				if networks.is_empty() && item.0.is_web3() {
-					None
+
+				if force_retain_web2_identity && item.0.is_web2() {
+					return Some((item.0.clone(), Vec::new()))
```
felixfaisal, line 108: This seems odd, Like we always know an `IdentityNetworkTuple` for a `web2` identity will always be `(Identity(//web2), vec![])` , is there a way we can improve this logic by changing/modifying the structs to support this usecase and any future usecases? This might be a silly question but I'm seeing if it's possible to improve the readability 😅

Kailai-Wang: I think it lies in the fact that we might use both web2 and web3 identities for an assertion building. I can imagine the best solution is have a `Vec<Network>` instead of `Vec<Web3Network>` to basically as web2 network types to the enum too.

This will be a breaking change tho, and only A1 will benefit from it (for now), so I didn't bother doing it 😂

BillyWooo (approved):
LGTM. 
I also agree not to change `Web3Network` into `Network`. Not worth the breaking change.

primitives/core/src/assertion/mod.rs
kziemianek: Yes, but if I understand correctly this will break Dynamic A1 assertion 😿

Kailai-Wang: It won't break dynamic A1 tho - it should still work as usual.

It's just the identities are already pre-processed before being passed into the assertion building, and this pre-processing is out of the scope of the contract. But again, A1 is really an edge case 😂

PR #2780, opened 2024-06-02
Making sure that the state is the same after migration

silva-fj:
As topic, `migrate-shard` is considered successful only if after migration the state is the same.


tee-worker/core-primitives/stf-state-handler/src/state_snapshot_repository.rs
```diff
@@ -281,6 +295,25 @@ where
...
 		Ok(self.snapshot_history.keys().cloned().collect())
 	}
+
+	fn compare_shards_state_file_size(
+		&self,
+		shard_a: &ShardIdentifier,
+		shard_b: &ShardIdentifier,
+	) -> Result<ShardStateFileComparison> {
+		let state_a_id = self.get_latest_snapshot_metadata(shard_a)?.state_id;
+		let state_size_a = self.file_io.state_file_size(shard_a, state_a_id)?;
```
BillyWooo: Maybe I am picky, just a very minor issue: would `state_a_size` a better name to keep consistency? 😅

BillyWooo:
We are discussing about simplify the migration step. Because current step, it actually is a memory copy/paste in internally. 
But it's nice to have this PR.

BillyWooo (approved):

PR #2604, opened 2024-03-20
Add ws support for DI server

Kailai-Wang:
### Context

As topic. It defines a `MaybeServerTlsStream` struct to wrap up the plain and rustls stream.

Whether launching a ws or wss server will depend on the value of command line arg `--trusted-external-address` or `--T`. Although it's called "external", I reckon it makes little sense when launching an "external" wss server while expecting a ws connection internally in our case. If this value is not present, `wss` will be used as fallback.

Similarly, `--untrusted-external-address` will fallback to `ws`.

Now CI uses `ws` by default.

I can see a potential refactoring of URL management, e.g. use `Url` by default, better stucture of "internal" and "external" addresses - but it's out of scope of this PR.

PR #2782, opened 2024-06-03
Add ws support for bitacross DI server 

kziemianek:
This is a copy of logic added to tee-worker in #2604


Kailai-Wang:
~~After this we'll need to re-launch the worker with ws, right~~

Kailai-Wang (approved):
Maybe we shall add some basic integration test too to bc-worker, in another issue tho

PR #2786, opened 2024-06-04
Add precompile integration unit tests

kziemianek:
Adds unit tests to check that precompiles can be succesfully called from Solidity code using `DynamicAssertion` smart contract.


Kailai-Wang (approved):
Cool, thanks!

PR #2787, opened 2024-06-04
fix: not format the enum assertion for metrics

felixfaisal:
We were previously directly converting the Assertion enum to string using `format!` and this was causing a mess when setting up the monitors, as we end up with a lot of variants because of unique labels. 

This way the `vc_build_time` metric will have 26 possible labels to choose from, and we can set up alert rules after this.


0xverin (commented):
Would you mind simply posting the changed format? Thanks.

kziemianek:
There is a bunch of tasks to migrate current assertions into dynamic assertions.
This means there will be only one variant or one variant with assertion id which will result in almost infinite labels set.
How we are going to deal with it then ? Maybe it doesn't make sense to measure it anymore ?

I think we should answer this question first before we put more effort into making use of this metric.

Kailai-Wang:
For DA shall we count the metrics of contract (addresses) used?

It's fine that we still need time to gradually improve the metric recording and usage, but we'll need to get it running (with basic chain-info metrics) first

kziemianek:
> For DA shall we count the metrics of contract (addresses) used?
> 
> It's fine that we still need time to gradually improve the metric recording and usage, but we'll need to get it running (with basic chain-info metrics) first

Once we become an open platform for VC generation logic I see no point in measuring that.

Kailai-Wang:
We may still find the statistic interesting - like the most used vc. It's rather indexer's work tho

felixfaisal:
> There is a bunch of tasks to migrate current assertions into dynamic assertions. This means there will be only one variant or one variant with assertion id which will result in almost infinite labels set. How we are going to deal with it then ? Maybe it doesn't make sense to measure it anymore ?
> 
> I think we should answer this question first before we put more effort into making use of this metric.

That's a very open ended question and quite broad, Need more clarity beyond just converting assertions, as far as the Dynamic Assertions are concerned, We can make use of `AssertionId` or `ContractAddress` for more granularity, that will not result in an infinite amount. 

I think as long as we are running the worker and creating new assertions, it's on us to measure and benchmark in order to optimise it or even drive business decisions based on it. There are many open ended questions to address even before we begin to become an open platform for VC generation logic.

felixfaisal:
> Would you mind simply posting the changed format? Thanks.

Yeah earlier it would something like `A8(BoundedVec[..]..)` now it will just be `A8`, The parameters would result in a new unique label everytime.

Kailai-Wang:
I'm OK to have it. Even if we switch to DA completely, the building time is still something good to be benchmarked. It can be helpful before the transition already - we should be able to know e.g. which VC takes too long and act accordingly.

Kailai-Wang:
@felixfaisal now that we add DA `id` to labels, will it spam the grafana monitoring? 😂  Or maybe grafana can be configured to only show something like top 5 in that specific view

felixfaisal:
> @felixfaisal now that we add DA `id` to labels, will it spam the grafana monitoring? 😂 Or maybe grafana can be configured to only show something like top 5 in that specific view

I don't think that's a problem, this only serves the metrics on to the metrics endpoint, the prometheus service pulls the metrics from this endpoint, and we can set the refresh rate accordingly, and prometheus has configurable retention length, So we have about 25 Assertions, and then we have a fixed number of contracts.  At some point the hardcoded assertions will be replaced by these fixed number of contracts, So I don't think that's an issue.

Kailai-Wang (approved):

PR #2788, opened 2024-06-05
Revert "Revert "fix: P-441 remove NoEligibleIdentity (#2739)" (#2777)"

wli-pro:
This reverts commit dbcac65f329dcd5fa5b503330f9e8fc996747de3.

### Labels

### Testing Evidences

### DIFF
--- a/primitives/core/src/error.rs
+++ b/primitives/core/src/error.rs
@@ -66,9 +66,6 @@ pub enum ErrorDetail {
 	// of verification data
 	#[codec(index = 10)]
 	VerifyWeb3SignatureFailed,
-	// error when trying to build vc but no eligible identity is found
-	#[codec(index = 11)]
-	NoEligibleIdentity,
 }
 
 impl fmt::Debug for ErrorDetail {
@@ -90,7 +87,6 @@ impl fmt::Debug for ErrorDetail {
 			ErrorDetail::WrongWeb2Handle => write!(f, "WrongWeb2Handle"),
 			ErrorDetail::UnexpectedMessage => write!(f, "UnexpectedMessage"),
 			ErrorDetail::VerifyWeb3SignatureFailed => write!(f, "VerifyWeb3SignatureFailed"),
-			ErrorDetail::NoEligibleIdentity => write!(f, "NoEligibleIdentity"),
 		}
 	}
 }
--- a/primitives/core/src/error.rs
+++ b/primitives/core/src/error.rs
@@ -66,6 +66,10 @@ pub enum ErrorDetail {
 	// of verification data
 	#[codec(index = 10)]
 	VerifyWeb3SignatureFailed,
+	// this is deprecated as this should not be emitted anymore to not leak the information, kept
+	// here so client side type definition do not need change
+	#[codec(index = 11)]
+	NoEligibleIdentity,
 }
 
 impl fmt::Debug for ErrorDetail {
--- a/primitives/core/src/error.rs
+++ b/primitives/core/src/error.rs
@@ -91,6 +91,7 @@ impl fmt::Debug for ErrorDetail {
 			ErrorDetail::WrongWeb2Handle => write!(f, "WrongWeb2Handle"),
 			ErrorDetail::UnexpectedMessage => write!(f, "UnexpectedMessage"),
 			ErrorDetail::VerifyWeb3SignatureFailed => write!(f, "VerifyWeb3SignatureFailed"),
+			ErrorDetail::NoEligibleIdentity => write!(f, "NoEligibleIdentity"),
 		}
 	}
 }
--- a/tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
+++ b/tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
@@ -86,7 +86,6 @@ export default {
                 UnexpectedMessage: "Null",
                 __Unused_WrongSignatureType: "Null",
                 VerifyWeb3SignatureFailed: "Null",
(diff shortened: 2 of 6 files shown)


tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
```diff
@@ -86,7 +86,6 @@ export default {
                 UnexpectedMessage: "Null",
                 __Unused_WrongSignatureType: "Null",
                 VerifyWeb3SignatureFailed: "Null",
-                NoEligibleIdentity: "Null",
```
jonalvarezz: it should be 
```
__Unused_NoEligibleIdentity
```

primitives/core/src/error.rs
```diff
@@ -66,9 +66,6 @@ pub enum ErrorDetail {
 	// of verification data
 	#[codec(index = 10)]
 	VerifyWeb3SignatureFailed,
-	// error when trying to build vc but no eligible identity is found
-	#[codec(index = 11)]
-	NoEligibleIdentity,
```
jonalvarezz: is it safe to remove? or do we need to keep this index unused? I'm thinking of upcoming errors we want to append

tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
Kailai-Wang: @jonalvarezz we intentionally removed it (instead of leaving a placeholder): https://github.com/litentry/litentry-parachain/pull/2739#discussion_r1605571350

jonalvarezz: Thanks @Kailai-Wang, I don't understand why the example in the other comment is wrong. Isn't it how we have been managing enum changes? Or is it because in this case, it is the last item of the enum?

Kailai-Wang: > Or is it because in this case, it is the last item of the enum?

Exactly :)

Kailai-Wang: And since you remove the entry in rust enum, it's better to do it consistently in ts

primitives/core/src/error.rs
Kailai-Wang: In fact, if we really want to retain the maximal compatibility, we could leave this enum unchanged (so does `definitions.ts`)

So this will just become an error type that will never be emitted.
Good thing is you don't have to change anything in SDK. @jonalvarezz @higherordertech

Kailai-Wang: I'm fine with either way - they both have arguments

Will leave it to you to decide

primitives/core/src/error.rs
```diff
@@ -66,7 +66,7 @@ pub enum ErrorDetail {
 	// of verification data
 	#[codec(index = 10)]
 	VerifyWeb3SignatureFailed,
-	// error when trying to build vc but no eligible identity is found
+	// this is deprecated as this should not be emitted anymore to not leak the information, kept here so client side type definition do not need change
 	#[codec(index = 11)]
 	NoEligibleIdentity,
```
wli-pro, line 72: This never-used error type is retained now, but with deprecated note

primitives/core/src/error.rs
```diff
@@ -90,7 +91,6 @@ impl fmt::Debug for ErrorDetail {
 			ErrorDetail::WrongWeb2Handle => write!(f, "WrongWeb2Handle"),
 			ErrorDetail::UnexpectedMessage => write!(f, "UnexpectedMessage"),
 			ErrorDetail::VerifyWeb3SignatureFailed => write!(f, "VerifyWeb3SignatureFailed"),
-			ErrorDetail::NoEligibleIdentity => write!(f, "NoEligibleIdentity"),
```
Kailai-Wang: Then you'll need to keep this

tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
```diff
@@ -86,7 +86,7 @@ export default {
...
+                __Unused_NoEligibleIdentity: "Null",
```
Kailai-Wang: Now that we keep `NoEligibleIdentity`, I suggest we leave it unchagned.

It's "unused" only from logic/code perspective, but the type is actually there.

Kailai-Wang (commented):
But I'm wondering if this PR is still up-to-date, because we don't send any "RequestVcFailed" events to parachian now...

So there's no information leak even for `NoEligibleIdentity`.

On the other side, it does help reduce the response time and alleviate the server load (so that it returns early when possible)

@higherordertech @jonalvarezz

tee-worker/client-api/parachain-api/prepare-build/interfaces/sidechain/definitions.ts
wli-pro: Sounds a plan, so the change becomes:
1 No logic will emit that, including StfError::RequestVCFailed, so to not leak it
2 NoEligibleIdentity and it's type definition still remains unchanged

BillyWooo:
Hi guys @higherordertech @Kailai-Wang , a bit confused. What's the conclusion? 
1. No NoEligibleIdentity event emit.
2. But still return early, if NoEligibleIdentity.
If so, the code logic doesn't match condition 2.

BillyWooo:
Don't need it anymore. Simply close it.

PR #2789, opened 2024-06-05
Use `skip_filtering` when getting eligible identities

Kailai-Wang:
### Context

It should be a better way than `force_retain_web2_identities`.

It's used as a parameter in `get_eligible_identities` to decide if we should simply pass those identities through and let assertion logic handle them.


wli-pro (approved):
LGTM

kziemianek:
Does this mean once we migrate all assertions to dynamic assertions we will remove `desired_web3networks` ?

Kailai-Wang:
> Does this mean once we migrate all assertions to dynamic assertions we will remove `desired_web3networks` ?

Yes if we decide to remove all code of legacy assertions

kziemianek:
This looks like a noteworthy change because it has impact on dynamic assertion results.

kziemianek (approved):

PR #2791, opened 2024-06-06
Avoid nested ErrorDetail in assertion building

Kailai-Wang:
### Context

A small PR as topic

The initial code wraps an extra layer of error even if `e` is already `ErrorDetail`

https://github.com/litentry/litentry-parachain/blob/62f5b3ad964f2783b368f149194466121ab38f13/tee-worker/litentry/core/assertion-build-v2/src/token_holding_amount/mod.rs#L49-L56


wli-pro (approved):
LGTM

PR #2795, opened 2024-06-10
Hotfix LIT/DOT Holder VC

Kailai-Wang:
### Context

It may fix other VCs incidentally too.

The root cause is we missed one place during the AbortStrategy refactoring, so it always returns `false`.

I also don't think we need metadata for `query_system_label` or `check_achainable_label` (btw what are the differences between these?), with metadata I could get response that consists of 10000 lines of JSON...

Additionally, I think `total_transactions` doesn't need metadata either

But please double-check it @higherordertech @zhouhuitian

### DIFF
--- a/tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
+++ b/tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
@@ -111,7 +111,7 @@ pub fn request_achainable(
 		)
 	})?;
 
-	Ok(false)
+	Ok(result)
 }
 
 pub fn request_uniswap_v2_or_v3_user(
--- a/tee-worker/litentry/core/data-providers/src/achainable.rs
+++ b/tee-worker/litentry/core/data-providers/src/achainable.rs
@@ -59,8 +59,11 @@ impl AchainableClient {
 	pub fn query_system_label(&mut self, address: &str, params: Params) -> Result<bool, Error> {
 		// TODO: double-check it with Zhouhui if we don't ever need metadata here
 		let body = ReqBody::new_with_false_metadata(address.into(), params);
-		self.post(SystemLabelReqPath::default(), &body)
-			.and_then(AchainableClient::parse)
+		let r = self
+			.post(SystemLabelReqPath::default(), &body)
+			.and_then(AchainableClient::parse);
+		debug!("r is {:?}", r);
+		r
 	}
 
 	fn parse_class_of_year(value: serde_json::Value) -> Result<String, Error> {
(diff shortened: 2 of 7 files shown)


Kailai-Wang:
This PR bumps the version too 

The (parachain) runtime version is bumped retroactively - our tee-prod has 9181 already, which was upgraded manually when we reverted the NoEligibleIdentity PR which resulted in a different error type

BillyWooo (approved):

tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
```diff
@@ -111,7 +111,7 @@ pub fn request_achainable(
 		)
 	})?;
 
-	Ok(false)
+	Ok(result)
```
jonalvarezz, line 114: 🙃

tee-worker/litentry/core/data-providers/src/achainable.rs
```diff
@@ -96,8 +97,13 @@ impl AchainablePost for AchainableClient {
 		let response = self
 			.client
 			.post_capture::<SystemLabelReqPath, ReqBody, serde_json::Value>(params, body);
-		debug!("ReqBody response: {:?}", response);
-		response.map_err(|e| Error::AchainableError(format!("Achainable response error: {}", e)))
+		match response {
+			Ok(r) => {
+				debug!("ReqBody response: {}", r);
```
jonalvarezz, line 102: do we want to keep this log? seems noisy to me, but ok if we find it valuable

jonalvarezz (approved):

tee-worker/litentry/core/data-providers/src/achainable.rs
Kailai-Wang: Yes only in dev - never in prod

tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
wli-pro: Well spot, but this issue was not introduced in AbortStrategy refactoring PR (because the AbortStrategy PR doesn't try to alter any existing logic in general) but it actually this was already like this before the AbortStrategy PR: https://github.com/litentry/litentry-parachain/pull/2707/files#diff-c23f6a3018245418b1a8d1c1dc7187aeae2c92785b32294c437e46849be35aaeR114, i.e. the original implementation might have problems since day 1.

wli-pro (approved):
LGTM

tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
Kailai-Wang: I see - thanks.

I'm wondering if it ever worked before - including our data provider test. cc @0xverin @BillyWooo

outofxxx (approved):

tee-worker/litentry/core/assertion-build/src/achainable/mod.rs
BillyWooo: @Kailai-Wang @higherordertech 
I think[ it worked before](https://github.com/litentry/litentry-parachain/pull/2707/files#diff-c23f6a3018245418b1a8d1c1dc7187aeae2c92785b32294c437e46849be35aaeL98) .

PR #2801, opened 2024-06-11
Wrong return case. When VC condition doesn't match, generate VC with false value.

BillyWooo:
very minor fix.


Kailai-Wang:
While I agree with the code adjustment, I think the title is not accurate: it should have nothing to do with EligibleIdentity

jonalvarezz (approved):

wli-pro (commented):
LGTM

PR #2802, opened 2024-06-11
add solidity nodereal client

kziemianek:
Adds Solidity library for integration with Nodereal JSON-RPC api.
`eth_getBalance` and `nr_getTokenBalance20` should be sufficient to fullfill requirements of eth and bnb token holding amount assertions.


0xverin (approved):
LGTM, Thanks!

PR #2804, opened 2024-06-12
fix: P-854 change precondition of BNB domain VC value being true

wli-pro:
to require amount to be greater than 0

### Labels

### Testing Evidences


Kailai-Wang (approved):
Thanks!

So it means even if `amount` is 0, the `value` was `true` based on the previous impl?

Kailai-Wang:
But this is not strictly related to P-854, is it?

In that issue, the user has **1** bnbDomain but VC says "not qualified".

wli-pro:
> `value` was `true` based on the previous impl?

Yes, which was wrong

wli-pro:
> But this is not strictly related to P-854, is it?
> 
> In that issue, the user has **1** bnbDomain but VC says "not qualified".

Actully this reported user has 1 but expired domain, hence the VC should really return false only.

PR #2805, opened 2024-06-12
fix: change precondition of LITStaking VC value being true

wli-pro:
to require amount to be greater than 0

### Labels

### Testing Evidences


jonalvarezz (approved):
any chance we can add tests to prevent it from happening again?

Kailai-Wang:
> any chance we can add tests to prevent it from happening again?

maybe @0xverin knows more - does our data provider test cover this case?

PR #2808, opened 2024-06-13
http get precompile

kziemianek:
* Adds HTTP get precompile. It returns JSON string so multiple values can be read from response.


wli-pro (approved):
LGTM

PR #2809, opened 2024-06-13
add json precompiles

kziemianek:
* Adds json precompiles so single values can be extracted from json string by solidity code


wli-pro (approved):
LGTM

PR #2810, opened 2024-06-13
fix: change precondition of A6 VC value true

wli-pro:
to require total followers to be greater than 0

### Labels

### Testing Evidences


jonalvarezz (approved):
Is there a (bad) pattern that took us into this from happenning? 🤔 



Anyway, thanks for fixing, Will

Kailai-Wang:
I'm also surprised we had some many bugs (and did it work before?) - why did we have `true` everywhere 🤔

PR #2813, opened 2024-06-15
Support Intel SDK 2.21 against rust-sdk v2.0.0

Kailai-Wang:
### Context

The target branch of this PR is `p-430-try-out-rust-sgx-sdk-200-branch`, not `dev`.

CI run: https://github.com/litentry/litentry-parachain/actions/runs/9526558116

Rust SDK revision: https://github.com/Kailai-Wang/incubator-teaclave-sgx-sdk/tree/v2.0.0-sdk-2.21, it is based on https://gitlab.com/dexlabs/incubator-teaclave-sgx-sdk/-/tree/v2.0.0-sgx-emm?ref_type=heads

Tested on a newly installed ubuntu 22 with Intel SDK 2.21, both HW and SW mode.

### Notes

1. It's found that `sgx_crypto` has some problems under `ucrypto` feature, so that the RsaKeyPair creation/de-/encryption will panic for untrusted (non-enclave) part, the trusted part looks good though.

Thus this PR tries to use a `lc-rsa-wrapper` crate to call `rsa` crate under the hood. It's mainly used in unit-tests and CLI

2. when experimenting with HW mode, I found that the current enclave config file caused initialisation failure - mainly the TCSNum configuration. I had to decrease it to 256 to make it work (512 wouldn't work), and the number of threads in the vc-task threadpool is down to 128 accordingly.

Edit:
It seems having 960 threads in vc-task-handler why keeping TCSNum = 256 still works (= compiles and runs), I don't know the exact performance penalty though.. I guess we'll need a follow-up issue for that 

From ChatGPT:
> In summary, you can run more threads in a thread pool than the number of TCSs, but only TCSNum threads will be able to execute concurrently inside the enclave. The excess threads will have to wait for a TCS to become available, which requires careful management to ensure efficient execution and to avoid performance bottlenecks.


silva-fj (approved):
LGTM

BillyWooo (approved):

PR #2817, opened 2024-06-17
http post precompile

kziemianek:
Adds HTTP POST precompile. It returns JSON string so multiple values can be read from response.


tee-worker/litentry/core/assertion-build/src/dynamic/contracts/tests/Post.sol
```diff
@@ -0,0 +1,42 @@
...
+contract Post {
+    function callPost(string memory url, string memory payload)
+        public
+        returns (bool, string memory)
+    {
+        HttpHeader[] memory headers = new HttpHeader[](0);
+        return Http.Post(url, payload, headers);
+    }
+
+    function callPostTwiceAndReturnSecondResult(
```
BillyWooo, line 32: maybe I lost the context, can you explain a bit what's the reason of dropping the first response?

kziemianek: This solidity code is used only for testing precompiles. 

In this case we want to prove that http precompile error doesn't revert solidity execution. We do this by calling precompile twice and asserting result.

tee-worker/litentry/core/evm-dynamic-assertions/src/precompiles/http_post.rs
```diff
@@ -1,10 +1,75 @@
-use crate::*;
+use crate::{precompiles::PrecompileResult, *};
 use itc_rest_client::http_client::SendHttpRequest;
 
 http_post_precompile_fn!(http_post_bool, Bool, as_bool);
 http_post_precompile_fn!(http_post_i64, Uint, as_i64);
 http_post_precompile_fn!(http_post_string, String, as_str);
 
+pub fn http_post<T: SendHttpRequest>(input: Vec<u8>, client: T) -> PrecompileResult {
```
Kailai-Wang, line 11: A dumb question:
is `http_post_xx` the convenient fn of `http_post` for different types (with json pointer)? If so why is `http_body` not (re-)used in their implementation body?

kziemianek: In general yes, it only differs becuase it extracts value using given pointer - but we cannot simply reause z`http_post` inside `http_post_xx` in order to reduce duplicated code. 

I've extracted the common part.

Kailai-Wang (approved):
Thanks!

PR #2824, opened 2024-06-18
P-862 add chain-aio-image

m1iktea:
### Context

To resolve this [task](https://linear.app/litentry/issue/P-862/prepare-a-new-docker-image-litentry-parachain-for-testing)

### Labels

- C1

### How (Optional)

I've added a new image litentry/litentry-chain-aio, based on the build logic of litentry-parachain. In the Dockerfile, I've added a new target to differentiate it from the regular litentry-parachain image.

The AIO (all-in-one) image integrates relay-chain and parachain, as well as automated registration features.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -426,21 +426,13 @@ jobs:
 
       - name: Save docker image
         run: |
-          docker save litentry/litentry-parachain:latest | gzip > litentry-parachain-dev.tar.gz
-          docker save litentry/litentry-chain-aio:latest | gzip > litentry-chain-aio-dev.tar.gz
+          docker save litentry/litentry-parachain:latest litentry/litentry-chain-aio:latest | gzip > litentry-parachain-dev.tar.gz
 
       - name: Upload parachain docker image
         uses: actions/upload-artifact@v4
         with:
           name: litentry-parachain-dev
           path: litentry-parachain-dev.tar.gz
-          if-no-files-found: error
-
-      - name: Upload chain-aio docker image
-        uses: actions/upload-artifact@v4
-        with:
-          name: litentry-chain-aio-dev
-          path: litentry-chain-aio-dev.tar.gz
           if-no-files-found: error
 
       - name: Fail early
@@ -914,16 +906,11 @@ jobs:
 
       - uses: actions/download-artifact@v4
         with:
-          name: litentry-chain-aio-dev
-
-      - uses: actions/download-artifact@v4
-        with:
           name: litentry-identity
 
       - name: Load docker image
         run: |
           docker load < litentry-parachain-dev.tar.gz
-          docker load < litentry-chain-aio-dev.tar.gz
           docker load < litentry-identity.tar.gz
           
       - name: Dockerhub login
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -434,6 +436,13 @@ jobs:
           path: litentry-parachain-dev.tar.gz
           if-no-files-found: error
 
+      - name: Upload docker image
```
BillyWooo: should not have the same name as previous step.

docker/entrypoint.sh
```diff
@@ -0,0 +1,135 @@
...
+
+PARACHAIN_BASEDIR="/opt/litentry/parachain"
+REPO_DIR="/code/litentry-parachain"
+
+check(){
+    if [ -z "$CHAIN" ]; then
+        export CHAIN="rococo"
+    fi
+
+    ALLOWED_VALUES=("rococo" "litmus" "litentry")
```
BillyWooo, line 14: I think we only 'rococo' is enough.

docker/Dockerfile
```diff
@@ -49,3 +49,59 @@ VOLUME ["/data"]
...
+    [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" && \
+    [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion" && \
+    nvm install 18 && \
+    nvm use 18 && \
+    apt-get clean && \
+    rm -rf /var/cache/apt/lists 
+
+RUN useradd -m -u 1000 -U -s /bin/sh -d /litentry litentry && \
+	mkdir -p /opt/litentry/parachain /code/litentry-parachain && \
+	chown -R litentry:litentry /opt/litentry
```
BillyWooo, line 88: I think here miss to change the ownership of `/code/litentry-parachain`.

BillyWooo (commented):
In general, it looks good 👍 .
I left some comments. 
Please fix the CI failure. Then we test it a bit.

docker/Dockerfile
m1iktea: The ownership change of /code/litentry-parachain must occur after the COPY operation. There is already a line for this after line 95.

BillyWooo (approved):

.github/workflows/ci.yml
```diff
@@ -422,18 +422,27 @@ jobs:
         if: needs.set-condition.outputs.rebuild_parachain == 'false'
         run: |
           docker pull litentry/litentry-parachain:latest
+          docker pull litentry/litentry-chain-aio:latest
 
       - name: Save docker image
         run: |
           docker save litentry/litentry-parachain:latest | gzip > litentry-parachain-dev.tar.gz
+          docker save litentry/litentry-chain-aio:latest | gzip > litentry-chain-aio-dev.tar.gz
```
Kailai-Wang: You could save multiple images to one tgz file (if you want), so that we once need upload/download once 😂

Kailai-Wang (approved):
Not much to complain, thank you!

.github/workflows/ci.yml
m1iktea: ok

PR #2825, opened 2024-06-18
Adding support to send batch extrinsics to the parachain 

silva-fj:
As topic: This PR adds the structure to send batch extrinsics to the parachain asynchronously. Starting with `vc_issued`


tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+			}
+		}
+		if !calls.is_empty() {
+			let extrinsic =
+				match extrinsic_factory.create_batch_extrinsic(calls.drain(..).collect(), None) {
+					Ok(extrinsic) => extrinsic,
+					Err(e) => {
+						error!("Failed to create extrinsic: {:?}", e);
+						continue
+					},
```
Kailai-Wang, line 49: If it errors out - does it mean the `calls` will be lost (as they are drained already)? There's no second chance for them to be sent.

Please also consider:
- `utility.batch` itself has a limit: `batched_calls_limit()`. It could be exceeded if we have a huge amount of `calls` within 6 sec
- the heavy (total) weight might constantly delay the parachain block production
- there should be a limit of byte array size going through the ocall

Please double-check those.
For the first two points an easy way is we do paged batch (like max 500 in one batch or so)

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+	OCallApi: EnclaveOnChainOCallApi,
+{
+	let task_receiver = init_parachain_extrinsic_sender_storage().map_err(|e| {
+		format!("Failed to initialize parachain extrinsic task sender storage: {:?}", e)
+	})?;
+	let mut calls = vec::Vec::new();
+
+	loop {
+		let start_time = time::Instant::now();
+		while start_time.elapsed() < time::Duration::from_secs(6) {
```
Kailai-Wang: Any specific reason for 6? Or is it just a random number that we feel comfortable to start with?

Kailai-Wang (commented):
Thanks! The logic looks fine, only have questions regarding the batch parameters

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
BillyWooo: We do have the problem of "huge amount of `calls` within 6 sec", based on the experience of last two campaign events.

BillyWooo (commented):
In general, LGTM. 
Except "too many calls", just one more point: are we considering to not use such global variables `GLOBAL_PARACHAIN_EXTRINSIC_TASK` in the future? 
maybe @Kailai-Wang have any idea?

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
silva-fj: > If it errors out - does it mean the calls will be lost (as they are drained already)? There's no second chance for them to be sent.

I thought about this, if there is an error it means there was an issue with the cached data (metadata or nonce), so I'm not sure how we could recover from that, we could add some sort of retry mechanism if its possible to recover, but if its not, the data will just grow until it runs out of memory.

> Please also consider:
utility.batch itself has a limit: batched_calls_limit(). It could be exceeded if we have a huge amount of calls within 6 sec
the heavy (total) weight might constantly delay the parachain block production
there should be a limit of byte array size going through the ocall
Please double-check those.
For the first two points an easy way is we do paged batch (like max 500 in one batch or so)

Yes I checked this, the limit is almost 11k (It should be possible to get this value from the metadata), but I think it makes sense to create bounded batches to be safe 👍🏼 

> We do have the problem of "huge amount of calls within 6 sec", based on the experience of last two campaign events.

I think for our use case it should be fine as these extrinsics are only for emitting events

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
silva-fj: I forgot to ask about this, for now I put what was mentioned in the issue but I wasn't sure what a sensible value for this would be

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
silva-fj: > I thought about this, if there is an error it means there was an issue with the cached data (metadata or nonce), so I'm not sure how we could recover from that, we could add some sort of retry mechanism if its possible to recover, but if its not, the data will just grow until it runs out of memory.

Actually, if `send_to_parentchain` fails it makes sense to retry them, I'll update that

silva-fj: After giving it some thought, I'm not sure a retry mechanism here would make sense unless we persist the calls somehow so it could retry again after restarting the worker for example. But I think this would be out of scope for this issue as the end result is not changing, currently if we have such errors, extrinsics are never sent

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+	let mut calls = vec::Vec::new();
+
+	loop {
+		let start_time = time::Instant::now();
+		while start_time.elapsed() < BATCH_EXTRINSIC_INTERVAL && calls.len() < MAX_BATCH_SIZE {
+			if let Ok(call) = task_receiver.recv() {
+				calls.push(call);
+			}
+		}
+		if !calls.is_empty() {
```
felixfaisal, line 42: So there is a possible scenario here, when you exit the above loop, that is the `recv()`, the vc task might still be sending extrinsics unaware that the receiver has exit the loop and this thread is in the process of submitting extrinsic, this will result in `Disconnected` error when the vc task tries submitting the extrinsic. That might not happen I guess 😅 
I'd be more comfortable if the while condition be convert to an `if` and the `if` condition be converted to `while`, So something like this: 
```rust
while let Ok(call) = task_receiver.recv() { 
      calls.push(call)
      if start_time.elapsed() > BATCH_EXTRINSIC_INTERVAL && .. { 
            // submit extrinsics here 
      }
}
```

felixfaisal: I thought about this, This is not a good idea too, We may be waiting for too long since this if condition is only triggered when it receives the message. So please ignore, i think the current logic should be fine

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+	OCallApi: EnclaveOnChainOCallApi,
+{
+	let task_receiver = init_parachain_extrinsic_sender_storage().map_err(|e| {
+		format!("Failed to initialize parachain extrinsic task sender storage: {:?}", e)
+	})?;
+	let mut calls = vec::Vec::new();
+
+	loop {
+		let start_time = time::Instant::now();
+		while start_time.elapsed() < BATCH_EXTRINSIC_INTERVAL && calls.len() < MAX_BATCH_SIZE {
```
felixfaisal, line 34: So I went through the logic, this particular condition - `start_time.elapsed() < BATCH_EXTRINSIC_INTERVAL && calls.len() < MAX_BATCH_SIZE` , it would suggest that at high traffic we will still be submitting a lot of extrinsics but in the form of batch extrinsics, so batches of 500 at a time correct me if I'm wrong (which more or less occupies the same weight as all the calls). This will still take up a lot of the block weight and not allow more important calls like `ParentchainBlockProcessed` to be present in the block, and yes the nonce increments will be fewer since we are submitting in batches. 

So we need to decouple these two conditions, We create a vector of all batched calls like this - `all_batched_calls: Vec<BoundedVec<500>>`. So when we hit capacity we insert the calls into it probably like this `all_batched_calls.insert(calls)`, and only when we hit the interval time, we submit these calls. So this max size should be such that it should take up only half of the block weight in the worst case, so that the chain is usable for other purposes and not spammed with worker extrinsics during high traffic. I'm just suggesting one logic, you can do this however you like.

felixfaisal: Based on this condition `while start_time.elapsed() < BATCH_EXTRINSIC_INTERVAL && calls.len() < MAX_BATCH_SIZE {`, In high traffic, We are still going submit extrinsics in batches of 500 to the chain and take up most of the available block weight. 
- We need to decouple these two conditions, When we hit max_batch_size we need to maintain an overflow vector and push it there 
- We should submit extrinsic only when we hit the interval, and the max_batch_size should be such that it takes up less than half of the total block weight so that during high traffic the chain is still usable for other activities 
- These extrinsics are not criticial to the system or UX, So it's fine if we submit only one batch every interval, the case in point is that the chain should be usable and there should be a decent amount of block weight left after it. 
- Calculating the weight will be a difficult thing, But since this is a batch of the same calls, it might be easier to come up with an okay number, the goal is to not spam the chain with so many calls, so it doesn't have to be very accurate if it does the job.

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
silva-fj: There should be no disconnect issue as the receiver instance is outside the loop, it is never dropped

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
silva-fj: Oh yes, the condition of the while loop should only consider the interval, good catch. I think I had it that way, not sure why I changed it.

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
Kailai-Wang: When sending fails due to nonce clash (e.g. too low nonce, it could casually happen due to e.g. too late parachain sync), retrying will likely fix it.

You are using `utility.batch`, so it's possible that only some calls fail. If we apply a retrying mechanism, do we retry the whole batch or only those failed ext?

Kailai-Wang: The parachain ext could mutate the parachain state, so it seems we would need them to be sent exactly once, not more, not less

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
silva-fj: if we take as an example (assuming the benchmarks are more or less accurate) the reference we have for `vc_issued` that is around 27M (the ref time), then 500 calls would take around 4% of the maxExtrinsic limit for normal class dispatch on the litentry parachain (If I read this correctly)

```
{
  baseBlock: {
    refTime: 392,184,000
    proofSize: 0
  }
  maxBlock: {
    refTime: 500,000,000,000
    proofSize: 5,242,880
  }
  perClass: {
    normal: {
      baseExtrinsic: {
        refTime: 113,638,000
        proofSize: 0
      }
      maxExtrinsic: {
        refTime: 324,886,362,000
        proofSize: 3,407,872
      }
      maxTotal: {
        refTime: 375,000,000,000
        proofSize: 3,932,160
      }
      reserved: {
        refTime: 0
        proofSize: 0
      }
    }
    operational: {
      baseExtrinsic: {
        refTime: 113,638,000
        proofSize: 0
      }
      maxExtrinsic: {
        refTime: 449,886,362,000
        proofSize: 4,718,592
      }
      maxTotal: {
        refTime: 500,000,000,000
        proofSize: 5,242,880
      }
      reserved: {
        refTime: 125,000,000,000
        proofSize: 1,310,720
      }
    }
    mandatory: {
      baseExtrinsic: {
        refTime: 113,638,000
        proofSize: 0
      }
      maxExtrinsic: null
      maxTotal: null
      reserved: null
    }
  }
}
```

I'm not 100% sure about these numbers but I think it should be fine. @Kailai-Wang Do you have some inputs on this? Is my assumption correct?

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
silva-fj: > You are using utility.batch, so it's possible that only some calls fail. If we apply a retrying mechanism, do we retry the whole batch or only those failed ext?

This is very good point, I think it would be a bit hard to retry the remaining calls. As soon as one of the calls in the batch fails an event is emitted (`BatchInterrupted`) with the index of the call and the error, then the batch call stops there, so we would have to listen to those events somehow and send another batch with the remaining calls from that index. 

Another option would be to use `batch_all` which is an atomic operation, so the transaction will rollback and fail if any of the calls failed

silva-fj: > When sending fails due to nonce clash (e.g. too low nonce, it could casually happen due to e.g. too late parachain sync), retrying will likely fix it.

Yes, you are right. The issue is that with the in current implementation there is no way to know that the call failed because of invalid nonce, we get an error only if the extrinsic could not be decoded in which case I don't think there is something we can do to recover.

Kailai-Wang: I think we'll need to have SendingStrategy here again, substrate-api-client already has some useful state indication, like if waiting for inclusion or not, if waiting for success or not.

But that's out of this PR's scope

tee-worker/litentry/core/parachain-extrinsic-task/receiver/src/lib.rs
Kailai-Wang: 500 should be a conservative estimation - we saw 1000+ `vc_issued` ext in one block without problems.

It's rather for (yet unknown) ext in the future if we plan to use this global thread to interact with parachain - but again it's not this PR's business

Kailai-Wang (approved):
I approve it as it's in any case an improvement to current structure.

> there should be a limit of byte array size going through the ocall

Please double-check it and/or change the limit

Shall we use this thread to send other extrinsics? Like teebag callbacks? My preference is yes.

Currently it's often used as:
```
		self.validator_accessor
			.execute_mut_on_validator(|v| v.send_extrinsics(parentchain_extrinsics))?;
```

Does it have the same effect as having this global thread?

silva-fj:
> there should be a limit of byte array size going through the ocall

The limit for `extrinsics_encoded_size` is `u32`, it is big enough

silva-fj:
> Shall we use this thread to send other extrinsics? Like teebag callbacks? My preference is yes.
> 
> Currently it's often used as:
> 
> ```
> 		self.validator_accessor
> 			.execute_mut_on_validator(|v| v.send_extrinsics(parentchain_extrinsics))?;
> ```
> 
> Does it have the same effect as having this global thread?

So, there are a couple of differences:
When using `validator_access.execute_mut_on_validator(|v| v.send_extrinsics(xts))?;`, after sending the extrinsics, we seal the `LightValidationState`, at the moment I'm not 100% sure what this implies. 

The other difference is the nature of the interval batching, if we are Ok with eventual consistency there should be no problem.

@Kailai-Wang Should I make the changes to send all extrinsics to the new thread as part of this PR? or should we create a follow up issue?

Kailai-Wang:
Please do it in a separate PR : )

PR #2826, opened 2024-06-18
Decrease ceremony runner tick interval to 1 ms

kziemianek:
* Decreases ceremony runner tick interval to 1 ms


Kailai-Wang:
Have you tried it locally how long it takes now? Or do we wait for deployment with this commit?

kziemianek:
It decreased from 2.7 s to 20 ms but times vary depends on run (network calls).

Previously: Ceremony took: 2.699441812s
Now: Ceremony took: 20.217815ms

BillyWooo (approved):
Nice.

BillyWooo:
Just an idea comes to my mind: according to your test, if we adjust the tick to e.g. 5ms, would it increase the efficiency?  @kziemianek

PR #2827, opened 2024-06-18
fix bitacross indirect-calls-executor tests

kziemianek:
* Fixes `itc-parentchain-indirect-calls-executor` package tests

`IndirectCallsFilter` was removed from `IndirectCallsExecutor` generics recently but mocks used in tests were not adjusted.


Kailai-Wang (approved):
Thanks!

This would have been detected if we run unit-tests for bitacross workers, right?

PR #2831, opened 2024-06-20
refactor: use println! instead of info!

felixfaisal:
This is a simple PR to use `println!` instead of `info!`, I've looked all the CLI commands, And only at two places I felt like we should be using `println!` instead of `info!`.


Kailai-Wang (commented):
Thank you!

Can you please print the hex format too?

Kailai-Wang (approved):

BillyWooo:
Ok. Just for record,  maybe @m1iktea we need to adapt our [deploy script](https://github.com/litentry/devops/blob/c9362cfb6284e0bb6ada4e24f42006dec9ab3874/worker_deploy/dev_deploy.sh#L784). Need to check.

BillyWooo (approved):

PR #2835, opened 2024-06-24
Remove shard_vault usage

Kailai-Wang:
### Context

More details in the ticket.

Currently we don't need a per-shard account acting as a proxy.


kziemianek (approved):
Looks good, thanks 👍🏼

PR #2838, opened 2024-06-24
Keep assertionText field as default empty

Kailai-Wang:
### Context

As topic - a minor PR to remove the current assertionText assignment, which is basically the debug formatter of assertion type.

@jonalvarezz - or do you want me to remove this field completely? If yes we (you) have to change the schema.

If the only purpose of this field is to "identify" the assertion, I'd like to remove it completely. As the client should conduct their own ways of identifying the credential if they have such needs (e.g. by adding extra fields).


jonalvarezz (approved):
if vc identification is left to clients entirely, then I see no reason to leave it either.

PR #2843, opened 2024-06-25
`Vc.issuer.id` should be the enclave account

Kailai-Wang:
### Context

As topic ... instead of vc_pubkey.

We need this value to find the mapped enclave in `teebag.enclaveRegistry`, see the linked issue.

@jonalvarezz - please check how can the new-sdk verify old vc, maybe by checking timestamp/block-number? Actually, this can be a more general question when there's breaking change, and we might need a more elegant solution


jonalvarezz (approved):
tested locally, looking good

PR #2845, opened 2024-06-26
P-872 basic tests for contracts

0xverin:
this pr:

- Copy the contract to the ts-tests directory for compilation and testing.

- Create assertionId and request vc


tee-worker/litentry/core/assertion-build/src/dynamic/contracts/A6.sol
```diff
@@ -18,7 +18,7 @@
 
 pragma solidity ^0.8.8;
 
-import "https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.0/contracts/utils/Strings.sol";
+import "./openzeppelin/Strings.sol";
```
wli-pro, line 21: Any benefit to hard code this imported code in our repo instead keeping them imported from remote during build time?

0xverin: Contracts import via http cannot be decoded and compiled by any compiler, so the contract will be flattened here.
This is supported in remix, but not in solc-related compilers.

tee-worker/ts-tests/integration-tests/package.json
```diff
@@ -7,7 +7,9 @@
         "format": "prettier --write .",
         "pretest": "eslint .",
         "test": "mocha --exit --sort -r ts-node/register --loader=ts-node/esm",
-        "check-types": "tsc --noEmit"
+        "check-types": "tsc --noEmit",
+        "cp-contracts": "rm -rf ./contracts/ &&  cp -r ../../litentry/core/assertion-build/src/dynamic/contracts ./contracts",
```
Kailai-Wang: Can we use `copy` instead of `cp`? 

It will work correctly no matter which working directory I'm in, right?

Alternatively if it's always run along with `compile-contracts`, we can move the commands there

tee-worker/cli/lit_tee_vc_contracts_test.sh
```diff
@@ -0,0 +1,83 @@
...
+pnpm install
+pnpm run build
+
+
+cd /
+ls
+
+cd /ts-tests
+curl -L https://foundry.paradigm.xyz | bash
+source /root/.bashrc
```
Kailai-Wang: How can I run this test locally?

tee-worker/litentry/core/assertion-build/src/dynamic/contracts/openzeppelin/Strings.sol
Kailai-Wang, line 1: Can you add in the README how/when to upgrade this third-party sol?

tee-worker/docker/lit-tee-vc-contracts-test.yml
```diff
@@ -0,0 +1,24 @@
...
+            dockerfile: build.Dockerfile
+            target: deployed-client
+        depends_on:
+            litentry-node:
+                condition: service_healthy
+            litentry-worker-1:
+                condition: service_healthy
+        networks:
+            - litentry-test-network
+        entrypoint: "bash -c '/usr/local/worker-cli/lit_tee_vc_contracts_test.sh contracts_depoly_request.test.ts 2>&1' "
```
Kailai-Wang: The filenames are confusing

`lit_tee_vc_contracts_test.sh` is already something specific to contract test, why do we need `contracts_depoly_request.test.ts` again?

tee-worker/ts-tests/integration-tests/compile.json
Kailai-Wang, line 1: We probably need README about how this file is generated/updated

tee-worker/ts-tests/integration-tests/contracts_depoly_request.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        );
+        teeShieldingKey = await getTeeShieldingKey(context);
+        aliceSubstrateIdentity = await context.web3Wallets.substrate.Alice.getIdentity(context);
+        alice = genesisSubstrateWallet('Alice');
+    });
+
+    step('loading tokenmapping contract bytecode', async function () {
+        const file = path.resolve('./', './compile.json');
+        const data = fs.readFileSync(file, 'utf8');
+        // contractBytecode = JSON.parse(data).bytecode.object;
```
Kailai-Wang: Remove this line please : )

tee-worker/ts-tests/integration-tests/contracts_depoly_request.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        const secret = '0x' + encryptedSecrets.toString('hex');
+
+        const assertionId = '0x0000000000000000000000000000000000000000';
+        const createAssertionEventsPromise = subscribeToEvents('evmAssertions', 'AssertionCreated', context.api);
+        await context.api.tx.evmAssertions.createAssertion(assertionId, contractBytecode, [secret]).signAndSend(alice);
+        const event = (await createAssertionEventsPromise).map((e) => e);
+        assert.equal(event.length, 1);
+    });
+
+    step('linking identities (alice)', async function () {
```
Kailai-Wang: We should extract the identity-linking code to some common place, or maybe even simply use CLI to link it (if we are not testing the linking itself)

It will increase the readability a lot

Kailai-Wang:
Please also write proper PR description

0xverin:
> Please also write proper PR description

~~Still in testing ci,  I will add remaining content after passed.~~

tee-worker/litentry/core/assertion-build/src/dynamic/contracts/A6.sol
wli-pro: > so the contract will be flattened here

Will flattened code break the solc-related complier? Will it break hardhat compiler too (if not we could use hardhat to test and remain the import from source maybe)?
I feel if we keep import from source it will be much easier to maintain. So trying to see if any workaround to not to  copy them in our own repo.

tee-worker/cli/lit_tee_vc_contracts_test.sh
0xverin: update in README.md

tee-worker/litentry/core/assertion-build/src/dynamic/contracts/A6.sol
0xverin: Hardhat does not support compiling with HTTP imports, only ESM import mode is supported.
The reason for flattening it is because consider that someone will compile in remix, otherwise we can unify esm import.

tee-worker/cli/lit_ts_integration_test.sh
```diff
@@ -62,10 +62,29 @@ cd  /client-api/sidechain-api
...
 
-cd /ts-tests
-pnpm install
+if [ "$TEST" = "assertion_contracts.test.ts" ]; then
+    cd /
+    ls assertion-contracts/
+    cp -r assertion-contracts /ts-tests/integration-tests/contracts
+
+    cd /ts-tests
+    curl -L https://foundry.paradigm.xyz | bash
```
kziemianek, line 76: is it safe to run foreign script that can be changed anytime inside CI ? will it be executed on our hosted runner used for releasing also?

tee-worker/ts-tests/integration-tests/compile.json
0xverin: Temporary test file, already deleted

tee-worker/docker/lit-tee-vc-contracts-test.yml
0xverin: Merged shell scripts, uniformly using lit_ts_integration_test.

tee-worker/ts-tests/integration-tests/contracts_depoly_request.test.ts
0xverin: This PR may be more about whether the contract can be compiled and deployed, and then get results through the deployed contract request VC. 

I think I will optimize this in the data-provider tests(or create a new issue to implement this.).

wli-pro (approved):

tee-worker/cli/lit_ts_integration_test.sh
Kailai-Wang: It won't be executed in releasing, plus we don't use a self-hosted runner now : )

But still, should we move the installation part to cli image?

0xverin:
@felixfaisal  is checking why the develperMembership pallet is missing in both CI and local environments. Once confirmed,  add Alice to the developer member, then merge this PR.

tee-worker/cli/lit_ts_integration_test.sh
BillyWooo: Yeah, moving it into cli image would make the scripts cleaner.

tee-worker/ts-tests/integration-tests/assertion_contracts.test.ts
```diff
@@ -0,0 +1,132 @@
...
+            process.env.WORKER_ENDPOINT!, // @fixme evil assertion; centralize env access
+            process.env.NODE_ENDPOINT! // @fixme evil assertion; centralize env access
+        );
+        teeShieldingKey = await getTeeShieldingKey(context);
+        aliceSubstrateIdentity = await context.web3Wallets.substrate.Alice.getIdentity(context);
+        alice = genesisSubstrateWallet('Alice');
+    });
+
+    step('loading tokenmapping contract bytecode', async function () {
+        const file = path.resolve('./', './contracts-build-info/TokenMapping.sol/TokenMapping.json');
```
BillyWooo, line 53: I know `TokenMapping.sol` is generated out by compiler. But originally how does it come out? derived from which sol file? or ?

Next question would be: how easy/difficult to extend this test structure to cover any other solidity files?

tee-worker/ts-tests/integration-tests/package.json
BillyWooo: I think here what Kai mean is: `cp-contracts` -> `copy-contracts`. Please update it. @0xverin

tee-worker/ts-tests/integration-tests/package.json
```diff
@@ -7,7 +7,9 @@
         "format": "prettier --write .",
         "pretest": "eslint .",
         "test": "mocha --exit --sort -r ts-node/register --loader=ts-node/esm",
-        "check-types": "tsc --noEmit"
+        "check-types": "tsc --noEmit",
+        "cp-contracts": "cp -r ../../litentry/core/assertion-build/src/dynamic/contracts ./contracts",
```
BillyWooo: A minor issue: `copy-contracts` would be better.

tee-worker/ts-tests/integration-tests/assertion_contracts.test.ts
0xverin: >I know TokenMapping.sol is generated out by compiler. But originally how does it come out? derived from which sol file? or ?

TokenMapping.sol comes from [TokenMapping.sol](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/litentry/core/assertion-build/src/dynamic/contracts/token_holding_amount/TokenMapping.sol), the foundry compilation will generate according to the contract name.

>Next question would be: how easy/difficult to extend this test structure to cover any other solidity files?
- We could easily add tests for other token contracts, such as directly adding `['bnb', 'ada'...] `to test.
- Regarding internal testing of the contract , we could write based on Foundry's tests, which is actually not difficult. However, this PR tends to focus more on deploying contract tests.

BillyWooo (approved):

PR #2846, opened 2024-06-26
fix: add origin check for assertions

felixfaisal:
This PR adds a `developer` council, a member of this council can call the evm_assertion extrinsic, anyone other than the members of this council cannot call this extrinsic. 

A few points/questions. 
- I can add a parachain account and idhub account to the developer committee, if there exists such accounts via the genesis file 
- I see that evm_assertions exists only in the rococo runtime, so I added this council only to the rococo runtime.


runtime/common/src/lib.rs
```diff
@@ -273,6 +275,22 @@ pub type EnsureRootOrTwoThirdsTechnicalCommittee = EitherOfDiverse<
...
+pub type EnsureRootOrHalfDeveloperCommittee = EitherOfDiverse<
+	EnsureRoot<AccountId>,
+	pallet_collective::EnsureProportionAtLeast<AccountId, DeveloperCommitteeInstance, 1, 2>,
+>;
+
+pub type EnsureRootOrTwoThirdsDeveloperCommittee = EitherOfDiverse<
+	EnsureRoot<AccountId>,
+	pallet_collective::EnsureProportionAtLeast<AccountId, DeveloperCommitteeInstance, 2, 3>,
+>;
+
```
Kailai-Wang: These are acutally not used, aren't they

runtime/common/src/lib.rs
```diff
@@ -234,6 +234,8 @@ pub type CouncilInstance = pallet_collective::Instance1;
 pub type TechnicalCommitteeInstance = pallet_collective::Instance2;
 pub type CouncilMembershipInstance = pallet_membership::Instance1;
 pub type TechnicalCommitteeMembershipInstance = pallet_membership::Instance2;
+pub type DeveloperCommitteeInstance = pallet_collective::Instance3;
+pub type DeveloperCommitteeMembershipInstance = pallet_membership::Instance3;
```
Kailai-Wang: Being picky: I'm wondering why not follow the order of declaration

Kailai-Wang (approved):
Thanks!

> I can add a parachain account and idhub account to the developer committee, if there exists such accounts via the genesis file

For dev-chain we can simply use Alice
For prod-chain genesis has no effect anyway

silva-fj (approved):

runtime/common/src/lib.rs
felixfaisal: Removed ✅

runtime/common/src/lib.rs
felixfaisal: Updated to follow the order ✅

PR #2849, opened 2024-06-28
P-906 fix: update limit param to 100 for the geniidata API for BRC20 VC

wli-pro:
and change it to pagination query

### Context

hotfix geniidata api breaking changes

### Labels

### Testing Evidences

image

### DIFF
--- a/tee-worker/litentry/core/data-providers/src/geniidata.rs
+++ b/tee-worker/litentry/core/data-providers/src/geniidata.rs
@@ -49,7 +49,7 @@ pub struct ResponseItem {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct ResponseData {
-	pub count: u64,
+	pub count: u32,
 	pub limit: String,
 	pub offset: String,
 	pub list: Vec<ResponseItem>,
@@ -68,8 +68,8 @@ impl RestPath<String> for GeniidataResponse {
 	}
 }
 
-// According to https://geniidata.readme.io/reference/get-brc20-tick-list-copy, the maximum limit is i32::MAX
-const GENIIDATA_QUERY_LIMIT: &str = "2147483647";
+// According to https://geniidata.readme.io/reference/get-brc20-tick-list-copy, the maximum limit is 100
+const GENIIDATA_QUERY_LIMIT: &str = "100";
 
 pub struct GeniidataClient {
 	client: RestClient<HttpClient<SendWithCertificateVerification>>,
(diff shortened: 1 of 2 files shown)


tee-worker/litentry/core/data-providers/src/geniidata.rs
```diff
@@ -68,8 +68,8 @@ impl RestPath<String> for GeniidataResponse {
 	}
 }
 
-// According to https://geniidata.readme.io/reference/get-brc20-tick-list-copy, the maximum limit is i32::MAX
-const GENIIDATA_QUERY_LIMIT: &str = "2147483647";
+// According to https://geniidata.readme.io/reference/get-brc20-tick-list-copy, the maximum limit is 100
```
Kailai-Wang, line 71: Hmm I don't see the limit is 100 tho, the doc only says `i32`

tee-worker/litentry/core/data-providers/src/geniidata.rs
```diff
@@ -98,18 +98,29 @@ impl GeniidataClient {
...
-					.client
-					.get_with::<String, GeniidataResponse>("".to_string(), query.as_slice())
-					.map_err(|e| {
-						DataProviderError::GeniiDataError(format!(
-							"GeniiData response error: {}",
-							e
-						))
-					})?;
-				all_items.extend(response.data.list);
+				let mut offset: u32 = 0;
```
Kailai-Wang, line 101: Why `offset` is `u32` while `GENIIDATA_QUERY_LIMIT` is `&str`? Seems inconsistent

tee-worker/litentry/core/data-providers/src/geniidata.rs
```diff
@@ -98,18 +98,29 @@ impl GeniidataClient {
...
+				loop {
+					let offset_str = offset.to_string();
+					let query = vec![
+						("limit", GENIIDATA_QUERY_LIMIT),
+						("offset", offset_str.as_str()),
+						("address", address),
+					];
+					let response = self
```
Kailai-Wang, line 110: It seems that the last round query will have offset > total data count, so we are sure that the response will have meaningful results, right? (instead of error out)

tee-worker/litentry/core/data-providers/src/geniidata.rs
wli-pro: Yeah, that's exactly the problem. What their doc says is not their runtime response allowed, see:
image

tee-worker/litentry/core/data-providers/src/geniidata.rs
wli-pro: Yes, the dp API simply returns empty array  when offset > total (instead error out):

image

tee-worker/litentry/core/data-providers/src/geniidata.rs
wli-pro: resolved by making GENIIDATA_QUERY_LIMIT to u32 too now.

tee-worker/litentry/core/data-providers/src/geniidata.rs
Kailai-Wang: So as per my understanding it's our limit in order to do paged query

Kailai-Wang (approved):

PR #2851, opened 2024-07-01
ws connection read messages fix

Jayanring:
- Read message repeatly when receive a readiness event


Jayanring:
https://linear.app/litentry/issue/P-894/bitacross-worker-stuck-when-handling-concurrent-requests

PR #2852, opened 2024-07-01
Continuous Websocket Message Reading on Readiness Event

Jayanring:
- Read message repeatly when receive a readiness event

### DIFF
--- a/bitacross-worker/core/tls-websocket-server/src/connection.rs
+++ b/bitacross-worker/core/tls-websocket-server/src/connection.rs
@@ -151,11 +151,11 @@ where
 					tungstenite::Error::AlreadyClosed => return Ok((true, false)),
 					_ => {
 						error!(
-						"Failed to read message from web-socket (connection {}): {:?}",
-						self.connection_token.0, e
+							"Failed to read message from web-socket (connection {}): {:?}",
+							self.connection_token.0, e
 						);
-						return Ok((false , false))
-					}
+						return Ok((false, false))
+					},
 				},
 			}
 			trace!("Read successful for connection {}", self.connection_token.0);
@@ -288,7 +288,7 @@ where
 			if connection_state.is_alive() {
 				loop {
 					if !read {
-						break;
+						break
 					}
 					(is_closing, read) = self.read_or_initialize_websocket()?;
 				}
--- a/bitacross-worker/core/tls-websocket-server/src/connection.rs
+++ b/bitacross-worker/core/tls-websocket-server/src/connection.rs
@@ -131,7 +131,7 @@ where
 	/// Read from a web-socket, or initiate handshake if websocket is not initialized yet.
 	///
 	/// Returns a boolean 'connection should be closed'.
-	fn read_or_initialize_websocket(&mut self) -> WebSocketResult<bool> {
+	fn read_or_initialize_websocket(&mut self) -> WebSocketResult<(bool, bool)> {
 		if let StreamState::Established(web_socket) = &mut self.stream_state {
 			trace!(
 				"Read is possible for connection {}: {}",
@@ -147,12 +147,15 @@ where
 						);
 					},
 				Err(e) => match e {
-					tungstenite::Error::ConnectionClosed => return Ok(true),
-					tungstenite::Error::AlreadyClosed => return Ok(true),
-					_ => error!(
+					tungstenite::Error::ConnectionClosed => return Ok((true, false)),
+					tungstenite::Error::AlreadyClosed => return Ok((true, false)),
+					_ => {
+						error!(
 						"Failed to read message from web-socket (connection {}): {:?}",
 						self.connection_token.0, e
-					),
+						);
+						return Ok((false , false))
+					}
 				},
 			}
 			trace!("Read successful for connection {}", self.connection_token.0);
@@ -161,12 +164,11 @@ where
 			self.stream_state = std::mem::take(&mut self.stream_state).attempt_handshake();
 			if self.stream_state.is_invalid() {
 				warn!("Web-socket connection ({:?}) failed, closing", self.connection_token);
-				return Ok(true)
+				return Ok((true, false))
 			}
 			debug!("Initialized connection {} successfully", self.connection_token.0);
 		}
-
-		Ok(false)
+		Ok((false, true))
 	}
 
 	fn handle_message(&mut self, message: Message) -> WebSocketResult<()> {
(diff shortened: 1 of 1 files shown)


bitacross-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -131,7 +131,7 @@ where
 	/// Read from a web-socket, or initiate handshake if websocket is not initialized yet.
 	///
 	/// Returns a boolean 'connection should be closed'.
-	fn read_or_initialize_websocket(&mut self) -> WebSocketResult<bool> {
+	fn read_or_initialize_websocket(&mut self) -> WebSocketResult<(bool, bool)> {
```
Kailai-Wang: In fact, I havne't found a place where it returns `Err`, so maybe `Result` is not necessary? But this can be another topic

bitacross-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -276,14 +278,20 @@ where
...
+		let mut read = true;
 
 		if event.readiness().is_readable() {
 			trace!("Connection ({:?}) is readable", self.token());
 
 			let connection_state = self.maybe_do_tls_read();
 
 			if connection_state.is_alive() {
-				is_closing = self.read_or_initialize_websocket()?;
+				loop {
```
Kailai-Wang: I think it's better to put the loop within `read_or_initialize_websocket` - basically looping over `read_message`

Kailai-Wang: ~~Will the loop cause `read_message` to block if nothing is in the socket's read buffer, or will it return `WouldBlock`?~~

Kailai-Wang (commented):
Good catch, thanks!

Should identity worker (in `tee-worker/`)  be changed too?

bitacross-worker/core/tls-websocket-server/src/connection.rs
Kailai-Wang: It uses mio's underlying tcp_stream so it should be non-blocking already, but please handle `WouldBlock` explicitly - e.g. ignore it as it's not an error

bitacross-worker/core/tls-websocket-server/src/connection.rs
BillyWooo: Yeah, there is no Err. But it looks like code is from upstream.

BillyWooo:
> Should identity worker (in tee-worker/) be changed too?

I think it should.

bitacross-worker/core/tls-websocket-server/src/connection.rs
Jayanring: Code is from [here](https://github.com/integritee-network/worker)

bitacross-worker/core/tls-websocket-server/src/connection.rs
Jayanring: Updated

kziemianek:
ws server is single threaded so draining connection will block it until it's finished.
This might be good enough for `bitacross-worker` but not for `tee-worker`.

However `Poll::Level` is used so mio events should be emitted as long as there is data in buffer and draining strategy should be used only in case of `Poll::Edge`. Currenlty we don't know why it behaves this way and this looks only like a workaround.

kziemianek (commented):
ws server is single threaded so draining connection will block it until it's finished.
This might be good enough for `bitacross-worker` but not for `tee-worker`.

However `Poll::Level` is used so mio events should be emitted as long as there is data in buffer and draining strategy should be used only in case of `Poll::Edge`. Currenlty we don't know why it behaves this way and this looks only like a workaround.

Kailai-Wang:
There _could_ be two possible reasons:
- the tungstenite `read_message` somehow reads all messages in one go and/or clear the socket reading state, so that mio thinks there's nothing left
- the re-register of mio is buggy (delay? race condition?) and doesn't trigger events even if there're still remaining data that were sent quickly after the first message

Either way - can this be reproduced?
If so we could try to dig further and find out when the first message is handled, if there's anything left in the socket just before re-registering

bitacross-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -131,31 +133,48 @@ where
...
+							tungstenite::Error::ConnectionClosed
+							| tungstenite::Error::AlreadyClosed => is_closing = true,
+							tungstenite::Error::Io(e)
+								if matches!(e.kind(), ErrorKind::WouldBlock) => {}, // no message to read
+							_ => {
+								error!(
+									"Failed to read message from web-socket (connection {}): {:?}",
+									self.connection_token.0, e
+								);
+							},
```
Kailai-Wang: Shall we close the connection too in this case? 🤔 

So basically we only allow `WouldBlock` - for any other error => we close the connection

bitacross-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -131,31 +133,48 @@ where
...
 				web_socket.can_read()
 			);
-			match web_socket.read_message() {
-				Ok(m) =>
-					if let Err(e) = self.handle_message(m) {
-						error!(
-							"Failed to handle web-socket message (connection {}): {:?}",
-							self.connection_token.0, e
-						);
+
```
kziemianek, line 143: Could you please put a comment that this is just a workaround and final solution will be applied in P-907 ?

Kailai-Wang:
Had a discussion with Kasper - we'll keep this workaround only for bitacross-worker.

Root cause is not exaclty known but it's likely the old `mio` crate is buggy - we see messages being kept in the read buffer in the socket but no mio event is triggered.

bitacross-worker/core/tls-websocket-server/src/connection.rs
```diff
@@ -131,31 +133,52 @@ where
...
+			// Looping over 'read_message' is merely a workaround for the unexpected behavior of mio event triggering.
+			// Final solution will be applied in P-907.
+			loop {
+				match web_socket.read_message() {
+					Ok(m) => messages.push(m),
+					Err(e) => {
+						match e {
+							tungstenite::Error::ConnectionClosed
+							| tungstenite::Error::AlreadyClosed => is_closing = true,
```
BillyWooo: maybe apply the similar changes from #2854

kziemianek (approved):

PR #2856, opened 2024-07-03
Deprecate `ScheduledEnclave` and introduce `AuthorizedEnclave`

Kailai-Wang:
### Context

This PR introduces a new `AuthorizedEnclave` aiming to supersede (and simplify) the use of the old `ScheduledEnclave`. It serves as a list of whitelisted MRENCLAVEs without any sidechain block number.

So the enclave upgrade flow becomes:
1. call `teebag.forceAddAuthorizedEnclave` to add the new MRENCLAVE
2. migrate shard for the worker
3. restart worker

We don't have to wait for anything. cc @BillyWooo @m1iktea 

The old `ScheduledEnclave` is kept for vc verification purpose. It will be discarded once we use litentry as the parentchain.

The `AuthorizedEnclave` is a bounded list, we only keep the latest `MaxAuthorizedEnclave` entries (5 for now).

More context and decison-making in the linked issue.


silva-fj (approved):
LGTM

Kailai-Wang:
I assume this is not a breaking change for the sdk as it doesn't use `ScheduledEnclave` or `AuthorizedEnclave` for vc verification

Kailai-Wang:
cc @jonalvarezz

jonalvarezz:
for the record, from the client side, the `scheduledEnclave` storage is no longer used for vc-validation. The validation logic instead trys to fetch the `enclaveRegistry` used to issue the vc and if not found (this is VC with no `parachainBlockNumber` data) it looks in all entries of `enclaveRegistry`. see https://github.com/litentry/client-sdk/pull/34

PR #2861, opened 2024-07-04
reuse websocket connection for ceremony rounds

kziemianek:
Workers will keep websocket connection open to each other and reuse it in order to perform musig2 round's communications.

I checked performance with script creating 100 ceremonies, here are times to complete:

**Previous**
ws:
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.236s
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.202s
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.120s

wss:
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	22.827s
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	21.951s
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	22.989s

**Current**
ws:
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.116s
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.122s
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.095s

wss:
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.056s
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.089s
ok  	github.com/litentry/brc20/ebrelayer/relayer/signer/tee	1.069s

**Times have significantly improved for wss connections.**


Jayanring (approved):

Kailai-Wang:
Two dumb questions:
- why not use `ws`? Either directly or via nginx
- Does `skip-ra` matter here? Will musig2 suceed even without `skip-ra`? If the answer is yes - how does the client verify the server's cert?

kziemianek:
Do not merge before release.

kziemianek:
> Two dumb questions:
> 
> * why not use `ws`? Either directly or via nginx
> * Does `skip-ra` matter here? Will musig2 suceed even without `skip-ra`? If the answer is yes - how does the client verify the server's cert?

* This implementation ensures almost the same performance for both ws and wss - so there is no need to worry about performance implication while choosing between ws or wss - the final decision needs to be always made while desiginig deployment environment.

* There is no such verification like in state provisioning server/client.

PR #2862, opened 2024-07-05
Adding  force-migrate-shard flag to run worker command

silva-fj:
### Context

This PR adds a new flag to migrate shard automatically before the worker is initialised


tee-worker/service/src/setup.rs
```diff
@@ -97,6 +97,17 @@ mod needs_enclave {
 		}
 	}
 
+	pub(crate) fn force_migrate_shard(enclave: &Enclave, &new_shard: &ShardIdentifier) {
+		match enclave.force_migrate_shard(new_shard.encode()) {
```
Kailai-Wang, line 101: Shall we exit the program if migration fails?

tee-worker/service/src/setup.rs
```diff
@@ -126,6 +137,20 @@ mod needs_enclave {
 	}
 }
 
+/// backs up shard directory and restores it after cleaning shards directory
+pub(crate) fn remove_old_shards(root_dir: &Path, new_shard_name: &str) -> ServiceResult<()> {
```
Kailai-Wang: Same - we'd better quit if it returns an error

Kailai-Wang (commented):
Thank you! Looks good overall, I'm just wondering if we should abort early if migration fails

tee-worker/service/src/setup.rs
```diff
@@ -126,6 +137,20 @@ mod needs_enclave {
...
+	let shard_backup = root_dir.join("shard_backup");
+	let shard_dir = root_dir.join(SHARDS_PATH).join(new_shard_name);
+
+	fs::rename(shard_dir.clone(), shard_backup.clone()).expect("Failed to backup shard");
+	remove_dir_if_it_exists(root_dir, SHARDS_PATH)?;
+	fs::create_dir_all(root_dir.join(SHARDS_PATH)).expect("Failed to create shards directory");
+	fs::rename(shard_backup, root_dir.join(SHARDS_PATH).join(shard_dir))
```
Kailai-Wang: Is this wrong?

`shard_dir` seems to have its base dir already

bitacross-worker/core-primitives/stf-state-handler/src/state_handler.rs
```diff
@@ -236,6 +236,21 @@ where
...
 		self.reset(state, &new_shard)
 	}
+
+	fn force_migrate_shard(&self, new_shard: ShardIdentifier) -> Result<Self::HashType> {
+		if self.shard_exists(&new_shard)? {
+			let (_, state_hash) = self.load_cloned(&new_shard)?;
+			return Ok(state_hash)
+		}
+		let old_shard = match self.list_shards()? {
+			shards if shards.len() == 1 => shards[0],
```
kziemianek, line 246: If it's the same shard (the worker was just restarted or code change was not inside enclave-runtime) does it still make sense to migrate it ? I think we should skip shard migration in such case.

tee-worker/service/src/setup.rs
Kailai-Wang: I tested it a bit, it seems `root_dir.join(SHARDS_PATH).join(shard_di)` will remove the duplicated path prefix automatically, but still it might be confusing - so I prefer using shard_dir directly

silva-fj: Good catch, it produces the same result that's why the test is ok:

```
println!(">> {:?}", root_dir.join(SHARDS_PATH).join(shard_dir.clone()));
println!(">> {:?}", root_dir.join(SHARDS_PATH).join(new_shard_name));

---- setup::tests::test_remove_old_shards stdout ----
>> "/home/franjs/projects/litentry-parachain/tee-worker/service/test_backup_shard/shards/new_shard"
>> "/home/franjs/projects/litentry-parachain/tee-worker/service/test_backup_shard/shards/new_shard"
```
I'll update it for clarity

bitacross-worker/core-primitives/stf-state-handler/src/state_handler.rs
Kailai-Wang: It does skip it - see a few lines above.

If we fall into this branch it means it has only 1 shard, but a different shard to the current one

kziemianek: True true, I was misleaded by log - it keeps on saying migration was succesfull.

Kailai-Wang: Ah it's also a good point - so log saying "migration successful" can actually mean "migration was skipped because the shard matches the current mrenclave"

Kailai-Wang: It can be improved later though - either we introduce additional `MigrationResult` enum, or simply add a log saying migration was skipped

Kailai-Wang (approved):
Thanks!

kziemianek (approved):

PR #2863, opened 2024-07-05
Add remove_vault in pallet-bitacross

Kailai-Wang:
### Context

As topic. It adds `remove_vault` to remove unneeded vaults in the pallet


BillyWooo (approved):
👍

kziemianek (changes requested):
If any keys are removed from vault we should remove them from bitacross-workers aswell through event handler.

Kailai-Wang:
> If any keys are removed from vault we should remove them from bitacross-workers aswell through event handler.

This is a good point - what how should this be handled exactly?
Is this registry used by the client only or by workers too?

Kailai-Wang:
As discussed, let's firstly make sure that we have some way to remove unneeded entries in the vault, then we'll take a deeper look how we shall improve the on-chain and worker logic

kziemianek (approved):

PR #2865, opened 2024-07-06
Refactoring force-migrate-shard

silva-fj:
This PR introduces the following changes:

- Now the new shard hex value needs to be pass along the `force-migrate-shard` flag.
- `mrenclave` command also prints the hex value.
- The enclave is re-initialized after force-migrate-shard.


silva-fj:
After discussion we want to take a different approach

PR #2866, opened 2024-07-06
Improve shard migration

silva-fj:
This PR introduces the following changes:

- `force-migrate-shard` is removed in favor `migrate-shard`, as there are some issues to migrate in one go with run command.
- improve `migrate-shard` to get new and old shard automatically.
- improve `mrenclave` command to also print hex value.


Kailai-Wang (approved):
Thanks - I updated CLI helper doc too

PR #2869, opened 2024-07-07
Support DCAP RA

Kailai-Wang:
### Context

Tested on an OVH host with:
```
RA_METHOD=dcap SGX_PRODUCTION=1 make
```

using Intel URL in `/etc/sgx_default_qcnl.conf`:
```
"pccs_url": "https://api.trustedservices.intel.com/sgx/certification/v4/"
```

Not making dcap as default yet, because our current workers still use IAS.


Kailai-Wang:
Next step would be testing it on an Azure host - but I don't expect any problems

pallets/teebag/src/types.rs
```diff
@@ -54,8 +54,8 @@ pub enum OperationalMode {
 #[derive(Encode, Decode, Default, Clone, Copy, PartialEq, Eq, RuntimeDebug, TypeInfo)]
 pub enum DcapProvider {
 	#[default]
-	MAA,
 	Intel,
+	MAA,
```
BillyWooo, line 58: Maybe we should remove MAA here?

BillyWooo (approved):
Thanks very much. LGTM.

silva-fj (approved):

PR #2870, opened 2024-07-07
feat: add a parachain event for completion of contract sync

felixfaisal:
This adds a callback extrinsic to signal the parachain that the assertion has been succesfully stored and if failure then it removes the assertion from the parachain storage so that it can be retried.


Kailai-Wang (approved):
Looks good from my side!

PR #2871, opened 2024-07-08
Add rpc method for checking `sign_bitcoin`

kziemianek:
Adds a RPC method to trigger `sign_bitcoin` on worker instance without having an access to relayer's private key.
`sign_bitcoin` triggered with `check` flag will return `true/false` representing musig2 signature verification result instead of signature itself. All ceremonies started with `check` flag uses the same fixed payload.

The general purpose of this method is to have a way to check system condition.


bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
```diff
@@ -136,10 +136,13 @@ pub struct MuSig2Ceremony<AK: AccessKey<KeyType = SchnorrPair>> {
 	second_round: Option<musig2::SecondRound<SignaturePayload>>,
 	ticks_left: u32,
 	agg_key: Option<PublicKey>,
+	// indicates whether it's check run - signature verification result is returned instead of signature
+	check: bool,
```
Kailai-Wang: Does `dry_run` sound better? Or at least `check_run` : ) 

`check` can be very ambiguous

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -157,6 +162,37 @@ where
...
+		let request = prepare_check_sign_bitcoin_request(
+			signing_key_repository_cloned.as_ref(),
+			shielding_key_cloned_2.as_ref(),
+			ocall_api_cloned.as_ref(),
+		);
+		async move {
+			if let Ok(request) = request {
+				let params = Params::Array(vec![jsonrpc_core::Value::String(request.to_hex())]);
+				let json_value = match request_bit_across_inner(params).await {
+					Ok(value) => value.to_hex(),
```
Kailai-Wang, line 176: So we don't have a determinisitic result, we simply sign `[0u8; 32]` - and if it returns _some_ value it means OK?

Kailai-Wang (commented):
Thanks!

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
kziemianek: It can have 3 results:

1. Ok(true) - means ceremony **was sucesfully completed** (Ok) and signature verification **passed** (true)
2. Ok(false) - means ceremony **was sucesfully completed** (Ok) and signature verification **did not pass** (false) 
3. Err(error) - means ceremony **was not sucesfully** completed

Kailai-Wang (approved):
Thanks!

We should let simple and relayer friends know about this

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
```diff
@@ -392,11 +397,22 @@ impl<AK: AccessKey<KeyType = SchnorrPair>> MuSig2Ceremony<AK> {
 					info!("Message {:?}", message);
 
 					info!("Verification result: ");
-					match verify_single(self.agg_key.unwrap(), signature, message) {
-						Ok(_) => info!("OK!"),
-						Err(_) => info!("NOK!"),
+					let result = match verify_single(self.agg_key.unwrap(), signature, message) {
+						Ok(_) => {
+							info!("OK!");
```
BillyWooo: maybe a bit picky, but can we push some context info here for log, as anyway we output log.
Same for the below "NOK".

kziemianek: i'll remove it

PR #2879, opened 2024-07-09
P-897 Pallet Staking: EVM Decimal Migration (Code view only)

wangminqi:
**Do not merge this PR.**
The PR create for code view purpose only.
The official PR is on p-895


runtime/litentry/src/migration/mod.rs
```diff
@@ -0,0 +1,748 @@
...
+		.expect("Storage query fails: ParachainStaking CandidatePool");
+		for bond in stored_data.0.iter_mut() {
+			bond.amount = bond.amount.saturating_mul(DECIMAL_CONVERTOR.into());
+		}
+		<CandidatePool<T>>::put(stored_data);
+		let weight = T::DbWeight::get();
+		frame_support::weights::Weight::from_parts(0, weight.write + weight.read)
+	}
+
+	pub fn repalce_delayed_payouts_storage() -> frame_support::weights::Weight {
```
silva-fj:
```suggestion
	pub fn replace_delayed_payouts_storage() -> frame_support::weights::Weight {
```

runtime/litentry/src/migration/mod.rs
```diff
@@ -0,0 +1,748 @@
...
+			new_delayed_payout.total_staking_reward =
+				new_delayed_payout.total_staking_reward.saturating_mul(DECIMAL_CONVERTOR.into());
+
+			<DelayedPayouts<T>>::insert(round, new_delayed_payout)
+		}
+		let weight = T::DbWeight::get();
+		migrated_count.saturating_mul(weight.write + weight.read)
+	}
+
+	pub fn repalce_staked_storage() -> frame_support::weights::Weight {
```
silva-fj:
```suggestion
	pub fn replace_staked_storage() -> frame_support::weights::Weight {
```

silva-fj (approved):
LGTM

kziemianek (approved):
I would suggest migrating also `AtStake` and `InflationConfig` just to have consistent decimal places everywhere.

The rest of the code looks fine.

Kailai-Wang:
Thanks! But we need to wait until other pallets' migration is done and merge them all at once?

wangminqi:
> Thanks! But we need to wait until other pallets' migration is done and merge them all at once?

Yes. This branch is for code view only. The true merge branch is on P-895. All decimal related migration PR will be merged to that branch first. Then we merge P-895 into DEV once we are ready.

wangminqi:
This branch has been merged to P-895.

runtime/litentry/src/migration/parachain_staking.rs
Kailai-Wang, line 1: I generally feel the migration is too verbose.

We should run as few loops as possible to minimise the calculation expense (weight).

I have an example here in this commit:
https://github.com/litentry/litentry-parachain/commit/cfaa499341b136e8f06d16fa42544aef2a970e98

There might be better ways (e.g. does `mutate` or `try_mutate` work out-of-box here?) - you could refer to substrate code sample.

Also I'm not sure about the effect of `assert!` here - does it panic if it's false? The migration should not panic, otherwise the chain has a chance to stall.

Kailai-Wang:
[1]
I generally feel the migration is too verbose.

We should run as few loops as possible to minimise the calculation expense (weight). We do have only 5000 accounts on Litentry, but many more storages.

I have an example here in this commit: [cfaa499](https://github.com/litentry/litentry-parachain/commit/cfaa499341b136e8f06d16fa42544aef2a970e98)

There might be better ways (e.g. does mutate or try_mutate work out-of-box here?) - please refer to substrate code sample - there's plenty.

[2]
Also I'm not sure about the effect of `assert!`  - does it panic if it's false? Same applies to other code that can panic.
The migration should not panic, otherwise the chain has a chance to stall.

See also the question that I asked before:
https://substrate.stackexchange.com/questions/4534/what-happens-if-on-runtime-upgrade-during-migration-fails

[3]
We should check the real on-chain storage before writing the code.
If the storage has no entry, there's no need to perform the migration ofc. It saves us time and reduces the runtime computation needs.

wangminqi:
> [1] I generally feel the migration is too verbose.
> 
> We should run as few loops as possible to minimise the calculation expense (weight). We do have only 5000 accounts on Litentry, but many more storages.
> 
> I have an example here in this commit: [cfaa499](https://github.com/litentry/litentry-parachain/commit/cfaa499341b136e8f06d16fa42544aef2a970e98)
> 
> There might be better ways (e.g. does mutate or try_mutate work out-of-box here?) - please refer to substrate code sample - there's plenty.
> 
> [2] Also I'm not sure about the effect of `assert!` - does it panic if it's false? Same applies to other code that can panic. The migration should not panic, otherwise the chain has a chance to stall.
> 
> See also the question that I asked before: https://substrate.stackexchange.com/questions/4534/what-happens-if-on-runtime-upgrade-during-migration-fails
> 
> [3] We should check the real on-chain storage before writing the code. If the storage has no entry, there's no need to perform the migration ofc. It saves us time and reduces the runtime computation needs.

Agreed. Let's remove these unnecessary assertion codes on branch-895.

wangminqi:
This branch is finished. No more maintaining.

PR #2882, opened 2024-07-09
fix: request-vc in cli with ii and stf fails to decode

felixfaisal:
The command used in the issue description did not include the `-d` flag indicating direct invocation which leads to a decoding issue. I felt a bit icky to have it default to true so I added a condition to decode and give a warning.


Kailai-Wang (approved):
Makes sense

kziemianek (approved):

PR #2883, opened 2024-07-09
Fixing state provisioning without skip_ra flag

silva-fj:
On `enclave_request_state_provisioning` an api for RA (remote attestation) is being called to get the enclave quote data, this is needed when using DCAP RA, This api requites to setup PCCS, which we don't have because we currently use IAS RA. So the call to `qe_get_target_info` fails as it cannot reach the service.

More context in the issue.


Kailai-Wang (approved):
Good catch, thanks!

However, this affects client's side request.
`skip-ra` is used on server side too, which will cause server to skip client RA's check.

Have you checked it?
Or to put it in another way - have you verified that state provisioning without `skip-ra` in your branch does work? Ideally with HW mode

BillyWooo (approved):
Nice catch. 👍

silva-fj:
> However, this affects client's side request. `skip-ra` is used on server side too, which will cause server to skip client RA's check.

Yes, this is why the check is not based on `skip_ra`, we still pass it down to `create_ra_report_and_signature` so the functionality will be the same when using the flag. 

The server side already has the same logic (only fetch when dcap is enabled)

https://github.com/litentry/litentry-parachain/blob/2cf19c239b38fb9e881bd605b49e1e3d82603cba/bitacross-worker/service/src/main_impl.rs#L138-L153

> Have you checked it? Or to put it in another way - have you verified that state provisioning without `skip-ra` in your branch does work? Ideally with HW mode

I'll check

silva-fj:
I tried to verify it but I don't have the `RA_API_KEY_FILE` in my environment so the request fails:

```bash
[DEBUG itp_attestation_handler::attestation_handler]     [Enclave] resp_string = HTTP/1.1 401 Access Denied
    Warning: 299 api.trustedservices.intel.com "The version of Attestation Service for Intel(R) Software Guard Extensions API you are using is scheduled to have restricted access after September 29, 2024, and will reach end of life April 2, 2025. For details and alternatives reference https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/resources/sgx-ias-using-epid-eol-timeline.html"
    Content-Length: 143
    Content-Type: application/json
    WWW-Authenticate: AzureApiManagementKey realm="https://api.trustedservices.intel.com/sgx/dev/attestation",name="Ocp-Apim-Subscription-Key",type="header"
    Date: Wed, 10 Jul 2024 08:43:38 GMT
    Connection: close

    { "statusCode": 401, "message": "Access denied due to invalid subscription key. Make sure to provide a valid key for an active subscription." }
```

how/where can I test this? @Kailai-Wang

Kailai-Wang:
@silva-fj I believe you'll need a real spid and key - but we are going to switch to dcap soon anyway :)

PR #2885, opened 2024-07-11
validation_data: support prettified challenge codes

jonalvarezz:
### Context

The challenge code is a raw payload in hex representation. However, Solana's `signMessage` doesn't have good support for non-utf-8 messages: Phantom wallet will refuse signing the message and OKX wallet will mess the encoding. 

These changes extend the validation_data logic to support both raw challenge code payload and prettified messages with a prefixed `Token: ` message.


Kailai-Wang (approved):
Looks ok from my side:

can you please:
1. add some comments in code showing why we need the message with prefix `Token: `?
2. ideally can get this covered in test

jonalvarezz:
cc @0xverin

PR #2894, opened 2024-07-15
restore connections to other musig2 signers

kziemianek:
This PR changes the way connections are handled to other musig2 signers:

1. Connections are initialized lazily - they are created right after ceremony creation
2. In case of request sending problem connection will be recreated and request will be sent again

There are no tests for failed requests retry because musig2 runner is undergoing heavy changes.


Jayanring:
It needs updating

Jayanring (changes requested):
This PR needs updating

Kailai-Wang:
@kziemianek can you please fix the PR on top of the recent multi-thread commit?

bitacross-worker/bitacross/core/bc-enclave-registry/src/lib.rs
```diff
@@ -251,4 +258,11 @@ impl EnclaveRegistryLookup for EnclaveRegistry {
 		let registry = self.registry.read().unwrap();
 		registry.iter().map(|(k, v)| (k.clone(), v.clone())).collect()
 	}
+
+	#[cfg(feature = "sgx")]
```
Kailai-Wang: A stupid question: why do we have `sgx` and `std` separately? They seem identical - do I overlook anything

bitacross-worker/bitacross/core/bc-musig2-event/src/lib.rs
```diff
@@ -206,25 +221,75 @@ pub fn process_event<OCallApi, SIGNINGAK, SHIELDAK, Responder>(
...
+	ceremony_registry: Arc<RwLock<CeremonyRegistry<BKR>>>,
+) where
+	ECL: EnclaveRegistryLookup,
+	BKR: AccessKey<KeyType = SchnorrPair>,
+{
+	loop {
+		let client = peers_map.lock().unwrap().get(&signer_id).cloned();
+		if let Some(mut client) = client {
+			if let Err(e) = client.send(&request) {
+				error!("Could not send request to signer: {:?}, reason: {:?}", signer_id, e);
```
Kailai-Wang, line 221: Will it spam the logs?

Kailai-Wang (approved):
Looks good to me

bitacross-worker/bitacross/core/bc-musig2-event/src/lib.rs
kziemianek: It can, currently 20 log entries per second at worst (20 threads).

We can either:
* change interval 
* have some kind of error deduplication - so we do not log the same error twice for request

In normal conditions this should not be logged at all.

bitacross-worker/bitacross/core/bc-enclave-registry/src/lib.rs
kziemianek: Thanks!

Usually we have different impls - but in this case they are identical so I've merged them

bitacross-worker/bitacross/core/bc-musig2-event/src/lib.rs
kziemianek: I've changed interval to 5 s.

Let's add deduplication/other impl preventing from logs being flooded in separate PR if we decide we need it

PR #2895, opened 2024-07-15
P-901: replace orml with pallet_assets

wangminqi:
Replace orml_token with pallet_assets.
Precompile impl of pallet_assets on Rococo is also included.

pallet_bridge and pallet_staking precompile updated since the precompile-utils change.

Should including runtime upgrade of storage clean. (do not need anymore, since only rococo which will be depreciated soon)


wangminqi:
Warning: the precompile utils pallet is not available in official frontier until 0.9.43. The one we use is a temporary version downgrade to 0.9.42.

wangminqi:
Disable integration ts test on precompile, it is just on Rococo.
Let's fix it after decimal migration solved.

precompiles/bridge-transfer/src/lib.rs
```diff
@@ -1,103 +1,54 @@
...
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
-//
-// Litentry is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-//
-// You should have received a copy of the GNU General Public License
-// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
```
Kailai-Wang: We still need this

Kailai-Wang (approved):
Thanks!

precompiles/bridge-transfer/src/lib.rs
wangminqi: Added back

PR #2896, opened 2024-07-15
Add pallet-score-staking

Kailai-Wang:
### Context

Initial version - refer to the comments in the pallet code for more.


pallets/score-staking/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			Round::<T>::put(r);
+			weight = weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+
+			// 2. calculate payout
+			let round_reward: BalanceOf<T> = (T::YearlyInflation::get() * T::YearlyIssuance::get() /
+				YEARS.into()) * Self::round_config().interval.into();
+
+			let total_stake = ParaStaking::Pallet::<T>::total();
+			let total_score =
+				Scores::<T>::iter_values().fold(0u32, |a, b| a.checked_add(b.score).unwrap_or(a));
```
wangminqi: Maybe instead of iterating, have an extra storage value to keep trace of total score?

pallets/score-staking/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			Ok(Pays::No.into())
+		}
+
+		// Intentionally use `Identity` type to lower the hurdle of mapping to the
+		// desired substrate account as it's handled on-chain instead of by client.
+		//
+		// Subject to requirement change though
+		#[pallet::call_index(4)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn update_score(
```
wangminqi, line 343: I think they will need a batch updating method too.

wangminqi (changes requested):

pallets/score-staking/src/lib.rs
Kailai-Wang: they are free to use native `utility.batch`

pallets/score-staking/src/lib.rs
Kailai-Wang: I can do it

Kailai-Wang: Improved

silva-fj (commented):
It looks good to me in general but I'm not sure if the unbounded iteration in the `on_initialize` hook would be a problem at some point

wangminqi (dismissed):
LGTM

Kailai-Wang:
> It looks good to me in general but I'm not sure if the unbounded iteration in the `on_initialize` hook would be a problem at some point

This can be a good point.

Maybe we should use `boundedVec` for scores for now. Later if we want, we can change it to top X candidates

wangminqi (commented):
Seems formula require some change. TBC

silva-fj (approved):

PR #2900, opened 2024-07-16
fix: P-927 refactor brc20 contract to configure tokens and networks 

wli-pro:
in a unified place, remove redundant network loops in function queryBalance in TokenQueryLogic, move url config to dp client

### Labels

### Testing Evidences

I will have follow up PR to test the new bitcoin network type via cli tests.

### DIFF
--- a/tee-worker/litentry/core/assertion-build/src/dynamic/contracts/token_holding_amount/TokenQueryLogic.sol
+++ b/tee-worker/litentry/core/assertion-build/src/dynamic/contracts/token_holding_amount/TokenQueryLogic.sol
@@ -23,26 +23,14 @@ import "../libraries/Utils.sol";
 import { TokenHoldingAmount } from "./TokenHoldingAmount.sol";
 import { NoderealClient } from "./NoderealClient.sol";
 import { GeniidataClient } from "./GeniidataClient.sol";
+
 abstract contract TokenQueryLogic is TokenHoldingAmount {
-	mapping(uint32 => string) internal networkUrls;
-	mapping(uint32 => bool) private queriedNetworks;
 	mapping(string => mapping(uint32 => string)) tokenAddresses;
 	mapping(string => string) internal tokenBscAddress;
 	mapping(string => string) internal tokenEthereumAddress;
 	mapping(string => uint32[]) internal tokenNetworks;
 
-	constructor() {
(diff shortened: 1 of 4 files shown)


tee-worker/litentry/core/assertion-build/src/dynamic/contracts/token_holding_amount/TokenQueryLogic.sol
```diff
@@ -57,57 +45,48 @@ abstract contract TokenQueryLogic is TokenHoldingAmount {
...
-								secrets,
-								_tokenContractAddress,
-								identityString
-							);
-						if (success) {
-							totalBalance += balance;
-						}
-					}
-					// Mark this network as queried
-					queriedNetworks[networks[i]] = true;
```
0xverin, line 97: Should `queriedNetworks` mapping be removed here? If the network is bsc and ethereum, what I mean is when execute param is :
```
identities = [

{

identity_type: 4,

value: "0xf7764F15Fd4F695fAD8dDd3e84F5293E6866FEFE",

networks: [7, 8],

},

]
```
in this case, will it be queried twice here?

wli-pro: As you said, same identity could has different networks, but we need sum the token balance from different network.
e.g.
1) There has an evm identity A has 2 networks ethereum and bsc.
2) We need generate TokenHoldingAmount VC for token TRX.
3) The total balance of token TRX for identity A is: (TRX token balance in network ethereum) + (TRX token balance in network bsc).

tee-worker/litentry/core/assertion-build/src/dynamic/contracts/token_holding_amount/TokenQueryLogic.sol
```diff
@@ -57,57 +45,48 @@ abstract contract TokenQueryLogic is TokenHoldingAmount {
 			.identityToString(network, identity.value);
 
 		if (identityToStringSuccess) {
-			string memory url;
-			uint32[] memory networks = tokenNetworks[tokenName];
 			uint256 totalBalance = 0;
 
-			for (uint32 i = 0; i < networks.length; i++) {
```
0xverin, line 64: Removing the for loop here, it should not be possible to query all network addresses of a token, I guess.

0xverin (approved):

PR #2901, opened 2024-07-16
P 893 enhance bitarcoss worker ceremony processing with multithreading

Jayanring:
This PR primarily about:
1. Reducing shared data access
2. Implementing more fine-grained controls over shared data.
3. Separating timeout, ceremony processing and network requests sending logics.
4. Utilizing multithreading for task processing and network requests sending.
5. Optimizing ceremony data sharing logic, now it sends response back once handled by websocket server


bitacross-worker/bitacross/core/bc-task-receiver/src/lib.rs
```diff
@@ -123,51 +146,59 @@ impl<
...
+			ceremony_command_tmp,
+			responder,
 		}
 	}
 }
 
 #[allow(clippy::type_complexity)]
-pub fn run_bit_across_handler_runner<SKR, EKR, BKR, S, H, O, RRL, ERL, SRL>(
-	context: Arc<BitAcrossTaskContext<SKR, EKR, BKR, S, H, O, RRL, ERL, SRL>>,
+pub fn run_bit_across_handler_runner<SKR, SIGNINGAK, EKR, BKR, S, H, O, RRL, ERL, SRL, Responder>(
```
kziemianek: Previously this function was responsible just for general task handling which was fine in my opinion.
Now it's getting coupled to musig2 ceremonies - can we avoid that ?

Previously:

* `bitacross-worker/litentry/core/direct-call/src/handler/sign_bitcoin.rs`
* `bitacross-worker/litentry/core/direct-call/src/handler/nonce_share.rs`
* `bitacross-worker/litentry/core/direct-call/src/handler/partial_signature_share.rs`
* `bitacross-worker/litentry/core/direct-call/src/handler/kill_ceremony.rs`

Where the only units from direct-call processing aware of musig ceremonies

Jayanring: The relationship between these modules has changed. Now, `bc-task-receiver` holds a thread that manages all ceremonies. I initially intended for` bc-task-receiver` to only receive tasks. However, since we still need a place to run a manager thread and have `bc-task-receiver` send tasks there, why not just let `bc-task-receiver `handle it directly?
Maybe I'd better change their name:
`bc-task-receiver` -> `bc-task-processor`
`bc-musig2-runner` -> `bc-musig2-event`

bitacross-worker/enclave-runtime/src/test/tests_main.rs
```diff
@@ -159,7 +159,7 @@ pub extern "C" fn test_main_entrance() -> size_t {
 		itc_parentchain::light_client::io::sgx_tests::sealing_creates_backup,
 
 		// test musig ceremony
-		bc_musig2_ceremony::sgx_tests::test_full_flow_with_3_ceremonies,
+		// bc_musig2_ceremony::sgx_tests::test_full_flow_with_3_ceremonies,
```
kziemianek: This is very important test, we must keep it.

Jayanring: I removed some test cause processing logic changed, I will try to restore this test

bitacross-worker/bitacross/core/bc-task-receiver/src/lib.rs
kziemianek: I would prefer to keep them separated - not they are mixed - for example it will be hard to extend in the future. 
I assume we sacrifice readability for performance ?

Renaming things won't change much in this case.

kziemianek: Could you encapsulate added logic in `handle_ceremony_command function` ?

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
```diff
@@ -28,74 +28,60 @@ use std::{format, string::String, sync::Arc};
...
-	IncorrectRound,
-	SecondRoundFinalizationError,
+	RoundFinalizationError,
 }
 
 // events come from outside and are consumed by ceremony in tick fn
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, Clone)]
 pub enum CeremonyCommand {
+	Init,
```
kziemianek: I'm wondering if we can avoid this command. Currently it just produces event - i think event should be created alongside ceremony.

bitacross-worker/bitacross/core/bc-musig2-runner/src/lib.rs
```diff
@@ -1,429 +0,0 @@
...
-	ClientFactory: RpcClientFactory + Send + Sync + 'static,
-	AK: AccessKey<KeyType = SchnorrPair> + Send + Sync + 'static,
-	ER: EnclaveRegistryLookup + Send + Sync + 'static,
-	OCallApi: EnclaveAttestationOCallApi + 'static,
-	SIGNINGAK: AccessKey<KeyType = ed25519::Pair> + Send + Sync + 'static,
-	SHIELDAK: AccessPubkey<KeyType = Rsa3072PubKey> + Send + Sync + 'static,
-	Responder: SendRpcResponse<Hash = H256> + 'static,
-{
-	let (responses_sender, responses_receiver) = channel();
-	std::thread::spawn(move || {
```
kziemianek, line 84: What do you think about keeping this thread and processing ceremony commands sent from rpc handlers instead of ticking ceremonies ?

```
		while let Ok(command) = musig2_commands.recv() {
			// there is a command so get ceremony from registry and execute it, eventually process all events
		}
```

Jayanring: I've already removed code about ticking, and this thread was kept actually, but merged into `bc-task-receiver` (now `bc-task-processor`), this module `bc-musig2-runner` (now `bc-musig2-event`) is not `runner` anymore.

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
Jayanring: I added this command to unify processing flow, now every request will generate a command, and all events are generated by these commands.
`DirectCall::SignBitcoin` -> `CeremonyCommand::Init` -> `CeremonyEvent::FirstRoundStarted`
`DirectCall::NonceShare` -> `CeremonyCommand::SaveNonce` -> `CeremonyEvent::SecondRoundStarted`
`DirectCall::PartialSignatureShare` -> `CeremonyCommand::SavePartialSignature` -> `CeremonyEvent::CeremonyEnded`

bitacross-worker/enclave-runtime/src/test/tests_main.rs
Jayanring: Done

bitacross-worker/bitacross/core/bc-task-receiver/src/lib.rs
Jayanring: OK

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
kziemianek: You made it work but it doesn't fit well after your code changes. 

I think you should made further changes:

1. remove command processing from ceremony
2. enrich `InitCeremony` with all fields required for ceremony creation
3. make `sign_bitcoin` direct call handler dumb, so it will only create and return `InitCeremony` command
4. create cereomony during `InitCeremony` command processing in `bc-task-processor` 

This will lead to more consistent design.

bitacross-worker/bitacross/core/bc-task-processor/src/lib.rs
```diff
@@ -0,0 +1,563 @@
...
+					Ok(client) => {
+						peers_map.lock().unwrap().insert(*identity.as_ref(), client);
+					},
+					Err(e) => error!("Could not connect to peer {}, reason: {:?}", address, e),
+				}
+			}
+		});
+
+	// process commands and events
+	let mut commands_to_process = vec![command];
```
kziemianek, line 365: if there is always single command to process, simplify this

kziemianek (changes requested):

bitacross-worker/bitacross/core/bc-task-processor/src/lib.rs
Jayanring: `ceremony_command_tmp` stores commands that are not ready to be processed when they are received. These commands will be moved to `commands_to_process` later, so the length of `commands_to_process` can be more than one.

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
Jayanring: > 1. remove command processing from ceremony

Could you explain more about this?

kziemianek: `handle_ceremony_command` can call `MuSig2Ceremony::new`, `ceremony.receive_nonce`, `ceremony.receive_partial_sign` directly.

`kill_ceremony.rs` handler can also be decoupled from `ceremony_registry` and `ceremony_commands` - it can return `CeremonyCommand::Kill` so it can be handled by `handle_ceremony_command`.

Then we have:
* commands are moved from `bc-musig2-ceremony` to `bc-task-processor`
* dumb direct call handlers - they only authorize requests and produce commands
* `bc-task-processor` is responsible of processing commands and managing `ceremony_registry`/`ceremony_commands` state

Jayanring: Done

bitacross-worker/bitacross/core/bc-task-processor/src/lib.rs
```diff
@@ -389,6 +412,80 @@ fn handle_ceremony_command<SKR, SIGNINGAK, EKR, BKR, S, H, O, RRL, ERL, SRL, Res
...
+		},
+	}
+}
+
+fn ceremony_process_command<AK: AccessKey<KeyType = SchnorrPair>>(
+	ceremony: &mut MuSig2Ceremony<AK>,
+	command: CeremonyCommand,
+) -> Option<CeremonyEvent> {
+	let event_ret = match command {
+		CeremonyCommand::InitCeremony(_, _) => unreachable!("InitCeremony should not reach here"),
```
kziemianek: Why this function cannot handle `InitCeremony` and `KillCeremony` command ? 🤔

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
```diff
@@ -123,37 +109,38 @@ pub fn generate_aggregated_public_key(mut public_keys: Vec<PublicKey>) -> Public
...
 	payload: SignBitcoinPayload,
 	// P-713: move to layer above, ceremony should be communication agnostic
 	aes_key: RequestAesKey,
 	me: SignerId,
 	signers: SignersWithKeys,
-	commands: Vec<CeremonyCommand>,
-	events: Vec<CeremonyEvent>,
 	signing_key_access: Arc<AK>,
+	agg_key: PublicKey,
+	check_run: bool,
```
kziemianek, line 121: please keep previous comment

bitacross-worker/bitacross/core/bc-task-sender/src/lib.rs
```diff
@@ -51,9 +51,33 @@ use std::{
 };
 
 #[derive(Debug)]
-pub struct BitAcrossRequest {
-	pub sender: oneshot::Sender<Result<BitAcrossProcessingResult, Vec<u8>>>,
-	pub request: AesRequest,
+pub enum BitAcrossRequest {
+	CreateSignTask(AesRequest, oneshot::Sender<Result<BitAcrossProcessingResult, Vec<u8>>>),
```
kziemianek: Into which category `DirectCall::KillCeremony` falls ?

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -512,15 +524,13 @@ pub enum BitacrossRequestError {
...
-	let request = AesRequest::from_hex(&payload)
-		.map_err(|e| format!("AesRequest construction error: {:?}", e))?;
+async fn bitacross_task_create_inner(params: Params) -> Result<RpcReturnValue, Vec<u8>> {
+	let request = get_request_from_params(params)?;
 
 	let bit_across_request_sender = BitAcrossRequestSender::new();
 	let (sender, receiver) = oneshot::channel::<Result<BitAcrossProcessingResult, Vec<u8>>>();
 
-	bit_across_request_sender.send(BitAcrossRequest { sender, request })?;
+	bit_across_request_sender.send(BitAcrossRequest::CreateSignTask(request, sender))?;
```
kziemianek: Currently this API is able to handle any `DirectCall`, why are you limiting it to 'signing tasks only' ? 
This was general api for submiting bitacross tasks - DirectCalls.
Now it's being coupled to 'signing' tasks only which is in my opinion a step back.

Please keep it as it is - or introduce new method for handling signing tasks only.

bitacross-worker/bitacross/core/bc-task-processor/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		}
+		{
+			let mut command_tmp_write = ceremony_command_tmp.write().unwrap();
+			command_tmp_write.retain(|_, &mut (_, create_time)| now - create_time < time_to_live);
+		}
+	});
+
+	let bit_across_task_receiver = init_bit_across_task_sender_storage();
+	let peers_map = Arc::new(Mutex::new(HashMap::<[u8; 32], DirectRpcClient>::new()));
+	let command_threads_count = 4;
```
kziemianek: Can this be parametrized through CLI args ?
It would be good to have an opportunity to tweak it without preparing new binary.

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
```diff
@@ -162,6 +162,18 @@ where
 		}
 	});
 
+	// btc sign task data share
+	io.add_sync_method("bitacross_btcDataShare", move |params: Params| {
```
kziemianek, line 166: This method simply processes any `DirectCall` - not only variants related to musig2 ceremony rounds.
For example i've tried to submit `SignEthereum` direct call - it was executed but signature was not returned.

Why you decided to add it ?

kziemianek (changes requested):

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
```diff
@@ -28,86 +28,72 @@ use std::{format, string::String, sync::Arc};
...
-	ContributionError,
-	IncorrectRound,
-	SecondRoundFinalizationError,
-}
-
-// events come from outside and are consumed by ceremony in tick fn
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, Clone)]
 pub enum CeremonyCommand {
+	InitCeremony([u8; 32], SignersWithKeys, bool),
```
kziemianek: It looks like this command is incomplete - it lacks of `SignBitcoinPayload`. 
Could you add it for the sake of simplicity?

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
Jayanring: > Optimizing ceremony data sharing logic, now it sends response back once handled by websocket server

There are two kinds of `DirectCall`, one is for submitting request from client (still `bitacross_submitRequest`), one is for communication between workers(now `bitacross_btcDataShare`).
You can check `bitacross_task_create_inner` and `bitacross_data_share_inner`, they are handled in different ways, the former waits for response sent from command processing thread, the latter not.
I think it's better to make them decoupled

bitacross-worker/bitacross/core/bc-task-sender/src/lib.rs
Jayanring: `KillCeremony` is also a kind of "data" shared by workers, so `ShareCeremonyData`

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
Jayanring: We can't see what concrete variant of `DirectCall` the request contains, because it's encrypted.

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
Jayanring: ~~`SignBitcoinPayload` is using by all kinds of commands, so I put it together with command in a tuple `Option<(CeremonyId, CeremonyCommand)>`.
If `InitCeremony` add `SignBitcoinPayload` then all variants should do so, and we need another function for taking out `SignBitcoinPayload` from variants.~~

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
kziemianek: So there is nothing specific to `btcDataShare` except the name here.

Jayanring: And different handling, `bitacross_task_create_inner `and `bitacross_data_share_inner`

bitacross-worker/bitacross/core/bc-task-processor/src/lib.rs
Jayanring: ~~It's better to do so, but could we improve this in another PR? I'm not familiar with code about passing configuration parameters for now.~~

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
Jayanring: Updated

bitacross-worker/bitacross/core/bc-musig2-ceremony/src/lib.rs
Jayanring: OK

bitacross-worker/bitacross/core/bc-task-processor/src/lib.rs
kziemianek: I'm fine with it as long as thread usage won't cause TCS crash on prod.

bitacross-worker/enclave-runtime/src/rpc/worker_api_direct.rs
Jayanring: I added `CeremonyRoundCall` to make this clearer

kziemianek (approved):
Looks fine, good job.

bitacross-worker/bitacross/core/bc-task-processor/src/lib.rs
```diff
@@ -0,0 +1,706 @@
...
+	std::thread::spawn(move || loop {
+		std::thread::sleep(Duration::from_secs(3));
+		let now = get_current_timestamp();
+		{
+			let mut ceremony_registry_write = ceremony_registry.write().unwrap();
+			ceremony_registry_write
+				.retain(|_, &mut (_, create_time)| now - create_time < time_to_live);
+		}
+		{
+			let mut command_tmp_write = ceremony_command_tmp.write().unwrap();
```
kziemianek, line 255: I think we got a small regression here. 
In case of ceremony timeout there is no reply sent back to rpc client

kziemianek (changes requested):

bitacross-worker/bitacross/core/bc-task-processor/src/lib.rs
Jayanring: working on it.

kziemianek (approved):

Kailai-Wang:
Thank for for getting all glitches fixed

PR #2902, opened 2024-07-16
fix: potential state mismatch caused by failed indirect trusted_call

felixfaisal:
TODO: 
- Need to update unit tests or add 
- Need to refactor some parts of the code 
- Should check no problems in CI especially multi-worker
- Should update RPC values for error variants 

So there are two types of failures in STF Execution 
### 1. Invalid Trusted Call 
- Does not read the sidechain state 
- Cannot mutate the sidechain state 
https://github.com/litentry/litentry-parachain/blob/9a188ee6b4a88c0d9ea345e9fdfa238040321bd5/tee-worker/core-primitives/stf-executor/src/executor.rs#L112
https://github.com/litentry/litentry-parachain/blob/9a188ee6b4a88c0d9ea345e9fdfa238040321bd5/tee-worker/core-primitives/stf-executor/src/executor.rs#L120
### 2. Stf Execution Failure 
- These read state of the sidechain state 
- Can potentially increment account nonce even in failure, which is a mutation to the sidechain state 
https://github.com/litentry/litentry-parachain/blob/9a188ee6b4a88c0d9ea345e9fdfa238040321bd5/tee-worker/core-primitives/stf-executor/src/executor.rs#L136

However both are treated the same as ` Ok(ExecutedOperation::failed(..))` and thereby not included in the block, 
Only successfully executed trusted operations are included in the block. 

### State Transition Problem 
Screenshot 2024-07-16 at 6 54 49 PM

A blockchain operates as a state transition function, Each transaction mutates the state creating a new state and this transaction is included in the block. However here we are mutating the state but we have no transaction to show for which is fundamentally against the concept of being a state transition function.

### DIFF
--- a/tee-worker/core-primitives/stf-executor/src/error.rs
+++ b/tee-worker/core-primitives/stf-executor/src/error.rs
@@ -17,13 +17,17 @@
 
 #[cfg(all(not(feature = "std"), feature = "sgx"))]
 use crate::sgx_reexport_prelude::*;
+use crate::RpcResponseValue;
 
 use itp_stf_primitives::error::StfError;
+use itp_types::parentchain::ParentchainCall;
 use sgx_types::sgx_status_t;
-use std::{boxed::Box, format};
+use sp_core::H256;
+use std::{boxed::Box, format, vec::Vec};
 
 pub type Result<T> = core::result::Result<T, Error>;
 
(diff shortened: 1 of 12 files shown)


felixfaisal:
Will re-open, right now, de-prioritizing

PR #2904, opened 2024-07-17
fix: A8 wrong network in VC

wli-pro:
P-926 added enum A8SupportedNetwork to matching ts types, removed unsupported network by dataprovider for A8

!!!! This is a breaking change !!!

### Context

1 This VC used achainable, and it only supported a subset of networks as below, instead all the network that we defined:
ethereum, polkadot, kusama, litmus, litentry, khala
image

### Labels

### Testing Evidences

Before the fix:
Screenshot 2024-07-17 at 5 18 48 pm

Screenshot 2024-07-17 at 5 19 11 pm

After the fix:
Screenshot 2024-07-17 at 5 19 37 pm

Screenshot 2024-07-17 at 5 20 00 pm

### DIFF
--- a/primitives/core/src/assertion/mod.rs
+++ b/primitives/core/src/assertion/mod.rs
@@ -143,7 +143,7 @@ pub enum Assertion {
 	Dynamic(DynamicParams)
 }
 
-pub const A8_SUPPORTED_NETWORKS: [Web3Network; 6] = [
+const A8_SUPPORTED_NETWORKS: [Web3Network; 6] = [
 	Web3Network::Polkadot,
 	Web3Network::Kusama,
 	Web3Network::Litentry,
--- a/primitives/core/src/assertion/mod.rs
+++ b/primitives/core/src/assertion/mod.rs
@@ -143,6 +143,15 @@ pub enum Assertion {
 	Dynamic(DynamicParams)
 }
 
+pub const A8_SUPPORTED_NETWORKS: [Web3Network; 6] = [
+	Web3Network::Polkadot,
+	Web3Network::Kusama,
+	Web3Network::Litentry,
+	Web3Network::Litmus,
+	Web3Network::Khala,
+	Web3Network::Ethereum,
+];
+
 impl Assertion {
 	// Given an assertion enum type, retrieve the supported web3 networks.
 	// So we limit the network types on the assertion definition level.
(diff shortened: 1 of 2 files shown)


primitives/core/src/assertion/mod.rs
```diff
@@ -164,7 +167,11 @@ impl Assertion {
 			Self::VIP3MembershipCard(..) |
 			Self::WeirdoGhostGangHolder => vec![Web3Network::Ethereum],
 			// total tx over `networks`
-			Self::A8(network) => network.to_vec(),
+			Self::A8(networks) => networks
```
BillyWooo, line 176: How do you think just add a filter to filter out types, which is NOT `Litentry, Litmus, Polkadot, Kusama, Khala, or Ethereum` ?
Then we don't need to create a new type `A8SupportedNetwork`.

Kailai-Wang: Does the client pass in the wrong networks? Shouldn't this be fixed at client side first?

But sure we could a second check logic here

And yes we can simply filter out unwanted types, or `retain` only desired ones

BillyWooo: @Kailai-Wang this is the leftover of refactoring of shifting network type from `link_identity` to `request_vc`. From IDHub point of view, they always wish the all supported networks. So they don't do any filtering at client side. We did the magic within `get_eligible_identities()`.

wli-pro: refactored to only filter instead creating new types.

primitives/core/src/assertion/mod.rs
```diff
@@ -143,6 +143,15 @@ pub enum Assertion {
 	Dynamic(DynamicParams)
 }
 
+pub const A8_SUPPORTED_NETWORKS: [Web3Network; 6] = [
```
BillyWooo: maybe a bit picky, it's better to not use `pub` here. It should be valid only within this file.

BillyWooo (approved):
LGTM

silva-fj (approved):

primitives/core/src/assertion/mod.rs
jonalvarezz: > this is the leftover of refactoring of shifting network type from link_identity to request_vc. From IDHub point of view, they always wish the all supported networks. So they don't do any filtering at client side. We did the magic within get_eligible_identities().

I doubt that's related.

The main motivation here was to have a struct to represent exactly the accepted values for A8. Clients must specify one of them to get the right assertion.

PR #2908, opened 2024-07-18
rm unused env in bc-worker

BillyWooo:
as title, very minor PR.


kziemianek (commented):
Could you remove entries from `config.json.eg` also ?

kziemianek (approved):

PR #2913, opened 2024-07-19
import sidechain block through trusted rpc

kziemianek:
The main goal of this PR is to move `sidechain_importBlock` from untrusted rpc endpoint to trusted. By doing so, block producer doesn't need to query all known peers for `untrusted_url` on each sidechain block broadcast.

It also:
* removes `rpc` ecall.
* removes 'rpc_methods' trusted json-rpc method


Kailai-Wang (approved):
Looks good from my side, I see it uses `ws` for block import - but that's the same logic as before.

A request: could you please propose a similar PR to upstream too?

tee-worker/enclave-runtime/src/initialization/mod.rs
```diff
@@ -210,20 +213,32 @@ pub(crate) fn init_enclave(
...
 		top_pool_author,
 		getter_executor,
 		shielding_key_repository,
 		ocall_api.clone(),
 		Some(state_handler),
 		data_provider_config,
 	);
+
+	#[cfg(feature = "sidechain")]
+	{
```
BillyWooo, line 228: In general, it looks good to me. I am just wondering: is it this block the fundamental change?

kziemianek:
I'll add license headers.

kziemianek:
It might still need a fix, see:
https://github.com/integritee-network/worker/pull/1605#discussion_r1687793725

tee-worker/enclave-runtime/src/initialization/mod.rs
kziemianek: The main change of this PR is in this file:
`tee-worker/service/src/worker.rs`

PR #2924, opened 2024-07-23
quick-fix: assertion contracts tests

0xverin:
### Context

fix https://github.com/litentry/litentry-parachain/actions/runs/10047804865/job/27782346298

Reason: ts-tests only set [one secret](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/ts-tests/integration-tests/assertion_contracts.test.ts#L68) when deploying contracts, that when requesting `bnb`, noderealclient needs to be used,it needs [secrets[1]](https://github.com/litentry/litentry-parachain/blob/dev/tee-worker/litentry/core/assertion-build/src/dynamic/contracts/token_holding_amount/TokenQueryLogic.sol#L67), which is out of index of array,that causes the secret to be unable to be decoded when requesting dynamic contracts.

Solution: Adding at least 3 secrets.

### DIFF
--- a/.github/file-filter.yml
+++ b/.github/file-filter.yml
@@ -37,6 +37,7 @@ tee_test: &tee_test
   - 'tee-worker/cli/*.sh'
   - 'docker/**'
   - 'tee-worker/docker/*.yml'
+  - 'tee-worker/litentry/core/assertion-build/src/dynamic/contracts/**'
 
 bitacross_src: &bitacross_src
   - 'bitacross-worker/**/*.rs'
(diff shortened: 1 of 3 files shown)


wli-pro (approved):

Kailai-Wang:
Please change `file-filter.yml` too

Kailai-Wang (changes requested):
Please change `file-filter.yml` too

0xverin:
> Please change `file-filter.yml` too

Is it necessary to optimize for separate testing in the future, modifying the contract only needs to run this assertion contracts tests, if necessary, I will create an issue to do it later.

Kailai-Wang (approved):

PR #2931, opened 2024-07-25
Adding register OIDC client call to identity management pallet

silva-fj:
This is part of the Litentry's OIDC POC. DApps that want to integrate litentry as OIDC provider need to register their redirect_uri's

There are a couple of TODO's for a later stage: 
- Add benchmarks for the new calls and use the calculated weights.
- Take a deposit when register oidc client.

### DIFF
--- a/runtime/litentry/src/lib.rs
+++ b/runtime/litentry/src/lib.rs
@@ -148,7 +148,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("litentry-parachain"),
 	authoring_version: 1,
 	// same versioning-mechanism as polkadot: use last digit for minor updates
-	spec_version: 9185,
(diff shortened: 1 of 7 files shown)


Kailai-Wang (approved):
Looks good, thanks

runtime/litentry/src/lib.rs
```diff
@@ -148,7 +148,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("litentry-parachain"),
 	authoring_version: 1,
 	// same versioning-mechanism as polkadot: use last digit for minor updates
-	spec_version: 9185,
+	spec_version: 9186,
```
Kailai-Wang, line 151: Actually we don't need to bump the version - I just bumped it 3 days ago
But small problem

PR #2934, opened 2024-07-26
Add basic bitacross worker ci integration tests

kziemianek:
Adds bitacross-worker CI enclave tests and integration test calling `bitacross_checkSignBitcoin` rpc method.


.github/workflows/ci.yml
```diff
@@ -511,6 +516,148 @@ jobs:
         if: failure()
         uses: andymckay/cancel-action@0.5
 
+  bitacross-build:
```
kziemianek, line 519: I'm totally unsure about caching here 🤔

Kailai-Wang: I suggest removing caching for now - the sole identity-worker mounts up to 3G for a single run, if we add bitacross, it would be another 3G (assume they generate different cache files)

The repo limit is 10G, so we can't afford even two (unique) runs for bitacross + identity

Kailai-Wang: We'll use unified cache once we have the same codebase for identity + bitacross

Kailai-Wang (changes requested):
Looks good in general!

Kailai-Wang (approved):

PR #2936, opened 2024-07-26
Update parachain dependencies to `polkadot-v1.1.0`

Kailai-Wang:
### Context

As topic -  we'll probably merge it after the decimal change.

It upgrades the parachain dep to `polkadot-v1.1.0` while not touching the worker dep, so the workers still use `polkadot-v0.9.42`.

It works because the worker can depend on different yet compatible versions of the same pallet: `pallet-teebag`, `core-primitives` etc.

I have to use `release-v0.9.19` branch for the pallets used in the worker to make sure no new versions of pallets are introduced. It means:
- we have to maintain two versions of pallets, but that's unavoidable
- we have to keep an eye on the potential incompatibility issue, e.g. the extrinsic bytes constructed within the enclave can't be parsed by parachain runtime, or storage type change between different versions.

It paves the way for the future parachain upgrades.
However, it doesn't mean we don't switch to sdk-v2.0. We'll still need it to eventually have the same substrate pallets, as well as newer crates

Particularly asking
@wangminqi to review the node/xcm/evm part
@kziemianek to review the worker changes


.github/workflows/ci.yml
```diff
@@ -332,15 +332,11 @@ jobs:
       - set-condition
       - sequentialise
     steps:
+      - uses: actions/checkout@v4
```
BillyWooo, line 335: should we always add `fetch-depth: 0` here?  And the following places.

Kailai-Wang: I don't think this is as important

Cargo.toml
```diff
@@ -268,9 +270,17 @@ pallet-evm-precompile-bridge-transfer = { path = "precompiles/bridge-transfer",
...
 pallet-evm-precompile-score-staking = { path = "precompiles/score-staking", default-features = false }
 pallet-evm-assertions = { path = "pallets/evm-assertions", default-features = false }
-# This is a temporary pallet copy from Astar with polkadot-0.9.43
-# The official release of this pallet will be in frontier repo
-precompile-utils = { git = "https://github.com/litentry/precompile-utils", branch = "main", default-features = false }
 
 [patch.crates-io]
 ring = { git = "https://github.com/betrusted-io/ring-xous", branch = "0.16.20-cleanup" }
+
+[patch."https://github.com/AstarNetwork/frontier"]
```
BillyWooo, line 277: I don't find anywhere else using `fc-consensus` from `https://github.com/AstarNetwork/frontier`. Why do we need this patch?

bitacross-worker/scripts/litentry/release/template/relay-alice.service
```diff
@@ -5,7 +5,7 @@ Description=Litentry Relaychain Alice
 Type=simple
 User=USER
 WorkingDirectory=/opt/litentry/parachain
-ExecStart=/opt/litentry/parachain/polkadot --base-path /opt/litentry/parachain/relay-alice --chain /opt/litentry/parachain/rococo-local-chain-spec.json --alice --port 30336 --ws-port 9946 --rpc-port 9936
+ExecStart=/opt/litentry/parachain/polkadot --base-path /opt/litentry/parachain/relay-alice --chain /opt/litentry/parachain/rococo-local-chain-spec.json --alice --port 30336 --rpc-port 9946
```
BillyWooo, line 8: I noticed there are quite some places that rpc-port change from 9936 to 9946. But old ws-port 9946 is just removed. Can we simply remove `--ws-port 9946` and not touch `--rpc-port 9936` ?  Any special consideration about port number?

.github/workflows/ci.yml
BillyWooo: It can save some disk space. Ok. Anyway, not that much. 140M is our repo current size.

BillyWooo:
> I have to use release-v0.9.18 branch

It's a typo or ? release-v0.9.19 is better? @Kailai-Wang

Kailai-Wang:
Yes it was 0.9.18 when I created the PR - later I updated it to use 0.9.19 branch

runtime/litmus/src/weights/mod.rs
```diff
@@ -17,12 +17,12 @@
...
 
 pub mod cumulus_pallet_xcmp_queue;
-pub mod frame_system;
+// pub mod frame_system;
 pub mod pallet_asset_manager;
 pub mod pallet_balances;
 pub mod pallet_bridge;
 pub mod pallet_bridge_transfer;
-pub mod pallet_collator_selection;
+// pub mod pallet_collator_selection;
```
silva-fj: why are these commented out here and in the other `*/weights/mod.rs` files? can they be removed?

silva-fj (approved):
🚀

runtime/litmus/src/weights/mod.rs
Kailai-Wang: I can't remember exactly why it was commented out.

I can uncomment them, it doesn't matter too much tho:
- the weightInfo uses `()`
- litmus will be removed soon anyway

node/src/command.rs
```diff
@@ -369,12 +367,22 @@ pub fn run() -> Result<()> {
 		Some(Subcommand::Key(cmd)) => cmd.run(&cli),
 
 		Some(Subcommand::ExportGenesisState(cmd)) => {
-			let runner = cli.create_runner(cmd)?;
-			runner.sync_run(|_config| {
-				let spec = cli.load_spec(&cmd.shared_params.chain.clone().unwrap_or_default())?;
-				let state_version = Cli::native_runtime_version(&spec).state_version();
-				cmd.run::<Block>(&*spec, state_version)
-			})
+			// TODO: is this the best way?
```
wangminqi, line 370: Simply remark here for discussion

runtime/litmus/src/weights/mod.rs
Kailai-Wang: I removed it completely.

node/src/command.rs
```diff
@@ -389,28 +397,21 @@ pub fn run() -> Result<()> {
...
-								cmd.run::<Block, LitmusParachainRuntimeExecutor>(config)
-							} else if config.chain_spec.is_litentry() {
-								cmd.run::<Block, LitentryParachainRuntimeExecutor>(config)
-							} else if config.chain_spec.is_rococo() {
-								cmd.run::<Block, RococoParachainRuntimeExecutor>(config)
-							} else {
-								Err(UNSUPPORTED_CHAIN_MESSAGE.into())
-							}
-						})
+						use crate::service::HostFunctions;
```
wangminqi, line 406: Why using HostFunctions instead of old RuntimeExecutor? 
It should not cause too much difference, but what if two parachain's spec_version are different.

wangminqi (changes requested):

node/src/command.rs
wangminqi: Solved

runtime/common/src/tests/mod.rs
```diff
@@ -14,15 +14,7 @@
...
-// mod transaction_payment;
-
-pub mod xcm_parachain;
-
 pub mod base_call_filter;
-
+pub mod setup;
 pub mod transaction_payment;
-
-pub mod orml_xcm;
```
wangminqi, line 28: Not request changing. But weird, when have we deleted this pallet?

runtime/litentry/src/weights/mod.rs
```diff
@@ -17,7 +17,7 @@
 #![allow(clippy::unnecessary_cast)]
 
 pub mod cumulus_pallet_xcmp_queue;
-pub mod frame_system;
+// pub mod frame_system;
```
wangminqi, line 20: Why disable system weight?

runtime/litentry/src/xcm_config.rs
```diff
@@ -299,13 +300,7 @@ parameter_types! {
...
 impl pallet_xcm::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
-	// We allow anyone to send any XCM to anywhere
-	// This is highly relied on if target chain properly filtered
-	// Check their Barriers implementation
-	// And for TakeWeightCredit
-	// Check if their executor's ShouldExecute trait weight_credit
-	type SendXcmOrigin =
-		FilterEnsureOrigin<RuntimeOrigin, LocalOriginToLocation, EnsureRootOrTwoThirdsCouncil>;
+	type SendXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
```
wangminqi, line 303: Since we do not have orml_xcm any more. Be aware that this impl will make us not able to make sovereign account do anything (even root).

wangminqi (changes requested):

wangminqi:
Anyway, nothing else crucial. I will approve first.

wangminqi (approved):

node/src/command.rs
Kailai-Wang: It's for benchmarking command only - the interface (trait bound) has changed. With the old Executor you'll get:
```
^^^^^ the trait `sc_executor::sp_wasm_interface::HostFunctions` is not implemented for `service::LitmusParachainRuntimeExecutor`
```

.github/workflows/ci.yml
Kailai-Wang: That's true.

Having `fetch-depth` will require more space - it explains why we don't need to add it if not required

Cargo.toml
Kailai-Wang: Can be indirect dep

bitacross-worker/scripts/litentry/release/template/relay-alice.service
Kailai-Wang: As spoken it replaces the old `ws-port`

runtime/common/src/tests/mod.rs
Kailai-Wang: we deleted it in one of the clean-up PRs, maybe this: https://github.com/litentry/litentry-parachain/commit/1de9f3828b6ae9b2169b359fffdde1d4c3df36c8

runtime/litentry/src/weights/mod.rs
Kailai-Wang: Because it has a different interface and needs to be re-benchmarked.

Having it uncommented will cause compilation error

PR #2937, opened 2024-07-26
parachain-api: expose link_identity_callback

jonalvarezz:
These changes make `TrustedCall::link_identity_callback` available for clients.

The `parachain-api` was built against `0.9.18-11` tag for publishing to not leak any other change from `dev` branch.

Published packages:
- parachain-api: https://www.npmjs.com/package/@litentry/parachain-api/v/0.9.18-11.2

### DIFF
--- a/tee-worker/client-api/parachain-api/CHANGELOG.md
+++ b/tee-worker/client-api/parachain-api/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.9.18-11.2] - 2024-07-26
+
+### Added
+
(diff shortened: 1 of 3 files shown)


Kailai-Wang (approved):
While this is OK to expose it, is it possible to use different versions of client-api in dev and prod?

Ideally we don't expose it on prod at all

jonalvarezz:
> While this is OK to expose it, is it possible to use different versions of client-api in dev and prod?
> 
> Ideally we don't expose it on prod at all

I see your point, but I don't see how to prevent it on clients. I mean, if we have different versions, a client could still choose to use the development version against a production worker.

We could programmatically choose which one to use in execution time on clients, but I don't think it is worth the effort. 

Maybe adding a check on client-sdk to throw an error if the network is not development?

PR #2941, opened 2024-07-29
p-895: Final Merge of Decimal Migration and Bridge change

wangminqi:
The following storage migration is so far included:
- [x] Parachain Staking Storage Migration
- [x] Balances Storage Migration
- [x] Bridge Storage Migation
- [x] Bounties, Treasury, Democracy, Preimage
- [x] Vesting, Identity, Proxy, Multisig

Other tasks:
- [x] Ts-test adaption
- [ ] ethereum smart contract migration


docker/bridge.dockerfile
```diff
@@ -1,3 +1,3 @@
 FROM golang:1.18
 
-RUN go install github.com/litentry/ChainBridge/cmd/chainbridge@dev
+RUN go install github.com/collab-ai-network/ChainBridge/cmd/chainbridge@sol0.8.19
```
Kailai-Wang: We should use litentry org (with a different branch)

runtime/common/src/tests/setup/mod.rs
```diff
@@ -42,8 +42,8 @@ pub fn charlie() -> AccountId {
 	AccountId::from(CHARLIE)
 }
 
-pub const PARA_A_USER_INITIAL_BALANCE: u128 = 500_000 * UNIT;
-pub const PARA_B_USER_INITIAL_BALANCE: u128 = 600_000 * UNIT;
+pub const PARA_A_USER_INITIAL_BALANCE: u128 = 500_000_000_000 * UNIT;
+pub const PARA_B_USER_INITIAL_BALANCE: u128 = 600_000_000_000 * UNIT;
```
Kailai-Wang, line 46: We should change the definition of `UNIT` instead of this

pallets/assets-handler/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+		},
+		/// A certain amount of asset tokens was minted
+		TokenBridgeIn {
+			asset_id: Option<AssetId<T>>,
+			to: T::AccountId,
+			amount: BalanceOf<T>,
+		},
+		TokenBridgeOut {
+			asset_id: Option<AssetId<T>>,
+			to: T::AccountId,
```
Kailai-Wang: you mean `from`?

runtime/litentry/src/migration/parachain_staking.rs
```diff
@@ -0,0 +1,748 @@
...
+		weight += Self::replace_delegator_state_storage();
+		weight += Self::replace_candidate_info_storage();
+		weight += Self::replace_delegation_scheduled_requests_storage();
+		weight += Self::replace_top_delegations_storage();
+		weight += Self::replace_bottom_delegations_storage();
+		weight += Self::replace_total_storage();
+		weight += Self::replace_candidate_pool_storage();
+
+		// No need for AtStake Migration since this is a snapshot, everything is good as long as it
+		// will not change proportion AtStake
```
Kailai-Wang, line 590: `AtStake` is referenced in 
```
for (account, snapshot) in <AtStake<T>>::iter_prefix(last_round) {
```
as well - could you double-check that is is unused?
Maybe becauase `collators` is never empty? 

Either way it would be safer if we also do the migration for it ofc

wangminqi: It is used. But everything relying on it is calculated based on its proportion calculation. Like AtStaked storing two balance number A and B, and everywhere using A, B by (A divide B), which making the migration of AtStake old snapshot meaningless.

runtime/common/src/tests/setup/mod.rs
wangminqi: What do you mean?

docker/bridge.dockerfile
wangminqi: Roger. I will create the corresponding branch in Litentry repo.

pallets/assets-handler/src/lib.rs
wangminqi: bridge assets Out of local chain.
bridge assets from local chain to remote chain.

I personally feel Out/From are equally "ambiguous".

docker/bridge.dockerfile
wangminqi: Changed. However, the lower level code still uses external repo, just like the trace of phala, chainsafe repo.

runtime/litentry/src/migration/parachain_staking.rs
wangminqi: You are correct, it does matter if collator is empty.

scripts/run-ts-test.sh
```diff
@@ -27,12 +27,13 @@ LITENTRY_PARACHAIN_DIR=${LITENTRY_PARACHAIN_DIR:-"/tmp/parachain_dev"}
...
+# Temporary disable ts test due to decimal change task.
+# if $bridge; then
+#     pnpm run test-bridge 2>&1 | tee -a "$LITENTRY_PARACHAIN_DIR/parachain_ci_test.log"
+# fi
+
+# if $evm; then
+#     pnpm run test-evm-transfer 2>&1 | tee "$LITENTRY_PARACHAIN_DIR/parachain_ci_test.log"
+#     pnpm run test-evm-contract 2>&1 | tee "$LITENTRY_PARACHAIN_DIR/parachain_ci_test.log"
+#     pnpm run test-precompile-contract 2>&1 | tee "$LITENTRY_PARACHAIN_DIR/parachain_ci_test.log"
+# fi
```
wli-pro: Should the tests be fixed as part of the same PR to get more test coverage of the current change?

node/res/chain_specs/litentry-plain.json
```diff
@@ -32,63 +32,63 @@
         "balances": [
           [
             "5ECCQzRn2nUqDesngcoi7FxLhJBXGSnoEmXiMDZXCj94chRP",
-            100000000000000
+            100000000000000000000
```
Kailai-Wang, line 35: It might be confusing why we change `litentry-plain.json` but leave `litentry.json` unchanged.

Not changing genesis in `litentry.json` is crucial.

But shall we keep it consistent here? Wdyt? Maybe we don't need `*-plain.json` anymore

node/res/chain_specs/litentry.json
```diff
@@ -16,7 +16,7 @@
   "protocolId": "litentry",
   "properties": {
     "ss58Format": 31,
-    "tokenDecimals": 12,
+    "tokenDecimals": 18,
```
Kailai-Wang, line 19: This probably has no effect. I assume the tokenDecimals in JSON is not used, please double-check the code:
```
fn default_parachain_properties() -> Option<Properties> {
	parachain_properties("LIT", 12, 31)
}
```

pallets/assets-handler/src/lib.rs
Kailai-Wang: `to` is worse tho 😂 

It's "bridging out", so it's from local chain to remote chain.
Users/devs expect a target address in remote chain with `to`.

pallets/assets-handler/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+		T::ExternalTotalIssuance::get()
+			.checked_sub(&pallet_balances::Pallet::<T>::total_issuance())
+			.map_or_else(|| 0u32.into(), |v| v)
+	}
+
+	// Native Token External Unlocked Total Balances of outside
+	#[pallet::storage]
+	#[pallet::getter(fn external_balances)]
+	pub type ExternalBalances<T: Config> =
+		StorageValue<_, BalanceOf<T>, ValueQuery, DefaultExternalBalances<T>>;
```
Kailai-Wang, line 109: I sort of dont' understand/forget why we need `ExternalBalances` at all - it seems to be always equal to total LIT circulation - total issuance on this chain, unless you explicitly set it.

But we won't set it because it will affect the inflation in pallet-parachain-staking.

If we only want to limit the maximal tokens that can be minted, we have `MaximumIssuance` already.

pallets/assets-handler/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+		match asset_info {
+			None => Err(Error::<T>::InvalidResourceId.into()),
+			// Native token
+			Some(AssetInfo { fee, asset: None }) => {
+				Self::deposit_event(Event::TokenBridgeOut {
+					asset_id: None,
+					to: who.clone(),
+					amount,
+					fee,
+				});
```
Kailai-Wang: Normally we emit the event at the end

runtime/common/src/tests/setup/mod.rs
Kailai-Wang: `UNIT` is already changed from 10^ 12 to 10^18 in `runtime/common/src/lib.rs`, we don't have to change anything here, same for other places

runtime/litentry/src/migration/balances_transaction_payment.rs
```diff
@@ -0,0 +1,383 @@
...
+		);
+		let pallet_prefix: &[u8] = b"Balances";
+		let storage_item_prefix: &[u8] = b"InactiveIssuance";
+		let stored_data = get_storage_value::<u128>(pallet_prefix, storage_item_prefix, b"")
+			.expect("Storage query fails: Balances InactiveIssuance");
+		<InactiveIssuance<T>>::put(stored_data.saturating_mul(DECIMAL_CONVERTOR));
+		let weight = T::DbWeight::get();
+		frame_support::weights::Weight::from_parts(0, weight.write + weight.read)
+	}
+	pub fn check_balances_account_storage() -> frame_support::weights::Weight {
```
Kailai-Wang: I'm not sure if we even want this fn (also other fields check).

We expect it to be empty, but `assert!` will panic if it's non-empty, won't it? I don't know how it can be non-empty, just somehow.

runtime/common/src/tests/setup/mod.rs
wangminqi: This is leftover change since we previously seperate balance runtime impl PR and pallet migration PR.
Yes, this can be removed. But it will not cause any further problem since it is for avoiding existential deposit balance issue.

scripts/run-ts-test.sh
wangminqi: Yes, Verin is on it, ts-test is not fixed yet. Meanwhile we fix other stuff.

node/res/chain_specs/litentry.json
wangminqi: I have applied change to your reference.
I do not know. It might effect if the frontend use chain_spec for displaying purposes.
I simply change as much deicmal 12 related impl as possible.

pallets/assets-handler/src/lib.rs
wangminqi: External Balances = total supply of non-local native token. And yes, it is supposed to target at around a total sum of 100 Million LIT. Which is used by parachain staking.

MaximumIssuance is for how much local bridge can be minted. No matter how much outside LIT want to get in parachain, at most this amount of LIT will be allowed to minted.

They are different.

runtime/litentry/src/migration/balances_transaction_payment.rs
wangminqi: No way it can be non-empty.
I have removed the runtime upgrade part but keep the pre/post code.

pallets/assets-handler/src/lib.rs
wangminqi: Done change

pallets/assets-handler/src/lib.rs
wangminqi: I see! You mean variable name. I have changed it to "from".

node/res/chain_specs/litentry-plain.json
wangminqi: I have no clue what litentry-plain used for. My change of this file is brutal.
We definitely need to regenerate all chain_spec files once code is settled.

runtime/litentry/src/migration/parachain_staking.rs
wangminqi: I decide to skip this migration. 
On one side it will be only triggered when the collator empty and a round begin.
On the other side,
AtStake is a double map. Let's ignore it. It will introduce extra procedure of key concat.

Kailai-Wang: > On one side it will be only triggered when the collator empty and a round begin.

Will it cause problem once the set is _somehow_ empty?
Will this storage get updated / overwritten once after the runtime upgrade (e.g. during the block production phase)? We need to make sure it eventually reaches the right state

node/res/chain_specs/litentry-plain.json
Kailai-Wang: No we shouldn't regenerate them - it will affect the genesis hash.

wangminqi:
We will merge it to DEV for building rococo node.

BillyWooo (approved):

PR #2952, opened 2024-08-01
feat: decrease minimum staking amount of delegator

wangminqi:
Just like tittle.


BillyWooo:
should we also change it for ROCOCO? even though it's not really used.

wangminqi:
> should we also change it for ROCOCO? even though it's not really used.

It is fine.

BillyWooo (approved):
LGTM

PR #2956, opened 2024-08-02
Optimize `bitacross-worker` performance

Jayanring:
This PR mainly about:
1. Remove encryption of requests
2. Handle messages from connections alternately(instead of letting one starves the rest)


silva-fj (approved):
LGTM

PR #2959, opened 2024-08-03
cleanup env before build docker image

BillyWooo:
small but important PR.
As title, cleanup env before building docker image. The github free runner has limited resources.


Kailai-Wang (commented):
I only recommend to do this cleanup wherever absolutely needed. It can take up to 8min and _might_ remove some packages that the build needs.

So it a job runs well without it, don't touch it.
Apply it to the job where "no space left" actually happens

Kailai-Wang (approved):

PR #2960, opened 2024-08-05
Balance account data migration fix

wangminqi:
For rococo, the initial migration of P-895/P9190, contains some error code which make account.frozen = account.reserve * 10^6
This storage error should be fixed in the following migration of rococo.

This migration will force all frozen amount to 0.


runtime/litentry/src/migration/balances_transaction_payment.rs
```diff
@@ -64,7 +64,11 @@ where
 			account_info.data.free = account_info.data.free.saturating_mul(DECIMAL_CONVERTOR);
 			account_info.data.reserved =
 				account_info.data.reserved.saturating_mul(DECIMAL_CONVERTOR);
-			account_info.data.frozen = account_info.data.reserved.saturating_mul(DECIMAL_CONVERTOR);
+
+			// Runtime migration is wrong at first. The below code has been mistakenly applied to
```
BillyWooo: I think it's better to remove this comment.

wangminqi: I might prefer keep it. Just as a reminder of history view purpose.

wangminqi: Oh sorry. I should remove it for litentry and litmus.

BillyWooo (approved):
LGTM. 
let's fix our mistake after CI passes successfully.

runtime/rococo/src/lib.rs
```diff
@@ -177,11 +177,7 @@ pub type Executive = frame_executive::Executive<
 	// it was reverse order before.
 	// See the comment before collation related pallets too.
 	AllPalletsWithSystem,
-	(
-		migration::ReplaceParachainStakingStorage<Runtime>,
-		migration::ReplaceBalancesRelatedStorage<Runtime>,
-		migration::ReplaceBridgeRelatedStorage<Runtime>,
-	),
```
Kailai-Wang, line 184: A question: why did we have only three migrations on rococo?

wangminqi: hahahaha. We will fix it in another PR. OMG

PR #2968, opened 2024-08-06
Small cleanup

Kailai-Wang:
### Context

As topic:
- removes bitacross-mimic
- fixes some cargo dep
- pin `serde_json` because:
   - the latest version introduces `memchr` and it seems to have conflict with substrate crates, see also https://substrate.stackexchange.com/questions/11793/more-errors-after-upgrade-to-sdk-1-13
   - the worker needs a lower version (for lock)


BillyWooo (approved):

Cargo.toml
```diff
@@ -69,7 +69,7 @@ scale-info = { version = "2.11", default-features = false, features = ["derive"]
 parity-scale-codec = { version = "3.6", default-features = false, features = ["derive", "max-encoded-len"] }
 paste = { version = "1.0" }
 serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
-serde_json = { version = "=1.0.120", default-features = false }
+serde_json = { version = "=1.0.103", default-features = false }
```
BillyWooo, line 72: This is kind of strange format.

BillyWooo: ok. I missed that grammar session. It's exact specific version.

PR #2969, opened 2024-08-07
p-836 Update vc correctness

0xverin:
### Context
this pr:

- Set gha as a scheduled task.
- Updated identity-linking addresses.
- Optimized the testing process

All addresses have been tested locally and can pass.
There may still be some missing VC tests, we can add and update at any time. Let this workflow run first.


.github/workflows/verify-correctness-of-vc-content-.yml
Kailai-Wang, line 1: An extra `-` at the end of this file name

.github/workflows/verify-correctness-of-vc-content-.yml
```diff
@@ -4,33 +4,40 @@ on:
...
-          # extracting parachain version and worker version from release tag
-          echo "PARACHAIN_TAG=$(echo ${{inputs.docker-tag}} | cut -d'-' -f1 | sed 's/p/v/')" >> $GITHUB_ENV
-          echo "WORKER_TAG=$(echo ${{inputs.docker-tag}} | cut -d'-' -f3 | sed 's/w/v/')" >> $GITHUB_ENV
+          if [ "${GITHUB_EVENT_NAME}" = "workflow_dispatch" ]; then
+            echo "DOCKER_TAG=${{ inputs.docker-tag }}" >> $GITHUB_ENV
+          else
+            echo "DOCKER_TAG=latest" >> $GITHUB_ENV
+          fi
+        env:
+          GITHUB_EVENT_NAME: ${{ github.event_name }}
```
Kailai-Wang: Does this work?

```
env:
  DOCKER_TAG: ${{ github.event.inputs.docker-tag }} || latest
```

.github/workflows/verify-correctness-of-vc-content-.yml
```diff
@@ -4,33 +4,40 @@ on:
...
         run: |
           docker pull parity/polkadot
-          docker pull litentry/identity-worker:$WORKER_TAG
-          docker pull litentry/identity-cli:$WORKER_TAG
-          docker pull litentry/litentry-parachain:$PARACHAIN_TAG
+          docker pull litentry/identity-worker:$DOCKER_TAG
+          docker pull litentry/identity-cli:$DOCKER_TAG
+          docker pull litentry/litentry-parachain:$DOCKER_TAG
 
       - name: Re-tag docker image
```
Kailai-Wang, line 28: Maybe only run it if `DOCKER_TAG` is not `latest`

Kailai-Wang:
TBH this only verifies each DP's validity, but not the VC logic.
E.g. our LIT holding amount VC consults multiple DPs and sums them (or quits early) -> what if DP works OK but our logic goes wrong?
So the testing unit should be VC, not DP...

We simply have:
- predefined IDGraph
- VC-type
- If the generated VC meets the expected result

That's it

I guess that's too much for this PR alone tho

.github/workflows/verify-correctness-of-vc-content.yml
```diff
@@ -4,33 +4,35 @@ on:
...
       - uses: actions/checkout@v4
 
       - name: Set ENV
-        run: |
-          # extracting parachain version and worker version from release tag
-          echo "PARACHAIN_TAG=$(echo ${{inputs.docker-tag}} | cut -d'-' -f1 | sed 's/p/v/')" >> $GITHUB_ENV
-          echo "WORKER_TAG=$(echo ${{inputs.docker-tag}} | cut -d'-' -f3 | sed 's/w/v/')" >> $GITHUB_ENV
+        env:
+          GITHUB_EVENT_NAME: ${{ github.event_name }}
+          DOCKER_TAG: ${{ github.event.inputs.docker-tag }} || latest
```
Kailai-Wang: I mean the `env` in the workflow level

Kailai-Wang (approved):

PR #2970, opened 2024-08-07
Update rococo bootnode and cleanup

Kailai-Wang:
### Context

As topic.
It should be possible to use `--chain=rococo` now instead of JSON


BillyWooo (approved):
LGTM

PR #2971, opened 2024-08-08
Add try-runtime check in release process

Kailai-Wang:
### Context

As topic, available as standalone CI too


.github/workflows/check-runtime-upgrade.yml
```diff
@@ -0,0 +1,136 @@
...
+    types: [released]
+
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: runtime.wasm release_tag
+        required: false
+
+env:
+  SUBWASM_VERSION: 0.19.1
```
BillyWooo, line 19: the latest version is v0.21.3. Do we consider using that?

BillyWooo, line 19: the latest version is v0.21.3. Do we consider using that? Any dependency about polkadot version?

Kailai-Wang: Not in this PR

Kailai-Wang: We'll need to double-check if the new version has breaking change

.github/runtime.json
```diff
@@ -0,0 +1,20 @@
...
+      "uri": "wss://rpc.rococo-parachain.litentry.io:443"
+    },
+    {
+      "name": "litmus",
+      "package": "litmus-parachain-runtime",
+      "path": "runtime/litmus",
+      "uri": "wss://rpc.litmus-parachain.litentry.io:443"
+    },
+    {
+        "name": "litentry",
```
BillyWooo: "uri" is the key to determinate whether one entry is valid or not?
Besides, small indentation issue here.

.github/workflows/coverage.yml
BillyWooo, line 1: not fully understand the changes within this file. Are we going to remove whole "coverage" report? If so, why not a new yml file and file name?

Kailai-Wang: Good catch - I used it to test GHA.

I'm going to revert it

.github/runtime.json
Kailai-Wang: yes

thanks - I'll fix it

PR #2974, opened 2024-08-08
Bump versions and use unified crate versions

Kailai-Wang:
### Context

This PR:
- bumps runtime versions to 9192 for upcoming upgrades
- use `0.1.0` as version number for every crate, including worker binaries, as we are not ready to publish any of them. An exception is `litentry-collator` binary, which is already publicly used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1198,7 +1198,7 @@ checksum = "e496a50fda8aacccc86d7529e2c1e0892dbd0f898a6b5645b5561b89c3210efa"
 
 [[package]]
 name = "core-primitives"
-version = "0.9.12"
+version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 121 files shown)


BillyWooo (commented):
For PR itself, it looks good to me. 
What's the reason behind it adjust all these versions?

Kailai-Wang:
> For PR itself, it looks good to me. What's the reason behind it adjust all these versions?

As I thought leaving them in different versions is more confusing ... either we don't update them at all, or always keep them updated.

BillyWooo (commented):
How about taking care of warnings from the build.Dockerfile?

BillyWooo (approved):

kziemianek (approved):

PR #2975, opened 2024-08-08
remove unnecessary lines

BillyWooo:
as title, minor change


Kailai-Wang (approved):
Thanks, it actually fixes a bug

BillyWooo:
> Thanks, it actually fixes a bug

Yes. There is a typo. I think copy/paste error.

PR #2976, opened 2024-08-08
Identity staking follow up 

silva-fj:
This PR adds a way to do periodic updates for the staking amount of IDGraphs.

More context in  the issue.


tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -201,17 +222,158 @@ impl ParentchainEventHandler {
...
+				Error::Other(
+					"Metadata retrieval for update_token_staking_amount_call_indexes failed".into(),
+				)
+			})?
+			.map_err(|_| Error::Other("Invalid metadata".into()))?;
+
+		for account_id in account_ids.iter() {
+			let default_id_graph = Vec::new();
+			let id_graph = accounts_graphs.get(account_id).unwrap_or(&default_id_graph);
+			let staking_amount: Balance = id_graph
```
kziemianek, line 331: Won't this lead to undesired competition given that identities (for example EVM) are free to generate and link ?

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -201,17 +222,158 @@ impl ParentchainEventHandler {
...
+					let delegator = delegator_states.get(identity)?;
+					Some(delegator.total)
+				})
+				.sum();
+			let call = OpaqueCall::from_tuple(&(
+				update_token_staking_amount_call_index,
+				account_id,
+				staking_amount,
+			));
+			extrinsic_sender
```
kziemianek, line 344: In multiworker setup this would lead to duplicate calls, have you considered executing this logic by sidechain block producer only ?

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -201,17 +222,158 @@ impl ParentchainEventHandler {
...
+				Error::Other(
+					"Metadata retrieval for complete_reward_distribution_call_indexes failed"
+						.into(),
+				)
+			})?
+			.map_err(|_| Error::Other("Invalid metadata".into()))?;
+
+		let complete_reward_distribution_call =
+			OpaqueCall::from_tuple(&(complete_reward_distribution_call_index.clone()));
+		extrinsic_sender.send(complete_reward_distribution_call).map_err(|_| {
```
kziemianek, line 362: This will also be sent by N workers in multiworker setup - please make sure rewards are not distributed more than once.

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
silva-fj: In this case we are only taking into account the identities that are delegators, we are filtering out all the others

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
silva-fj: This is a very good point, I have not considered it 👍🏼

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
silva-fj: Calling this multiple times won't have any effect on the distribution, it's just to trigger an event. But it would be better to call it once 👍🏼

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -142,6 +147,10 @@ pub enum TrustedCall {
 	send_erroneous_parentchain_call(Identity),
 	#[codec(index = 25)]
 	maybe_create_id_graph(Identity, Identity),
+	#[codec(index = 25)]
+	update_token_staking_amounts(Identity, Vec<(AccountId, Balance)>),
+	#[codec(index = 27)]
+	reward_distribution_completed(Identity),
```
kziemianek: Can these two calls be merged into one ?

We assume they are executed together and maybe there is strict ordering requirement ?

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -142,6 +147,10 @@ pub enum TrustedCall {
 	send_erroneous_parentchain_call(Identity),
 	#[codec(index = 25)]
 	maybe_create_id_graph(Identity, Identity),
+	#[codec(index = 25)]
+	update_token_staking_amounts(Identity, Vec<(AccountId, Balance)>),
```
kziemianek: I think this call can be removed from the pool due to limits for example.
Maybe we should make sure there is no way to delay / prevent execution by regular rpc clients.

tee-worker/app-libs/stf/src/trusted_call.rs
silva-fj: After the discussions, we are going back to putting all the logic in the event_handler

tee-worker/app-libs/stf/src/trusted_call.rs
silva-fj: Same here. After the discussions, we are going back to putting all the logic in the event_handler

kziemianek (approved):

pallets/score-staking/src/lib.rs
```diff
@@ -68,6 +68,10 @@ pub trait AccountIdConvert<T: Config> {
 	fn convert(account: AccountId32) -> T::AccountId;
 }
 
+pub trait TokenStakingAuthorizer<T: Config> {
+	fn can_update_staking(sender: &T::AccountId) -> bool;
+}
+
```
Kailai-Wang: Hmm there's a `EnsureEnclaveSigner` that does similar things - maybe use that (and origin control)?

pallets/score-staking/src/lib.rs
```diff
@@ -445,6 +478,53 @@ pub mod pallet {
...
+					});
+					Ok(Pays::No.into())
+				},
+				None => Err(Error::<T>::UserNotExist.into()),
+			}
+		}
+
+		#[pallet::call_index(10)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn complete_reward_distribution(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
Kailai-Wang, line 511: Hmm strange - where is the token distribution process? (I thought it would be done within this extrinsic)

pallets/score-staking/src/lib.rs
silva-fj: I'll look for it 👍🏼

pallets/score-staking/src/lib.rs
silva-fj: This is just to trigger an event that marks the distribution for the round as completed. (Currently we don't do anything with the `token_staking_amount`, will there be another phase/issue to come?)

Kailai-Wang: No I mean "distribution for the round as completed" should be triggered by the worker - when it finishes with staking amount upload for all accounts, it should trigger the distribution

silva-fj: I have updated the code to make the distribution on `update_token_staking_amount`, this extrinsic will be kept to signal that the distribution for the round is completed.

BillyWooo (approved):

PR #2980, opened 2024-08-09
Add a burn method to pallet-account-fix

Kailai-Wang:
### Context

As topic, I believe a simple `burn_from` is enough.
Please note if `burn_from` fails, no balance change would occur.


wangminqi (approved):

wangminqi:
LGTM

PR #2988, opened 2024-08-13
P-805 add required pallets to Litentry

BillyWooo:
As title, there are some pallets, which are only available for rococo. We are going to enable it on Litentry parachain now.


runtime/litentry/src/lib.rs
```diff
@@ -1220,33 +1290,45 @@ impl Contains<RuntimeCall> for NormalModeFilter {
...
 			// BridgeTransfer
 			RuntimeCall::BridgeTransfer(_) |
-			// Utility
-			RuntimeCall::Utility(_) |
-			// Session
-			RuntimeCall::Session(_) |
-			// ParachainStaking
-			RuntimeCall::ParachainStaking(_) |
+			// XTokens::transfer for normal users
+			RuntimeCall::XTokens(orml_xtokens::Call::transfer { .. }) |
```
wangminqi: I do not think we want Xtoken enabled. It is not sure ready yet.

wangminqi (changes requested):

wangminqi (approved):
LGTM

PR #2992, opened 2024-08-15
Clear the score when the user fully unstakes

Kailai-Wang:
### Context

A small adjustment based on the new requirement.

When a user fully unstakes, his reward will become 0 according to the current logic, which is fine. However, his score shouldn't participate in the reward calculation, which is unfair to other users.

This PR fixes it in a brutal way - there's room to improve: both structural and algorithmetic way - but the time is limited, we can do the improvement later.

This PR simply clears the user's score for whoever has 0 staking when we calculate the rewards.
We can't simply remove the entry from `Scores` as the user might still have unclaimed reward.


wangminqi (approved):
LGTM

silva-fj (approved):

PR #2994, opened 2024-08-16
change sidechain block slot duration from 6s to 12s

BillyWooo:
tiny PR as title
We are suffering not enough time to sync parachain block and produce sidechain block.

### DIFF
--- a/tee-worker/core-primitives/settings/src/lib.rs
+++ b/tee-worker/core-primitives/settings/src/lib.rs
@@ -90,5 +90,5 @@ pub mod worker {
 pub mod sidechain {
 	use core::time::Duration;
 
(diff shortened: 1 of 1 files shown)


Kailai-Wang:
Let's add more contextual content to both title and PR description 🙌 
Have we tried dwellir endpoint?

BillyWooo:
> Let's add more contextual content to both title and PR description 🙌 Have we tried dwellir endpoint?

Yes. Tried dwellir endpoint. Situation is worse. The worker can't even start up successfully. Didn't look into details.

Kailai-Wang (approved):

BillyWooo:
After a discussion, we decide to first try another solution: disable sidechain slot duration check.

PR #2995, opened 2024-08-16
disable sidechain slot duration check

BillyWooo:
minor PR, as title
The worker is suffering from syncing parachain block and producing sidechain block on production env. So disable the slot duration checking for the moment. Only skip checking when it's single worker。

### DIFF
--- a/tee-worker/enclave-runtime/src/top_pool_execution.rs
+++ b/tee-worker/enclave-runtime/src/top_pool_execution.rs
@@ -185,10 +185,13 @@ fn execute_top_pool_trusted_calls_internal() -> Result<()> {
 		&mut LastSlot,
 	)? {
 		Some(slot) => {
-			if slot.duration_remaining().is_none() {
-				warn!("No time remaining in slot, skipping AURA execution");
-				return Ok(())
-			}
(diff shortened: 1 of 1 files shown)


tee-worker/enclave-runtime/src/top_pool_execution.rs
```diff
@@ -185,10 +185,13 @@ fn execute_top_pool_trusted_calls_internal() -> Result<()> {
...
 	)? {
 		Some(slot) => {
-			if slot.duration_remaining().is_none() {
-				warn!("No time remaining in slot, skipping AURA execution");
-				return Ok(())
-			}
+			// FIXME: To disable this checking is a temp solution for the moment.
+			// Later we need to investigate and double check again whether this is
+			// a good solution.
+			// if slot.duration_remaining().is_none() {
```
kziemianek: Can we keep this check and run it only if there are more than 1 enclave registered ?

Kailai-Wang: "Later we need to investigate" => I doubt if we'll do that...

kziemianek (approved):

PR #3004, opened 2024-08-20
add missing dcap arg within dockerfile

BillyWooo:
There was no RA_METHOD args within dockerfile. So the parameter from yml file is not passed into dockerfile when building release docker images. 
Ref: 

https://github.com/litentry/litentry-parachain/blob/f138676e152adb07fc1dc02572e1cae2d482ed51/.github/workflows/create-release-draft.yml#L219

https://github.com/litentry/litentry-parachain/blob/f138676e152adb07fc1dc02572e1cae2d482ed51/.github/workflows/create-release-draft.yml#L292


Kailai-Wang:
Shall we use DCAP in regular CI too?

BillyWooo:
> Shall we use DCAP in regular CI too?

Maybe not. Because it requires hardware. Regular CI runners don't have hardware.
Or do you mean to set up a self-host runner?

Kailai-Wang:
I think we should let it run with DCAP RA - you could double-check how these RA extrinsics are generated. It's a skip-ra variant, but it's a dcap version of skip-ra

BillyWooo:
> I think we should let it run with DCAP RA - you could double-check how these RA extrinsics are generated. It's a skip-ra variant, but it's a dcap version of skip-ra

Not really, this PR doesn't change any logic within ci.yml. So our general CI is not affected. DCAP feature will only be enabled  when building release package.

Kailai-Wang (approved):

PR #3006, opened 2024-08-20
Wait untill sidechain blocks are broadcasted

kziemianek:
Current impl performs broadcasting (rpc calls) in background making slot worker thread not waiting for the result. This cause recently observed problem with sidechain block production.

New impl should also broadcast block to peers in parallel using threads dedicated for handling blocking operations.


Kailai-Wang (approved):
Good finding!

PR #3008, opened 2024-08-21
Updating score staking on parachain unstaking events

silva-fj:
This PR adds a small improvement in the staking reward calculation by updating the scores whenever a user tokens are unstaked instead of iterating all scores on every round to update them. More context in the issue.


pallets/parachain-staking/src/mock.rs
```diff
@@ -149,6 +149,13 @@ impl Config for Test {
 	type OnNewRound = ();
 	type WeightInfo = ();
 	type IssuanceAdapter = ();
+	type ScoreUpdater = ();
```
Kailai-Wang: It can be a bit confusing that readers see the word `Score` in parachain-staking pallet, as it's completely unrelated.

Can we have a better name for the hook? Like `OnAllDelegationRemoved` or so

pallets/parachain-staking/src/delegation_requests.rs
```diff
@@ -248,6 +248,7 @@ impl<T: Config> Pallet<T> {
 				<DelegationScheduledRequests<T>>::insert(collator, scheduled_requests);
 				if leaving {
 					<DelegatorState<T>>::remove(&delegator);
+					let _ = T::ScoreUpdater::clear_score_for(&delegator);
```
Kailai-Wang: If I search `<DelegatorState<T>>::remove` in the code, there's a fourth place in `types.rs` - can you check if we need to trigger the hook there too?

Kailai-Wang: Can you also check what if the user unstakes all the amount (`bondLess`) instead of revoking the delegation? Will the hook be triggered too?

silva-fj: let me check 👍🏼

silva-fj: > If I search <DelegatorState<T>>::remove in the code, there's a fourth place in types.rs - can you check if we need to trigger the hook there too?

Yes, the delegator state is also removed when a bottom delegation is kicked out. I've updated the code to trigger the call.

> Can you also check what if the user unstakes all the amount (bondLess) instead of revoking the delegation? Will the hook be triggered too?

The are checks that prevent delegators and candidates to decrease the bond with an amount that is greater than the bonded amount, so in these cases we don't need to trigger it

Kailai-Wang: Does it mean the user can only unstake all money with the call to `revoke`?

silva-fj: From the delegator's perspective it can be done with revoke and leave_delegators. The stake is also returned on `leave_candidates` and when the list of delegators is full and another user bonds more, the bottom delegation is kicked out

pallets/score-staking/src/tests.rs
```diff
@@ -382,3 +411,27 @@ fn claim_works() {
 		);
 	});
 }
+
+#[test]
+fn clear_score_for_works() {
```
Kailai-Wang: I assume this test-case needs to be renamed too

runtime/litentry/src/lib.rs
```diff
@@ -867,6 +867,7 @@ impl pallet_parachain_staking::Config for Runtime {
 	type OnNewRound = ();
 	type WeightInfo = weights::pallet_parachain_staking::WeightInfo<Runtime>;
 	type IssuanceAdapter = AssetsHandler;
+	type OnAllDelegationRemoved = pallet_score_staking::Pallet<Runtime>;
```
Kailai-Wang: Does `... = ScoreStaking` work? For consistency

pallets/parachain-staking/src/lib.rs
```diff
@@ -194,6 +195,8 @@ pub mod pallet {
 		type WeightInfo: WeightInfo;
 		/// The source for adjusted inflation base.
 		type IssuanceAdapter: IssuanceAdapter<BalanceOf<Self>>;
+		/// ScoreStaking updater
```
Kailai-Wang: Comment needs to be adjusted

Kailai-Wang (approved):
Thanks!

runtime/litentry/src/lib.rs
silva-fj: Yes, updated 👍🏼

PR #3012, opened 2024-08-23
ts-tests: avoid hardcoded runtime versions

jonalvarezz:
Manually dumping the runtime version in our tests was easy to forget and led to another 1.5h CI before realizing it.

These changes parse the spec files from the Rust source to statically get the versions.

### Additional changes
* I had to modify some CI related files to ensure vc tests followed our regular file structure


tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -146,11 +146,35 @@ export async function assertVc(context: IntegrationTestContext, subject: CorePri
...
-        'Check VC runtime version: it should equal the current defined versions'
+    const specPaths = ['../../../runtime/litentry/src/lib.rs', '../../app-libs/sgx-runtime/src/lib.rs'].map(
+        (p) => path.resolve(process.cwd(), p) // resolve relative to working directory: integration-tests
     );
+    try {
+        const [nodeVersion, sidechainVersion] = await Promise.all(
+            specPaths.map((spec) => readFile(spec, { encoding: 'utf-8' }))
+        ).then((specs) =>
+            specs.map((spec) => {
+                const version = (spec.match(/spec_version:\s*(\d+)/) || [])[1];
```
jonalvarezz, line 157: a bit flaky, but wanted to give it a try before moving the spec versions into json files.

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
```diff
@@ -146,11 +146,35 @@ export async function assertVc(context: IntegrationTestContext, subject: CorePri
 
     // step 7
     // check runtime version is present
-    assert.deepEqual(
-        vcPayloadJson.issuer.runtimeVersion,
-        { parachain: 9195, sidechain: 109 },
-        'Check VC runtime version: it should equal the current defined versions'
+    const specPaths = ['../../../runtime/litentry/src/lib.rs', '../../app-libs/sgx-runtime/src/lib.rs'].map(
```
jonalvarezz, line 149: to confirm: are these the right sources?

jonalvarezz:
btw: I'm working on getting CI to pass. This approach is based on relative paths (file system) but the file structure is different in CI because of docker custom mounting volumes.

jonalvarezz:
alright, CI looks good now

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
Kailai-Wang: Our CI uses rococo I believe

tee-worker/ts-tests/integration-tests/common/utils/assertion.ts
Kailai-Wang: That's fine : )

tee-worker/cli/lit_ts_integration_test.sh
```diff
@@ -75,7 +75,13 @@ if [ "$TEST" = "assertion_contracts.test.ts" ]; then
...
 fi
 
-cd /ts-tests
+# VC assertions requires our regular file structure
+if [[ "$TEST" == *"vc"* ]]; then
+    cd /app/tee-worker/ts-tests
+else
+    cd /ts-tests
+fi
+
```
Kailai-Wang, line 84: Aren't they the same path?

both `<parahcain-dir>/tee-worker/ts/-tests` - If I calculate it right

jonalvarezz: yes, but we need the entire folder structure for this work both locally and on CI: https://github.com/litentry/litentry-parachain/blob/e5f109f9ab6ae0935f86c3120c6e8a29c84b5e93/tee-worker/ts-tests/integration-tests/common/utils/assertion.ts#L149

what's a better comment to leave in the to signal it? 🤔

Kailai-Wang: Then shall we always use `cd /app/tee-worker/ts-tests`? And only mount `/app` without mounting `ts-tests` alone?

Kailai-Wang: Btw another way is to query the runtime version via RPC - that should be easy too 

Actually it should be slightly more accurate: imagine you code has bumped the runtime version but your chain runtime hasn't been updated yet

jonalvarezz: > Then shall we always use cd /app/tee-worker/ts-tests? And only mount /app without mounting ts-tests alone?

Yes, I'm in with this one but wanted to leave it for a follow-up pull requests as multiple files need update.

> Btw another way is to query the runtime version via RPC - that should be easy too
> 
> Actually it should be slightly more accurate: imagine you code has bumped the runtime version but your chain runtime hasn't been updated yet

Do you want me to close this PR and use the RPC better? If so, would you mind pointing out the RPC method name for me, please?

jonalvarezz: cc @Kailai-Wang

Kailai-Wang: Should be this one (use `specVersion`):
image

Kailai-Wang: And yes please I suggest we close this PR and open a new one using RPC : )

PR #3018, opened 2024-08-26
Remove Litmus from runtime

Kailai-Wang:
### Context

This PR removes Litmus runtime integration.

One thing that I'm not very sure is `Litmus` in the Assertion Web3Network, I keep it for now for backwards compatibility while removing some of the usage (e.g. when calculating LIT balance).

I'd like to remove it completely, but unsure about the consequence.
Did we have some problems when removing `SubstrateTestNetwork`?

Please take a look @jonalvarezz @higherordertech


jonalvarezz:
> I'd like to remove it completely, but unsure about the consequence.
> Did we have some problems when removing `SubstrateTestNetwork`?

Yes, I remember we had to put it back considering the effort of a migration

jonalvarezz (approved):
lgtm 

I agree that it is safer to do the Web3Network clean up in a more controlled set of changes, if any

silva-fj (approved):

PR #3030, opened 2024-08-29
CollabAI implementation

wangminqi:
TBA


wangminqi:
unittest of aiusd part merged.

wangminqi:
unittest of guardian part merged.

wangminqi:
unittest of curator part merged.

parachain/pallets/collab-ai/aiusd-convertor/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		// Lock target_asset_id token and mint AIUSD
+		#[pallet::call_index(0)]
+		#[pallet::weight({195_000_000})]
+		#[transactional]
+		pub fn mint_aiusd(
+			origin: OriginFor<T>,
+			target_asset_id: AssetIdOf<T>,
```
Kailai-Wang: Do you actually mean `source_asset_id`? It's a conversion from _some-asset_ to AIUSD, right?

parachain/pallets/collab-ai/aiusd-convertor/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+			aiusd_amount: AssetBalanceOf<T>,
+		) -> DispatchResult {
+			let beneficiary = ensure_signed(origin)?;
+			if let Some(ratio) = EnabledTokens::<T>::get(&target_asset_id) {
+				ensure!(
+					EnabledTokens::<T>::contains_key(&target_asset_id),
+					Error::<T>::AssetNotEnabled
+				);
```
Kailai-Wang: reduntant `ensure!` - you are already in `if let` branch

parachain/pallets/collab-ai/aiusd-convertor/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+				let aiusd_id = T::AIUSDAssetId::get();
+				ensure!(
+					InspectFungibles::<T>::asset_exists(aiusd_id.clone())
+						&& InspectFungibles::<T>::asset_exists(target_asset_id.clone()),
+					Error::<T>::InvalidAssetId
+				);
+				// It will fail if insufficient fund
```
Kailai-Wang: Comment is wrong? It can (only) fail because of overflow

parachain/pallets/collab-ai/aiusd-convertor/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+			decimal_ratio: AssetBalanceOf<T>,
+		},
+		AssetDisabled {
+			asset_id: AssetIdOf<T>,
+		},
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		// Lock target_asset_id token and mint AIUSD
```
Kailai-Wang: Why lock, not burn? Burn + mint seems more natural.

If you want to actually swap two assets in a "pool", maybe [pallet-assert-conversion](https://github.com/paritytech/polkadot-sdk/tree/release-polkadot-v1.1.0/substrate/frame/asset-conversion) is better? Your `ConvertingFeeAccount` seems like a pool-like account

parachain/pallets/collab-ai/common/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+pub const INVESTING_POOL_INDEX_SHIFTER: u128 = 1_000_000_000_000_000;
+pub const INVESTING_POOL_START_EPOCH_SHIFTER: u128 = 1_000;
+pub const INVESTING_POOL_END_EPOCH_SHIFTER: u128 = 1;
+
+pub struct InvestingPoolAssetIdGenerator<AssetId>(PhantomData<AssetId>);
+impl<AssetId: From<u128> + Into<u128>> InvestingPoolAssetIdGenerator<AssetId> {
+	/// Create a series of new asset id based on pool index and reward epoch
+	/// Return None if impossible to generate. e.g. overflow
+	pub fn get_all_pool_token(pool_index: InvestingPoolIndex, epoch: u128) -> Option<Vec<AssetId>> {
```
Kailai-Wang, line 190: This vec can be very large if epoch is large, right? Why do we need such a large assetId collection? It can be very inefficient for iterations

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+//
+//! # Curator Pallet
+//!
+//! - [`Config`]
+//! - [`Call`]
+//!
+//! ## Overview
+//!
+//! The Curator pallet handles the administration of general curator and proposed staking pool.
```
Kailai-Wang, line 24: Is "staking pool" and "investing pool" the same thing? We should use onle of them consistently

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+		Twox64Concat,
+		CuratorIndex,
+		(InfoHash, BlockNumberFor<T>, T::AccountId, CandidateStatus),
+		OptionQuery,
+	>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		CuratorRegisted {
```
Kailai-Wang, line 108: Typo: `Registered`

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+
+	#[pallet::type_value]
+	pub fn DefaultForCuratorIndex() -> CuratorIndex {
+		1u128
+	}
+
+	/// The next curator index, the number of (public) curator that have been made so far + 1.
+	#[pallet::storage]
+	#[pallet::getter(fn public_curator_count)]
+	pub type PublicCuratorCount<T> =
```
Kailai-Wang, line 85: Hmm why is it called `PublicCuratorCount`, instead of e.g. `NextCuratorIndex`:
- it's uncommon that "count" is 1 even for no items
- is there "private" curator?

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+		CuratorIndexNotExist,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Registing a curator legal info
+		#[pallet::call_index(0)]
+		#[pallet::weight({195_000_000})]
+		#[transactional]
+		pub fn regist_curator(origin: OriginFor<T>, info_hash: InfoHash) -> DispatchResult {
```
Kailai-Wang, line 142: `register`? (Other places too)

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn public_curator_count)]
+	pub type PublicCuratorCount<T> =
+		StorageValue<_, CuratorIndex, ValueQuery, DefaultForCuratorIndex>;
+
+	/// The public curator to index
+	#[pallet::storage]
+	#[pallet::getter(fn public_curator_to_index)]
+	pub type PublicCuratorToIndex<T: Config> =
+		StorageMap<_, Twox64Concat, T::AccountId, CuratorIndex, OptionQuery>;
```
Kailai-Wang, line 92: If anyone can register as a curator, we'd better use Blake256 Hahser here

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+				PublicCuratorToIndex::<T>::get(&who).ok_or(Error::<T>::CuratorNotRegistered)?;
+
+			// Update curator
+			// But if banned, then require extra reserve
+			CuratorIndexToInfo::<T>::try_mutate_exists(
+				curator_index,
+				|maybe_info| -> Result<(), DispatchError> {
+					let info = maybe_info.as_mut().ok_or(Error::<T>::CuratorIndexNotExist)?;
+
+					if info.3 == CandidateStatus::Banned {
```
Kailai-Wang, line 191: If makes sense to put the curator info into a struct: `.3` is not that readable, and maybe error-prone: imagine there's more than one field that has the same type

Kailai-Wang: So for banned curators, everytime I want to update my info, I have to reserve `MinimumCuratorDeposit` balance?

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+				PublicCuratorToIndex::<T>::take(&who).ok_or(Error::<T>::CuratorNotRegistered)?;
+
+			// Update curator
+			// But if banned, then require extra reserve
+			CuratorIndexToInfo::<T>::try_mutate_exists(
+				curator_index,
+				|maybe_info| -> Result<(), DispatchError> {
+					let info = maybe_info.as_ref().ok_or(Error::<T>::CuratorIndexNotExist)?;
+
+					if info.3 != CandidateStatus::Banned {
```
Kailai-Wang, line 228: So for banned curators, the item will be removed but deposit never returned, right?

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+		#[pallet::weight({195_000_000})]
+		pub fn judge_curator_status(
+			origin: OriginFor<T>,
+			curator: T::AccountId,
+			status: CandidateStatus,
+		) -> DispatchResult {
+			T::CuratorJudgeOrigin::ensure_origin(origin)?;
+			let curator_index = PublicCuratorToIndex::<T>::get(curator.clone())
+				.ok_or(Error::<T>::CuratorNotRegistered)?;
+			CuratorIndexToInfo::<T>::try_mutate_exists(
```
Kailai-Wang, line 251: I think `try_mutate` is enough

parachain/pallets/collab-ai/curator/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+		if let Some(curator_index) = PublicCuratorToIndex::<T>::get(&account) {
+			if let Some(info) = CuratorIndexToInfo::<T>::get(curator_index) {
+				if info.3 == CandidateStatus::Verified {
+					return true;
+				}
+			}
+		}
+
+		false
+	}
```
Kailai-Wang, line 297: could be merged into something like `assert_curator_status`

parachain/pallets/collab-ai/guardian/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+	#[pallet::type_value]
+	pub fn DefaultForGuardianIndex() -> GuardianIndex {
+		1u128
+	}
+
+	/// Next guardian index, the number of (public) guardian that have been made so far + 1.
+	#[pallet::storage]
+	#[pallet::getter(fn public_guardian_count)]
+	pub type PublicGuardianCount<T> =
+		StorageValue<_, GuardianIndex, ValueQuery, DefaultForGuardianIndex>;
```
Kailai-Wang, line 86: Similar comments to `curator` pallet

parachain/pallets/collab-ai/guardian/src/lib.rs
Kailai-Wang, line 1: Can this be merged into `curator`? Their ways of registration, removal, verification are quite similar, we need need some `enum Role` and maintain the mapping to metadata/config

parachain/pallets/collab-ai/guardian/src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+						status,
+					});
+					Ok(())
+				},
+			)?;
+			Ok(())
+		}
+
+		/// Anyone can vote for guardian
+		/// However if voter is not participating the staking pool
```
Kailai-Wang, line 294: I don't think the impl reflects that

parachain/pallets/collab-ai/investing-pool/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+	pub type InvestingPoolSetting<T: Config> = StorageMap<
+		_,
+		Twox64Concat,
+		InvestingPoolIndex,
+		PoolSetting<T::AccountId, BlockNumberFor<T>, BalanceOf<T>>,
+		OptionQuery,
+	>;
+
+	// investing pools' stable token reward waiting claiming
+	// Pool id, epcoh index => epoch reward
```
Kailai-Wang: typo: `epoch`

parachain/pallets/collab-ai/investing-pool/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+		AccountIdConversion, AtLeast32BitUnsigned, CheckedAdd, CheckedDiv, CheckedMul, One, Zero,
+	},
+	ArithmeticError, Perquintill, Saturating,
+};
+use sp_std::{fmt::Debug, prelude::*};
+
+use pallet_collab_ai_common::*;
+
+#[derive(PartialEq, Eq, Clone, Encode, Debug, Decode, TypeInfo)]
+pub struct InvestingWeightInfo<BlockNumber, Balance> {
```
Kailai-Wang, line 48: Why is this struct related to `Weight` or `WeightInfo`? 🤔

parachain/pallets/collab-ai/investing-pool/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+
+	// Asset id of AIUSD
+	#[pallet::storage]
+	#[pallet::getter(fn aiusd_asset_id)]
+	pub type AIUSDAssetId<T: Config> = StorageValue<_, AssetIdOf<T>, OptionQuery>;
+
+	// Asset id of CAN
+	#[pallet::storage]
+	#[pallet::getter(fn can_asset_id)]
+	pub type CANAssetId<T: Config> = StorageValue<_, AssetIdOf<T>, OptionQuery>;
```
Kailai-Wang, line 263: Why do they need to have their own storages? (Instead of retrieving from `pallet-assets`)

parachain/pallets/collab-ai/investing-pool/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+		EpochNotExist,
+		NoAssetId,
+		TypeIncompatibleOrArithmeticError,
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		/// Weight: see `begin_block`
+		fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+			Weight::zero()
```
Kailai-Wang, line 354: empty impl?

parachain/pallets/collab-ai/investing-pool/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+			pool_id: InvestingPoolIndex,
+			setting: PoolSetting<T::AccountId, BlockNumberFor<T>, BalanceOf<T>>,
+			admin: T::AccountId,
+		) -> DispatchResult {
+			T::PoolProposalPalletOrigin::ensure_origin(origin)?;
+
+			// Create all asset token categories
+			let asset_id_vec: Vec<AssetIdOf<T>> =
+				InvestingPoolAssetIdGenerator::get_all_pool_token(pool_id, setting.epoch)
+					.ok_or(ArithmeticError::Overflow)?;
```
Kailai-Wang: As previously mentioned, I fail to understand why we need a different token asset for each epoch 😂

parachain/pallets/collab-ai/investing-pool/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+			}
+
+			ensure!(
+				frame_system::Pallet::<T>::block_number() <= setting.start_time,
+				Error::<T>::PoolAlreadyStarted
+			);
+			ensure!(
+				!InvestingPoolSetting::<T>::contains_key(pool_id),
+				Error::<T>::PoolAlreadyExisted
+			);
```
Kailai-Wang: Traditionally let's check before write

parachain/pallets/collab-ai/investing-pool/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+	// None means TypeIncompatible Or Overflow
+	fn weight_force(&self, n: BlockNumber) -> Option<u128> {
+		let e: u128 = n.checked_sub(&self.effective_time)?.try_into().ok()?;
+		let s: u128 = self.amount.try_into().ok()?;
+		e.checked_mul(s)
+	}
+}
+
+#[derive(PartialEq, Eq, Clone, Encode, Debug, Decode, TypeInfo)]
+pub struct PoolSetting<AccountId, BlockNumber, Balance> {
```
Kailai-Wang: We have a `PoolStetting` already in common - which is the valid one?

parachain/pallets/collab-ai/investing-pool/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+					Self::deposit_event(Event::<T>::RewardUpdated {
+						pool_id,
+						epoch,
+						amount: reward,
+					});
+					Ok(())
+				},
+			)?;
+
+			// Mint AIUSD into reward pool
```
Kailai-Wang, line 411: Shall we check the current block number?
Does it mean we can set reward for future epochs? If this is not a problem, why not mint the rewards already when the pool is created/started where the epoch is known?

parachain/pallets/collab-ai/pool-proposal/src/lib.rs
```diff
@@ -0,0 +1,638 @@
...
+
+			Weight::zero()
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Curator propose a investing pool
+		///
+		/// max_pool_size: At most this amount of raised money curator/investing pool willing to take
```
Kailai-Wang, line 333: In which currency?

Kailai-Wang (commented):
I only did a very rough review.
I didn't look deep into `investing-pool` and `pool-proposal` as those two pallets are closely related to requirements/context.

parachain/pallets/collab-ai/pool-proposal/src/lib.rs
wangminqi: max_pool_size will be pallet_assets balance type

parachain/pallets/collab-ai/aiusd-convertor/src/lib.rs
wangminqi: fixed.

parachain/pallets/collab-ai/aiusd-convertor/src/lib.rs
wangminqi: fixed.

parachain/pallets/collab-ai/aiusd-convertor/src/lib.rs
wangminqi: changed

parachain/pallets/collab-ai/common/src/lib.rs
wangminqi: we create a series of tokens, so if user claimed one epoch reward, it will be converted into one equal amount of another type, until epoch overlapping the ending epoch => leads to nothing.

We actually do not iterate through all asset ids in logic.

parachain/pallets/collab-ai/investing-pool/src/lib.rs
wangminqi: This design is for each token can be considered as ERC20, so the dex can be used directly to support.

Tracking epoch reward claim status is hard without destroying the token's fungible property.

parachain/pallets/collab-ai/investing-pool/src/lib.rs
wangminqi: moved

parachain/pallets/collab-ai/investing-pool/src/lib.rs
wangminqi: leftover code, common Poolsetting deleted

parachain/pallets/collab-ai/investing-pool/src/lib.rs
wangminqi: You are 100% correct. But I plan to add one extra pallet ( responsible for epcoh reward submitting). 
But it will not be in this PR

parachain/pallets/collab-ai/aiusd-convertor/src/lib.rs
wangminqi: Switch to ConvertingFeeAccount to ConvertingPool.

We use lock since we want to make sure that the bridge on remote chains (ethereum, bsc etc.) have enough token to support the swap. Otherwise we will lose track of distribution of assets.

Kailai-Wang: Do you mean we need something to track the liquidity so that we can't e.g. bridge out more assets than what exists in the pool?

Then I still think pallet-asset-conversion is a better solution, when bridging in, you are actually adding liquidity to the pool and swapping the for AIUSD, it's also guaranteed that when all assets are drained, you can't do the swap anymore.

Actually it should be used for our chain bridge too 🤔

parachain/pallets/collab-ai/common/src/lib.rs
Kailai-Wang: I think we do, in `create_investing_pool`:

```
// Create all asset token categories
let asset_id_vec: Vec<AssetIdOf<T>> =
InvestingPoolAssetIdGenerator::get_all_pool_token(pool_id, setting.epoch)
	.ok_or(ArithmeticError::Overflow)?;
for i in asset_id_vec.iter() {
<T::Fungibles as FsCreate<<T as frame_system::Config>::AccountId>>::create(
	*i,
	admin.clone(),
	true,
	One::one(),
)?;
}
```

> we create a series of tokens, so if user claimed one epoch reward, it will be converted into one equal amount of another type, until epoch overlapping the ending epoch => leads to nothing.

Maybe I still don't get the requirements, probably we need a sync

wangminqi:
A runtime upgrade of paseo have been executed at commit [66f40ee]

parachain/runtime/paseo/src/lib.rs
```diff
@@ -1036,6 +1039,87 @@ impl pallet_group::Config<VCMPExtrinsicWhitelistInstance> for Runtime {
 	type GroupManagerOrigin = EnsureRootOrAllCouncil;
 }
 
+parameter_types! {
+	pub const MinimumCuratorDeposit: Balance = 100 * DOLLARS;
+	pub const MinimumGuardianDeposit: Balance = 20 * DOLLARS;
+	// Declare the official AIUSDAssetId
+	pub const AIUSDAssetId: u128 = 1000;
+	pub OfficialGapPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 10 * MINUTES, "ROCOCO_OFFICIALGAPPERIOD");
```
BillyWooo, line 1047: Maybe we should change the name from `ROCOCO` to `Paseo`. There are two more below.
Besides, for non-prod, is `10*MINUTES` the best choice? is there any concern to run non-prod env faster?

wangminqi: Actually the rococo name is still remained everywhere, not only brought in by the collabAI part code. I will raise another PR to replace the name.

wangminqi: the offficial gap period is the time between a pool proposal mature and an official investing pool begin accounting.
In reality, it should be equal to the approximate time of GPU data center setup.
In testing mode, any timeframe will work.

BillyWooo (approved):
LGTM

PR #3035, opened 2024-09-02
Add `SignTon` direct call

kziemianek:
* Adds `Ed25519` keypair repository dedicated for signing Ton transactions
* Adds `SignTon` direct call
* Adjust `bitacross` pallet for storing worker's Ton public key
* `init-wallet` command can selectively init wallets

I've manually checked signing correctness with implementation using `tweetnacl` npm package.


Kailai-Wang (approved):
Thanks, not much to complain

PR #3040, opened 2024-09-02
ts-tests: avoid hardcoded runtime version

jonalvarezz:
update the vc assertions to check the runtime version dynamically instead.

### DIFF
--- a/tee-worker/enclave-runtime/src/rpc/common_api.rs
+++ b/tee-worker/enclave-runtime/src/rpc/common_api.rs
@@ -9,7 +9,7 @@ use crate::{
 use base58::FromBase58;
 use codec::{Decode, Encode};
 use core::result::Result;
-use ita_sgx_runtime::{Runtime, System};
+use ita_sgx_runtime::{Runtime, System, VERSION};
 use ita_stf::{aes_encrypt_default, AesOutput, Getter, TrustedCallSigned};
 use itc_parentchain::light_client::{concurrent_access::ValidatorAccess, ExtrinsicSender};
 use itp_ocall_api::EnclaveAttestationOCallApi;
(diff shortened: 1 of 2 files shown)


tee-worker/enclave-runtime/src/rpc/common_api.rs
```diff
@@ -233,8 +233,18 @@ pub fn add_common_api<Author, GetterExecutor, AccessShieldingKey, OcallApi, Stat
...
+		Ok(json!({
+			// "specName": runtime_version.spec_name,
+			// "implName": runtime_version.impl_name,
+			"authoringVersion": runtime_version.authoring_version,
+			"specVersion": runtime_version.spec_version,
+			"implVersion": runtime_version.impl_version,
+			"apis": runtime_version.apis,
+			"transactionVersion": runtime_version.transaction_version,
+			"stateVersion": runtime_version.state_version,
+		}))
```
jonalvarezz: What's a better way of achieving this for [`RuntimeVersion`](https://docs.rs/sp-runtime/7.0.0/sp_runtime/enum.RuntimeString.html)? 

I see that it implements a `serde::Serialize` trait:

https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/primitives/version/src/lib.rs#L159

but `json!(runtime_version)` will fail with `the trait Serialize is not implemented for sp_version::RuntimeVersion`

And it also has this `Display` implementation:

https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/primitives/version/src/lib.rs#L261


cc @kziemianek

Kailai-Wang: Hmm it seems that all rpc return value is scale & hex encoded (using `to_hex()`), see other methods.

Maybe it's better to follow this to be consistent

Kailai-Wang (approved):

tee-worker/enclave-runtime/src/rpc/common_api.rs
jonalvarezz: Done in https://github.com/litentry/litentry-parachain/pull/3040/commits/75991ca658c6fd1f74b8515856b9b08cff997f90

PR #3041, opened 2024-09-03
Refactoring ts integration tests to get worker url from the parachain

silva-fj:
The key changes in this PR are:

- We no longer use a hardcoded worker url, instead we get the url from the registered enclaves in the parachain.
- We select a random enclave from the registry.

More context in the linear issue

### DIFF
--- a/tee-worker/ts-tests/integration-tests/common/utils/context.ts
+++ b/tee-worker/ts-tests/integration-tests/common/utils/context.ts
@@ -1,4 +1,4 @@
-import { WsProvider, ApiPromise } from 'parachain-api';
+import { WsProvider, ApiPromise, PalletTeebagEnclave } from 'parachain-api';
 import { cryptoWaitReady } from '@polkadot/util-crypto';
 import { hexToString } from '@polkadot/util';
 import WebSocketAsPromised from 'websocket-as-promised';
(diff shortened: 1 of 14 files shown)


tee-worker/ts-tests/integration-tests/common/utils/context.ts
```diff
@@ -63,6 +61,22 @@ export async function initIntegrationTestContext(
...
+            return enclaves;
+        }
+        return [...enclaves, enclave.unwrap()];
+    }, [] as PalletTeebagEnclave[]);
+
+    if (identityEnclaves.length === 0) {
+        throw new Error('No identity worker found');
+    }
+
+    return identityEnclaves[Math.floor(Math.random() * identityEnclaves.length)].url.toHuman() as string;
```
jonalvarezz, line 77: maybe the `.toString()` saves from forcing the type

jonalvarezz (approved):
lgtm

0xverin (approved):

tee-worker/ts-tests/integration-tests/common/utils/context.ts
silva-fj: it needs to be casted as well because doing `toString()` to `AnyJson` returns `string | undefined`

PR #3043, opened 2024-09-04
remove redundant wrong dcap parameter from dockerfile

BillyWooo:
As title, this bug was introduced during the two sequential different commits. `dcap` is currently enabled by default from Makefile. So don't need this param in dockerfile any more. 

TODO: could consider to remove `ias` mode code and only keep `dcap`?


jonalvarezz (approved):
Thanks @BillyWooo , 

I tried the following:
* log into `azure-dev-2`
* checkout to git branch `fix_dcap_wrong_param`
* build worker with `make`
* run with `./local-setup/launch.py --parachain standalone`

it worked 🙌

PR #3048, opened 2024-09-06
Adjust the parameter input of enclave endpoint

0xverin:
### Context

`tee-dev` and `tee-staging` currently still get the enclave URL through teebag as `localhost`.
After confirming with @simple.xu, there may be other issues with using the enclave domain here. In order to avoid being blocked,  simply adjust the parameters of the enclave endpoint.

image

This results in ci errors:
https://github.com/litentry/litentry-parachain/actions/runs/10701200078


tee-worker/ts-tests/integration-tests/common/utils/context.ts
```diff
@@ -27,7 +27,10 @@ export async function initWorkerConnection(endpoint: string): Promise<WebSocketA
     return wsp;
 }
 
-export async function initIntegrationTestContext(substrateEndpoint: string): Promise<IntegrationTestContext> {
+export async function initIntegrationTestContext(
+    substrateEndpoint: string,
```
Kailai-Wang: can you rename `substrateEndpoint` to `parachainEndpoint`? Should be more clear 🙌

tee-worker/ts-tests/integration-tests/vc_correctness.test.ts
```diff
@@ -43,7 +43,7 @@ describe('Test Vc (direct invocation)', function () {
     const errorArray: { id: string; index: number; assertion: any; error: any }[] = [];
     this.timeout(6000000);
     before(async () => {
-        context = await initIntegrationTestContext(nodeEndpoint);
+        context = await initIntegrationTestContext(nodeEndpoint, enclaveEndpoint);
```
Kailai-Wang, line 46: There're more places using `initIntegrationTestContext` - do they need to be adjusted too, or only this test?

0xverin: Currently only vc correctness is testing tee-dev, the rest are ci staging tests, no need to change.

tee-worker/ts-tests/integration-tests/common/utils/context.ts
0xverin: And I also renamed all the variable/env names related to the endpoint.
for example:
`workerEndpoint`=>`enclaveEndpoint`
`substrateEndpoint/nodeEndpoint `=> `parachainEndpoint`

Kailai-Wang (approved):
WorkerEndpoint vs EnclaveEndpoint -> I slightly favor worker 😂 
But I'm fine with either

0xverin:
> WorkerEndpoint vs EnclaveEndpoint -> I slightly favor worker 😂 But I'm fine with either

I see client-sdk is using "enclave" for naming, so I kept it consistent.
image

jonalvarezz:
Worker, Enclave, Sidechain, Node, Parachain, Mainnet, parentchain.

I've seen we use them interchangeably 😂. I know it isn't _technically_ wrong, but we might want to align how we want to communicate to avoid confusion in the community.

PR #3050, opened 2024-09-06
Clean up STF request_vc call

Kailai-Wang:
### Context

As topic.

We only use `author_requestVc` to send vc request, so both STF rust and ts part are removed.

It includes a bunch of clippy warning fixes too.


tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -132,8 +128,7 @@ pub enum TrustedCall {
 		Option<RequestAesKey>,
 		H256,
 	),
-	#[codec(index = 21)]
-	request_vc_callback(Identity, Identity, Assertion, Vec<u8>, Option<RequestAesKey>, bool, H256),
+	// index = 21 was `request_vc_callback`, which was deprecated after native requestVc was introduced
```
kziemianek: following TCs doesn't seem to be requested by anyone from outside, maybe we can adjust indices ?

If somebody uses old indices it should still fail at decoding phase.

tee-worker/app-libs/stf/src/trusted_call.rs
```diff
@@ -781,109 +775,14 @@ where
...
-						calls,
-						node_metadata_repo,
-						Some(who),
-						e.to_vcmp_error(),
-						req_ext_hash,
-					);
-					e
-				})?;
-				Ok(TrustedCallResult::Streamed)
+			TrustedCall::request_vc(_signer, _who, _assertion, _maybe_key, _req_ext_hash) => {
```
kziemianek, line 777: Have you considered just removing this variant? 
We got only one client and the sooner the better

Kailai-Wang: It's there just to construct the request so that we can reuse some code (e.g. enc/dec)
it should have a different category enum (`NativeRequest`) to `TrustedCall` - but I'll leave it to the issue regarding `NativeRequest`

tee-worker/app-libs/stf/src/trusted_call.rs
Kailai-Wang: That's true  .. I'm fine with either way

Kailai-Wang: I adjusted the index

tee-worker/app-libs/stf/src/trusted_call.rs
kziemianek: Ah I forgot direct calls are coupled to stf 🫤

jonalvarezz (approved):
My only concern is that we will no longer able to use Polkadot.js App to request VCs, right?  

However, since we are working on these series of playground tools, that part should be covered. so, nothing against merging.

Kailai-Wang:
Yes it was discussed before - we have your sandbox :)

felixfaisal (approved):
LGTM

PR #3061, opened 2024-09-11
Adding support for email identity

silva-fj:
This PR adds support for email identities.

Important notes: 
We have some limitations due to our current rust tool chain version and the compatibility with sgx environment, so we can't use the available libraries to send emails (via SMTP). After exploring some options available I found [SendGrid](https://sendgrid.com/) to be the most convenient option (IMO). It provides an api we can use to send emails via http, the services are reliable and the pricing is reasonable. So, I have implemented a sengrid mailer. Having said that, we could easily migrate to other options if needed in the future.

We may as well need some design for the email template, for now I've added an example template that looks like this:
 
Screenshot 2024-09-11 at 4 42 33 PM

The flow is the following:
- User makes an RPC request to `identity_requestEmailVerification` with the prime identity and the email address.
- The user receives an email with the verification code.
- The user makes a request to link the identity sending the email and verification code as validation data.

We need to create/setup an account in SendGrid and setup the email address that is going to be used to send the verification templates, then the following env variables need to be defined:

- `SENDGRID_API_KEY`
- `SENDGRID_FROM_EMAIL`

This PR (https://github.com/litentry/litentry-parachain/pull/3063) needs to be merged first

> [!IMPORTANT]
Please note that similar to the OAuth2 verification, we need to implement "native" requests for this to work with multi-worker setup, as verification codes are stored in memory and not synced between workers.


tee-worker/litentry/core/identity-verification/src/web2/email/mailer/sendgrid_mailer.rs
```diff
@@ -0,0 +1,97 @@
...
+
+		Self { client: RestClient::new(http_client, base_url), from: from_email }
+	}
+}
+
+impl Mailer for SendGridMailer {
+	fn send(&mut self, mail: Mail) -> Result<(), String> {
+		let sendgrid_mail = SendGridMail::new(self.from.clone(), mail);
+		self.client
+			.post(String::default(), &sendgrid_mail)
```
kziemianek, line 92: This means sendgrid account owner knows what emails were linked and when ?

silva-fj: This is an interesting point. The SendGrid account owner could know what emails requested the verification code in the last few days, but this is true for any service. We would have to setup and manage our own SMTP server if we want to prevent this, but even if we do that it would be very difficult to convince users that this is really the case. There has to be some level of trust for this solution. Another solution could be to add the option to verify with ZK proofs, so users can decide what to use. The tradeoff is that the UX is really bad

tee-worker/litentry/core/identity-verification/src/web2/email/mailer/mod.rs
```diff
@@ -0,0 +1,14 @@
+pub mod sendgrid_mailer;
+pub(crate) mod template;
+
+use std::string::String;
```
Kailai-Wang: How about using `alloc` instead of `std` wherever possible?

tee-worker/litentry/core/identity-verification/src/web2/mod.rs
```diff
@@ -200,6 +212,44 @@ pub fn verify(
...
+			let stored_verification_code = match email::VerificationCodeStore::get(&account_id) {
+				Ok(data) => data.ok_or_else(|| {
+					Error::LinkIdentityFailed(ErrorDetail::StfError(ErrorString::truncate_from(
+						std::format!(
+							"no verification code found for {}",
+							account_id_to_string(&account_id)
+						)
+						.as_bytes()
+						.to_vec(),
+					)))
```
Kailai-Wang, line 232: There's an `IntoErrorDetail` trait, can be something like:
```
impl IntoErrorDetail for Error {
	fn into_error_detail(self) -> ErrorDetail {
		ErrorDetail::StfError(ErrorString::truncate_from(format!("{}", self).into()))
	}
}
```

Kailai-Wang (approved):
I haven't tested it but the logic looks good!

Do you know if we still have SendGrid account? @jonalvarezz 
I remember we had it for MCP

tee-worker/litentry/core/identity-verification/src/web2/mod.rs
silva-fj: I've improved the error handling below a bit (implemented `IntoErrorDetail` for the `VerificationCodeStoreError`), but in this case it's an `Option` that is converted to `Result`, so there is no error to implement the trait.

PR #3074, opened 2024-09-18
Identity staking improvement

silva-fj:
Updated: This PR improves the logic of the identity token distribution, the distribution is being done in batches updating the rewards for each account only once per round.

### DIFF
--- a/pallets/score-staking/src/lib.rs
+++ b/pallets/score-staking/src/lib.rs
@@ -174,7 +174,7 @@ pub mod pallet {
 		ScoreRemoved { who: Identity },
 		ScoreCleared {},
 		RewardDistributionStarted { round_index: RoundIndex },
-		RewardDistributionCompleted { distributed: BalanceOf<T> },
+		RewardDistributionCompleted { round_index: RoundIndex },
 		RewardClaimed { who: T::AccountId, amount: BalanceOf<T> },
 	}
 
@@ -460,8 +460,6 @@ pub mod pallet {
 			let n = Self::round_config().stake_coef_n;
 			let m = Self::round_config().stake_coef_m;
 
-			let mut all_user_reward = BalanceOf::<T>::zero();
-
 			for (a, mut p) in Scores::<T>::iter() {
 				let default_staking = BalanceOf::<T>::zero();
 				let id_graph_staking = id_graphs_staking_map.get(&a).unwrap_or(&default_staking);
@@ -480,15 +478,23 @@ pub mod pallet {
 				p.last_round_reward = user_reward;
 				p.total_reward += user_reward;
 				p.unpaid_reward += user_reward;
-				all_user_reward += user_reward;
 				Scores::<T>::insert(&a, p);
 			}
 
+			Ok(Pays::No.into())
+		}
+
+		#[pallet::call_index(10)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn complete_reward_distribution(
+			origin: OriginFor<T>,
+			round_index: RoundIndex,
+		) -> DispatchResultWithPostInfo {
+			let _ = T::TEECallOrigin::ensure_origin(origin)?;
+
 			LastTokenDistributionRound::<T>::put(round_index);
 
-			Self::deposit_event(Event::RewardDistributionCompleted {
-				distributed: all_user_reward,
-			});
+			Self::deposit_event(Event::RewardDistributionCompleted { round_index });
 
 			Ok(Pays::No.into())
 		}
--- a/pallets/score-staking/src/lib.rs
+++ b/pallets/score-staking/src/lib.rs
@@ -159,38 +159,17 @@ pub mod pallet {
 	#[pallet::event]
 	#[pallet::generate_deposit(pub (crate) fn deposit_event)]
 	pub enum Event<T: Config> {
-		PoolStarted {
-			start_block: BlockNumberFor<T>,
-		},
+		PoolStarted { start_block: BlockNumberFor<T> },
 		PoolStopped {},
-		ScoreFeederSet {
-			new_score_feeder: Option<T::AccountId>,
-		},
-		RoundConfigSet {
-			new_config: RoundSetting,
-		},
-		ScoreUpdated {
-			who: Identity,
-			new_score: Score,
-		},
-		ScoreRemoved {
-			who: Identity,
-		},
+		ScoreFeederSet { new_score_feeder: Option<T::AccountId> },
+		RoundConfigSet { new_config: RoundSetting },
+		ScoreUpdated { who: Identity, new_score: Score },
+		ScoreRemoved { who: Identity },
 		ScoreCleared {},
-		TokenStakingAmountUpdated {
-			account: T::AccountId,
-			amount: BalanceOf<T>,
-		},
-		RewardDistributionStarted {
-			total: BalanceOf<T>,
-			distributed: BalanceOf<T>,
-			round_index: RoundIndex,
-		},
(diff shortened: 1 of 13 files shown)


pallets/score-staking/src/lib.rs
```diff
@@ -275,43 +254,17 @@ pub mod pallet {
...
-					.saturating_div(total_score.into())
-					.saturating_mul(num_integer::Roots::nth_root(&user_stake_u128.pow(n), m))
-					.saturating_div(num_integer::Roots::nth_root(&total_stake_u128.pow(n), m));
-				let user_reward = user_reward_u128.saturated_into::<BalanceOf<T>>();
-
-				p.last_round_reward = user_reward;
-				p.total_reward += user_reward;
-				p.unpaid_reward += user_reward;
-				all_user_reward += user_reward;
+				p.last_round_reward = round_reward;
```
Kailai-Wang: Hmm ... why do we need to update `last_round_reward` in this hook? 
It seems we don't need this loop

pallets/score-staking/src/lib.rs
```diff
@@ -491,14 +444,31 @@ pub mod pallet {
 						return Err(Error::<T>::TokenStakingAmountAlreadyUpdated.into());
 					}
 					let last_round = Round::<T>::get();
-					s.last_round_reward += amount;
```
Kailai-Wang, line 494:
```
let last_round = Round::<T>::get();
```

The round_interval is configurable, what if `Round::<T>::get()` is bigger than `round_index`? Shall we always use `round_index` here to assign `s.last_token_distributed_round`?

pallets/score-staking/src/lib.rs
```diff
@@ -491,14 +444,31 @@ pub mod pallet {
...
+						.map(|s| s.total)
+						.unwrap_or_default()
+						.saturated_into::<u128>();
+					let user_reward_u128 = round_reward_u128
+						.saturating_mul(s.score.into())
+						.saturating_div(total_score.into())
+						.saturating_mul(num_integer::Roots::nth_root(&user_stake_u128.pow(n), m))
+						.saturating_div(num_integer::Roots::nth_root(&total_stake_u128.pow(n), m));
+					let mut user_reward = user_reward_u128.saturated_into::<BalanceOf<T>>();
+					user_reward += amount;
```
Kailai-Wang: Why do we add `user_reward` by `amount`?

Kailai-Wang (changes requested):
After a second thought:

iterating over all score users in a single extrinsic is still better than what we have now (= updating them one by one), because it makes sure we have a common and consistent state for calculation: imagine the state (e.g. scores, total-staking) changes between updating user A and B.

pallets/score-staking/src/lib.rs
silva-fj: You are right, I'll remove this in the new logic

tee-worker/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -315,53 +315,37 @@ where
...
-				)
-			})?
-			.map_err(|_| Error::Other("Invalid metadata".into()))?;
-
-		let complete_reward_distribution_call =
-			OpaqueCall::from_tuple(&(complete_reward_distribution_call_index.clone()));
-		extrinsic_sender.send(complete_reward_distribution_call).map_err(|_| {
-			Error::Other("Failed to send complete_reward_distribution_call extrinsic".into())
-		})?;
+		let id_graphs_staking: Vec<(AccountId, Balance)> = account_ids
```
Kailai-Wang, line 326: I'm just concerned that the `account_ids` is unbounded - can we do it in batch?

silva-fj: Yes, it makes sense. 👍🏼

BillyWooo (approved):
LGTM

pallets/score-staking/src/lib.rs
```diff
@@ -477,41 +431,70 @@ pub mod pallet {
...
+			let round_reward_u128 = round_reward.saturated_into::<u128>();
+
+			let total_stake_u128 = ParaStaking::Pallet::<T>::total().saturated_into::<u128>();
+			let total_score = Self::total_score();
+			let n = Self::round_config().stake_coef_n;
+			let m = Self::round_config().stake_coef_m;
+
+			for (a, mut p) in Scores::<T>::iter() {
+				let default_staking = BalanceOf::<T>::zero();
+				let id_graph_staking = id_graphs_staking_map.get(&a).unwrap_or(&default_staking);
```
Kailai-Wang, line 465: Shall we skip (continue) if `&a` is not found?

pallets/score-staking/src/lib.rs
```diff
@@ -477,41 +431,70 @@ pub mod pallet {
...
+				let user_stake_u128 = ParaStaking::Pallet::<T>::delegator_state(&a)
+					.map(|s| s.total)
+					.unwrap_or_default()
+					.saturated_into::<u128>()
+					+ (*id_graph_staking).saturated_into::<u128>();
```
Kailai-Wang, line 470: Why do we add it again?  IDGraph should contain the prime key already IIRC

pallets/score-staking/src/lib.rs
```diff
@@ -116,6 +119,8 @@ pub mod pallet {
 		type AccountIdConvert: AccountIdConvert<Self>;
 		// For extrinsics that should only be called by origins from TEE
 		type TEECallOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+		#[pallet::constant]
+		type MaxIDGraphAccountsPerCall: Get<u16>;
```
Kailai-Wang, line 123: Sorry my bad.

We have `MaxScoreUserCount` already and we do iterate over them all in the original `on_initialize` hook. Based on that it should be no problem if we do the whole loop in an extrinsic, so we don't need this constant and batched call in the worker.

But we do need to specify the (max) weights in the extrinsic weight macro - this is the important

Kailai-Wang: If we call `distribute` in batch and again we might fall into the problem of inconsistent calculation context

Kailai-Wang: I'd still propose a different paradigm, instead of calling `distribute` ext that accepts a vector with potentially 1M entries (that's the `MaxScoreUserCount` runtime binding), can we:
- update the staking amount in batch, with a batch size 500 (1000 might be a bit aggressive, 750 is the limit that we found out last time with batch balance.transfer)
- when it's done, call `distribute`
- within `distribute`, we iterate **all** entrieis in `Scores` and do the reward calculation/distribution in this atomic extrinsic

How does this sound to you?

Kailai-Wang (changes requested):
I apologize for my last comment, I think it's better to do batch update but one single distribution, see my comment inline

pallets/score-staking/src/lib.rs
silva-fj: oh, I didn't know the IDGraph also contains the prime identity. I'll check it out 👍🏼

pallets/score-staking/src/lib.rs
silva-fj: No worries, that's actually what I had in mind with the initial implementation (when I added a new field to the `ScorePayment`), but back then It was not clear to me how the logic was supposed to be. I'll refactor it to do that

pallets/score-staking/src/lib.rs
silva-fj: All this code will not be needed in the new implementation 👍🏼

PR #3075, opened 2024-09-19
fix: increase the size limit value to 128

wli-pro:
to avoid exceeding the limit while link identity

### Labels

### Testing Evidences

https://github.com/user-attachments/assets/1cf34875-07a8-44dc-85bb-cbbd29f5c500


BillyWooo (approved):
Verified locally.

PR #3079, opened 2024-09-22
Restructure the repo files

Kailai-Wang:
### Context

As topic, this is now the tree structure:

```
.
├── LICENSE
├── Makefile
├── README.md
├── common
├── local-setup
├── parachain
└── tee-worker
```

And in `tee-worker`:
```
└── tee-worker
    ├── bitacross
    └── identity
```

A root-dir can only have one workspace definition.
It would be very chaotic and incredibly long deps if I have one big `Cargo.toml` that specifies deps for both parachain and workers.

So I'm going to have a workspace definition for parachain and worker, respectively. The `common` will then come up as standalone crates.

> [!NOTE]  
> Workers still have isolated directories - I don't want to introduce too many changes in one PR. A bigger PR will follow up with workspace-dependencies for workers.

> [!WARNING]
> Please note the changes (in case anything needs to be changed in the deployment script) @m1iktea


jonalvarezz (approved):
Thanks Kai, 

Since it is mostly restructuring, I tested it by checking out the code locally, building the node and the identity worker and launching in standalone mode. It looks good. I could do all basic operations like retrieve/ link identity and request vc.

silva-fj (approved):
This is a nice improvement 👍🏼

PR #3082, opened 2024-09-23
Improving Identity staking rewards distribution

silva-fj:
This is the identity staking rewards distribution follow up. The main logic is as follows:

- An event is emitted to trigger the rewards distribution on each round
- The TEE Worker gets the staking amount for each account stored in `Scores` and calls the `distribute_rewards` for each account
- Then TEE Worker calls `complete_rewards_distribution` to indicate that the distribution for the round to all users is completed


parachain/pallets/score-staking/src/lib.rs
```diff
@@ -445,6 +427,70 @@ pub mod pallet {
...
+		pub fn distribute_rewards(
+			origin: OriginFor<T>,
+			round_index: RoundIndex,
+		) -> DispatchResultWithPostInfo {
+			let _ = T::TEECallOrigin::ensure_origin(origin)?;
+			ensure!(
+				round_index > LastTokenDistributionRound::<T>::get(),
+				Error::<T>::RoundRewardsAlreadyDistributed
+			);
+			let round_reward: BalanceOf<T> = (T::YearlyInflation::get() * T::YearlyIssuance::get()
```
felixfaisal: Is there any specific reason why we don't use `saturated_*` or `checked_*` only in this line.

parachain/pallets/score-staking/src/lib.rs
```diff
@@ -445,6 +427,70 @@ pub mod pallet {
...
+				/ YEARS.into()) * Self::round_config().interval.into();
+			let round_reward_u128 = round_reward.saturated_into::<u128>();
+			let total_stake_u128 = ParaStaking::Pallet::<T>::total().saturated_into::<u128>();
+			let total_score = Self::total_score();
+			let n = Self::round_config().stake_coef_n;
+			let m = Self::round_config().stake_coef_m;
+
+			for (a, mut p) in Scores::<T>::iter() {
```
felixfaisal: This loop is quite expensive, it's going to be O(n), 
Have we researched using something like `on_idle` hook for this, that way we make sure to use the idle blockspace and that way we allow the chain to be used for other activities and not prioritise the distribution of tokens?

silva-fj: I had the same concern in the original PR (the current `on_initialize` hook), iterating over the entries could be a problem eventually. The issue with `on_idle` is that it's only executed if there is weight left, and the logic implemented needs to consider the remaining weight as it should not exceed it. But the main problem with using a hook is that we have no access to the IDGraphs to calculate the total staking for each user.

Another solution could be to go back to calculating the user's rewards one by one (this was the initial proposal of the [PR](https://github.com/litentry/litentry-parachain/pull/3074) that was closed), but in order to have a consistent state when calculating the rewards for each users we also keep track of the `total_score` and `total_staking` of the last round and use that in the calculation. What do you think @Kailai-Wang ?

felixfaisal: Another approach we could explore is, if we calculate the users reward when the user claims rewards. If we know the total reward of that round, and how much is staked by the user, then we can simply calculate the reward the user has to receive(provided we have storage items for it). This is just an approach, I can't think of exact details. I think this is what polkadot staking or the frame staking pallet does. I simply want to avoid an O(N) loop if possible.

silva-fj: With the logic that we have to calculate the rewards we would need to keep track of values like `total_stake`, `total_score` for every user each round as well as the global values we use, so we will end up in a similar situation (or even worst). I think the initial approach should be fine, It's just that back then we didn't think about how to keep the state consistent between rounds to do the calculations, keeping track of the global values ​​would solve that.

parachain/pallets/score-staking/src/lib.rs
silva-fj: I think it was just for convenience as we would need to do type conversions to use the safe methods. In this case it's fine, we are in control of those values and it won't overflow

BillyWooo (dismissed):
Please, hold on until new release goes out.

silva-fj:
This PR is getting too old and a lot has changed. I'll close it but keep it as reference in case we need to implement this in the future

PR #3083, opened 2024-09-23
Propagate release-v0.9.19 into dev

jonalvarezz:
as topic


jonalvarezz:
Closing as it needs conflict resolution and we shouldn't push the fix to release-v0.9.19

PR #3096, opened 2024-09-25
Hot fix create release draft

BillyWooo:
as title


.github/workflows/create-release-draft.yml
```diff
@@ -79,7 +79,6 @@ jobs:
           - rococo
           - paseo
           - litentry
-    name: ${{ matrix.chain }}
```
Kailai-Wang, line 82: Do we have a reason of deleting this? It's good for display

.github/workflows/create-release-draft.yml
```diff
@@ -148,7 +147,6 @@ jobs:
       - name: Save docker image
         run: |
           docker save litentry/litentry-parachain:${{ env.RELEASE_TAG }} | gzip > litentry-parachain-dev.tar.gz
-          docker save litentry/litentry-chain-aio:${{ env.RELEASE_TAG }} | gzip > litentry-chain-aio.tar.gz
```
Kailai-Wang, line 151: Hmm if we don't need it in release - then we should not build it IMO

felixfaisal (commented):
This is not good practice to directly merge to `dev` without an approval or without the CI check passing. Despite the title being `hot fix`, It is expected to at least make sure CI is passing.

PR #3097, opened 2024-09-26
Use workspace dependencies for tee-worker(s)

Kailai-Wang:
### Context

> [!WARNING]
> It's a breaking change for bitacross client due to the `DirectRequestStatus` type change  

As topic.

This PR tries to unify the cargo dependencies of identity-worker and bitacross-worker to form a workspace-based configuration.

Please note:
- the common crates are under `tee-worker/common/`
- `enclave-runtime` needs to have its own workspace for sgx-specific patches
- some crates aren't merged, as they are differently implemented and/or meant to be specific for each worker, e.g. sgx-runtime, stf-executor, enclave-api ... see e.g. `identity/core` and `identity/core-primitives` for those crates. It's still possible to merge them into a "common" crate, but bigger code change would be needed
- as a result, the workspace would contain two crates with the same name (e.g. `ita-stf`), to overcome this, I prepended the crate name with the worker type, e.g. `id-ita-stf` vs `bc-ita-stf`.
- `enum DirectRequestStatus` is defined with different entry orders in id- and bc-worker. I follow the one in id-worker to not break anything from IDHub. Do we need to change anything for bc-worker? cc @kziemianek 
- some changes can be seen as an attempt (example) to merge the crates, e.g. `HandleParentchainEvents`, `FilterEvents` which contains fn for both workers

fixes P-242
fixes P-552


Kailai-Wang:
fixes P-242
fixes P-552

jonalvarezz (approved):
what a great effort, thanks @Kailai-Wang!

lgtm

kziemianek:
@Kailai-Wang this will be a breaking change for a client:
we need to:
* update definition [here](https://github.com/litentry/litentry-parachain/blob/b4df5a6577c64ecb404ba023fc166c5f25f5dd55/tee-worker/bitacross/example/client/definitions.json#L19-L31)
* let relayer team know so they can adjust their code

silva-fj (approved):
It looks good overall 👍🏼 
I can feel the pain of making this changes 😆

Kailai-Wang:
> @Kailai-Wang this will be a breaking change for a client: we need to:
> 
> * update definition [here](https://github.com/litentry/litentry-parachain/blob/b4df5a6577c64ecb404ba023fc166c5f25f5dd55/tee-worker/bitacross/example/client/definitions.json#L19-L31)
> * let relayer team know so they can adjust their code

Thanks @kziemianek !
Here it is: https://github.com/litentry/litentry-parachain/pull/3097/commits/c96fc21d33a99796564795b10f5929e30fb7d383
The only thing that I'm not that sure is if `H256` type is natively supported - I assume yes, as I don't see errors when compiling and running it (bc-worker should not get `TrustedOperationStatus` anyway)

kziemianek:
> > @Kailai-Wang this will be a breaking change for a client: we need to:
> > 
> > * update definition [here](https://github.com/litentry/litentry-parachain/blob/b4df5a6577c64ecb404ba023fc166c5f25f5dd55/tee-worker/bitacross/example/client/definitions.json#L19-L31)
> > * let relayer team know so they can adjust their code
> 
> Thanks @kziemianek ! Here it is: [c96fc21](https://github.com/litentry/litentry-parachain/commit/c96fc21d33a99796564795b10f5929e30fb7d383) The only thing that I'm not that sure is if `H256` type is natively supported - I assume yes, as I don't see errors when compiling and running it (bc-worker should not get `TrustedOperationStatus` anyway)

It should be supported [see](https://pkg.go.dev/github.com/itering/scale.go/types#H256).

PR #3098, opened 2024-09-26
Adding omni-account pallet

silva-fj:
As the changes for moving the IDGraphs on-chain are getting too big, I have divided the issue in multiple sub-tasks to make it easier to review and get earlier feedback. This is the first part, the second part will have the changes in the identity worker

Fixes: P-1081, P-1083, P-1087


parachain/pallets/omni-account/src/identity_context.rs
```diff
@@ -0,0 +1,50 @@
...
+
+// The context associated with the (litentry-account, did) pair
+#[derive(Clone, Eq, PartialEq, Default, Debug, Encode, Decode, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+#[codec(mel_bound())]
+pub struct IdentityContext<T: Config> {
+	// the block number at which the identity is linked
+	pub link_block: BlockNumberFor<T>,
+	// a list of web3 networks on which the identity should be used
+	pub web3networks: Vec<Web3Network>,
```
Kailai-Wang: I suggest we dismiss this - it's a good opportunity to get rid of this field now

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+				identity: identity_hash,
+				web3networks,
+			});
+
+			Ok(())
+		}
+
+		#[pallet::call_index(4)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn remove_identity(
```
kziemianek: I belive this should not be possible.

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+			let id_graph_link = id_graph_links
+				.iter_mut()
+				.find(|(id, _)| id == &private_identity)
+				.ok_or(Error::<T>::IdentityNotFound)?;
+
+			let public_identity_hash = H256::from(blake2_256(&public_identity.encode()));
+			id_graph_link.0 = public_identity;
+			IDGraphs::<T>::insert(who.clone(), id_graph_links);
+			let private_identity_hash = H256::from(blake2_256(&private_identity.encode()));
+			LinkedIdentities::<T>::remove(private_identity_hash);
```
kziemianek: Does this mean it allows for linking the same identity to two different id graphs ?
One id graph will contain it as a public while another as a private 🤔

parachain/pallets/omni-account/src/lib.rs
kziemianek: It's called by the worker only so as long as guards remain there its fine 👍🏼

parachain/pallets/omni-account/src/lib.rs
silva-fj: Yes. The logic to prevent this will be on the worker's side as it's the only place we can check the private identities

kziemianek: Just don't remove `private_identity_hash` .

If I understand flow correctly private identity needs to be linked first.

kziemianek: But worker is eventually consistent with parachain storage, so I belive there is a way to link the same identity twice (private and public variant) using two different (?) workers ?

silva-fj: Good point, I think if we make the worker send the identity_hash instead of calculating in the pallet we could prevent this. What do you think?

kziemianek: Yes, it can be a hash of identity's DID, just something that uniquely identifies it - no matter if it's private or public form.

common/primitives/core/src/identity.rs
```diff
@@ -524,6 +524,18 @@ impl From<[u8; 33]> for Identity {
     }
 }
 
+#[derive(Encode, Decode, TypeInfo, Clone, PartialEq, Eq, RuntimeDebug)]
+pub enum OmniAccountIdentity {
```
Kailai-Wang: The name can be a bit misleading - it's meant to be one entry of the IDGraph (I assume).

Maybe `SubIdentity`, or `MemberIdentity`? Hmm - `MemberIdentity` might be better I'm not sure

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+		type TEECallOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+		/// The maximum number of identities an id graph can have.
+		#[pallet::constant]
+		type MaxIDGraphLength: Get<u32>;
+	}
+	pub type IDGraphLinks<T> =
+		BoundedVec<(OmniAccountIdentity, IdentityContext<T>), <T as Config>::MaxIDGraphLength>;
+
+	#[pallet::storage]
+	pub type LinkedIdentities<T: Config> =
```
Kailai-Wang: Maybe call it `LinkedIdentityHashes?`

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+		#[pallet::call_index(0)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn link_identity(
+			origin: OriginFor<T>,
+			who: Identity,
+			identity: OmniAccountIdentity,
+			web3networks: Vec<Web3Network>,
+		) -> DispatchResult {
+			let _ = T::TEECallOrigin::ensure_origin(origin)?;
+			let identity_hash = H256::from(blake2_256(&identity.encode()));
```
kziemianek: If we ever decide to introduce significant change to `OmniAccountIdentity` or `Identity` it will require storage migration as encoding is coupled to data structure. 

Given that some identities are private and encrypted form can only be produced by enclaves the process is getting even more complicated.

What do you think of using hash of DID instead?

silva-fj: Yeah I think it's a good idea 👍🏼

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+		BoundedVec<(OmniAccountIdentity, IdentityContext<T>), <T as Config>::MaxIDGraphLength>;
+
+	#[pallet::storage]
+	pub type LinkedIdentities<T: Config> =
+		StorageMap<Hasher = Blake2_128Concat, Key = H256, Value = ()>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn id_graphs)]
+	pub type IDGraphs<T: Config> =
+		StorageMap<Hasher = Blake2_128Concat, Key = Identity, Value = IDGraphLinks<T>>;
```
Kailai-Wang: I favor `T::AccountId` over `Identity` in this case - it shows explictly that the OmniAccount is a native substrate account under the hood.

So whichever Identity you use as a prime key to log onto our service, we'll derive a substrate account for you (and show it in the UI), in case users want to check balance/lookup tx their should use this account, our tooling UI can help with that.

One more thing - users can even unlink their prime identity after the OmniAccount (if we support the remove_identity UserOp) is set and only use MemberIdentities to control the OmniAccount

Kailai-Wang: We'll need another storage, either in `IDGraphs` or as a separate storage to record:
- in which block number the OmniAccount is created
- the whole IDGraph fingerprint (hash)

This can be done in another PR though

parachain/pallets/omni-account/src/identity_context.rs
```diff
@@ -0,0 +1,51 @@
+use crate::{BlockNumberFor, Config, Web3Network};
+use parity_scale_codec::{Decode, Encode};
+use scale_info::TypeInfo;
+use sp_std::vec::Vec;
+
+#[derive(Clone, Eq, PartialEq, Default, Debug, Encode, Decode, TypeInfo)]
+pub enum IdentityStatus {
```
Kailai-Wang: Maybe we don't need this status either - it's been rarely/never used

It could be useful to prevent some member identity from accessing the OmniAccount in contingency (e.g. when I know some account is compromised) - but the hacker could have updated this setting using the compromised account

Or we leave as it is hoping the hacker doesn't know about it and OmniAccount owner might have time window to take actions
Wdyt?

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+			Ok(())
+		}
+
+		#[pallet::call_index(5)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn make_identity_public(
+			origin: OriginFor<T>,
+			who: Identity,
+			private_identity: OmniAccountIdentity,
+			public_identity: OmniAccountIdentity,
```
Kailai-Wang: It seems we only need the `public_identity` here?  `private_identity` can be derived from public identity?

Kailai-Wang (commented):
Thank you! Yes it makes sense to build it step by step, I left some comments

common/primitives/core/src/identity.rs
silva-fj: Yes I was not sure about this name either, I had `IDGraphIdentity` before and changed it. I like `MemberIdentity` as well

parachain/pallets/omni-account/src/identity_context.rs
silva-fj: I think it could be a nice feature, we could make it similar to a Multisig to only allow re-activating an identity if at least two members (or more) of the OmniAccount approve it or something like this. (The same for removing an identity if we want to support this at some point)

parachain/pallets/omni-account/src/lib.rs
silva-fj: Yes I'll remove this together with the changes to send the identity_hash from the worker

parachain/pallets/omni-account/src/identity_context.rs
silva-fj: But if we don't want to support this, or at least now and we are removing the web3networks, I think we can get rid of the identity context all together, I think the link_block was used to sort the IDGraph but now we don't need to sort it

Kailai-Wang: Multisig-ish control is a good idea.

Yes it seems we don't need this Context at the moment - the `link_block` was designed for potential future feature like if we want to ask users to "renew" the ownership in X blocks.

BillyWooo (dismissed):
Please hold on until new release goes out.

Kailai-Wang:
@silva-fj let's resolve the TODOs asap and try to merge it to not block the sub-issues : )

silva-fj:
> @silva-fj let's resolve the TODOs asap and try to merge it to not block the sub-issues : )

Done

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use frame_support::pallet_prelude::*;
+use frame_system::pallet_prelude::*;
+use sp_core::H256;
+use sp_core_hashing::blake2_256;
+use sp_std::vec::Vec;
+
+pub trait AccountIdConverter<T: Config> {
+	fn convert(account_id: T::AccountId) -> Identity;
+}
```
Kailai-Wang: Does it work?

It seems `Identiy` -> `T::AccountId` is a one-way conversion, it's like hash fn - you can't convert it back, can you?

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		/// The origin which can manage the pallet.
+		type TEECallOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+		/// The maximum number of identities an id graph can have.
+		#[pallet::constant]
+		type MaxIDGraphLength: Get<u32>;
+		/// AccountId converter
+		type AccountIdConverter: AccountIdConverter<Self>;
+	}
+	pub type IDGraphLinks<T> = BoundedVec<(H256, MemberIdentity), <T as Config>::MaxIDGraphLength>;
```
Kailai-Wang: Shall we use a struct to replace `(H256, MemberIdentity)`? Like
```
struct MemberAccount {
  id: MemberIdentity,
  hash: H256,
}
```

Maybe better for future expansion, but it's a flavor problem

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn link_identity(
+			origin: OriginFor<T>,
+			who: T::AccountId,
+			identity_hash: H256,
+			identity: MemberIdentity,
+			maybe_id_graph_hash: Option<H256>,
```
Kailai-Wang, line 120: I think we need to record both old and new IDGraph:
- old: the current IDGraph
- new: the pre-computed IDGraph if the identity is successfully linked

And how to compute IDGraphHash: maybe the merkle root of all member identity hashes? The hash of encoded IDGraph could change as you make some identity public?

parachain/pallets/omni-account/src/lib.rs
silva-fj: We can do `Identity::from(account_id)` if `account_id` is a substrate account but because at the pallet level rust does not know the concrete type of `T::AccountId`, we can't use it directly so I've added this as a workaround to get the substrate Identity.

Kailai-Wang: That is true, but only for the case of substrate.

I see that it's mainly used to compute `prime_identity_hash` - but what if the prime identity is btc/evm? So it means in `link_identity`, `who` should be of `Identity` type I guess?

parachain/pallets/omni-account/src/lib.rs
silva-fj: Here is the idea of the current implementation: Worker A and Worker B both want want to update the IDGraph, they both have the current `id_graph_hash`, if Worker B updates it first then the hash will change so the call from Worker A will be rejected. Does it makes sense?

> And how to compute IDGraphHash: maybe the merkle root of all member identity hashes? The hash of encoded IDGraph could change as you make some identity public?

This is a good point, using only the identity hashes makes sense

silva-fj: To keep it simple I've updated the `id_graph_hash` calculation to be the hash of all the identity_hashes

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+	pub type IDGraphMembers<T> = BoundedVec<MemberAccount, <T as Config>::MaxIDGraphLength>;
+
+	#[pallet::storage]
+	pub type LinkedIdentityHashes<T: Config> =
+		StorageMap<Hasher = Blake2_128Concat, Key = H256, Value = ()>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn id_graphs)]
+	pub type IDGraphs<T: Config> =
+		StorageMap<Hasher = Blake2_128Concat, Key = Identity, Value = IDGraphMembers<T>>;
```
Kailai-Wang: There might be some misunderstanding

I mean the key of `IDGraphs` is still `T::AccountId`, but when linking identities, `who` should be of `Identity` type as it's used to compute hash

Kailai-Wang: does it make sense?

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn link_identity(
+			origin: OriginFor<T>,
+			who: Identity,
+			identity_hash: H256,
+			identity: MemberIdentity,
```
Kailai-Wang: We can use `MemberAccount` type here

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+		) -> DispatchResult {
+			let _ = T::TEECallOrigin::ensure_origin(origin)?;
+			ensure!(
+				!LinkedIdentityHashes::<T>::contains_key(identity_hash),
+				Error::<T>::IdentityAlreadyLinked
+			);
+			let mut id_graph_members = match IDGraphs::<T>::get(&who) {
+				Some(id_graph_members) => {
+					let current_id_graph_hash =
+						IDGraphHashes::<T>::get(&who).ok_or(Error::<T>::PrimeIdentityNotFound)?;
```
Kailai-Wang: I think it's possible that `IDGraphHashes` doesn't have `who` yet - e.g. for the first time linking

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+					if LinkedIdentityHashes::<T>::contains_key(who_hash) {
+						return Err(Error::<T>::IdentityAlreadyLinked.into());
+					}
+					let mut id_graph_members: IDGraphMembers<T> = BoundedVec::new();
+					id_graph_members
+						.try_push(MemberAccount {
+							id: MemberIdentity::from(who.clone()),
+							hash: who_hash,
+						})
+						.map_err(|_| Error::<T>::IDGraphLenLimitReached)?;
```
Kailai-Wang: Don't we need to push both PrimeIdentity and MemberAccount?

Kailai-Wang (commented):
Thanks, we are close to merging it.

The biggest change that I requeset is to use `T::AccountId` as the key in IDGraphs - we can use `Identity` when linking identities, but the underlying storage type has to be `T::AccountId`.

I'm also a bit confused by the logic in `link_identity`, but we can merge it first, I have some ideas to make some adjustments to align with future usage - I'll raise PRs later.

Wdyt?

Another thing is: does it make sense that we stop calling it identities in storage name and extrinsic names? Like `link_account` instead of `link_identity` 😂 Maybe not now

parachain/pallets/omni-account/src/lib.rs
silva-fj: Yes, it should be ok. We still need to receive `who` as `Identity` in the `link_identity` call, but I think it's fine.

parachain/pallets/omni-account/src/lib.rs
silva-fj: In this branch (`Some`) of the `match` expression, the IDGraph already exists as there is a value. The `None` branch is the first time an identity will be linked so the IDGraph is created. I can extract the logic to make it more descriptive, like:

```rust
let mut id_graph_members = match IDGraphs::<T>::get(&who) {
    Some(id_graph_members) => {
        Self::validate_id_graph_hash(maybe_id_graph_hash)?;
        id_graph_members
    },
    None => Self::create_id_graph(who, identity)
};
```

parachain/pallets/omni-account/src/lib.rs
silva-fj: In this part of the code we are only creating the IDGraph (for the first time), this is why we only add the prime identity. In the code below is where we add identity (MemberAccount). This is why the `id_graph_members` is initialized as `mut`, this way we don't have to repeat the logic depending on the branch, we just add the new member one time for both cases

silva-fj:
> Wdyt?
> 
> Another thing is: does it make sense that we stop calling it identities in storage name and extrinsic names? Like `link_account` instead of `link_identity` 😂 Maybe not now

I was thinking the same, we started to call them accounts and members. I think the names should reflect the same language/terms as how the business will call them. I think if we are going to keep calling this `link_identity` we should change everything to `Identity`, otherwise we should change the extrinsics to something like `add_account` or `link_account`

Kailai-Wang:
Yes I was thinking of the same, but then the request (trusted_call) would be called as `link_account` too, which requires adjustments in SDK

what do you think? @jonalvarezz

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+		/// Identity is already linked
+		IdentityAlreadyLinked,
+		/// IDGraph len limit reached
+		IDGraphLenLimitReached,
+		/// Identity not found
+		IdentityNotFound,
+		/// Invalid identity
+		PrimeIdentityInvalid,
+		/// Prime identity not found
+		PrimeIdentityNotFound,
```
kziemianek: Is the concept of prime identity still valid, mostly it refers to `who` ?
Only this and variant above mentions it.

Maybe it could be changed to:
`PrimeIdentityInvalid` -> `InvalidIdentity` 
`PrimeIdentityNotFound` -> `UnknownIdGraph`

parachain/pallets/omni-account/src/lib.rs
kziemianek: Why don't you use `IDGraphHashMissing` ?

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+	pub trait Config: frame_system::Config {
+		/// The event type of this pallet.
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		/// The origin which can manage the pallet.
+		type TEECallOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+		/// The maximum number of identities an id graph can have.
+		#[pallet::constant]
+		type MaxIDGraphLength: Get<u32>;
+	}
+	pub type IDGraphMembers<T> = BoundedVec<MemberAccount, <T as Config>::MaxIDGraphLength>;
```
kziemianek: `OmniAccountMembers` or maybe `SubAccounts` ?

I see no reference to omni account inside this pallet - what it really is ?

`IdGraph` -> `OmniGraph` ?
`Identity` -> `SubAccount` ?

The domain changes and it should imply vocabulary change aswell.

Do we really still link identity or maybe add identity or add (sub) account ?
What does control omni account ? Is it identity, account or maybe sub account as described in notion page ?

parachain/pallets/omni-account/src/lib.rs
silva-fj: I guess it could be ok, but `IDGraphHashMissing` has a different meaning. It means that the the user wanted to update an existing IDGraph but it didn't provide the IDGraph hash (maybe_id_graph_hash param). In theory this error should never happen because if the IDGraph for `who` exists, the IDGraphHash should also exist. So, this would be a logic error

parachain/pallets/omni-account/src/lib.rs
silva-fj: You are right, we are going to have a discussion about the language/vocabulary we will use. This issue started as just a storage replacement for the current implementation (moving the storage to on-chain), but it has been evolving on the fly to adjust it with the new direction of the product and we ended up with a mix of old names (Link, Identity, etc) and new names.

It's the same for the other comment:
>Is the concept of prime identity still valid, mostly it refers to who ?
Only this and variant above mentions it.
>
>Maybe it could be changed to:
PrimeIdentityInvalid -> InvalidIdentity
PrimeIdentityNotFound -> UnknownIdGraph

We still are not 100% clear on the names we should use, but the general architecture should not change much. So we will improve the names in follow up PR's once the business defines the correct vocabulary. This will also impact the SDK

parachain/pallets/omni-account/src/lib.rs
silva-fj: I've refactored the code to make it clear

Kailai-Wang (approved):
For me it can be merged : )

kziemianek (approved):

PR #3105, opened 2024-09-28
Move pallet-parentchain to worker directory

Kailai-Wang:
### Context

As topic - it's a pallet used in workers only, so it makes sense to move it there


BillyWooo (approved):
LGTM

kziemianek (approved):

PR #3111, opened 2024-10-01
Script improvment

BillyWooo:
some improvement of scripts.
Remove dockerfile warnings.
Only build all-in-one image for "latest" tag.


parachain/scripts/build-docker.sh
```diff
@@ -87,13 +76,43 @@ docker build ${NOCACHE_FLAG} --pull -f ./parachain/docker/Dockerfile \
...
 
 # Show the list of available images for this repo
 echo "Image is ready"
 echo "------------------------------------------------------------"
-docker images | grep ${GITREPO}
\ No newline at end of file
+docker images | grep ${GITREPO}
+
+# ===================================================================================
+if [ -z "$TAG" ] || [ "$TAG" = "latest" ]; then
```
kziemianek: If you want to build all-in-one image for "latest" tag only shouldn't first predicate be removed ?

parachain/scripts/build-docker.sh
```diff
@@ -59,26 +59,15 @@ docker build ${NOCACHE_FLAG} --pull -f ./parachain/docker/Dockerfile \
...
-
-# Tag it with latest if no tag parameter was provided
-[ -z "$2" ] && docker tag ${GITUSER}/${GITREPO}:${TAG} ${GITUSER}/${GITREPO}:latest
-
-# Show the list of available images for this repo
-echo "Image is ready"
-echo "------------------------------------------------------------"
-docker images | grep ${GITREPO}
+    --target builder \
+    --tag local-builder:latest .
```
Kailai-Wang, line 63: I don't fully understand this - why do we need to build `local-builder` first? Or what's the advantage comparing to the current impl?

Kailai-Wang (commented):
Thanks, only one question:

parachain/scripts/build-docker.sh
BillyWooo: In case we need to build all-in-one image, this can save the build time.
While splitting the original parachain image building into two steps doesn't create much overlap. We have this similar operation when building worker images, which works fie.

Kailai-Wang (approved):

PR #3112, opened 2024-10-02
fix: set a default name for job identity-multi-worker-test

wli-pro:
set a default name for job identity-multi-worker-test  while identity-multi-worker-test is skipped, so  we don't see ${{ matrix.test_name }}

### Labels

### Testing Evidences


Kailai-Wang (approved):
Nice, thanks!

Kailai-Wang:
Acutally, it still didn't work as far as I saw

and it seems to be an unsolved github problem: https://github.com/orgs/community/discussions/13261

PR #3113, opened 2024-10-02
chore: P-1044 merge all identity tests into one

wli-pro:
### Context
As title

### Labels

### Testing Evidences


jonalvarezz (approved):
having all the tests in a single place makes sense, thanks @higherordertech 

Because of how ts-tests were created, I understand we need to clean up storage with `clear_linked_identities`. I guess it is ok to have development-only features, but I'd have preferred to re-work ts-tests to for example use random accounts each time instead of requiring a cleanup on every run. I understand it'd be out of scope, too.

Kailai-Wang:
Thanks!

It's good to go once CI passes - please note we expect the tests to be adjusted again to align with the omni-account usage (very soon, but the effort should not that huge)

wli-pro:
> having all the tests in a single place makes sense, thanks @higherordertech
> 
> Because of how ts-tests were created, I understand we need to clean up storage with `clear_linked_identities`. I guess it is ok to have development-only features, but I'd have preferred to re-work ts-tests to for example use random accounts each time instead of requiring a cleanup on every run. I understand it'd be out of scope, too.

Spot on, I updated the tests to not rely on cli to clear idgraph but just randomize new account for each test case now.

tee-worker/identity/ts-tests/integration-tests/common/helpers.ts
```diff
@@ -115,3 +130,16 @@ export function stfErrorToString(stfError: StfError): string {
 
     return stfError.type;
 }
+
+export async function cleanIDGraphs(context: IntegrationTestContext, teeShieldingKey: KeyObject) {
```
kziemianek: It looks like it's not used anymore, can it be removed ?

tee-worker/identity/ts-tests/integration-tests/common/di-utils.ts
```diff
@@ -347,6 +347,26 @@ export async function createSignedTrustedGetterIdGraph(
     return parachainApi.createType('Getter', { trusted: getterSigned });
 }
 
+export async function createSignedTrustedCallCleanIDGraphs(
```
kziemianek: Remove if not used ?

tee-worker/identity/ts-tests/integration-tests/common/utils/identity-helper.ts
```diff
@@ -146,15 +146,15 @@ export async function buildValidations(
     signerIdentitity: CorePrimitivesIdentity,
     linkIdentity: CorePrimitivesIdentity,
     startingSidechainNonce: number,
-    network: 'ethereum' | 'substrate' | 'bitcoin' | 'solana',
+    network: 'evm' | 'ethereum' | 'substrate' | 'bitcoin' | 'solana',
```
kziemianek: Can 'ethereum' be removed ?

kziemianek (approved):

PR #3114, opened 2024-10-02
add support for metadata hash extention

BillyWooo:
As title. 

Enable CheckMetadataHash extension, guide [here](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_metadata_hash/index.html) : 
- the metadata generation is only enabled for release package with `--features=on-chain-release-build` 
- test locally using - https://github.com/erkkonet/test-metadatahash

Test evidence:

image

### DIFF
--- a/parachain/Cargo.lock
+++ b/parachain/Cargo.lock
@@ -5613,6 +5613,7 @@ dependencies = [
  "fp-self-contained",
  "frame-benchmarking",
  "frame-executive",
+ "frame-metadata-hash-extension",
  "frame-support",
  "frame-system",
  "frame-system-benchmarking",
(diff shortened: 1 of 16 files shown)


.github/workflows/create-release-draft.yml
```diff
@@ -131,7 +131,7 @@ jobs:
 
       - name: Build docker image
         run: |
-          ./parachain/scripts/build-docker.sh production ${{ env.RELEASE_TAG }}
+          ./parachain/scripts/build-docker.sh production ${{ env.RELEASE_TAG }} --features=on-chain-release-build
```
Kailai-Wang: Why not use `metadata-hash` directly?

Kailai-Wang: Can we have this enabled by default for both `release` and `production` build? It seems that we are testing one variant in general CI (without this feature) but actually release another variant

I ask this because it can be a breaking change for the worker, see https://github.com/integritee-network/worker/pull/1603

BillyWooo: > Why not use `metadata-hash` directly?

@Kailai-Wang  How about we just add it all by default without add a new feature? From the [guide](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_metadata_hash/index.html), it recommends to separate to save building time. But it actually turns out there is only less than 2 mins difference.

BillyWooo: Regarding `release` and `prodcution`, I think `release` is for `production`.  Our current `release` is actually used for testing purpose. Do you think it's worth to rename current `release` to some different name?

Kailai-Wang:
Do you know what to do with this PR? 🤔

BillyWooo:
replaced by #3280

PR #3116, opened 2024-10-04
Adding abstractions for  the omni-account in the worker

silva-fj:
This PR introduces some abstractions related to the omni-account in the worker.

The main changes in this PR are: 
- New primitive types (related to omni-account).
- Litentry core abstractions: `OmniAccountRepository` is an abstraction to request data from the parachain and `InMemoryStore` is an abstraction to store and manage the OmniAccount Members in RAM.
- Optional `header` parameter to `get_storage_keys`.

TODO: 
- [ ] Replace usage of state key repository with the new key repository (aes256 key) for the on-chain data encryption.


Kailai-Wang:
> Replace usage of state key repository with the new key repository (aes256 key)

"Replacing" it probably incurs a state migration for currently encrypted state, which we might want to avoid when possible.

Can we just additionally use the aes256 without touching the old state key repository?

common/primitives/core/src/omni_account.rs
```diff
@@ -0,0 +1,47 @@
+use crate::{Hash, Identity, Vec};
```
Kailai-Wang: Could you please add licence header?

Same for other new files : )

silva-fj:
> > Replace usage of state key repository with the new key repository (aes256 key)
> 
> "Replacing" it probably incurs a state migration for currently encrypted state, which we might want to avoid when possible.
> 
> Can we just additionally use the aes256 without touching the old state key repository?

Oh, yes. That's exactly what I mean. Currently I'm using the concrete `EnclaveStateKeyRepository` type to make CI pass, but once we have the new aes256 key repository available we just need to replace the concrete type as the interface should be the same (`key_repository::AccessKey`), we also need to update the bounds to use `Aes256` instead of `Aes` for the `KeyType`. I'm not thinking of touching anything related to the current state encryption

tee-worker/identity/litentry/core/omni-account/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use alloc::{collections::btree_map::BTreeMap, vec::Vec};
+use itp_types::parentchain::{AccountId, Hash, Header, ParentchainId};
+use litentry_primitives::MemberAccount;
+
+pub type OmniAccounts = BTreeMap<AccountId, OmniAccountMembers>;
+
+#[derive(Debug, Clone)]
+pub struct OmniAccountMembers {
+	pub member_accounts: Vec<MemberAccount>,
```
Kailai-Wang: I think internally stored member_accounts have to be clear text, otherwise you won't be able to compute the hash and pass it as parameter when e.g. `add_account`

Kailai-Wang (commented):
Thanks!

I'm a bit confused about all the changes in `StfExecutor`, why would we need those? I naively assume in the in-RAM AccoutStore has nothing to do with sidechain state (thus stf), so mutation of it doesn't need to go through the stf executor. Is it the case?

tee-worker/identity/litentry/core/omni-account/src/repository.rs
```diff
@@ -0,0 +1,84 @@
...
+		let account_store_hash = member_accounts.hash();
+
+		Ok(OmniAccountMembers { member_accounts, hash: account_store_hash })
+	}
+
+	fn get(&self) -> Result<OmniAccounts, Error> {
+		let account_store_key_prefix = storage_prefix(b"OmniAccount", b"AccountStore");
+		let account_store_storage_keys_response = self
+			.ocall_api
+			.get_storage_keys(account_store_key_prefix.into())
```
kziemianek: This should get keys at `&self.header` or it might be querying different state here and in following `get_multiple_storages_verified`.

tee-worker/identity/litentry/core/omni-account/src/lib.rs
silva-fj: My understanding is this: The user wants to add an account, so it sends the trusted call to the worker with the `Identity` of the account to be added, with that `Identity` the worker construct the `MemberAccount` to be sent to the parachain (`add_account` extrinsic). The hash is computed from the `Identity` received in the trusted call and the `MemberIdentity` would be `MemberIdentity::Private(identity_encrypted)`. Does it makes sense?

silva-fj:
> Thanks!
> 
> I'm a bit confused about all the changes in `StfExecutor`, why would we need those? I naively assume in the in-RAM AccoutStore has nothing to do with sidechain state (thus stf), so mutation of it doesn't need to go through the stf executor. Is it the case?

The idea was that when handling a trusted call, we are able to:
1. Query the latest state of the `AccountStore` to be sure we have the latest hash before sending the extrinsic to mutate it. (We could rely on the `InMemoryStore` for this), this is the reason for injecting the `ocall_api` and the parentchain `header`.
2. We need to be able to encrypt the data before sending the extrinsics and decrypt the data when handling trusted getters. So we inject the encryption key repository.

Does it makes sense? Do you had another approach in mind?

silva-fj:
Another approach could be:
- We keep the data unencrypted in the `InMemoryStore`, so we won't need to decrypt it for the trusted getters.
- We rely on the `InMemoryStore` to get the current `AccountStore` and hash when handling the trusted calls.
- We create a channel/thread that would be responsible for encrypting the data and sending the extrinsics to the parachain. So when handling the trusted call we verify and send the message to this channel. The problem with this is that we won't be able to wait until it's included in the block.

This way we won't need to inject anything into the executor. What do you guys think? Any other ideas? @Kailai-Wang @kziemianek

tee-worker/identity/litentry/core/omni-account/src/in_memory_store.rs
```diff
@@ -0,0 +1,77 @@
...
+		let omni_account_members = STORE
+			.read()
+			.map_err(|_| {
+				log::error!("[InMemoryStore] Lock poisoning");
+				Error::LockPoisoning
+			})?
+			.get(&owner)
+			.cloned();
+
+		omni_account_members.ok_or(Error::NotFound)
```
kziemianek: `None` can naturally reflect that something is not present.

tee-worker/identity/litentry/core/omni-account/src/repository.rs
```diff
@@ -0,0 +1,95 @@
...
+			"AccountStore",
+			&owner,
+			&StorageHasher::Blake2_128Concat,
+		);
+		let storage_entry = self
+			.ocall_api
+			.get_storage_verified(storage_key, &self.header, &ParentchainId::Litentry)
+			.map_err(|_| Error::OCallApiError("Failed to get storage"))?;
+		let member_accounts: Vec<MemberAccount> =
+			storage_entry.value().to_owned().ok_or(Error::NotFound)?;
```
kziemianek: Same here, you can use `None` instead of error

kziemianek (approved):

tee-worker/identity/litentry/core/omni-account/src/in_memory_store.rs
silva-fj: Yes, this was to avoid returning `Result<Option<Vec<MemberAccount>>>` so it would be easier to handle. But I guess it's ok

PR #3117, opened 2024-10-07
Add omni account origin

Kailai-Wang:
### Context

As topic, other with two extra extrinsics:
- `dispatch_as_omni_account`
- `dispatch_as_signed`

```
// 1. to decouple `TEECallOrigin` and extrinsic that should be sent from `OmniAccount` origin only
// 2. allow other pallets to specify ensure_origin using this origin
// 3. leave room for more delicate control over OmniAccount in the future (e.g. multisig-like control)
```

### DIFF
--- a/parachain/pallets/omni-account/src/lib.rs
+++ b/parachain/pallets/omni-account/src/lib.rs
@@ -1,3 +1,19 @@
+// Copyright 2020-2024 Trust Computing GmbH.
+// This file is part of Litentry.
+//
+// Litentry is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Litentry is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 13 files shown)


parachain/pallets/omni-account/src/lib.rs
```diff
@@ -111,14 +176,56 @@ pub mod pallet {
...
 		#[pallet::call_index(0)]
 		#[pallet::weight((195_000_000, DispatchClass::Normal))]
+		pub fn dispatch_as_omni_account(
+			origin: OriginFor<T>,
+			account_hash: H256,
+			call: Box<<T as Config>::RuntimeCall>,
+		) -> DispatchResult {
+			let _ = T::TEECallOrigin::ensure_origin(origin)?;
+			let omni_account =
+				LinkedIdentityHashes::<T>::get(account_hash).ok_or(Error::<T>::IdentityNotFound)?;
```
silva-fj, line 189: Does this mean that `who` sent to `link_identity` is the OmniAccount? So, the OmniAccount should exist before linking any account, right?

Kailai-Wang: no - nothing changed for `link_identity`

`LinkedIdentityHashes` contains now the mapping member_account_hash <> omni-account

Kailai-Wang: I mean `who` should be the omni-account even in the current impl

silva-fj (approved):

kziemianek (approved):

PR #3120, opened 2024-10-08
Use encoded Identity for hashing

Kailai-Wang:
### Context

As topic, we don't need to convert to string first.


silva-fj (approved):
Cool, I didn't know `Encode` had this method

PR #3121, opened 2024-10-08
Adding OmniAccount core primitives

silva-fj:
These types will be shared between the omni-account pallet and the identity worker


Kailai-Wang (approved):
Looks fine, thanks

PR #3123, opened 2024-10-10
chore: P-1044 remove all di identity ts tests from ci

wli-pro:
so the followup [PR](https://github.com/litentry/litentry-parachain/pull/3113) to remove these ts tests will pass ci checks.

### Labels

### Testing Evidences

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -709,15 +709,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


Kailai-Wang:
I think it's enough to have https://github.com/litentry/litentry-parachain/pull/3113

It seems we don't need this one.

It didn't show as "passed" just because repo settings needed status report of those tests - now it's changed

PR #3124, opened 2024-10-10
Use cargo patch and decouple more deps

Kailai-Wang:
### Context

So that we don't have to maintain `release-v0.9.19` branch anymore for primitives.

We'll use the same code, just compiled with different (overriden) substrate deps in parachain and worker.

To achieve that, I needed to de-couple some pallet deps, so that worker doesn't depend on any of the parachain pallet. The copied `Delegator` struct is because of that.


kziemianek (approved):
👍🏼

PR #3126, opened 2024-10-11
OmniAccount derivation

Kailai-Wang:
### Context

As topic - it uses consistent method to derive an `OmniAccount` from a given `Identity`, so that the derived OmniAccount has no private key even for substrate/evm users.

To make the difference, the mapping from substrate/evm accounts is called `to_native_account()`.

It also simplifies some extrinsic parameter: e.g. `add_acount` doesn't require AccountStoreHash anymore.


parachain/pallets/omni-account/src/lib.rs
```diff
@@ -218,37 +195,29 @@ pub mod pallet {
 		#[pallet::weight((195_000_000, DispatchClass::Normal))]
 		pub fn add_account(
 			origin: OriginFor<T>,
-			who: Identity,
-			member_account: MemberAccount,
-			maybe_account_store_hash: Option<H256>,
+			who: Identity,                 // `Identity` used to derive OmniAccount
```
silva-fj, line 198: This means that we keep a similar concept of `PrimeIdentity`, right?

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -287,80 +258,51 @@ pub mod pallet {
...
+		/// make a member account public in the AccountStore
+		/// we force `Identity` type to avoid misuse and additional check
 		#[pallet::call_index(4)]
 		#[pallet::weight((195_000_000, DispatchClass::Normal))]
-		pub fn publicize_account(
-			origin: OriginFor<T>,
-			member_account_hash: H256,
-			public_identity: MemberIdentity,
-		) -> DispatchResult {
+		pub fn publicize_account(origin: OriginFor<T>, member_account: Identity) -> DispatchResult {
```
silva-fj, line 265: The `hash()` in `MemberAccount` is quite useful 👍🏼

silva-fj (approved):

parachain/pallets/omni-account/src/lib.rs
Kailai-Wang: Good question - on the contrary, actuall y I don't want to introduce the concept of prime identity.

`who` should have been of `T::AccountId` type (which is the type of OmniAccount), but we need to create the AccountStore if not present. For that we do need an `Identity` type.

I feel this is not very consistent, any better ideas? Or we introduce an explicit `create_account_store` extrinsic, which is called when the user logs in for the first time? I'm just a bit concerned by the UX (user needs ti sign the `create_account_store` request)

silva-fj: I think we should have an extrinsic to create the omni account, which internally creates the account store, adding the first account.  It would make it easier to understand for the users.

It made sense for the "link_identity" concept not to have a "create_id_graph", as users would just link another identity(s) to it's main (prime) identity, but for "add_account", we need to answer the question "add account to what?", so it feels natural to first create the omni account. Similar to a Multisig, you first create the Multisig, then you can add/remove signatories.

Kailai-Wang: The logic makes sense.

I just want minimal steps from UX perspective. Does `create_account_store` require user signature, or do we create it for users on the background as soon as user logs into our webapp? If the latter, is it then a public call?

Kailai-Wang: also cc @jonalvarezz

silva-fj: In the worst case scenario for new users, the UX for calling `create_account_store`, should be the same as calling `add_account` for the first time. I think we should require the signature, but this could be part of the "login" mechanism

jonalvarezz: I like the idea of having matching api signatures as Francisco suggests.

Kailai-Wang: I'll add an `create_account_store` interface, however, it should only be called when we really need to create a store (e.g. when adding accounts), not upon every login.

When to call which will be worker's logic then

PR #3127, opened 2024-10-11
Add `create_account_store` and adjust call origins

Kailai-Wang:
### Context

As topic.

We should (logically) separate `create_account_store` and `add_account`, because creating store requires Identity, while adding account not. And mutation to AccountStore should follow a consistent origin check.

When to call which is decided by the worker logic.


silva-fj (approved):
👍🏼

PR #3129, opened 2024-10-15
Setting up native tasks for identity worker

silva-fj:
This PR adds an initial setup to process `TrustedCall`s directly outside the stf executor, similar to how `request_vc` are handled. This is a temporary step, (we already have an issue to unify how to handle requests/tasks). This will allow us to start adding `TrustedCall`s related to the OmniAccount. Later we'll clean up the architecture.

### DIFF
--- a/tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
+++ b/tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
@@ -123,6 +123,6 @@ pub struct DirectCallRequestResult {
 	pub result: Result<DirectCallResult, DirectCallErrorDetail>,
 }
 
-pub enum DirectCallTask {
+pub enum NativeRequest {
 	// TODO: Define the tasks
 }
--- a/tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
+++ b/tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
@@ -1,0 +1,128 @@
+use codec::{Decode, Encode};
+use ita_sgx_runtime::Hash;
+use ita_stf::{Getter, TrustedCallSigned};
+use itp_extrinsics_factory::CreateExtrinsics;
+use itp_node_api::metadata::{provider::AccessNodeMetadata, NodeMetadataTrait};
+use itp_ocall_api::{EnclaveAttestationOCallApi, EnclaveMetricsOCallApi, EnclaveOnChainOCallApi};
+use itp_sgx_crypto::{key_repository::AccessKey, ShieldingCryptoDecrypt, ShieldingCryptoEncrypt};
+use itp_stf_executor::traits::StfEnclaveSigning as StfEnclaveSigningTrait;
+use itp_top_pool_author::traits::AuthorApi as AuthorApiTrait;
+use lc_data_providers::DataProviderConfig;
+use litentry_primitives::AesOutput;
+use sp_core::{ed25519::Pair as Ed25519Pair, H256};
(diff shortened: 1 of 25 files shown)


tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
```diff
@@ -0,0 +1,128 @@
...
+	ExtrinsicConstructionFailed(String), // Stringified itp_extrinsics_factory::Error
+	ExtrinsicSendingFailed(String),      // Stringified sgx_status_t
+}
+
+// split this type and move it to the right place once we have the new architecture
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub struct DirectCallResult {
+	pub account_store_hash: H256,
+	pub mutated_account_store: AesOutput,
+}
```
Kailai-Wang: It sounds like the "DirectCall" is only about omni-account handling, even if that's true, I don't think these fields are required for every such call handling.

Basically it should be more generalised - let's use empty result `()` for now. Unlike vc-request, we don't know yet if rpc call should return immediately or wait until some status is available (e.g. finalised?)

tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
```diff
@@ -0,0 +1,128 @@
...
+
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub struct DirectCallRequestResult {
```
Kailai-Wang: Maybe we don't need this extra wrapper - can we use `Result< , >` directly?

tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
```diff
@@ -0,0 +1,128 @@
...
+	pub result: Result<DirectCallResult, DirectCallErrorDetail>,
+}
+
+pub enum DirectCallTask {
```
Kailai-Wang: Let's call it `NativeRequest` for now

Kailai-Wang (commented):
The skeleton looks fine

Let's call them `Native ....`, and the directory as `core/native-task/` to be consistent with `stf-task` and `vc-task`

tee-worker/identity/sidechain/rpc-handler/src/direct_top_pool_api.rs
```diff
@@ -120,6 +121,24 @@ pub fn add_top_pool_direct_rpc_methods<R, TCS, G>(
 		Ok(json!(json_value))
 	});
 
+	io_handler.add_sync_method("author_SubmitDirectCallRequest", move |params: Params| {
```
Kailai-Wang: `submit` starts with small letter `s` : )

tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
silva-fj: ah yes, this type was for an example and it should not be here. The idea was to have a result type for each trusted call in `trusted_call_result.rs` as we currently have until we move things that are not related to `stf`

tee-worker/identity/litentry/core/direct-call-request/receiver/src/types.rs
silva-fj: I think it would be better, I was using the same structure we have for `vc` (`RequestVcResultOrError`) but we probably don't need this.

Kailai-Wang (approved):

PR #3131, opened 2024-10-16
One-time command to upload the current idgraph to parachain

Kailai-Wang:
### Context

As topic.

We just stop the worker and do `RUST_LOG=info ./litentry-worker upload-id-graph -m <shard>` - it will not return before all txs are finalised on-chain.

`-m <shard>` is optional if the shard (mrenclave) is the same as where IDGraph is stored.

Locally tested ok

image


tee-worker/identity/enclave-runtime/src/rpc/common_api.rs
```diff
@@ -506,6 +516,105 @@ pub fn add_common_api<Author, GetterExecutor, AccessShieldingKey, OcallApi, Stat
 			Err(_) => Ok(json!(compute_hex_encoded_return_error("Could not parse params"))),
 		}
 	});
+
+	let local_state = state.clone();
+
+	io_handler.add_sync_method("omni_UploadIDGraph", move |params: Params| {
+		debug!("worker_api_direct rpc was called: omni_UploadIDGraph");
```
kziemianek: I would suggest putting this in startup code, guarded by CLI argument.

Currently anybody can call this method - it will flood parachain.
Another thing is: upload should be triggered **after** sidechain's imp pallet becomes immutable and **before** parachain storage becomes mutable - otherwise there will be conflicts/outdated account stores.

kziemianek (commented):
If we decide to keep it as it is we need to call it ultimately one more time once sidechai's IMP becomes immutable (trusted call are not touching idgraps) and before IMP pallet storage is removed so I assume there will be worker version supporting this command and disabling IMP's related trusted calls.

tee-worker/identity/enclave-runtime/src/rpc/common_api.rs
Kailai-Wang: I didn't go for this as I wanted to save the communication between non-enclave and enclave,  as it's a one-time thing. `omni_UploadIDGraph` could be called multiple times - e.g. right before we switch to the new on-chain IDGraph, the whole process should be done within 1 minute - like it's hard for any abusers to predict when we are going to do the upgrade.

Let me change it to start-up sub-command. In that way we'll stop worker, do the upload (so that nothing changes at the same time), and restart it with newer version that supports on-chain Graph

BillyWooo: `start-up sub-command` that would be a better solution. Completely offline. No worries for any potential hacks.

tee-worker/identity/enclave-runtime/src/initialization/global_components.rs
```diff
@@ -106,6 +109,7 @@ pub type EnclaveGetter = Getter;
 pub type EnclaveTrustedCallSigned = TrustedCallSigned;
 pub type EnclaveStf = Stf<EnclaveTrustedCallSigned, EnclaveGetter, StfState, Runtime>;
 pub type EnclaveStateKeyRepository = KeyRepository<Aes, AesSeal>;
+pub type EnclaveIDGraphKeyRepository = KeyRepository<Aes256Key, Aes256Seal>;
```
silva-fj: How about `EnclaveAccountStoreKeyRepository`?

tee-worker/bitacross/enclave-runtime/src/initialization/global_components.rs
```diff
@@ -68,7 +68,7 @@ use itp_sgx_crypto::{
 	ecdsa::{Pair as EcdsaPair, Seal as EcdsaSeal},
 	key_repository::KeyRepository,
 	schnorr::{Pair as SchnorrPair, Seal as SchnorrSeal},
-	Aes, AesSeal, Ed25519Seal, Rsa3072Seal,
+	Aes, AesSeal, Ed25519Seal, IDGraphKey, IDGraphKeySeal, Rsa3072Seal,
```
silva-fj: Do these types exist?

tee-worker/identity/enclave-runtime/src/initialization/global_components.rs
Kailai-Wang: Yes I changed it locally - haven't pushed it yet, thanks for checking

tee-worker/bitacross/enclave-runtime/src/initialization/global_components.rs
Kailai-Wang: No - that's why compilation failed : ) I fixed it locally too

tee-worker/identity/enclave-runtime/src/rpc/common_api.rs
Kailai-Wang: I updated it

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -171,7 +172,7 @@ pub mod pallet {
 				who: omni_account,
 				result: result.map(|_| ()).map_err(|e| e.error),
 			});
-			Ok(())
+			Ok(Pays::No.into())
```
kziemianek, line 167: I've also noticed it missing but in my case I had to:
https://github.com/litentry/litentry-parachain/pull/3132/files#diff-960f61f16faf996e829d1decb06b54fb404525cc02cdc585b9e8db86d596014dR344

Otherwise all extrinsincs signed by newly generated account were rejected (low funds)

kziemianek (approved):
I've checked it locally - works fine for small amount of idgraphs.

silva-fj (approved):

PR #3133, opened 2024-10-17
AccountStore in-memory state initialization and synchronization

silva-fj:
As topic, The main changes in this PR are the following:

- In memory AccountStore initialization: During worker's initialization it gets the state from the parachain and load the account stores into memory, with the block height. (this happens only once)
- In memory state sync on every relevant omni-account pallet event (if not processed yet, based on the block number)


tee-worker/identity/service/src/main_impl.rs
```diff
@@ -162,6 +162,11 @@ pub(crate) fn main() {
 		enclave_metrics_receiver,
 	)));
 
+	// init in-memory store, it should be done after the o-call bridge is initialized
+	if let Err(e) = enclave.init_in_memory_state() {
```
Kailai-Wang, line 166: Should it be called `init_in_memory_store` according to the comment? Or is it actually a more generic call for potential similar initialisation in the future?

Kailai-Wang (approved):
I've got pretty much nothing to complain, thanks!

Have you tried it locally?
- I had some problem with nested ecall -> ocall, in some old issue though
- Does the paged retrieval work correctly?

tee-worker/identity/service/src/main_impl.rs
silva-fj: Yes, the idea was to make it generic in case we need something similar in the future

silva-fj:
> I've got pretty much nothing to complain, thanks!
> 
> Have you tried it locally?
> 
> * I had some problem with nested ecall -> ocall, in some old issue though
> * Does the paged retrieval work correctly?

I ran it locally but it's not fully tested as we currently don't have an easy way to generate enough data for testing

Kailai-Wang:
I think a (so far) easy way is to use cli to batch request vcs (which will generate new IDGraphs if it's a new account), and then use `upload-id-graph` to upload it on-chain 😂  

Then you can see if it can be downloaded

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -119,13 +119,25 @@ pub mod pallet {
...
 	pub enum Event<T: Config> {
 		/// An account store is created
-		AccountStoreCreated { who: T::AccountId },
+		AccountStoreCreated { who: T::AccountId, account_store: MemberAccounts<T> },
 		/// Some member account is added
-		AccountAdded { who: T::AccountId, member_account_hash: H256 },
+		AccountAdded {
+			who: T::AccountId,
+			member_account_hash: H256,
+			account_store: MemberAccounts<T>,
```
kziemianek: Why does it need to contain all accounts ? Does it make sense to have only added account here ?
For a long run it doesn't seem to be efficient.

If we decide to keep it, maybe worker is only interested in `AccountStoreChanged` event ? 
It could be emitted in case of every account store modification (4 event types reduced to 1)

tee-worker/identity/app-libs/parentchain-interface/src/integritee/event_handler.rs
```diff
@@ -346,10 +377,58 @@ where
...
+		}
+
+		if let Ok(events) = events.get_account_made_public_events() {
+			debug!("Handling AccountMadePublic events");
+			events
+				.into_iter()
+				.try_for_each(|event| {
+					debug!("found AccountMadePublic event: {:?}", event);
+					handled_events.push(hash_of(&event));
+					Self::update_account_store(event.who, event.account_store, block_number)
```
kziemianek: In Case of `AccountStoreUpdated` event these 4 branches could be reduced to 1.

parachain/pallets/omni-account/src/lib.rs
silva-fj: I had a similar thought, the idea is that we don't have to manually mutate the in-memory state in the worker as well. Also, since we don't re-validate the data cached it would be safer in case the worker gets out of sync for some reason as the last event for an account would always keep it up to date. If the concern is that users could have too many accounts I think we could reduce the limit to something more realistic, we currently have the same limit as the IDGraphs (64) but for the omni-account it probably does not make much sense.

I understand the idea to reduce the events to just one, but I think having separate events also makes it easier for us to understand the user's behaviour (if we ever want to track this)

kziemianek: There can be 5 event types, 4 indicating action ( ex `AccountAdded`) and one indicating sideeffect (`AccountStoreChanged`).

But I'm still more concerned about event size in this case, if there is alternative we should explore it.

silva-fj: > There can be 5 event types, 4 indicating action ( ex AccountAdded) and one indicating sideeffect (AccountStoreChanged).

True, that's a good idea.

> But I'm still more concerned about event size in this case, if there is alternative we should explore it.

Yeah that's a valid concern. Let's see what else we can do, maybe having an aggressive cache re-validation  or simply request the data on demand would not be that bad? There are some tradeoff we should consider.

But I think in reality, let's say on average users have 4 or even 10 accounts, I don't think that would be too much of an issue

Kailai-Wang: Yes I'd say we are safe for now - it's a boundedVec too, we can always limit the size

kziemianek (approved):

PR #3134, opened 2024-10-17
Skip sidechain block import confirmation

Kailai-Wang:
### Context

As topic.

It should be caused by [failed teebag.sidechainBlockImported ext](https://litentry.statescan.io/#/extrinsics?spec=9201&section=teebag&method=sidechainBlockImported&time_dimension=block&page=1).

For now I simply skip the confirmation as I don't see it bringing a lot of value at the moment


BillyWooo (approved):
”sidechain_block_imported“ does not bring much value here. We can skip it. 👍

kziemianek (approved):

PR #3138, opened 2024-10-18
Makefile build flag is not passed inside enclave-runtime

BillyWooo:
Small, but very important PR. 
We all overlooked this 😿 

I would suggest to make a following PR to remove all IAS related code. There won't be different mode, but only dcap.


Kailai-Wang:
I don't fully understand this, can you please elaborate why this is the root cause for SGX crash?

It seems that we indeed haven't applied `dcap` feature for enclave compilation, but questions are:
1. why did it always work locally in your case? Allegedly it was verified multiple times
2. from the code - this should affect TLS server/client only - so primarily multi-worker situation. How is it related to the enclave registration and crash?

Kailai-Wang:
cc @kziemianek as you once made some changes to enclave-runtime/Makefile too

BillyWooo:
> I don't fully understand this, can you please elaborate why this is the root cause for SGX crash?
> 
> It seems that we indeed haven't applied `dcap` feature for enclave compilation, but questions are:
> 
> 1. why did it always work locally in your case? Allegedly it was verified multiple times
> 2. from the code - this should affect TLS server/client only - so primarily multi-worker situation. How is it related to the enclave registration and crash?

The crash will only be triggered from “create_ra_report_and_signature”, which won't be executed during initialization.

Kailai-Wang (approved):
Thank you!

Shall we apply the same change to bc-worker too?

Kailai-Wang:
@m1iktea please block external access to port 2001 and 3443 for identity-worker, and 4545 too if you don't use it for healthcheck.

It will help us not accept unneeded connections...

BillyWooo:
> Thank you!
> 
> Shall we apply the same change to bc-worker too?

Yes. We should. Let me make a separate PR for that. 
I would like to merge and get  new id-worker release out ASAP.

m1iktea:
> @m1iktea please block external access to port 2001 and 3443 for identity-worker, and 4545 too if you don't use it for healthcheck.
> 
> It will help us not accept unneeded connections...

I have disabled public network access in the firewall, but I did not remove these two ports from the startup parameters.

PR #3143, opened 2024-10-24
Adding request_intent trusted call

silva-fj:
This PR adds the first trusted call related to the omni-account, here are some of the most important changes:

- Since we are not using the sidechain we are now using the nonce of the parentchain for the call verifications.
- The `dispatch_as_omni_account` call takes the `nonce` of the omni account. (Included in the TrustedCall), so this part of the verification happens on-chain.


common/primitives/core/src/intent.rs
```diff
@@ -1,16 +1,23 @@
...
 #[derive(Encode, Decode, Debug, Clone, PartialEq, Eq, MaxEncodedLen, TypeInfo)]
 pub enum Intent {
     #[codec(index = 0)]
     TransferEthereum(TransferEthereum),
     #[codec(index = 1)]
     CallEthereum(CallEthereum),
+    #[codec(index = 2)]
+    SystemRemark(BoundedVec<u8, ConstU32<MAX_REMARK_LEN>>),
+    #[codec(index = 3)]
+    TransferNative(TransferNative),
```
Kailai-Wang, line 20: While this is fine for the demo, how about having a more generic `CallNative(Box<RuntimeCall>)`?

tee-worker/identity/enclave-runtime/src/rpc/common_api.rs
```diff
@@ -506,6 +515,48 @@ pub fn add_common_api<Author, GetterExecutor, AccessShieldingKey, OcallApi, Stat
 			Err(_) => Ok(json!(compute_hex_encoded_return_error("Could not parse params"))),
 		}
 	});
+
+	io_handler.add_sync_method(
+		"omni_account_requestEmailVerificationCode",
```
Kailai-Wang: let's call it `omni_requestEmailVerificationCode`

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -173,11 +174,15 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			member_account_hash: H256,
 			call: Box<<T as Config>::RuntimeCall>,
+			nonce: <T as frame_system::Config>::Nonce,
```
Kailai-Wang: It can be confusing if one only reads the parachain code, it's unclear why a `nonce` is required here (and what value it should be), AFAICT it's only relevant if we want to verify/add member account, right?

Shall we make it optional and add a comment when it's required? Like `maybe_expected_nonce: Option<Nonce>`

Kailai-Wang (approved):
In general looks good, thanks. I left comments regarding `nonce`

parachain/pallets/omni-account/src/lib.rs
silva-fj: Actually, we probably don't need this at all

PR #3145, opened 2024-10-25
add running instructions

kziemianek:


tee-worker/omni-executor/README.md
```diff
@@ -1,3 +1,19 @@
...
+Gramine is required for running inside TEE, please refer to [installation options](https://gramine.readthedocs.io/en/stable/installation.html).
+
+1. `make SGX=1` to build and sign application
+2. `RUST_LOG=info gramine-sgx omni-executor -- <parentchain-rpc-url> <ethereum-rpc-url>`
+
+
+## Running whole setup locally
+
+Build omni-executor docker image first `make build-docker`.
+Start local environment using `docker-compose up` command.
```
Kailai-Wang: Can we use `docker compose` instead of `docker-compose` in Makefile? The former is the standard (and more preferred) now

Kailai-Wang: Shall we use `make start-local` here to be more consistent?

Kailai-Wang (commented):
Thanks!

So `make start-local` will not start gramine, but just a native rust (image), right?

Btw `make start-gramine-server` requires eth and parachain node URL

kziemianek:
> Thanks!
> 
> So `make start-local` will not start gramine, but just a native rust (image), right?
> 
> Btw `make start-gramine-server` requires eth and parachain node URL

Yes, we can adjust it to use `gramine-sgx` or `gamine-direct`.
Current way was the most straightforward.

tee-worker/omni-executor/Makefile
```diff
@@ -51,7 +51,7 @@ endif
 
 .PHONY: start-gramine-server
 start-gramine-server: all
-	$(GRAMINE) tee-bridge
+	$(GRAMINE) omni-executor ${ARGS}
```
BillyWooo, line 54: There is renaming above "omni-executor" -> "executor-worker". Here it still uses "omni-executor"?

Kailai-Wang (approved):

BillyWooo (approved):

PR #3148, opened 2024-10-28
Add runtime-api call to get OmniAccount

Kailai-Wang:
### Context

As topic, locally tested OK


parachain/pallets/omni-account/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use core_primitives::Identity;
+use parity_scale_codec::{Decode, Encode};
+
+sp_api::decl_runtime_apis! {
+	pub trait OmniAccountApi<AccountId>
+	where
+		AccountId: Encode + Decode,
```
silva-fj: NIT: we could do `use parity_scale_codec::Codec;`

```suggestion
		AccountId: Codec,
```

silva-fj (approved):
👍🏼

BillyWooo (approved):

jonalvarezz (commented):
Thank you, Kai! 🚀

I can see it locally from the p-js explorer: 
Screenshot 2024-10-28 at 11 31 01

However, I can't reach it programmatically. It should be on `api.call`, shouldn't it? 
Screenshot 2024-10-28 at 11 32 51

I also tried forcing a client-api update off your branch to get the latest types, but i can't see it either:
Screenshot 2024-10-28 at 11 33 48


I'm not sure if this is really part of the scope of this PR, but wanted to bring it up in case we are missing anything.

Kailai-Wang:
@jonalvarezz `api.rpc` requires integration in the client (node), runtime-api has/should have the advantage that it's exposed automatically by substrate-based node.

Can you try JSON-RPC with `state_call` or `api.rpc.state.call`? Method should be something like `OmniAccountApi_omni_account`

jonalvarezz:
> @jonalvarezz `api.rpc` requires integration in the client (node), runtime-api has/should have the advantage that it's exposed automatically by substrate-based node.
> 

Yeh exactly, I was expecting it to be exposed automatically, but I can't find it. I'm not sure what I'm missing.



> Can you try JSON-RPC with `state_call` or `api.rpc.state.call`? Method should be something like `OmniAccountApi_omni_account`

Sweet, this one worked:

Screenshot 2024-10-28 at 18 20 01

```ts
const account = await api.rpc.state.call('OmniAccountApi_omni_account', api.createType('CorePrimitivesIdentity', {
  Substrate:
    '0x3a2647fe0ecd1fe0d450fd8650cd8bf3536ecac77b6e8ec2df1b5f5fc9d62931',
}).toHex());
```

I guess we can live without this types meanwhile 🤔

Kailai-Wang:
Cool then I'll merge it

PR #3149, opened 2024-10-29
Fix dockerfile warning

m1iktea:
### Context

fx dockerfile warning, https://linear.app/litentry/issue/DO-54/clean-up-dockerfile-warning-message-in-pr

### Labels

- `C1-noteworthy`

### DIFF
--- a/tee-worker/identity/build.Dockerfile
+++ b/tee-worker/identity/build.Dockerfile
@@ -25,7 +25,7 @@ LABEL maintainer="Trust Computing GmbH <info@litentry.com>"
 ENV SGX_SDK=/opt/sgxsdk
 ENV PATH="$PATH:${SGX_SDK}/bin:${SGX_SDK}/bin/x64:/opt/rust/bin"
 ENV PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:${SGX_SDK}/pkgconfig"
-ENV LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${SGX_SDK}/sdk_libs"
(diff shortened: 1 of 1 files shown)


tee-worker/identity/build.Dockerfile
```diff
@@ -25,7 +25,7 @@ LABEL maintainer="Trust Computing GmbH <info@litentry.com>"
 ENV SGX_SDK=/opt/sgxsdk
 ENV PATH="$PATH:${SGX_SDK}/bin:${SGX_SDK}/bin/x64:/opt/rust/bin"
 ENV PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:${SGX_SDK}/pkgconfig"
-ENV LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${SGX_SDK}/sdk_libs"
+ENV LD_LIBRARY_PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:${SGX_SDK}/sdk_libs"
```
BillyWooo, line 28: This LD_LIBRARY_PATH is less important than the following one. 
"Sensitive data should not be used in the ARG or ENV commands" -> please follow the info: `https://docs.docker.com/go/dockerfile/rule/secrets-used-in-arg-or-env/` ; then there is suggestion about how to: `https://docs.docker.com/build/building/secrets/`

m1iktea: Regarding this warning, I believe it is because GitHub is treating the SGX_COMMERCIAL_KEY environment variable as a secret key that needs to be kept confidential. However, in reality, this is just the filename of a key file, so no further modifications are needed.

BillyWooo:
After investigate [build secrete](https://docs.docker.com/build/building/secrets/), it's not worth to spend much resource here just to solve the warnings.

PR #3150, opened 2024-10-29
`omni-executor` enclave registration

kziemianek:
It uses dcap attestation method. Currently always built with sgx debug = 1


tee-worker/omni-executor/Makefile
```diff
@@ -73,6 +73,6 @@ start-local:
 stop-local:
 	docker compose down
 
-.PHONY: get-omni-pallet-metadata
-get-omni-pallet-metadata:
-	subxt metadata --url http://localhost:9944 --allow-insecure --pallets OmniAccount > parentchain/artifacts/rococo-omni-account.scale
+.PHONY: get-pallet-metadata
+get-pallet-metadata:
```
Kailai-Wang, line 77: Where exactly is this target used? I see we do refer to this file in the listener, but the build order / depedency doesn't reflect that

tee-worker/omni-executor/parentchain/listener/src/lib.rs
```diff
@@ -130,3 +147,60 @@ pub async fn create_listener<EthereumIntentExecutorT: IntentExecutor + Send + Sy
...
+	#[cfg(feature = "gramine-quote")]
+	{
+		use std::fs;
+		use std::fs::File;
+		use std::io::Write;
+		let mut f = File::create("/dev/attestation/user_report_data").unwrap();
+		let content = signer.public_key().0;
+		f.write_all(&content).unwrap();
+
+		quote = fs::read("/dev/attestation/quote").unwrap();
```
Kailai-Wang, line 181: Do we know how to generate/mount this?

tee-worker/omni-executor/parentchain/listener/src/transaction_signer.rs
```diff
@@ -0,0 +1,118 @@
...
+		let genesis_hash = client.get_genesis_hash().await.unwrap();
+
+		let nonce = *self
+			.nonce
+			.read()
+			.map_err(|e| {
+				error!("Could not read nonce: {:?}", e);
+				RecoverableError
+			})
+			.unwrap();
```
Kailai-Wang, line 91: Is there any methods from `client` that retrieves the nonce? It seems we don't need to store it actually

Kailai-Wang (approved):
Looks good, thanks
Have you tried it with real dcap data?

tee-worker/omni-executor/parentchain/listener/src/transaction_signer.rs
kziemianek: Yes, there is todo in line 59, i'll improve that in separate pr.

tee-worker/omni-executor/parentchain/listener/src/lib.rs
kziemianek: its exposed by gramine
https://gramine.readthedocs.io/en/stable/attestation.html#low-level-dev-attestation-interface

tee-worker/omni-executor/Makefile
kziemianek: Currently we need to manually run it. Basically every time these pallets changes. It's not convinient and I think we will need to another strategy.

PR #3151, opened 2024-10-29
Fixing verification code email

silva-fj:
- Updating template
- Fixing content payload


Kailai-Wang (approved):
Thanks - one simple question is shall we display different text for different actions, e.g. login, authorize to execute intent etc

PR #3153, opened 2024-10-30
3rd party ethereum node integration adjustments

kziemianek:
Some adjustments needed to make ethereum work with 3rd party node - infura for example.

* install required ca-certificates in docker image
* set tx gas price
* get nonce instead of account


Kailai-Wang (approved):

tee-worker/omni-executor/ethereum/intent-executor/src/lib.rs
```diff
@@ -43,22 +43,28 @@ impl IntentExecutor for EthereumIntentExecutor {
...
 		.unwrap();
 		let wallet = EthereumWallet::from(signer);
 		let provider = ProviderBuilder::new()
 			.with_recommended_fillers()
 			.wallet(wallet)
 			.on_http(self.rpc_url.parse().map_err(|e| error!("Could not parse rpc url: {:?}", e))?);
-		let account =
-			provider.get_account(provider.signer_addresses().next().unwrap()).await.unwrap();
+		let nonce = provider
+			.get_transaction_count(provider.signer_addresses().next().unwrap())
```
Kailai-Wang, line 55: Dumb question: what's the diff?

Kailai-Wang (approved):

tee-worker/omni-executor/ethereum/intent-executor/src/lib.rs
kziemianek: `eth_getAccount` is not listend in ethereum api documentation https://ethereum.github.io/execution-apis/api-documentation/ , it worked for locally run anvil node but failed while I was testing the integration with infura's api.

PR #3155, opened 2024-10-31
Introducing add_account trusted call

silva-fj:
As topic.

Most important changes:
- New trusted call `add_account`: As this is a native task, we no longer need a two steps process for the identity verification.
- The verification message now uses the parachain nonce.
- `send_to_parentchain` method of the `ocall_api` has been refactored to allow waiting for other extrinsic statuses and to be able to return the extrinsic result.
- Native task response (new trusted calls results) now include the extrinsic report, so clients can query the on-chain data.


Kailai-Wang (commented):
Could you have a very basic integration (ts) test for it? Or you could ask @0xverin to do it

silva-fj:
> Could you have a very basic integration (ts) test for it? Or you could ask @0xverin to do it

Yes, but I think we should add the tests in a separate PR as it would require quite some changes to prepare the test suit (at least for the first test, once we have the utilities and helpers in place it would be trivial to add other test cases). Also, we could reuse what @jonalvarezz has been working on related to the omni account and the new types like `TrustedCallAuthenticated`, `TCAuthentication`, etc. Since we need to create the definitions and generate the new ts types as well

tee-worker/common/core-primitives/node-api/api-client-types/src/lib.rs
```diff
@@ -74,6 +79,103 @@ pub type ParentchainSignature = Signature<ParentchainSignedExtra>;
 /// Signature type of the [UncheckedExtrinsicV4].
 pub type Signature<SignedExtra> = Option<(Address, PairSignature, SignedExtra)>;
 
+/// Simplified TransactionStatus to allow the user to choose until when to watch
+/// an extrinsic.
+// Indexes must match the substrate_api_client::TransactionStatus::as_u8
```
Kailai-Wang, line 83: Dumb Q: why is that? 👀  Where is it actually force-casted?

tee-worker/identity/enclave-runtime/src/ocall/on_chain_ocall.rs
```diff
@@ -47,14 +59,23 @@ impl EnclaveOnChainOCallApi for OcallApi {
...
+				response.as_mut_ptr(),
+				response_size as u32,
 			)
 		};
 
 		ensure!(rt == sgx_status_t::SGX_SUCCESS, rt);
 		ensure!(res == sgx_status_t::SGX_SUCCESS, res);
 
-		Ok(())
+		let decoded_response: Vec<ExtrinsicReport<H256>> = Decode::decode(&mut response.as_slice())
```
Kailai-Wang, line 74: Is it possible that we actually receive more bytes than `response_size` due to encoding?

tee-worker/bitacross/service/src/ocall_bridge/worker_on_chain_ocall.rs
```diff
@@ -144,101 +161,102 @@ where
...
+				)
+			})?;
+		let extrinsics: Vec<OpaqueExtrinsic> = Decode::decode(&mut extrinsics_encoded.as_slice())
+			.map_err(|_| {
+			OCallBridgeError::SendExtrinsicsToParentchain("Could not decode extrinsics".to_string())
+		})?;
+		let mut extrinsic_reports: Vec<ExtrinsicReport<H256>> = Vec::new();
+		let parentchain_id = ParentchainId::decode(&mut parentchain_id.as_slice())?;
+		debug!(
+			"Enclave wants to send {} extrinsics to parentchain: {:?}. await each inclusion: {:?}",
```
Kailai-Wang: Do you want to adjust the comment too?

tee-worker/identity/litentry/core/native-task/receiver/src/lib.rs
```diff
@@ -143,13 +130,106 @@ pub fn run_native_task_receiver<
...
+	let authentication_valid = match tca.authentication {
+		TCAuthentication::Web3(signature) => verify_tca_web3_authentication(
+			&signature,
+			&tca.call,
+			tca.nonce,
+			&mrenclave,
+			&request.shard,
+		),
+		TCAuthentication::Email(verification_code) =>
+			verify_tca_email_authentication(&tca.call, verification_code),
```
Kailai-Wang, line 211: Shall we double-check that the `sender_identity` is actually an email identity in `verify_tca_email_authentication`?

tee-worker/identity/litentry/core/native-task/receiver/src/types.rs
```diff
@@ -123,8 +135,5 @@ pub enum NativeTaskError {
 	ExtrinsicConstructionFailed(String), // Stringified itp_extrinsics_factory::Error
 	ExtrinsicSendingFailed(String),      // Stringified sgx_status_t
 	InvalidRequest,
-}
-
-pub enum NativeRequest {
-	// TODO: Define the tasks
```
Kailai-Wang, line 129: I see this is commented out, does it mean we are re-using TrustedCall enum for now?

Kailai-Wang (commented):
Thanks, it looks good in general, I left some comments for (id-)worker, ditto for bc-worker.

The most important thing if the `response_size` is sufficient.

The correctness of `add_account` will need to be verified by ts-test

tee-worker/common/core-primitives/node-api/api-client-types/src/lib.rs
silva-fj: This code is actually copied from `substrate-api-client` so we can implement `Encode`/`Decode` on these types, I forgot to add the comments.

tee-worker/identity/litentry/core/native-task/receiver/src/lib.rs
silva-fj: This check is implicit as the verification code will not be found if the identity is different. We insert `Identity::from_email(&email).hash()` as part of the key

tee-worker/identity/litentry/core/native-task/receiver/src/types.rs
silva-fj: Initially this was the equivalent of `Web2IdentityVerificationRequest`, but since we don't need to add the web2 accounts in two steps anymore it's not needed. Yes, for now we are re-using  the`TrustedCall` enum for the native tasks, as we do for `request_vc`

tee-worker/identity/enclave-runtime/src/ocall/on_chain_ocall.rs
silva-fj: Good question, I tried calculating the size with the encoded values but when testing it locally I got an error because the buffer was not big enough, but after changing it to this it worked fine. Let me revisit this

tee-worker/common/core-primitives/node-api/api-client-types/src/lib.rs
```diff
@@ -74,6 +79,105 @@ pub type ParentchainSignature = Signature<ParentchainSignedExtra>;
 /// Signature type of the [UncheckedExtrinsicV4].
 pub type Signature<SignedExtra> = Option<(Address, PairSignature, SignedExtra)>;
 
+// The following types are copied from the substrate-api-client crate to be able to encode/decode them
+/// Simplified TransactionStatus to allow the user to choose until when to watch
+/// an extrinsic.
+// Indexes must match the substrate_api_client::TransactionStatus::as_u8
```
kziemianek, line 83: Can we have a test case so we can prove they match ?

kziemianek (approved):

tee-worker/identity/litentry/core/native-task/receiver/src/lib.rs
Kailai-Wang: I didn't find the code `Identity::from_email(&email).hash()`, this is 

```
pub fn verify_tca_email_authentication(call: &TrustedCall, verification_code: String) -> bool {
	let identity_hash = call.sender_identity().hash();
	match OmniAccountStore::get_omni_account(identity_hash) {
		Ok(Some(account_id)) => match VerificationCodeStore::get(&account_id, identity_hash) {
			Ok(Some(code)) => code == verification_code,
			_ => false,
		},
		_ => false,
	}
}
```

I mean here we don't check if the `call.sender_identity()` is actually an email identity, does it mean theoretically I (as the client) could wrongly construct a TrustedCall whose first identity is non-email identity while claiming this is an email-TCA?

tee-worker/identity/litentry/core/native-task/receiver/src/types.rs
Kailai-Wang: I see - it's fine for now

But let's keep in mind that in the future we might need them different names or enum types

tee-worker/identity/litentry/core/native-task/receiver/src/lib.rs
silva-fj: > I didn't find the code Identity::from_email(&email).hash()

ah, that was an example of the end result. The actual code is this:
```rust
let email_identity = Identity::from_email(&email);

match VerificationCodeStore::insert(
    omni_account,
    email_identity.hash(),
    verification_code.clone(),
) { 
    ...
}
```
and can be found [here](https://github.com/litentry/litentry-parachain/blob/9d797a513fadae2b05eebc6b4a311d38b4e7b05d/tee-worker/identity/enclave-runtime/src/rpc/common_api.rs#L521)

> I mean here we don't check if the call.sender_identity() is actually an email identity, does it mean theoretically I (as the client) could wrongly construct a TrustedCall whose first identity is non-email identity while claiming this is an email-TCA?

Yes, the client could construct a TrustedCall with another variant of Identity but the verification will never pass. When the authentication is not valid we return an error that the authentication verification failed, this is a similar logic of what we currently have if the signature verification fails, except that currently we don't say why the operation failed :
```rust
if !trusted_call.verify_signature(&mrenclave.m, &shard) {
    error!("TrustedCallSigned: bad signature");
    return Ok(ExecutedOperation::failed(operation_hash, top_or_hash, vec![], vec![]))
}
```
Do you think we should return different errors based on the authentication type (Web2, Web3)?

tee-worker/identity/litentry/core/native-task/receiver/src/types.rs
silva-fj: Yeah, we need to do some clean up when doing the migration or when unifying the api

tee-worker/identity/enclave-runtime/src/ocall/on_chain_ocall.rs
silva-fj: I've implemented `MaxEncodedLen` on the types so we get the maximum possible size when encoded

tee-worker/common/core-primitives/node-api/api-client-types/src/lib.rs
silva-fj: done

tee-worker/identity/litentry/core/native-task/receiver/src/lib.rs
Kailai-Wang: I see, so `VerificationCodeStore` is like a double map

The type isn't that critical

Kailai-Wang (approved):
I'm fine with it, we are good to go after conflicts are resolved

PR #3168, opened 2024-11-06
P 993 improve enclaveidentifier managing mechanism

BillyWooo:
As title, this PR moves EnclaveIdentifier under root/admin control. 

So in production env,  only Root/Admin can update the whitelist of EnclaveIdentifier. And only the whitelisted EnclaveIdentifier can register enclaves. 

This is a noteworthy change for deployment. 

Besides the feature implementation, there are some path fix (originally coming from folder structure change).


parachain/pallets/teebag/src/lib.rs
```diff
@@ -700,21 +713,41 @@ impl<T: Config> Pallet<T> {
...
 	}
 
 	pub fn add_enclave(sender: &T::AccountId, enclave: &Enclave) -> Result<(), Error<T>> {
-		match EnclaveRegistry::<T>::get(sender) {
-			Some(old_enclave) => ensure!(
-				old_enclave.worker_type == enclave.worker_type,
-				Error::<T>::UnexpectedWorkerType
-			),
-			None => Self::add_enclave_identifier(enclave.worker_type, sender)?,
+		match Self::mode() {
```
silva-fj, line 734: How about removing the duplicated check of the worker_type? something like:

```rust
if let Some(old_enclave) = EnclaveRegistry::<T>::get(sender) {
    ensure!(old_enclave.worker_type == enclave.worker_type, Error::<T>::UnexpectedWorkerType);
}

match Self::mode() {
    OperationalMode::Production | OperationalMode::Maintenance => {
        ensure!(
            EnclaveIdentifier::<T>::get(enclave.worker_type).contains(sender),
	    Error::<T>::EnclaveIdentifierNotExist
         );
    },
    OperationalMode::Development => {
        if EnclaveRegistry::<T>::get(sender).is_none() {
	    Self::add_enclave_identifier_internal(enclave.worker_type, sender)?;
	}
    },
};
```
it would also make the logic easier to read

silva-fj (approved):
LGTM! I just left a comment with a small suggestion

Kailai-Wang (approved):

PR #3171, opened 2024-11-07
Add option to specify start block

kziemianek:
Adds option to specify `start_block` for `omni-executor`'s parentchain listener. 

also:
* rpc client errors are logged
* wait 1s in case of sync error
* system signal handling


BillyWooo (approved):
LGTM

PR #3173, opened 2024-11-08
Adding integration tests for the new trusted calls (omni account)

silva-fj:
As topic, this PR adds integration tests for the omni account trusted calls as well as bug fixes and some clean up.


tee-worker/identity/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
```diff
@@ -25,7 +32,7 @@ export default {
         },
         TrustedCallAuthenticated: {
             call: "TrustedCall",
-            nonce: "Index",
+            nonce: "u32",
```
jonalvarezz: it should work with `Index` too. I mean, without changing anything else – did you find any issues? it'd be good to keep the type consistent with the other `nonce` in these definitions.

jonalvarezz (approved):

tee-worker/identity/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
silva-fj: Nope, there is no issue. That's why I changed it, to be consistent with the `TrustedCallSigned`:
```
TrustedCallSigned: {
    call: "TrustedCall",
    index: "u32",
    signature: "LitentryMultiSignature",
},
```
In the rust code the field `index` is called `nonce`

silva-fj: We can also change both to `Index`. What do you think?

Kailai-Wang: Historically it was called `index` in `frame_system::Config`, now it's `nonce`

Kailai-Wang (approved):
Great, thanks!
We are very close to switch-over, I guess only vc-issuance is left

tee-worker/identity/client-api/parachain-api/prepare-build/interfaces/trusted_operations/definitions.ts
jonalvarezz: Right, I was confused by the enclave nonce's getter and `system.accountNextIndex` – both return `Index`.

> We can also change both to Index. What do you think?

Sounds good to me

jonalvarezz:
Here is the client part: https://github.com/litentry/litentry-parachain/pull/3174

I used this branch as base. We may want to merge it first and then this one so it goes in the same commit.

I tested locally and it looks good

PR #3176, opened 2024-11-13
Enabling email authentication for TrustedCallAuthenticated

silva-fj:
as topic

Published packages
client-sdk:https://www.npmjs.com/package/@litentry/client-sdk/v/1.0.0-next.3

### DIFF
--- a/tee-worker/identity/client-sdk/packages/client-sdk/src/lib/type-creators/request.ts
+++ b/tee-worker/identity/client-sdk/packages/client-sdk/src/lib/type-creators/request.ts
@@ -55,7 +55,7 @@ export async function createRequestType(
       call,
       nonce,
       authentication: api.createType('TCAuthentication', {
-        Web3: signature,
+        [signer.isEmail ? 'Email' : 'Web3']: signature,
       }),
     });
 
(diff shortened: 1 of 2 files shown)


tee-worker/identity/client-sdk/packages/client-sdk/src/lib/type-creators/request.ts
```diff
@@ -55,7 +55,7 @@ export async function createRequestType(
       call,
       nonce,
       authentication: api.createType('TCAuthentication', {
-        Web3: signature,
+        [signer.isEmail ? 'Email' : 'Web3']: signature,
```
silva-fj, line 58: I'm not changing the signature name as it would be a breaking change, for now it would work passing the email verification code as it is also a string

0xverin: Would it be better to use an enum here, take a parameter, and map out the type? What if more types are added in the future?

silva-fj: Yes. Well, the proper fix would be change the `createRequestType` signature to properly support `TrustedCallAuthenticated`, but it would be a breaking change and after the demo some types will probably change so we need to refactor this again anyways.

0xverin (approved):

PR #3181, opened 2024-11-19
P 1139 omniexecutor ci and deployment for dev

BillyWooo:
- create litentry/omni-executor docker image
- add skeleton for CI test
- update Makefile, move executable and manifest with same folder level, as well as signature file.
- use Alice as default sign account. Will update this later with predefined account; or randomly generated account for testing purpose.


.github/workflows/ci.yml
```diff
@@ -285,7 +293,7 @@ jobs:
         working-directory: ./tee-worker
         shell: bash
         run: |
-          for d in . identity/enclave-runtime bitacross/enclave-runtime; do
+          for d in . identity identity/enclave-runtime; do
```
Kailai-Wang: Hmm why this change?

.github/workflows/ci.yml
```diff
@@ -586,26 +663,62 @@ jobs:
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          # use the docker driver to access the local image
+          # we don't need external caches or multi platforms here
+          # see https://docs.docker.com/build/drivers/
+          driver: docker
+
+      - name: Build local builder
```
Kailai-Wang: Maybe we don't need this intermediate image while still having two stages in Dockerfile

It's built for other two workers for cache purpose

.github/workflows/ci.yml
```diff
@@ -943,6 +1056,74 @@ jobs:
...
+        run: |
+          make launch-network-litentry
+
+      - name: Integration omni executor test ${{ matrix.test_name }}
+        working-directory: ./tee-worker/omni-executor/docker
+        if: needs.set-condition.outputs.run_omni_executor_test == 'true'
+        timeout-minutes: 40
+        run: |
+          docker compose -f docker-compose.yml
+        # docker compose -f docker-compose.yml -f ${{ matrix.test_name }}.yml up --no-build --exit-code-from ${{ matrix.test_name }} ${{ matrix.test_name }}
```
Kailai-Wang: We can remove commented lines

tee-worker/omni-executor/parentchain/listener/src/key_store.rs
```diff
@@ -34,9 +34,16 @@ impl SubstrateKeyStore {
...
+		// Secret seed:       0xe5be9a5092b81bca64be81d212e7f2f9eba183bb7a90954f7b76361f6edb5c0a
+		// Public key (hex):  0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d
+		// Account ID:        0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d
+		// Public key (SS58): 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
+		// SS58 Address:      5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
 		Ok([
-			45, 219, 105, 155, 49, 74, 164, 131, 153, 192, 15, 213, 225, 179, 167, 129, 12, 160,
-			229, 37, 133, 168, 141, 233, 98, 117, 254, 112, 139, 210, 76, 6,
+			229, 190, 154, 80, 146, 184, 27, 202, 100, 190, 129, 210, 18, 231, 242, 249, 235, 161,
+			131, 187, 122, 144, 149, 79, 123, 118, 54, 31, 110, 219, 92, 10,
```
Kailai-Wang, line 46: It's better to use explict expression than magic numbers

Kailai-Wang (approved):
In general it looks good

.github/workflows/ci.yml
BillyWooo: previously `tee-check` checks `enclave-runtime` from both `identity` and `bitacross`. But didn't check other files within `identity` and `bitacross`. 
I changed it to two jobs, each of them check files both within `enclave-runtime` and outside.

.github/workflows/ci.yml
BillyWooo: I think we still need two stages. Actually I tried to merge them into one step. But the `builder` stage uses `rust:1.82-bookworm`, which is quite big. The final image pushed is also very big. That's the reason I revert back to two stages.

.github/workflows/ci.yml
BillyWooo: 😂  These lines I left there on purpose. The above line is just placeholder. So it can easily to extend in the future.

tee-worker/omni-executor/parentchain/listener/src/key_store.rs
BillyWooo: Yeah. I will improve it.

.github/workflows/ci.yml
Kailai-Wang: I think it does.

`for d in . identity/enclave-runtime bitacross/enclave-runtime; do`

does the workspace-wide check as they belong to the same workspace

.github/workflows/ci.yml
Kailai-Wang: We still have two stages in Dockerfile, but we don't need to build a `local-builder` first IMO

.github/workflows/ci.yml
BillyWooo: Yes. You are right they are in the same workspace. I reverted the changes.

.github/workflows/ci.yml
BillyWooo: I have removed the `local-builder`.

BillyWooo:
tested on tee-dev.

PR #3183, opened 2024-11-22
Adding native `request_vc` and `request_batch_vc` trusted call handlers 

silva-fj:
As topic this PR adds native task handlers for `request_vc` and `request_batch_vc`. Also the result types for the native tasks have been refactored, new integration tests have been added and a new call have been added to `vc-management` pallet (`on_vc_issued`)

Updated and released client packages:
- `parachain-api`: https://www.npmjs.com/package/@litentry/parachain-api/v/0.9.20-next.9
- `client-sdk`: https://www.npmjs.com/package/@litentry/client-sdk/v/1.0.0-next.6


silva-fj:
I have updated the client packages: `parachain-api` and `client-sdk`

parachain/pallets/vc-management/src/lib.rs
```diff
@@ -161,6 +161,12 @@ pub mod pallet {
 			detail: ErrorDetail,
 			req_ext_hash: H256,
 		},
+		VCHasBeenIssued {
```
kziemianek: `VCIssued` and `VCHasBeenIssued` is very confusing. Maybe one event definition but with `OmniAccountOrIdGraphHash` ?

kziemianek (approved):

parachain/pallets/vc-management/src/lib.rs
silva-fj: This definition is temporary, after we migrate the IDGraphs and sunset the old code we should have a single `VCIssued` event

Kailai-Wang: What's the consequence of directly changing it? I think we can make it work by firstly runtime-upgrade the chain, then the worker

silva-fj: The only difference is that the current call and event expect `id_graph_hash` and this one `AccountId` (OmniAccount), we can directly update the current call and event but we also need to update the worker (old code) not to call `vc_issued`. The idea was that we can run both versions at the same time without modifying the current functionality

Kailai-Wang: I see, then maybe we can make it simple and call it `VcIssued2` or `VcIssuedNew` - I assume eventually this will replace the old call name and we are going to have only 1.

Also, we only need an OmniAccount identitifer and assertion, right? It seems we don't need the `Identity` and `reqExtHash`

silva-fj: I guess `Identity` is probably not needed, I was just replicating what we currently have (in the tests we make assertions on the identity that requested the vc). `reqExtHash` is needed for the client to identify the event. 

I'll update the name of the event if you guys prefer it, the idea was to called as the original after the migration

PR #3189, opened 2024-12-04
Adding support for oauth2 trusted call authentication

silva-fj:
As topic, this PR adds support for OAuth2 authentication.

Note that for this kind of authentication the client is required to get a new authorization code from google for every request. To solve this issue, a new type of authentication will be introduced in [this issue](https://linear.app/litentry/issue/P-1202/create-a-new-trustedcall-to-request-authorisation-token)


tee-worker/identity/litentry/core/native-task/receiver/src/authentication_utils.rs
```diff
@@ -0,0 +1,106 @@
...
+	payload: &OAuth2Data,
+) -> bool {
+	let mut google_client = GoogleOAuth2Client::new(
+		data_providers_config.google_client_id.clone(),
+		data_providers_config.google_client_secret.clone(),
+	);
+	let code = payload.code.clone();
+	let redirect_uri = payload.redirect_uri.clone();
+	let Ok(token) = google_client.exchange_code_for_token(code, redirect_uri) else {
+		return false;
```
kziemianek: Would be good to log something usefull in case of error, what do you think?

kziemianek (approved):

PR #3190, opened 2024-12-04
Upgrade to polkadot-sdk v1.11.0

Kailai-Wang:
### Context

As topic.

### Some notes:

- xcm needs an overhaul, we are way behind (it's v4 now), plus we don't use it actively, so I turn most of the settings "off", as well as integration tests - @wangminqi please double-check the config to make sure there's no token transfer via XCM possible
- rococo runtime is removed
- orml is dismissed, it seems the [forked branch](https://github.com/litentry/open-runtime-module-library/tree/litentry-polkadot-v1.11.0) would still introduce compile errors. The original orml uses crates.io based deps. The `OrderedSet` is copied over manually @wangminqi 
- `sp_core_hashing` introduces std feature unintentionally, thus superseded by `sp_core::hashing`
- the node code is based on cumulus template code, some more RPC methods are introduced
- rust toolchain is upgraded to 1.81
- the finalisation delay is fixed
- feature propagation is fixed using [zepter](https://github.com/ggwpez/zepter)

Due to re-org, sometimes (especially in the dev) you can see tx status `Retracted`, but eventually it will be finalised. To deal with that in the enclave registration, I've forked https://github.com/Kailai-Wang/substrate-api-client/tree/polkadot-v0.9.42-litentry to allow `Retracted` status - base repo has something similar too: https://github.com/scs/substrate-api-client/pull/807

As a (partial) consequence, I need to wait for `Finalized` status (instead of `InBlock`) in the account funding in the worker initialisation.

See also: https://substrate.stackexchange.com/questions/8872/when-is-a-transaction-retracted/8873#8873

### Main files to review:
- files under `parachain/node/`
- litentry/paseo runtime configurations
- worker changes


parachain/runtime/litentry/src/xcm_config.rs
```diff
@@ -82,152 +68,103 @@ pub type LocationToAccountId = (
...
-	type Weigher = XcmWeigher;
-	// Litentry: This is the tool used for calculating that inside XcmExecutor vm, how to transfer
-	// asset into weight fee. Usually this is in order to fulfull Barrier
-	// AllowTopLevelPaidExecutionFrom requirement. Currently we have not implement the asset to fee
-	// rule for Foreign Asset, so pure cross chain transfer from XCM parachain will be rejected no
-	// matter.
-	type Trader = Traders;
+	type Barrier = Barrier;
+	type Weigher = FixedWeightBounds<UnitWeightCost, RuntimeCall, MaxInstructions>;
+	type Trader = UsingComponents<IdentityFee<Balance>, SelfLocation, AccountId, Balances, ()>;
```
wangminqi, line 167: This comment is just for recording purpose.

So this impl will make us support native token only for now.

Kailai-Wang: Yes this is the goal (for now)

parachain/node/src/chain_specs/litentry.rs
```diff
@@ -55,50 +54,41 @@ struct GenesisInfo {
...
-			)
-		},
-		Vec::new(),
-		None,
-		Some("litentry"),
-		None,
-		default_parachain_properties(),
+	ChainSpec::builder(
+		WASM_BINARY.expect("WASM binary was not built, please build it!"),
 		Extensions { relay_chain: "rococo-local".into(), para_id: LITENTRY_PARA_ID },
```
BillyWooo, line 59: Within this file, there are still three `rococo`. Should we also change it? I think it's just left overs.

wangminqi:
XCM part LGTM

parachain/node/src/chain_specs/litentry.rs
Kailai-Wang: No it's good

BillyWooo (approved):
LGTM. Thanks.

Kailai-Wang:
There seems to be some race conditions regarding parachain ext submission - it can be related to the missing parachain events that we found earlier.

I'm trying to add more logs to debug it, **so don't merge it now**

kziemianek (approved):

silva-fj (approved):

PR #3191, opened 2024-12-05
Adding clean scripts

0xverin:
Minor optimization:

Sometimes when we clear the cache of ts-tests or client-api, we need to switch directories back and forth to delete node_modules. Adding a script can simplify the process.


Kailai-Wang (commented):
Thanks - can you also write in README (or comment in code) when should this be done?

Kailai-Wang (approved):

PR #3192, opened 2024-12-06
Adding support for auth token authentication for trusted calls

silva-fj:
As topic, this PR adds a new trusted call to request auth token that can be used with `TCAuthentication`.
This new authentication mechanism will allow users to make multiple requests (intents) without having to re-authenticate. The token have a block number expiration (This is the initial version, it is flexible enough to extend it to support other ways of expiration as well as add control on what can be done with the token).


BillyWooo (approved):
LGTM to me. If we can provider some ci test case, that would be even better.

Kailai-Wang (approved):
Looks good to me!

silva-fj:
> LGTM to me. If we can provider some ci test case, that would be even better.

I'll create an issue to add some integration tests 👍🏼

PR #3195, opened 2024-12-10
fix: P-1203 unified npm package version

wli-pro:
### Context

Fixing conflicting npm package.

### Labels

### Testing Evidences


Kailai-Wang:
@0xverin please review it : )

0xverin (commented):
1. After deleting `pnpm-lock.yaml` and `node_modules`, there are still many unmet peers when running pnpm install. I think the lock file should be deleted(at least at this update) and then update the package for testing, otherwise, it's impossible to know the correct version range of the package, as the locked version is used during pnpm install.
image


2. ts-tests still have unmet peers (tested without deleting pnpm-lock.yaml).
image

3. Can you find a way to run all the tests in CI (like adding a console.log in tests to triggle ci or something), the CI passed, but actually none of the tests were run?I am worried that with version modifications, some methods may become ineffective. thank you.

wli-pro:
> 1. After deleting `pnpm-lock.yaml` and `node_modules`, there are still many unmet peers when running pnpm install. I think the lock file should be deleted(at least at this update) and then update the package for testing, otherwise, it's impossible to know the correct version range of the package, as the locked version is used during pnpm install.
> 
> image
> 2. ts-tests still have unmet peers (tested without deleting pnpm-lock.yaml).
> 
> image
> 3. Can you find a way to run all the tests in CI (like adding a console.log in tests to triggle ci or something), the CI passed, but actually none of the tests were run?I am worried that with version modifications, some methods may become ineffective. thank you.

fixed now.

0xverin:
> > 1. After deleting `pnpm-lock.yaml` and `node_modules`, there are still many unmet peers when running pnpm install. I think the lock file should be deleted(at least at this update) and then update the package for testing, otherwise, it's impossible to know the correct version range of the package, as the locked version is used during pnpm install.
> > 
> > image
> > 2. ts-tests still have unmet peers (tested without deleting pnpm-lock.yaml).
> > image
> > 3. Can you find a way to run all the tests in CI (like adding a console.log in tests to triggle ci or something), the CI passed, but actually none of the tests were run?I am worried that with version modifications, some methods may become ineffective. thank you.
> 
> fixed now.

I also fixed the issues after the solc upgrade In parachain ts-tests.

0xverin (approved):

PR #3203, opened 2024-12-13
Remove `bitacross-worker`

kziemianek:


common/primitives/core/src/teebag/types.rs
```diff
@@ -70,8 +70,9 @@ pub enum AttestationType {
 #[derive(Encode, Decode, Clone, Copy, Default, PartialEq, Eq, RuntimeDebug, TypeInfo)]
 pub enum WorkerType {
     #[default]
+    #[codec(index = 0)]
     Identity,
-    BitAcross,
+    #[codec(index = 2)]
```
Kailai-Wang: Shall we use `index = 1` to avoid any surprises? Or will it affect any existing storages? At least not on prod I assume

Kailai-Wang (approved):
Thanks!

Kailai-Wang:
We'd need to resolve conflicts too

PR #3204, opened 2024-12-13
Implement access control for AccountStore members

silva-fj:
As topic, this PR introduces an access control mechanism for the OmniAccount (AccountStore) members.


parachain/pallets/omni-account/src/lib.rs
```diff
@@ -396,6 +482,55 @@ pub mod pallet {
...
+					// If member has default permission, they can set permissions to any value
+					if member_permissions.contains(&T::Permission::default()) {
+						return Ok(());
+					}
+					// an account can only set permissions to the same or less permissions
+					if !new_permissions.iter().all(|p| member_permissions.contains(p)) {
+						return Err(Error::<T>::NoPermission);
+					}
+				},
+				_ => return Ok(()),
```
BillyWooo, line 557: Just wondering: is it better to return error here?

BillyWooo (approved):
In general, LGTM. Left one comment.

parachain/pallets/omni-account/src/lib.rs
silva-fj: Returning `Err` means the member account does not have permissions, but that's not the case if it reaches this point. 

The permissions for the call are checked at the beginning of the method, the extra checks for `add_account` and `set_permissions` are to make sure the caller cannot set permissions or add an account with more permissions than it (the caller) has. For all the other calls the permissions have already been checked.

BillyWooo: The reason I ask is: `ensure_permission` is only called by `add_account` and `set_permissions`.

silva-fj: `ensure_permissions` is called on `dispatch_as_signed` and `dispatch_as_omni_account` (in this case the call to dispatch could be `add_account` or `set_permissions` for example)

common/primitives/core/src/omni_account.rs
```diff
@@ -57,3 +57,13 @@ impl OmniAccountConverter for DefaultOmniAccountConverter {
...
 }
+
+// This type must be kept in sync with the `Permission` type used in the runtime.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub enum OmniAccountPermission {
+    All,
+    AccountManagement,
+    RequestNativeIntent,
+    RequestEthereumIntent,
+    RequestSolanaIntent,
```
Kailai-Wang: I would suggest that we detailise it:

for eth/sol intent, there're basically two categories: balance transfer and contract call. Can we set maximum amount to transfer, as well as whitelist contract address/call?

It can be defaulted as `u32::MAX` (or u128) and `any`

Kailai-Wang: If it's too much we could have another PR for it

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -396,6 +482,55 @@ pub mod pallet {
...
+
+			match call.is_sub_type() {
+				Some(Call::add_account { permissions: ref new_account_permissions, .. }) => {
+					// If member has default permission, they can add accounts with any permission
+					if member_permissions.contains(&T::Permission::default()) {
+						return Ok(());
+					}
+					match new_account_permissions {
+						Some(new_permissions) => {
+							// an account can only add another account with the same or less permissions
```
Kailai-Wang, line 533: Do you consider using `is_superset()`?

common/primitives/core/src/omni_account.rs
silva-fj: Yeah we could add any permissions we want, these are just examples. I assume this will be driven by the product, we can create an issue defining the permissions we are going to need

parachain/pallets/omni-account/src/lib.rs
silva-fj: I did but since `is_superset` expects a single entry and we need to compare it with multiple entries, we'll still have to keep part of the logic here to make sure we check all the permissions. (I wanted to support multiple permissions instead of just one to make it more flexible and easier to extend)

Kailai-Wang (approved):

Kailai-Wang:
Thanks, we are good to go once conflicts are resolved

PR #3205, opened 2024-12-13
Send parentchain extrinsics as mortal

Kailai-Wang:
### Context

Try to mitigate the invalid tx situation by specifying them as mortal - the mortality at least gives the tx pool a chance to consider a tx as invalid once expired (and thus prune it from tx pool).

We should rarely need immortal extrinsic - so the current impl is definitely flawed.

In my local tests, it looks better - I don't see missing `ParentchainBlockProcessed` events anymore.

It doesn't solve all the problems tho, some tx will still be marked as invalid in the first submission with no reason being specified.

Ideally the `validator_accessor` and `extrinsic_factory` could be merged (as creating parentchain extrinsic requires parentchain state), but for now no big change is made until we rewrite it.


silva-fj (approved):
LGTM!

PR #3206, opened 2024-12-14
feat: add conviction vote support

wangminqi:
This impl add conviction vote related logic to paseo.


parachain/runtime/paseo/src/governance_v2/tracks.rs
```diff
@@ -0,0 +1,125 @@
...
+use pallet_referenda::Curve;
+const TRACKS_DATA: [(u16, pallet_referenda::TrackInfo<Balance, BlockNumber>); 4] = [
+	(
+		0,
+		pallet_referenda::TrackInfo {
+			// Name of this track.
+			name: "root",
+			// A limit for the number of referenda on this track that can be being decided at once.
+			// For Root origin this should generally be just one.
+			max_deciding: 5,
```
silva-fj, line 38: Why 5 on the root track? Shouldn’t this be 1?

silva-fj (approved):
LGTM

parachain/runtime/paseo/src/governance_v2/tracks.rs
wangminqi: I just think in the past time we had multiple root operations simultaneously.
But you are correct, 1 will more fit DAO concept. Maybe we can rediscuss the parameter when impl it on mainnet.

PR #3209, opened 2024-12-17
Add back bitacross-worker for paseo

Kailai-Wang:
### Context

As topic - it's still required for some service.

### DIFF
--- a/common/primitives/core/src/teebag/types.rs
+++ b/common/primitives/core/src/teebag/types.rs
@@ -68,6 +68,7 @@ pub enum AttestationType {
 pub enum WorkerType {
     #[default]
     Identity,
+    BitAcross,
     OmniExecutor,
 }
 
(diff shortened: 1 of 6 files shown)


Kailai-Wang:
btw @kziemianek

I just noticed if we want to remove the bc-worker cleanly, we have to do proper migration, otherwise we have wrong type info and the old bc-workers appear under `OmniExecutor` type in teebag:
image

silva-fj (approved):

kziemianek:
We either explicitly assing indices as [here](https://github.com/litentry/litentry-parachain/pull/3203/commits/91778e5fdc5c201bfe201a8a4c9f1c4b321b3d15#diff-678b237d8a5227b3c6ea4aa540fcdfc13da5ffa6e0c93a435720cd3b1478513bL71) or write migration / do it by hand.

It happens becuase if indices are not explicitly set the `OmniExecutor` variant gets index 1 in new version which was previously assigned to bitacross - thus bitacross are displayed as omni executor workers

PR #3211, opened 2024-12-18
Use custom txpool to workaround the invalid tx

Kailai-Wang:
### Context

As topic.

A proper fix is using fork-aware txpool as done here: https://github.com/paritytech/polkadot-sdk/pull/4639

However we are using `stable2407`  branch which means direct cherry-picking will result in many incompatible issues. I also tried ported versions [from tanssi](https://github.com/moondance-labs/polkadot-sdk/tree/tanssi-polkadot-stable2407-tx-pool-fix-wip) and it didn't work so well.

The workaround is taken from the comment in https://github.com/paritytech/polkadot-sdk/issues/1202 (from Frequency) - in my tests it looked good without `--delayed-best-block`


BillyWooo (approved):
Thanks. LGTM.

silva-fj (approved):
So, the workaround is not to call `remove_invalid` from the wrapped pool? Are invalid transactions removed at some point?

silva-fj:
> So, the workaround is not to call `remove_invalid` from the wrapped pool? Are invalid transactions removed at some point?

nvm, I read the comments. They should be removed during revalidation

Kailai-Wang:
Yes the tx is not lost tho, it's forked, so the client should still get valid response back

PR #3214, opened 2024-12-19
add one more cleanup target

BillyWooo:
very tiny PR to detach zombienet to run in background.

### DIFF
--- a/parachain/scripts/launch-network.sh
+++ b/parachain/scripts/launch-network.sh
@@ -68,7 +68,7 @@ if [ -f "$PARACHAIN_BIN" ]; then
   echo "found one, version:"
   ./litentry-collator --version
 else
-  echo "not here, copying from docker image if we are on Linux ..." 
+  echo "not here, copying from docker image if we are on Linux ..."
   if [ $(uname -s) = "Linux" ]; then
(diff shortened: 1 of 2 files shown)


parachain/scripts/launch-network.sh
```diff
@@ -80,7 +80,7 @@ fi
 print_divider
 
 echo "launching zombienet network (in background), dir = $ZOMBIENET_DIR ..."
-$ZOMBIENET_BIN -d $ZOMBIENET_DIR -l silent spawn config.toml &
+nohup $ZOMBIENET_BIN -d $ZOMBIENET_DIR -l silent spawn config.toml > /dev/null 2>&1 &
```
Kailai-Wang, line 83: Hmm why do we need nohup? Isn't this process running in the background already?

BillyWooo: Without nohup, there is still dependency on that specific terminal, which logged in. 
It blocks the script to finish completely, even though it looks like everything inside script is already executed successfully.

Kailai-Wang: I see, so it's rather for jenkins instead of local setup?

BillyWooo: Yes. But it won't affect local setup. I have tested it.

Kailai-Wang (approved):

PR #3216, opened 2024-12-22
Optimize parachain events handling

kziemianek:
Iterate parachain's events only once and keep only events worker is interested in.

This should reduce time spent by event handler during parachain's event processing. It doesn't guarantee that sidechain block events will come out every 6 s.


Kailai-Wang (approved):
Good catch, thanks!

Some small improvements that we could make:
- we could define a collection of `impl StaticEvent` (to simplify the code) and match `any` of them in `FilterableEvents::new() - it should have better readability and maintainability
- we could clean-up the interested events a bit - e.g. we don't actually care about `ParentchainBlockProcessed` events - but including them in the interested events would actually push them into the post-processing including merkle root calculation

PR #3217, opened 2024-12-23
Handling AccountStoreUpdated events in the omni-executor

silva-fj:
This is the first of a series of PRs to migrate the IDHub worker to gramine. As discussed we will start migrating the main functionality to the omni-executor.

This PR contains the following:
- Preparation for handling other types of events in the omni-executor 
- Updated metadata scale file and generated rust interface from it
- Small improvements
- A storage layer have been added to save AccountStores


tee-worker/omni-executor/Makefile
```diff
@@ -81,4 +81,8 @@ stop-local:
 
 .PHONY: get-pallet-metadata
 get-pallet-metadata:
-	subxt metadata --url http://localhost:9944 --allow-insecure --pallets OmniAccount,Teebag > parentchain/artifacts/rococo-omni-account.scale
+	subxt metadata --url http://localhost:9944 --allow-insecure --pallets OmniAccount,Teebag > parentchain/api-interface/artifacts/rococo-omni-account.scale
```
Kailai-Wang: We don't have rococo anymore, it's either Litentry or Paseo (depending on how you launch the parachain)

Maybe it's just an intermediate file, but we hope to be accurate when possible : )

tee-worker/omni-executor/parentchain/api-interface/README.md
```diff
@@ -0,0 +1,43 @@
+# Parentchain API Interface
+
+This crate contains auto-generated Substrate runtime API definitions for the OmniAccount and Teebag pallets.
+
+## Overview
+
+The crate exports the generated interface from a scale encoded metadata file. The interface is used to interact with the parentchain's runtime API.
+
+## Generation Process
```
Kailai-Wang: Could you write a bit **when** it should be re-generated/updated?

Also does it imply a potential inconsistency if e.g. the interface is generated against local chain but live chain has an older interface so worker won't work along?

tee-worker/omni-executor/parentchain/listener/src/lib.rs
```diff
@@ -137,22 +138,18 @@ where
 
 	perform_attestation(client_factory, signer, &transaction_signer).await?;
 
-	let intent_event_handler = IntentEventHandler::new(
+	let account_store_storage = Arc::new(AccountStoreStorage::new());
```
Kailai-Wang, line 141: It's initiated from the disk, which means we can't skip any parentchain blocks.

Shall we combine it with the download of AccountStore at the program start?

Kailai-Wang (approved):
Looks good from my side - I'm glad that we start to add storages to it

@kziemianek could you take a look too?

tee-worker/omni-executor/Makefile
silva-fj: True, I'll change it to use a generic name 👍🏼

tee-worker/omni-executor/parentchain/api-interface/README.md
silva-fj: Correct, we have to keep the version in mind. But this should not be an issue, we can re-generate the interface as part of our build process using the target chain endpoint on CI, it won't compile if the types are not compatible. 

I'll add some comments in the docs

tee-worker/omni-executor/parentchain/listener/src/lib.rs
silva-fj: Yes, that's the plan. We already have an [issue](https://linear.app/litentry/issue/P-1251/download-all-accountstores-if-they-dont-exists-when-starting-the-omni) for it

tee-worker/omni-executor/Makefile
```diff
@@ -79,6 +79,10 @@ start-local:
...
 	docker compose -f $(OMNI_DIR)/docker/docker-compose.yml down
 
-.PHONY: get-pallet-metadata
-get-pallet-metadata:
-	subxt metadata --url http://localhost:9944 --allow-insecure --pallets OmniAccount,Teebag > parentchain/artifacts/rococo-omni-account.scale
+.PHONY: get-metadata
+get-metadata:
+	subxt metadata --url http://localhost:9944 --allow-insecure --pallets OmniAccount,Teebag > parentchain/api-interface/artifacts/metadata.scale
+
+.PHONY: generate-api-interface
```
BillyWooo, line 86: should this target depend on `get-metadata` ? (always get updated `metadata.scale` ?)

tee-worker/omni-executor/parentchain/api-interface/README.md
```diff
@@ -0,0 +1,48 @@
...
+This crate contains auto-generated Substrate runtime API definitions for the OmniAccount and Teebag pallets.
+
+## Overview
+
+The crate exports the generated interface from a scale encoded metadata file. The interface is used to interact with the parentchain's runtime API.
+
+#### When to regenerate the interface?
+Whenever the types of the relevant pallets (OmniAccount, Teebag) change, the interface should be regenerated. This can be done by following the steps in the Generation Process section.
+
+Please note that the metadata fetched from the local node may not be the same as the metadata of the live chains. Therefore, when running the worker agains the live chains, the metadata should be fetched from the live chain and the interface should be regenerated.
```
BillyWooo: typo: `... running the worker agains` -> `against`.

tee-worker/omni-executor/parentchain/api-interface/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Generation Process
+
+The API interface can be generated using the following commands:
```
BillyWooo: would it be better to merge these two steps? `make generate-api-interface` can depend on `make get-metadata`.

silva-fj: The idea was that eventually we could have commands to get metadata for different chains, e.g: `local` or `dev`. But we can merge them for now and change it later if we need to

tee-worker/omni-executor/Makefile
silva-fj: Yep 👍🏼

PR #3220, opened 2024-12-26
Storage Version Migration

wangminqi:
P-1245
Our Litentry, Paseo are Storage version outdated. It will block tools/CI like try-runtime.


parachain/runtime/litentry/src/migration/mod.rs
```diff
@@ -1 +1,362 @@
...
+// Litentry is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+// By the time of this migration on Litentry 9220 (polkadot stable2407)
+// The current        storage version:      pallet version:
```
Kailai-Wang, line 18: Hmm I think we only have `StorageVersion` now, no `PalletVersion`?

As I saw there was once a migration from `PalletVersion` to the new `StorageVersion`: https://paritytech.github.io/polkadot-sdk/master/src/frame_support/migrations.rs.html#255-261

parachain/runtime/litentry/src/migration/mod.rs
```diff
@@ -1 +1,362 @@
...
+	// Does not matter if we do have old storage or not
+	// Should simply works
+	pallet_multisig::migrations::v1::MigrateToV1<Runtime>,
+	// Preimage V0 => V1
+	// Migrate old StatusFor and PreimageFor Storage into new Storage
+	pallet_preimage::migration::v1::Migration<Runtime>,
+	// Balances V0 => V1
+	// The official pallet migration is not need since we do not have any XCM deactive accounts
+	// But our onchain inactiveIssuance storage of pallet_balance is non-negative
+	// TODO: Where does this number come from?
```
Kailai-Wang, line 81: I believe it's just calculated from total_issuance - active_issuance.

I googled a bit:

How Does It Emerge?
inactive issuance can accumulate in the following scenarios:

Locks Applied by Other Pallets: Pallets like pallet-staking or pallet-democracy can introduce locks on tokens, marking them as inactive until certain conditions are met (e.g., unlocking after a referendum).
Misconfiguration or Bugs: An incorrect runtime implementation or upgrade could inadvertently increase inactive issuance.
Custom Pallets or Extensions: Developers may implement additional features that lock balances in ways not accounted for during runtime upgrades or storage migrations.
When tokens are locked via these mechanisms, they are no longer counted as "active issuance" but still exist in the chain's total issuance.

parachain/runtime/litentry/src/migration/mod.rs
```diff
@@ -1 +1,362 @@
...
+		let on_chain_version = pallet_balances::Pallet::<T>::on_chain_storage_version();
+
+		if on_chain_version == 0 {
+			// Remove the old `StorageVersion` type.
+			frame_support::storage::unhashed::kill(&frame_support::storage::storage_prefix(
+				pallet_balances::Pallet::<T>::name().as_bytes(),
+				"StorageVersion".as_bytes(),
+			));
+
+			InactiveIssuance::<T>::kill();
```
Kailai-Wang: IMO we should not touch it - just upgrading the storage version is enough

parachain/runtime/litentry/src/migration/mod.rs
```diff
@@ -1 +1,362 @@
...
+	// Preimage V0 => V1
+	// Migrate old StatusFor and PreimageFor Storage into new Storage
+	pallet_preimage::migration::v1::Migration<Runtime>,
+	// Balances V0 => V1
+	// The official pallet migration is not need since we do not have any XCM deactive accounts
+	// But our onchain inactiveIssuance storage of pallet_balance is non-negative
+	// TODO: Where does this number come from?
+	BalancesUpdateStorageVersionResetInactive<Runtime>,
+	// Democracy V0 => V1
+	// This migration only effects onging proposal/referedum, NextExternal
```
Kailai-Wang, line 84: Does it mean it's better to wait until there's no on-going ref/proposals to do this migration?

parachain/runtime/litentry/src/migration/mod.rs
```diff
@@ -1 +1,362 @@
...
+	// V3 to V4
+	// XCMP QueueConfig has different default value
+	// Migration targeting at changing storage value to new default value if old value matched
+	// Our current Paseo setting has already hard-coded
+	// This migration should have no effect except bumping storage version
+	cumulus_pallet_xcmp_queue::migration::v4::MigrationToV4<Runtime>,
+	// V4 to V5
+	// Did nothing to storage
+	// Just checking MaxActiveOutboundChannels is not exceeded
+	// Our current Paseo Storage is 0
```
Kailai-Wang: Do you mean Litentry?

parachain/runtime/litentry/src/migration/mod.rs
```diff
@@ -1 +1,362 @@
...
+	// The referedum info's proposal hash is migrated if the hash is in old form (In our case, even for an onging one it will do nothing)
+	pallet_democracy::migrations::v1::v1::Migration<Runtime>,
+	// Bounties V0 => V4
+	// The official migration does nothing but change pallet name and bump version
+	// So we just bump version storage instead
+	BountiesUpdateStorageVersion<Runtime>,
+	// V3 to V4
+	// XCMP QueueConfig has different default value
+	// Migration targeting at changing storage value to new default value if old value matched
+	// Our current Paseo setting has already hard-coded
```
Kailai-Wang: Do you mean Litentry?

Kailai-Wang (commented):
Thank you!

Just to double-check: we'll need to apply this migration first, then try-runtime cli will work?

parachain/runtime/litentry/src/migration/mod.rs
wangminqi: Yes, apply this migration first then try-runtime feature.

pallet version is indeed storage version. And truth is we are missing/wrong on both.

There are few pallets (weird) using a special "storageVersion" StorageMap holding the info. But palletVersion query onchain is not bonding to this storage at all. (see my comment on transactionPayment, vesting pallets)

The PalletVersion migration you providing is mostly likely not going to work for us. It just clean the Pallet version storage (if any) and assign the latest storage version without doing any specific migration.

parachain/runtime/litentry/src/migration/mod.rs
wangminqi: Okay, I will remove this part of code for security purpose.

parachain/runtime/litentry/src/migration/mod.rs
wangminqi: Technically it should be the same whether or not there is any onging ones.

see extra comment
// The referedum info's proposal hash is migrated if the hash is in old form (In our case, even for an onging one it will do nothing)

new ongoing ref/proposal should already have new form, so it will do nothing.


But from a ghost story perspective, of course it will be better if no onging ones.

parachain/runtime/litentry/src/migration/mod.rs
wangminqi: Yes, updated.

parachain/runtime/litentry/src/migration/mod.rs
wangminqi: Yes, updated.

parachain/runtime/litentry/src/migration/mod.rs
Kailai-Wang: Can you remove it? Then we are good to go I guess

Kailai-Wang:
This comment is not yet resolved:
https://github.com/litentry/heima/pull/3220/files/a7230bb2d845ee2b2e97ec7e71d1bb8b4e7d62ab#r1905144095

Nor is this PR approved :(

PR #3223, opened 2025-01-07
Add pallet-omni-bridge

Kailai-Wang:
### Context

fixes P-1261

It's a rewriting/refactoring of the old bridge pallet for chainsafe, it tries to be as chainsafe-friendly as possible as we are still going to use chainsafe contracts on eth side (due to time pressure), main changes:
- use a single pallet and decouple it with pallet_assets
- allows bridging of both native token and local assets (XCM assets are not supported for now)
- clean up already finalised payouts automatically (*)

(*) please check the code logic - **theoretically** it might happen that the cleanup takes too much time (as it would go through too many loops), but in practice, there's very low chance that it would happen, it would imply the majority of relayers malfunctions.

Still, we could add some improvements later to bound the execution time.


parachain/pallets/omni-bridge/src/lib.rs
```diff
@@ -0,0 +1,710 @@
...
+
+const MODULE_ID: PalletId = PalletId(*b"hm/ombrg");
+const DEFAULT_RELAYER_THRESHOLD: u32 = 1;
+
+// TODO: maybe using xcm Location is better
+//       but we'd need enums for all foreign types, or use GeneralIndex
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+pub enum ChainType {
+	Heima,         // this chain
+	Ethereum(u32), // with chain id
```
kziemianek, line 51: If adding new variant here is the only change needed to be made in order to support new chain type I would suggest using something more generic

kziemianek (approved):

BillyWooo (approved):

parachain/pallets/omni-bridge/src/lib.rs
```diff
@@ -0,0 +1,710 @@
...
+}
+
+// We assume "chain + asset-id" can uniquely identify an asset that can
+// be bridged out/in
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+pub struct ChainAsset<AssetKind> {
+	pub chain: ChainType,
+	pub asset: AssetKind,
```
felixfaisal, line 59: I think we are currently using different pallets to register/create assets, it would be great to streamline creating assets for bridging via this bridge such that it's easier to create and manage assets accordingly.

felixfaisal (approved):

parachain/pallets/omni-bridge/src/lib.rs
Kailai-Wang: This might require another pallet - something like asset manager/registry.

I intentionally decouple it from this bridge pallet (initially I have something like `create_local_asset` and then removed it), this pallet handles bridging only but not creation/deletion of assets.

It only cares if an asset exists and can be mutated.

Kailai-Wang: So far we are good with `Balances` and `Assets` - I can imagine if we have XCM assets we might need a dedicated pallet to manage them all

PR #3226, opened 2025-01-09
Utility batch test and zero delay impl

wangminqi:
We will change unstaking delay to 0.

Tests will be added for Poc that utility.batch o(f unstake and execute extrinsic) is functional in one single tx.


wangminqi:
Since EVM user can not benefits from utility batch function, we have add conditional logic when delay = 0.
This will twist code logic a little bit, but benefits the frontend workload.

parachain/pallets/parachain-staking/Cargo.toml
```diff
@@ -35,6 +35,7 @@ substrate-fixed = { workspace = true }
 similar-asserts = { workspace = true }
 sp-core = { workspace = true, features = ["std"] }
 sp-io = { workspace = true, features = ["std"] }
+pallet-utility = { workspace = true, features = ["std"] }
```
Kailai-Wang: In fact, we don't need this utility, right?

parachain/pallets/parachain-staking/src/lib.rs
```diff
@@ -978,21 +984,13 @@ pub mod pallet {
...
 		pub fn schedule_leave_candidates(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
 			let collator = ensure_signed(origin)?;
-			let mut state = <CandidateInfo<T>>::get(&collator).ok_or(Error::<T>::CandidateDNE)?;
-			let (now, when) = state.schedule_leave::<T>()?;
-			let mut candidates = <CandidatePool<T>>::get();
-			if candidates.remove(&Bond::from_owner(collator.clone())) {
-				<CandidatePool<T>>::put(candidates);
+			let _ = Self::candidate_schedule_revoke(collator.clone())?;
+			if T::LeaveCandidatesDelay::get() == 0u32 {
+				Self::candidate_execute_schedule_revoke(collator)
```
Kailai-Wang, line 989: Can we call the old `Self::execute_leave_candidates()` directly? (so we don't need to move it into another fn)

For other fn too

The origin check will pass anyway (even if it wouldn't, we could dispatch extrisnic internally too without origin check)

Kailai-Wang (commented):
Thanks, the logic looks good

I'm just wondering why we created extra wrapper fn for some extrinsics, they seem redundant. Could you please double-check them? @wangminqi

parachain/pallets/parachain-staking/src/lib.rs
wangminqi: Extrinsic triggering an extrinsic may behave weird compared to an internal function.
Not sure if it is proper to do so.

parachain/pallets/parachain-staking/Cargo.toml
wangminqi: You are right. I will remove this.

Kailai-Wang (approved):

PR #3228, opened 2025-01-10
Initialize storage on omni executor startup

silva-fj:
As topic, when starting the omni-executor it downloads the AccountStore from the parentchain and saves the in the storage.

Some modules have been reorganized for reuse.


tee-worker/omni-executor/parentchain/primitives/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "parentchain-primitives"
+version = "0.1.0"
+authors = ['Trust Computing GmbH <info@litentry.com>']
+edition.workspace = true
+
+[dependencies]
+executor-core = { path = "../../executor-core" }
```
Kailai-Wang: Do you consider using workspace dep?

tee-worker/omni-executor/parentchain/storage/src/lib.rs
```diff
@@ -1,4 +1,51 @@
...
+		log::error!("Could not create client: {:?}", e);
+	})?;
+
+	init_account_store_storage(&client).await?;
+
+	Ok(())
+}
+
+async fn init_account_store_storage(client: &SubxtClient<CustomConfig>) -> Result<(), ()> {
+	let storage_query = parentchain_api_interface::storage().omni_account().account_store_iter();
```
Kailai-Wang: There's no concept of storage proof right? Like we don't have to verify anything that we downloaded

Kailai-Wang (approved):
Looks good, thanks

tee-worker/omni-executor/parentchain/primitives/Cargo.toml
silva-fj: We probably should. I was just trying to be consistent with what we currently have. I'll update the dependencies 👍🏼

tee-worker/omni-executor/parentchain/storage/src/lib.rs
silva-fj: Well, it is still available in the `legacy` modules. The rpc method to get the proofs is there and some types to verify the values. Do you think we need to do this? As long as we connect the worker to a node that we trust its ok

Kailai-Wang: Can wet try to understand the necessity to verify the proof? Maybe it needs a bit research

But in general, given that it's a one-time thing and it stores secret data, I do think it's plausible to verify it

silva-fj: Done 👍🏼

PR #3231, opened 2025-01-15
P 1164 auto fund sign account when omni executor is running in

BillyWooo:
instead of original idea to auto fund sign account, here use keystore to store generated account. 
some minor update about makefile, gitignore, etc.


Makefile
```diff
@@ -104,12 +104,15 @@ fmt: fmt-cargo fmt-taplo fmt-ts
 fmt-cargo:
 	@cd parachain && cargo fmt --all
 	@cd tee-worker && cargo fmt --all
+	@cd tee-worker/identity && cargo fmt --all
```
Kailai-Wang: We don't need this line I think

tee-worker/omni-executor/executor-worker/src/cli.rs
```diff
@@ -1,10 +1,49 @@
...
+#[command(propagate_version = true)]
 pub struct Cli {
+	#[command(subcommand)]
+	pub cmd: Commands,
+}
+
+#[derive(Subcommand)]
+pub enum Commands {
+	Run(RunArgs),
+	GenKey(GenKeyArgs),
```
Kailai-Wang, line 14: What does `GenKey` do exactly? It seems we don't need this command

Kailai-Wang (commented):
Thank you

maybe I don't quite get the goal of this PR - IIUC even with keystore, it's not guaranteed that the signer in the keystore is funded?

tee-worker/omni-executor/executor-worker/src/cli.rs
BillyWooo: It's used for the first time execution to generate a signer account. Then user can fund that account. Then start over.

BillyWooo:
> it's not guaranteed that the signer in the keystore is funded

At the development phase, an "Alice" account is provided by default. So it's by default we assume it has funds inside. At the same time, we get rid of "development" feature. There is no different versions of code for "development" and "production".

Makefile
BillyWooo: yes, you are right.

kziemianek (approved):
Looks fine, however it's not auto funding executor's account.

BillyWooo:
> Looks fine, however it's not auto funding executor's account.

Yeah, by default it's 'Alice'. It's supposed to be funded. In production mode, then it's some real account which should be funded with real asset. Can't make it auto here.

tee-worker/omni-executor/.gitignore
```diff
@@ -3,6 +3,8 @@ target/
 .idea/
 **/*.bin
 
+storage
```
silva-fj: Thanks, I forgot to add it. The path was renamed to `storage_db`
```suggestion
storage_db
```

PR #3233, opened 2025-01-16
Creating RPC API and base structure to handle native calls in the omni-executor

silva-fj:
This PR introduces the base structure to start migrating the core functionality of the ID worker (OmniAccount and VC), setting up the RPC api and the native task handler.

It also integrates some of the types and code from the legacy code. (via imports or copying) required to handle AesRequest and the interfaces for authentication.


tee-worker/omni-executor/rpc-server/src/methods/submit_native_aes_request.rs
```diff
@@ -0,0 +1,122 @@
...
+		})
+		.expect("Failed to register submitNativeRequest method");
+}
+
+fn get_native_call_from_aes_request<'a>(
+	request: &mut AesRequest,
+	ctx: Arc<RpcContext>,
+) -> Result<NativeCall, ErrorObject<'a>> {
+	#[cfg(feature = "gramine-quote")]
+	if request.mrenclave != ctx.mrenclave {
```
kziemianek: can we use `[0; 32]` for environment where real mrenclave is not present ?

silva-fj: Yes, it makes sense. This method does not need to know about gramine 👍🏼

common/primitives/core/Cargo.toml
```diff
@@ -17,7 +17,7 @@ serde = { version = "1.0", default-features = false, features = ["derive", "allo
 serde_json = { version = "=1.0.133", default-features = false }
 strum = { version = "0.26", default-features = false }
 strum_macros = { version = "0.26", default-features = false }
-webpki = { version = "=0.102.0-alpha.3", git = "https://github.com/rustls/webpki", rev = "da923ed", package = "rustls-webpki", default-features = false, features = ["alloc", "ring"] }
+webpki = { version = "=0.102.0-alpha.3", git = "https://github.com/rustls/webpki", rev = "da923ed", package = "rustls-webpki", default-features = false, features = ["alloc", "ring"], optional = true }
```
Kailai-Wang, line 20: Why this? Is it because worker needs teebag primitives but don't want to introduce webpki? Or any dep conflict with this crate? (I remember we need to pin this version for no-std support)

tee-worker/omni-executor/crypto/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "crypto"
```
Kailai-Wang: I think this crate and the `storage` crate below need a new name, otherwise it's too general

Common paradigm is using project as prefix, like `sp-`, `subxt-` etc

Shall we call it `hm-crypto` (heima-crypto) if it's a crate that can be used by repos, or `hoa-crypto` (heima-omni-account crypto) if it's omni-account specific?

tee-worker/omni-executor/parentchain/primitives/src/signature/bitcoin_signature.rs
Kailai-Wang, line 1: I think those files don't need to end with `_signature.rs` as they are under `signature/` already

tee-worker/omni-executor/rpc-server/src/methods/submit_aes_request.rs
```diff
@@ -0,0 +1,115 @@
...
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub struct AuthenticatedCall {
+	pub call: NativeCall,
+	pub nonce: Nonce,
+	pub authentication: Authentication,
+}
+
+pub fn register_submit_aes_request(module: &mut RpcModule<RpcContext>) {
+	module
+		.register_async_method("native_submitAesRequest", |params, ctx, _| async move {
```
Kailai-Wang, line 41: Would you add another `native_submitPlainRequest` which doesn't need to be encrypted with aes?

Also a question:

rpc methods start with `native_` stand for a nativeCall request, right? A `NativeCall` isn't equal to `AuthenticatedCall`?

Like we should differentiate between aes-encrypted `nativeCall`, and authenticated `nativeCall`, and even aes-encrypted authenticated `nativeCall`:
- aes means we want the payload to be encrypted/private
- authenticated means we want to verify the origin of the requester

I'm wondering if there're better structure/names to make these more readable

Kailai-Wang (approved):
The skeleton looks good, thanks!

common/primitives/core/Cargo.toml
silva-fj: Yes, there are some conflicts with this crate so I made it optional to be able to use the `core-primitives` in the worker

tee-worker/omni-executor/crypto/Cargo.toml
silva-fj: Adding a prefix sounds good. I was not happy with the names either 👍🏼

tee-worker/omni-executor/rpc-server/src/methods/submit_aes_request.rs
silva-fj: The `AuthenticatedCall` is the equivalent of `TrustedCallSigned` or `TrustedCallAuthenticated` in the identity worker and `NativeCall` is the equivalent of `TrustedCall`. 

How about if we call them `NativeCallAuthenticated` and `NativeCall`? 

Now that you mentioned this about the encrypted requests, why do we need them for direct calls? Would `native_submitNativeCallAuthenticated` be enough for these calls? The client would only need to send the `AuthenticatedNativeCall` encoded.

Also, here we are skipping another wrapper (`TrustedOperation`). If it's OK to remove the need for encrypting the requests for direct calls it would simplify even more the client side. What do you think @Kailai-Wang ?

kziemianek (approved):

silva-fj:
I'll wait for https://github.com/litentry/heima/pull/3235 and https://github.com/litentry/heima/pull/3238 reviews to get them merged here first to avoid merging conflicts.

PR #3235, opened 2025-01-22
Adding support for email authentication in the omni-executor

silva-fj:
This PR adds support for email authentication, includes some infrastructure changes as well as some refactoring. 

Note: This PR uses https://github.com/litentry/heima/pull/3233 branch as base as it is not merged yet


tee-worker/omni-executor/Cargo.toml
```diff
@@ -55,13 +56,14 @@ ethereum-intent-executor = { path = "ethereum/intent-executor" }
...
 executor-storage = { path = "executor-storage" }
+heima-identity-verification = { path = "heima/identity-verification" }
 heima-primitives = { package = "core-primitives", path = "../../common/primitives/core", default-features = false, features = ["std"] }
 native-task-handler = { path = "native-task-handler" }
 parentchain-api-interface = { path = "parentchain/api-interface" }
 parentchain-listener = { path = "parentchain/listener" }
-parentchain-primitives = { path = "parentchain/primitives" }
 parentchain-rpc-client = { path = "parentchain/rpc-client" }
 parentchain-storage = { path = "parentchain/storage" }
+primitives = { path = "primitives" }
```
Kailai-Wang, line 70: I prefer `oa_primitives` (omni-account-primitives) or `oe_primitives` (omni-executor-primitives)

tee-worker/omni-executor/executor-storage/src/account_store.rs
```diff
@@ -1,33 +1,29 @@
...
 use parentchain_api_interface::omni_account::storage::types::account_store::AccountStore;
 use parity_scale_codec::{Decode, Encode};
+use primitives::AccountId;
 use rocksdb::DB;
-use std::path::Path;
-use subxt_core::utils::AccountId32 as AccountId;
+use std::sync::Arc;
 
-const STORAGE_NAME: &str = "account_store_storage";
+const STORAGE_NAME: &[u8; 21] = b"account_store_storage";
```
Kailai-Wang, line 8: I'm thinking if this random length would make the db indexing a bit inefficient - imagine we have another storage type later, whose storage name might be of other length, hence the encoded storage key too

Shall we follow what parity does in the module/pallet id, and always use [u8; 8] for it? Then you'll need to use abbreviations like `b"oa/accst"`

tee-worker/omni-executor/executor-storage/src/lib.rs
```diff
@@ -1,31 +1,53 @@
...
+	Ok(db)
 }
 
-async fn init_account_store_storage(client: &mut SubxtClient<CustomConfig>) -> Result<(), ()> {
+async fn init_omni_account_storages(
+	client: &mut SubxtClient<CustomConfig>,
+	storage_db: Arc<StorageDB>,
+) -> Result<(), ()> {
+	let account_store_storage = AccountStoreStorage::new(storage_db.clone());
+	let member_omni_account_storage = MemberOmniAccountStorage::new(storage_db.clone());
```
Kailai-Wang, line 50: The `page_size` below - it should be either per-crate constant or a parameter in this init fn 🙌

tee-worker/omni-executor/executor-storage/src/account_store.rs
Kailai-Wang: Alternatively it's fine to have variant length, but shall we use the hash of the encoded value as the storage key? (Same reason as above) - maybe something like Blake128Concat or so

Kailai-Wang (approved):
Looks good in general, thanks!

tee-worker/omni-executor/executor-storage/src/account_store.rs
silva-fj: I'll crate an issue for this

tee-worker/omni-executor/Cargo.toml
silva-fj: I'll tackle this in a separate PR 👍🏼

PR #3237, opened 2025-01-24
P 1277  Avoid depending on moonbeam/astar deps

BillyWooo:
remove moonbeam/astar direct dependency; 
copy rpc related code and put it into repo `litentry/moonbeam-vendor-rpc`. 
Also check repo: https://github.com/litentry/moonbeam-vendor-rpc


parachain/scripts/clean-network.sh
```diff
@@ -19,15 +19,15 @@ esac
...
 ZOMBIENET_PID=$(pidof $ZOMBIENET_BIN)
 
-if [ -z $ZOMBIENET_PID ]; then
+if [ -z "$ZOMBIENET_PID" ]; then
   # the network might not be started with zombienet
   killall polkadot || true
   killall litentry-collator || true
 else
-  kill -2 $ZOMBIENET_PID
+  sudo kill -2 $ZOMBIENET_PID
```
Kailai-Wang: I think we don't need sudo here

Kailai-Wang (approved):
Thank you!

parachain/scripts/clean-network.sh
BillyWooo: wired. I was getting something like `./scripts/clean-network.sh: line 27: kill: (388645) - Operation not permitted`. 
But now it's disappeared. 
I will remove this `sudo`.

PR #3240, opened 2025-01-27
Change litentry chain spec with heima

wangminqi:
As discussing here, https://github.com/litentry/heima/pull/3236
This PR will only focus on chain spec and minor the effect.

Other rebanding effect will rely on p-1259


parachain/runtime/litentry/src/lib.rs
```diff
@@ -226,8 +226,8 @@ impl_opaque_keys! {
 #[sp_version::runtime_version]
 pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// It's important to match `litentry-parachain-runtime`, which is runtime pkg name
-	spec_name: create_runtime_str!("litentry-parachain"),
-	impl_name: create_runtime_str!("litentry-parachain"),
+	spec_name: create_runtime_str!("heima-parachain"),
+	impl_name: create_runtime_str!("heima-parachain"),
```
Kailai-Wang: I don't think this is allowed to be changed

parachain/node/src/chain_specs/litentry.rs
```diff
@@ -108,8 +108,8 @@ pub fn get_chain_spec_staging() -> ChainSpec {
 pub fn get_chain_spec_prod() -> ChainSpec {
 	get_chain_spec_from_genesis_info(
 		include_bytes!("../../res/genesis_info/litentry.json"),
-		"Litentry",
-		"litentry",
+		"Heima",
+		"heima",
```
Kailai-Wang: This won't be compatible with generated JSON file, we have two options here:

1. we keep the `id` unchanged (still as `litentry-*`), minimal changes required - also for collators, but `is_litentry()` might be a bit confusing
2. we change the `id` to `heima-*`, then we'll need to rename the `is_litentry()` fn as well as change the database path

Kailai-Wang (commented):
Can you also make changes to paseo too?

Basically we want to apply the change on paseo first to play it safe

Kailai-Wang:
I committed my change directly to @wangminqi 's branch - now asking for a review from @litentry/parachain

silva-fj (approved):

BillyWooo (approved):

PR #3244, opened 2025-01-29
Adding support for google oauth2 authentication in the omni-executor

silva-fj:
This PR introduces support for google oauth2 authentication for native calls as well as some refactoring: Renamed the `primitives` crate to `executor-primitives` and updated all relevant references in the codebase.


tee-worker/omni-executor/heima/identity-verification/src/web2/google/mod.rs
```diff
@@ -0,0 +1,102 @@
+mod oauth2_client;
```
kziemianek: Can this be moved out of identity-verificaiton / heima context ? To me it looks like just general oauth

kziemianek (approved):

tee-worker/omni-executor/heima/identity-verification/src/web2/google/mod.rs
silva-fj: Sure, in the identity worker we have this in `litentry/core/data-providers` but since we are not sure if we are going to support the identity/vc related stuff in the omni-executor I didn't create that structure. I'll create an `oauth-providers` crate as we might add support for others

PR #3245, opened 2025-01-30
Adding native_submitPlainRequest RPC method

silva-fj:
This pull request introduces a new rpc method to submit plain (unencrypted) requests for native calls as well some small refactoring.


tee-worker/omni-executor/rpc-server/src/request.rs
```diff
@@ -22,6 +22,12 @@ use executor_primitives::ShardIdentifier;
 use parity_scale_codec::{Decode, Encode};
 use std::fmt::Debug;
 
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct PlainRequest {
+	pub shard: ShardIdentifier,
```
kziemianek: Why do we need this ?

silva-fj: oh, I forgot to check the mrenclave. thanks 👍🏼

kziemianek: I mean, why do we need `shard` at all :) 
Is there any sharding, because i'm a little bit confused - maybe this should be named differently?

silva-fj: ah yes. Well, initially it was discussed that we wanted to be as close as possible to the identity worker and reuse as much as possible the old code/logic so I've been trying not to change to many things. I guess we can just call it `mrenclave` and if we ever need to support sharding we can just refactor it

kziemianek: It's scale encoded so name doesn't matter for clients

silva-fj: true

tee-worker/omni-executor/rpc-server/src/methods/submit_plain_request.rs
```diff
@@ -0,0 +1,87 @@
...
+			let Ok(request) = PlainRequest::from_hex(&hex_request) else {
+				return Err(ErrorCode::ServerError(INVALID_PLAIN_REQUEST_CODE).into());
+			};
+			let join_handle = task::spawn_blocking({
+				let ctx = ctx.clone();
+				let aes_request = request.clone();
+				|| handle_plain_request(aes_request, ctx, Handle::current())
+			});
+			let (native_call, auth_type) = join_handle.await.map_err(|e| {
+				log::error!("Failed to handle AES request: {:?}", e);
```
kziemianek: AES -> plain ?

tee-worker/omni-executor/rpc-server/src/methods/submit_plain_request.rs
```diff
@@ -0,0 +1,87 @@
...
+		.register_async_method("native_submitPlainRequest", |params, ctx, _| async move {
+			let Ok(hex_request) = params.one::<String>() else {
+				return Err(ErrorCode::ParseError.into());
+			};
+			let Ok(request) = PlainRequest::from_hex(&hex_request) else {
+				return Err(ErrorCode::ServerError(INVALID_PLAIN_REQUEST_CODE).into());
+			};
+			let join_handle = task::spawn_blocking({
+				let ctx = ctx.clone();
+				let aes_request = request.clone();
```
kziemianek: `aes_request` -> `plain_request` ?

kziemianek (approved):
It would be good to have have tests (authentication for example).

silva-fj:
> It would be good to have have tests (authentication for example).

Yep, we'll do in a separate PR

PR #3247, opened 2025-02-02
feat: P-1222 added graph getter for OmniAccount in client-sdk

wli-pro:
### Context

As title.

The implementation is replicated from the heima test code: https://github.com/litentry/heima/blob/cdc652d6e52d4661a17f90c76619ec037c372edf/tee-worker/identity/ts-tests/integration-tests/common/utils/native-request-helpers.ts#L219-L225

### Labels

### Testing Evidences


tee-worker/identity/client-sdk/packages/client-sdk/src/lib/requests/get-omni-graph.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const accountStore = await api.query.omniAccount.accountStore(omniAccountAddress);
+
+    if (accountStore.isNone) {
+      return null;
+    }
+
+    return accountStore.unwrap() as MemberAccount[];
+  } catch (e) {
+    // swallow - empty omni graph
```
kziemianek, line 30: Is `accountStore.unwrap()` the only call that can throw an error ?
If not it might ignore it (for example query error) and return null so user will think that account doesn't exists

tee-worker/identity/client-sdk/packages/client-sdk/src/lib/requests/get-omni-graph.ts
```diff
@@ -0,0 +1,34 @@
...
+import { encodeAddress } from '@polkadot/util-crypto';
+
+/**
+ * Retrieve the omni account graph for a given identity.
+ * @param api - Polkadot API instance
+ * @param who - Litentry identity to query
+ * @returns Array of member accounts or null if not found
+ * @throws Error if API calls fail
+ */
+export async function getOmniGraph(
```
silva-fj, line 13: This getter should probably get the data from the worker (TrustedGetter) as private members are encrypted. But since we are rewriting the id worker in the omni-executor maybe it would be better to wait a bit for this.

wli-pro: https://github.com/litentry/heima/pull/3250

wli-pro:
closed this one and will have the same in onmi executor sdk instead.

PR #3250, opened 2025-02-05
Implementing OmniAccount related native calls in the omni-executor

silva-fj:
This PR add the OmniAccount related native calls. This is the initial migration of the `TrustedCall`s that are related to the OmniAccount, with only small modifications to adapt them to the omni-executor's code.

Currently only Web3 and Email accounts are supported on `add_account`. More web2 identities will be added later.
This PR also includes some small refactoring as well as some infrastructure code needed. 

The scale file (metadata) have been updated to include `System` and `Balances` pallet types.


wli-pro (approved):
LGTM

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -129,38 +144,237 @@ async fn handle_native_call<
...
+		},
+		NativeCall::add_account(
+			sender_identity,
+			identity,
+			validation_data,
+			public_account,
+			permissions,
+		) => {
+			let omni_account_storage = MemberOmniAccountStorage::new(ctx.storage_db.clone());
+			let Some(omni_account) = omni_account_storage.get(&sender_identity.hash()) else {
```
wli-pro, line 222: MemberOmniAccountStorage stores a mapping where the key is MemberAccount and the value is OmniAccount, meaning MemberOmniAccountStorage.get(MemberAccount) = OmniAccount.
For NativeCall::add_account:
The identity represents MemberAccount, so sender_identity should correspond to OmniAccount.
Should omni_account_storage.get(&sender_identity.hash()) be omni_account_storage.get(&identity.hash())?

silva-fj: The `identity` is the member account that the `sender_identity` wants to add, so it's not yet a member of the OmniAccount. You need an existing member to get the OmniAccount

PR #3251, opened 2025-02-06
Creating client-api type definitions for the omni executor

silva-fj:
As topic, this PR adds the initial type definitions for the omni-executor, so we can write ts-tests and/or start working on an sdk package for the omni-executor.

I've moved the `client-api` workspace one level up (outside the identity worker) to be easily found from the `omni-executor` project


tee-worker/client-api/parachain-api/prepare-build/interfaces/omniExecutor/definitions.ts
```diff
@@ -0,0 +1,96 @@
+export default {
```
0xverin, line 1: I am not familiar with this part. Will we add tests later, or do we already have tests for these definitions?

0xverin (approved):
Overall looks fine.  Thanks!

tee-worker/client-api/parachain-api/prepare-build/interfaces/omniExecutor/definitions.ts
silva-fj: We will use these definitions to write some tests in the omni-executor

PR #3252, opened 2025-02-07
Parametrize `sgx.debug`

kziemianek:
Parametrizes `sgx.debug` in Gramine manifest.
This should be set to `false` for prod deployments.
Changing this flag requires fresh deployment (for example enclave will be unable to unseal existing files).


Kailai-Wang (approved):
Thank you, currently only the omni-executor is affected, isn't it?

PR #3254, opened 2025-02-12
P 1275 omni bridge precompile

wangminqi:
omni bridge precompile

### DIFF
--- a/parachain/precompiles/omni-bridge/src/lib.rs
+++ b/parachain/precompiles/omni-bridge/src/lib.rs
@@ -24,3 +24,63 @@ use sp_runtime::traits::Dispatchable;
 
 use sp_core::{H256, U256};
 use sp_std::{marker::PhantomData, vec::Vec};
+
+pub struct OmniBridgePrecompile<Runtime>(PhantomData<Runtime>);
+
+type BridgeBalanceOf<Runtime> = <Runtime as pallet_omni_bridge::Config>::Balance;
+type BridgeAssetKind<Runtime> = <Runtime as pallet_omni_bridge::Config>::AssetKind;
+
+#[precompile_utils::precompile]
+impl<Runtime> BOmniBridgePrecompile<Runtime>
+where
+	Runtime: pallet_omni_bridge::Config + pallet_evm::Config,
+	Runtime::RuntimeCall: Dispatchable<PostInfo = PostDispatchInfo> + GetDispatchInfo,
+	Runtime::RuntimeCall: From<pallet_omni_bridge::Call<Runtime>>,
+	<Runtime::RuntimeCall as Dispatchable>::RuntimeOrigin: From<Option<Runtime::AccountId>>,
+	BridgeBalanceOf<Runtime>: TryFrom<U256> + Into<U256>,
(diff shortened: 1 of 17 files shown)


wangminqi:
only ts-test require small adjust. All other code should be ready fro review.

parachain/precompiles/omni-bridge/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+	Runtime::RuntimeCall: Dispatchable<PostInfo = PostDispatchInfo> + GetDispatchInfo,
+	Runtime::RuntimeCall: From<pallet_omni_bridge::Call<Runtime>>,
+	<Runtime::RuntimeCall as Dispatchable>::RuntimeOrigin: From<Option<Runtime::AccountId>>,
+	BridgeBalanceOf<Runtime>: TryFrom<U256> + Into<U256>,
+{
+	#[precompile::public("payIn(uint256,uint8,bool,uint256,bytes)")]
+	fn pay_in(
+		handle: &mut impl PrecompileHandle,
+		amount: U256,
+		dest_id: u8,
```
Kailai-Wang, line 51: We can live with it for now - is there a more general way of mapping the `ChainType` in pallet code?

Imagine we want to bridge to some non-evm chain later

parachain/precompiles/omni-bridge/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+		native: bool,
+		asset_id: U256,
```
Kailai-Wang, line 53: Actually the same problem:

we either have `native == true`, or non-native with `asset_id`, it's actually an enum

Kailai-Wang (approved):

Kailai-Wang:
@0xverin ideally you could take a look at the test too

parachain/precompiles/omni-bridge/src/lib.rs
wangminqi: in evm contract, there is no ethereum struct can 100% consistently support rust enum.
we have { native, id(u32) }, this enum in rust carry extra parameter u32.

parachain/precompiles/omni-bridge/src/lib.rs
wangminqi: Then we should define the type id for all different types of chain category, and hard code mapping it in precompile.

wangminqi: for evm interface, adding one extra parameter by that time.

parachain/precompiles/omni-bridge/src/lib.rs
wangminqi: dest_id share this problem too.
{ heima, Ethereum(chain_id) } carry extra chain_id info.

0xverin (approved):
Ts tests part looks good!

wangminqi:
Still confirming ts-test, let me merge manually later.

wangminqi:
Yes, it is caused by failed to decode event data. 

ready to merge. TODO fix in future.

PR #3255, opened 2025-02-12
Adding integration tests for the omni-executor

silva-fj:
As topic, this PR introduces the initial setup for integrations tests (Not yet configured on CI), this includes:
 
- Integration tests for the OmniAccount related calls
- Type definitions adjustments
- Bug fixes

To run the tests:
- launch the parachain as usual
- run the omni-executor worker: 
```bash
RUST_LOG=info cargo run --  run "ws://localhost:9944" "http://localhost:8545" "http://localhost:8546" "2010"
```
- Add funds to the signer account of the omni-executor
- run the integration tests from **tee-worker/omni-executor/ts-tests**: 
```bash
OMNI_WORKER_ENDPOINT=ws://localhost:2010 pnpm --filter integration-tests test omni_account.test.ts
```


tee-worker/omni-executor/ts-tests/integration-tests/omni_account.test.ts
```diff
@@ -0,0 +1,255 @@
+import { step } from 'mocha-steps';
```
0xverin, line 1: Thanks!The entire standard test steps is very clear. Can we add error test cases later, we can creat a new issue and raise a new pr for it? This is actually very important, as error tests can often catch some unexpected bugs.

tee-worker/omni-executor/ts-tests/integration-tests/utils/signer.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export interface Signer {
+    getAddressRaw(): Uint8Array;
+    sign(message: HexString | string | Uint8Array): Promise<Uint8Array>;
+    type(): KeypairType;
+    getAddressInSubstrateFormat(): Uint8Array;
+    getIdentity(api: ApiPromise): Promise<CorePrimitivesIdentity>;
+}
+
+export class SubstrateSigner implements Signer {
```
0xverin, line 26: This part should be the same as the ts-tests in identity, right?Temporarily copying it over like this is fine.But can we create an issue to move these common utils are in one folder/workspace?. Having too much duplicate code is actually not good for maintenance.

0xverin (commented):
Thanks!The ts-tests section looks generally good.
However, there is an problem that we have copied a lot of duplicate code from identity‘s ts-tests.
We should create an issue to address this.
As the duplicate code increases, it becomes inconvenient to maintain. Imagine if a fn has a bug, we might need to change the same fn across several files.

BillyWooo:
> Thanks!The ts-tests section looks generally good. However, there is an problem that we have copied a lot of duplicate code from identity‘s ts-tests. We should create an issue to address this. As the duplicate code increases, it becomes inconvenient to maintain. Imagine if a fn has a bug, we might need to change the same fn across several files.

Yeah, that's a good point. We can discuss offline. We need to take it into consideration that how identity-work develops in the future.

tee-worker/omni-executor/ts-tests/integration-tests/omni_account.test.ts
silva-fj: Yes, definitely. This is just the initial setup 👍🏼

tee-worker/omni-executor/ts-tests/integration-tests/utils/signer.ts
silva-fj: Correct. The assumption is that eventually we will replace the identity worker with the omni-executor so the old code won't have to be maintained. I didn't want to make premature abstractions before we know what will happen in the future, but yeah the general idea is to have a single code base to maintain so there will be some refactoring for sure

0xverin (approved):

PR #3257, opened 2025-02-13
Setting the mrenclave to the rpc server context in the omni-executor

silva-fj:
As topic.

The `dcap_types` have been extracted from `common/primitives/core/src/teebag/sgx_verify/mod.rs` so we can import them in the omni-executor (there are some conflicts with the `webpki` crate).


common/primitives/core/src/teebag/dcap_types.rs
```diff
@@ -0,0 +1,301 @@
+use crate::{MrEnclave, MrSigner, QuotingEnclave, SgxBuildMode};
```
kziemianek, line 17: This file is missing a license header.

tee-worker/omni-executor/omni-executor.manifest.template
```diff
@@ -42,6 +43,7 @@ sgx.trusted_files = [
 
 sgx.allowed_files = [
   "file:storage_db/",
+  "file:local/",
```
kziemianek: What's being stored here ?

silva-fj: `log` and `keystore`. I have updated the manifest to protect the `storage_db` and the `keystore`

kziemianek (approved):

PR #3259, opened 2025-02-16
Fix the parsing of precompile omnibridge event data

0xverin:
Follow up [precompile bridge tests](https://github.com/litentry/heima/blob/dev/parachain/ts-tests/integration-tests/precompile-contract.test.ts#L201)


wangminqi (approved):
awesome

wangminqi:
PS: This changes no runtime

PR #3263, opened 2025-02-20
Add vesting precompile

wangminqi:
Add vesting precompile


parachain/precompiles/vesting/VestingInterface.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: GPL-3.0-only
```
0xverin, line 1: Is it possible to add a test case in parachain ts-tests?

wangminqi: Sure. I will add one.

parachain/runtime/litentry/src/precompiles.rs
```diff
@@ -118,6 +119,12 @@ pub type PrecompilesSetAt<R> = (
 	PrecompileAt<AddressU64<1026>, ECRecoverPublicKey, (CallableByContract, CallableByPrecompile)>,
 	PrecompileAt<AddressU64<1027>, Ed25519Verify, (CallableByContract, CallableByPrecompile)>,
 	// Litentry precompiles (starts from 0x5000):
+	// ParachainStaking: pallet_parachain_staking = 15 + 20480
```
Kailai-Wang: Hmm, why not 11 + 20480 to make it consistent? 🤔

Kailai-Wang (commented):
CI still seems to fail

parachain/runtime/litentry/src/precompiles.rs
wangminqi: corrected

Kailai-Wang (approved):

PR #3269, opened 2025-02-26
Onmi sdk request auth token method

wli-pro:
### Context
As titile

### Labels

### Testing Evidences


wli-pro:
https://linear.app/litentry/issue/P-1338/implements-request-auth-token-in-omni-client-sdk

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/enclave.ts
```diff
@@ -0,0 +1,333 @@
+import WebSocket from 'isomorphic-ws';
+import { TypeRegistry } from '@polkadot/types';
+import { Index } from '@polkadot/types/interfaces';
+import { compactStripLength, hexToU8a, u8aToString } from '@polkadot/util';
+import { ApiPromise, identity, LitentryIdentity, omniExecutor } from '@litentry/parachain-api';
```
wli-pro: @litentry/parachain-api is a shared package used by identity worker too. Will deal with the rebranding naming sepererately.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/enclave.ts
```diff
@@ -0,0 +1,333 @@
...
+   */
+  getConnectionState(): ConnectionState {
+    return this.#currentState;
+  }
+
+  async getNonce(
+    /** Litentry Parachain API instance from Polkadot.js */
+    api: ApiPromise,
+    /** The user's omniAccount.  Use `createLitentryIdentityType` helper to create this struct */
+    omniAccount: LitentryIdentity,
```
silva-fj: Maybe using `LitentryIdentity` for the OmniAccount could be a bit confusing as you could think you can pass any member identity of the OmniAccount. How about if we expect the same types the `ApiPromise` methods expect for AccountId? ( `AccountId | AccountId32 | string | Uint8Array`)?

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/get-last-registered-enclave.ts
```diff
@@ -0,0 +1,32 @@
...
+  enclave: CorePrimitivesTeebagTypesEnclave;
+}> {
+  const identifiers = (await api.query.teebag.enclaveIdentifier(workerType)) as unknown as AccountId32[];
+  const latestEnclaveId = identifiers[identifiers.length - 1];
+
+  if (latestEnclaveId.isEmpty) {
+    throw new Error('[omni-sdk] No Enclave identifiers found');
+  }
+
+  const maybeEnclave = (await api.query.teebag.enclaveRegistry(latestEnclaveId)) as unknown as Option<any>;
```
silva-fj: We could get this with a single query like `await api.query.teebag.enclaveRegistry.entries()`, then we filter the entries by `workerType`.  (and maybe sort by `last_seen_timestamp` if we really need the last registered).

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/request_auth_token.request.ts
```diff
@@ -0,0 +1,109 @@
...
+      jsonrpc: '2.0',
+      method: 'native_submitAesRequest',
+      params: [request.toHex()],
+    };
+
+    const hexString = await enclave.send<string>(rpcRequest);
+
+    const [, data] = compactStripLength(hexToU8a(hexString));
+
+    const result = api.createType('NativeCallResponse', data) as unknown as NativeCallResponse;
```
silva-fj:
```suggestion
    const result: NativeCallResponse = api.createType('NativeCallResponse', data);
```

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/request_auth_token.request.ts
```diff
@@ -0,0 +1,109 @@
...
+      authentication: args.authentication,
+      call,
+      nonce,
+      shard: shardU8,
+    });
+
+    // send the request to the Enclave
+    const rpcRequest: JsonRpcRequest = {
+      jsonrpc: '2.0',
+      method: 'native_submitAesRequest',
```
silva-fj: I think it's fine to use `PlainRequest` for direct calls, the connection should be secure anyways (tls). The `native_submitAesRequest` was implemented because we were more or less replicating the identity worker but I think it's not needed. What do you think @Kailai-Wang ?

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/enclave.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  /**
+   * Retrieve the Enclave's Shard from the Parachain.
+   *
+   * The Enclave registry contains the information of the registered TEE workers. These TEE Workers share the
+   * same Enclave's Shard value.
+   *
+   * @see Test it by yourself https://polkadot.js.org/apps/?rpc=wss://tee-dev.litentry.io#/chainstate
+   */
+  async getShard(api: ApiPromise): Promise<`0x${string}`> {
```
silva-fj: I don't think we will have the concept of `shards` in the omni-executor. We could call this `getMrEnclave`

silva-fj (commented):
Overall it looks pretty good. I just left a couple of comments

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/enclave.ts
wli-pro: omni-client-api is added now so that we don't need touch the existing tee-worker/client-api that's used for identity worker for now.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/request_auth_token.request.ts
wli-pro: The createType returns a value of type Codec, if not using the type convert, the below ts error will show:
```
Type 'Codec' is missing the following properties from type 'NativeOperationResponse': isErr, asErr, isOk, asOk, and 11 more.
```

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/enclave.ts
wli-pro: See the MemberAccount defined in common/primitives/core/src/omni_account.rs using Identity as the public item type.
```
pub enum MemberAccount {
    Public(Identity),
    Private(Vec<u8>, Hash),
}
```
So we shouldn't change it to the type to AccountId | AccountId32 | string | Uint8Array, the Identity contains identity type that we will use to request, e.g. to determine if it is a web3 identity.
Now I change LitentryIdentity to Identity.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/get-last-registered-enclave.ts
wli-pro: resolved now.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/enclave.ts
wli-pro: resolved.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/enclave.ts
silva-fj: What I meant was that the OmniAccount is only an `AccountId` and it should not be confused with the `Identity` nor the `MemberIdentity` types. It's a regular native account of the parachain (That is controlled by one or more identities)

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/request_auth_token.request.ts
silva-fj: I think something is off, we do this in other places. e.g: https://github.com/litentry/heima/blob/a910008b2a676fe34a1e0a289db9102a5504443e/tee-worker/omni-executor/ts-tests/integration-tests/utils/type_creators.ts#L109

tee-worker/omni-client-api/parachain-api/prepare-build/interfaces/omniExecutor/definitions.ts
```diff
@@ -0,0 +1,152 @@
+export default {
+    types: {
+        OmniAesRequest: {
```
silva-fj: This was called like this to prevent naming conflicts with the legacy type but now we can call it `AesRequest`.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/request_auth_token.request.ts
Kailai-Wang: Does the executor start a TLS server? Or just non-TLS but we rely on the nginx reverse proxy? If the latter I suggest we still use `native_submitAesRequest`; if former then we can submit plain request

It depends on how gramine handles https probably

@silva-fj @wli-pro

tee-worker/omni-client-api/parachain-api/prepare-build/interfaces/omniExecutor/definitions.ts
wli-pro: resolved now.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/request_auth_token.request.ts
wli-pro: resolved now.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/enclave.ts
wli-pro: Ok, previously I misunderstood the omniAccount param, now I realized it should be the member account of the omniAccount.
So now I changed the param name from `omniAccount` to `member` as below, and now the type of the `member` is Identity
```
export async function requestAuthToken(
  api: ApiPromise,
  data: {
    member: Identity;
    expiresAt: number;
  },
)
```

Change here: https://github.com/litentry/heima/pull/3269/commits/edf1db3e3d600e3f12907cdef0a701e075588062#diff-1e2dfb90c0ac59677b6c2e567533474a09f59298939c879fea913b0eedf78245R36

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/request/request_auth_token.request.ts
silva-fj: Currently the server is not configured to enable TLS. We could revisit this in the future

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/request-auth-token.request.ts
```diff
@@ -0,0 +1,118 @@
...
+  },
+): Promise<{
+  payloadToSign?: string;
+  send: (args: { authentication: AuthenticationData }) => Promise<{
+    token: string;
+  }>;
+}> {
+  const { member, expiresAt } = data;
+
+  const [nonce, mrEnclave] = await Promise.all([getNonce(api, member), enclave.getMrEnclave(api)]);
```
silva-fj: According to the `getNonce` logic, this will fail if the user logs in with web2 identities which should be a common case, we should use the nonce of the OmniAccount.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/get-nonce.request.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Retrieves the nonce for a given identity.
+ * 
+ * @param api Polkadot.js API instance
+ * @param account The account to get the nonce for
+ * @returns Promise resolving to the nonce value
+ * @throws Error if api is not initialized or account is invalid
+ */
+export async function getNonce(
+  api: ApiPromise,
+  account: Identity | string | Uint8Array | AccountId,
```
silva-fj: The member identities are used to control the OmniAccount but transactions on-chain will not be made with them, so we only need to get the nonce of the OmniAccount. This is why I suggested to only allow `AccountId | AccountId32 | string | Uint8Array`

silva-fj (commented):
Thanks, it looks better now. Just one minor issue to fix and we are ready to get it merged 👍🏼

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/request-auth-token.request.ts
wli-pro: resolved now.

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/get-nonce.request.ts
wli-pro: resolved now

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/get-nonce.request.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Retrieves the omni account nonce for a given identity.
+ *
+ * @param api Polkadot.js API instance
+ * @param identity The identity to get the nonce for
+ * @returns Promise resolving to the nonce value
+ */
+export async function getOmniAccountNonceWithIdentity(api: ApiPromise, identity: Identity): Promise<Index> {
+  return await getOmniAccountNonce(api, toOmniAccount(api.registry, identity));
```
silva-fj: There is a small issue here, the OmniAccount is only derived from the identity that created the AccountStore (the first member account) so if the user decides to use another member identity to login and make the requests, the OmniAccount will still be the one derived from the first member. 

You can get the OmniAccount with any member identity from the on-chain storage: https://github.com/litentry/heima/blob/78a663b503072b749a4f912aed37c1586fb6a248/parachain/pallets/omni-account/src/lib.rs#L154

Alternatively we could add a runtime api to get the OmniAccount nonce from any member identity

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/utils/identity.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Converts the given identity to the omni account.
+ * 
+ * @reference https://github.com/litentry/heima/blob/dev/common/primitives/core/src/identity.rs#L406
+ *
+ * @param identity - The identity object to convert
+ * @returns An omni account of type AccountId
+ */
+export function toOmniAccount(registry: Registry, identity: Identity): AccountId {
```
silva-fj: This is only valid if the identity is the one that created the AccountStore, if the AccountStore already exists on chain and has multiple members you have to query the OmniAccount from the chain

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/get-nonce.request.ts
wli-pro: resolved now

tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/utils/identity.ts
wli-pro: use toHash() now instead.

silva-fj (approved):

PR #3271, opened 2025-02-28
Setting up skeleton to execute cross chain intents

silva-fj:
This pull request includes some refactoring as well as initial skeleton to support cross-chain intent execution:

- Moved intent execution to the `request_intent` native call handler.
- Refactored project structure by moving intent executors into a new `intent-executors` directory and updating dependencies accordingly.
-  Added a new cross-chain intent executor to handle cross-chain swaps. 
- Modified the `Intent` enum to include a new variant `CrossChainSwap` and added related structures `Asset` and `SwapOrder`.

The logic and services needed to actually execute the swaps will be added later.


common/primitives/core/src/intent.rs
```diff
@@ -47,3 +51,27 @@ pub struct TransferSolana {
     pub to: [u8; 32],
     pub value: u64,
 }
+
+#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq, MaxEncodedLen, TypeInfo)]
+pub struct Asset {
+    pub id: AssetId,
```
kziemianek, line 57: This is u128 - does it mean we will have an asset registry ?
Maybe we need erc20 token addreses ?

common/primitives/core/src/intent.rs
```diff
@@ -47,3 +51,27 @@ pub struct TransferSolana {
...
+    Address33(Address33),
+}
+
+#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq, MaxEncodedLen, TypeInfo)]
+pub struct SwapOrder {
+    pub from_asset: Asset,
+    pub from_amount: u64,
+    pub to_asset: Asset,
+    pub to_min_amount: u64,
+    pub to_address: Option<MultiAddress>,
```
kziemianek, line 76: We might need slippage tolerance here. Can be added later.

kziemianek (approved):

common/primitives/core/src/intent.rs
silva-fj: Good question, I intentionally set it like this to open a discussion. We may need to have some sort of registry we can use to map the assets with the CEX (e.g binance convert api)

common/primitives/core/src/intent.rs
silva-fj: The `to_min_amount` should be calculated with the slippage and the quote from the CEX/DEX we'll use. (the min amount of the output token is what the raydium sdk uses for Swap Instructions, I took it as a guide for now)

common/primitives/core/src/intent.rs
kziemianek: I can imagine we have a registry which maps (network, CA) -> token id on binance somewhere in binance module

tee-worker/omni-executor/parentchain/listener/src/event_handler.rs
```diff
@@ -201,26 +122,6 @@ impl<
...
-				handle_intent_requested_event(
-					&self.ethereum_intent_executor,
-					&self.solana_intent_executor,
-					&self.rpc_client_factory,
-					self.transaction_signer.clone(),
-					intent_requested,
-				)
-				.await?;
-			},
 			AccountStoreUpdated::EVENT => {
```
kziemianek, line 123: It handles only this variant `AccountStoreUpdated::EVENT` while it logs `log::debug!("Got IntentRequested event: {:?}", event.id);`. 

Which component is going to handle `IntentRequested` events ?

kziemianek (changes requested):

tee-worker/omni-executor/parentchain/listener/src/event_handler.rs
silva-fj: >  It handles only this variant AccountStoreUpdated::EVENT while it logs log::debug!("Got IntentRequested event: {:?}", event.id);.

I didn't see that log, it's removed now.

> Which component is going to handle IntentRequestd events ?

We are executing the intents directly when handling `NativeCall::request_intent`. We still send the extrinsic to the chain to log the event, but since the omni-executor handles the intents now, there is no need to wait for those events.

kziemianek: It's [here](https://github.com/litentry/heima/blob/1bd93e6b0a70eb314ea8927c70debeb7a36f48d6/tee-worker/omni-executor/parentchain/listener/src/event_handler.rs#L98)

kziemianek: > We are executing the intents directly when handling NativeCall::request_intent. We still send the extrinsic to the chain to log the event, but since the omni-executor handles the intents now, there is no need to wait for those events.

But from pallet user point of view he can request intent by calling pallet directly, should this call be restricted to enclaves only ?

silva-fj: >It's [here](https://github.com/litentry/heima/blob/1bd93e6b0a70eb314ea8927c70debeb7a36f48d6/tee-worker/omni-executor/parentchain/listener/src/event_handler.rs#L98)

What I meant was that I removed it. (https://github.com/litentry/heima/pull/3271/commits/7ead1f0ae4aed8849e8b8b05ca304cdbc224b38e)

>But from pallet user point of view he can request intent by calling pallet directly, should this call be restricted to enclaves only ?

It is already restricted. The `request_intent` call  only supports `OmniAccountOrigin`, so it can only be called from `dispatch_as_omni_account` (which only supports `TEECallOrigin`)

silva-fj: Now that I think about it, I don't think we even need the `request_intent` call anymore, we are already logging every call dispatched for omni accounts and as well as emitting events when intents are executed

kziemianek (approved):

PR #3277, opened 2025-03-04
Setup new google cloud server for benchmark

BillyWooo:
- fix the google cloud server.
- re-run pallet benchmark for most of the pallets.
- something worth to notice:`add_enclave()` includes `add_enclave_identifier_internal()` inside.


parachain/runtime/litentry/src/weights/pallet_treasury.rs
```diff
@@ -148,4 +103,8 @@ impl<T: frame_system::Config> pallet_treasury::WeightInfo for WeightInfo<T> {
 			.saturating_add(T::DbWeight::get().writes((3_u64).saturating_mul(p.into())))
 			.saturating_add(Weight::from_parts(0, 5206).saturating_mul(p.into()))
 	}
+	fn spend_local() -> Weight { todo!() }
+	fn payout() -> Weight { todo!() }
+	fn check_status() -> Weight { todo!() }
+	fn void_spend() -> Weight { todo!() }
```
Kailai-Wang, line 109: Hmm what's wrong with pallet-treasury? 👀

Kailai-Wang (approved):
Thanks!

parachain/docker/Dockerfile
```diff
@@ -4,9 +4,8 @@ ARG PROFILE
 # ==========================
 # stage 1: build
 # https://hub.docker.com/_/rust
-# our host is based on bullseye/sid (ubuntu 20)
 # ==========================
-FROM rust:bullseye as builder
+FROM rust:1.81 as builder
```
Kailai-Wang: Shall we revert this?

parachain/runtime/litentry/src/weights/pallet_treasury.rs
BillyWooo: pallet_treasury benchmark is skipped for the moment. There is some error within the dependency crate. Spent some time investigate, but didn't succeed. But for the building process, these functions need to be kept here.

parachain/docker/Dockerfile
BillyWooo: Yes, you are right. 👍

PR #3279, opened 2025-03-05
Adding library to the omni-executor for the binance api (Convert)

silva-fj:
This PR introduces a new library to be able to interact with the Binance API (Convert). 

Note: We need to setup a test account to properly test the library. I have not looked into it yet.


tee-worker/omni-executor/binance-api/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn test_sign_request() {
+		let api_key =
```
Kailai-Wang, line 151: How are these api keys and secrets generated? Is it some known secrets (for tests) or did you generate them locally?

Kailai-Wang (approved):
Looks good, thanks!

tee-worker/omni-executor/binance-api/src/lib.rs
silva-fj: These are known credentials, used in binance api documentation as an example to verify that the signature is generated correctly

PR #3280, opened 2025-03-06
P-949 Enable metadata hash for runtime

Kailai-Wang:
### Context

As topic - this feature is enabled by default.

We'll need a follow-up client and runtime upgrade.


parachain/runtime/litentry/src/lib.rs
```diff
@@ -230,7 +231,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("heima"),
 	authoring_version: 1,
 	// same versioning-mechanism as polkadot: use last digit for minor updates
-	spec_version: 9233,
+	spec_version: 9240,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
```
BillyWooo: do we need to bump `transaction_version` ?

Kailai-Wang: I think it doesn't need to - but safer if any third party relies on it

I'll bump it

BillyWooo (approved):
Still not clear why CI from #3114 failed. 
But anyway it's nice to see this issue closed.

PR #3281, opened 2025-03-07
Improve omni-account creation and permissions

silva-fj:
This pull request introduces a few improvements for the omni-account management. The most important changes include adding checks for existing account stores and ensuring permissions can only be set under specific conditions.


Kailai-Wang (approved):
Looks good, thanks!

PR #3282, opened 2025-03-10
P 1350 accounting contract on dest chain

felixfaisal:
Ready for review 
base of the contract - https://linear.app/litentry/issue/P-1350/accounting-contract-on-dest-chain#comment-093d83d5

Pre-requisites: 
1. Install solana-cli 
2. Install anchor 0.30.1 

You can test the sol program locally by performing the following: 
1. Start sol validator - ` solana-test-validator --mint 8QQds7P14EL1ZFjPLsTg2AHZaQHfNGH1EDW8wMhJmxaX --reset` 
4. Run the test - `anchor test --skip-local-validator` 
I have to start this separately because in tests the upgrade authority is always set 0x11..111, so I can't set admin in tests because of that.


solana/accounting-contract/programs/accounting-contract/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub admin_account: Account<'info, AdminAccount>,
+    pub system_program: Program<'info, System>,
+}
+
+#[derive(Accounts)]
+pub struct SetAdmin<'info> {
+    #[account(init_if_needed, payer = caller, space = 8 + AdminAccount::INIT_SPACE, seeds = [b"admin"], bump)]
+    pub admin_account: Account<'info, AdminAccount>,
+    #[account(mut)]
+    pub caller: Signer<'info>,
```
Kailai-Wang: Is it possible to name it as `signer` consistently?

In code I see both `ctx.accounts.signer.key` and `ctx.accounts.caller.key` and I once thought `ctx.accounts.caller.key` doesn't require signing (while it's actually a `Signer` type)

solana/accounting-contract/programs/accounting-contract/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+                    from: ctx.accounts.depositor.to_account_info(),
+                    to: ctx.accounts.treasury.to_account_info(),
+                },
+            ),
+            amount,
+        )?;
+        Ok(())
+    }
+
+    pub fn withdraw_funds(ctx: Context<WithdrawFunds>, amount: u64) -> Result<()> {
```
Kailai-Wang, line 48: A dumb question:
Is `ctx.accounts.beneficiary` actually part of the fn params and can be different in each call? (So that it's not firm like constructor params)

solana/accounting-contract/programs/accounting-contract/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+            .treasury
+            .to_account_info()
+            .try_borrow_mut_lamports()? -= amount;
+        **ctx
+            .accounts
+            .beneficiary
+            .to_account_info()
+            .try_borrow_mut_lamports()? += amount;
+
+        let payout_request = &mut ctx.accounts.pay_out_request;
```
Kailai-Wang, line 104: Also a dumb question: why do we need to mutate the payout request? 🤔

Kailai-Wang (commented):
Thank you for completing the prototype - I have left a few (stupid) questions

Kailai-Wang:
Two more things:
- have you tried to deploy it on solana testnet? If so maybe you could have some deployment script too (CI is even better)
- do you also have the worker code to interact with the contract? (mainly `create_pay_request`)

These two are not strictly needed in this PR tho - but for completeness we should have them (on another PR maybe)

solana/accounting-contract/programs/accounting-contract/src/lib.rs
silva-fj: Why do we need the `PayoutRequest` account?

solana/accounting-contract/programs/accounting-contract/src/lib.rs
silva-fj: Yes, you can set the account that is going to be used when building the transaction in the client side

solana/accounting-contract/programs/accounting-contract/src/lib.rs
felixfaisal: Done, Set everything to signer for consistency ✅

solana/accounting-contract/programs/accounting-contract/src/lib.rs
felixfaisal: So the account cannot be part of the `fn` parameter if we're going to use it to transfer funds to it or from it, so we cannot do it via the `Pubkey`, and we need `AccountInfo`. 

In Solana, we need to be upfront all the accounts we will be using and need to pass them as part of the context for that particular instruction, so yeah `ctx.accounts.beneficiary` is part of the context constructed when calling this function which is modifiable when building the transaction.  

This is how it is constructed for reference - https://github.com/litentry/heima/blob/46bcfa001e058d87ac28b460c3692406a1439144/solana/accounting-contract/tests/src/test_initialize.rs#L107

solana/accounting-contract/programs/accounting-contract/src/lib.rs
felixfaisal: 1. Why do we need to mutate the payout request? 
So `Payout_Request` is a storage item that I'm maintaining to store the `beneficiary` , `nonce` and `amount`, so when someone submits a payout_request, this storage item gets created and updated. 
2. Why do we need the PayoutRequest account?
While you can track individual transactions via the explorer, i felt it was a bit tedious and a storage item could be easier since there was a lot of emphasis on having the ability to audit these cross-chain transactions, If we feel we don't need this storage item then we can remove it.

felixfaisal:
> Two more things:
> 
> * have you tried to deploy it on solana testnet? If so maybe you could have some deployment script too (CI is even better)
> * do you also have the worker code to interact with the contract? (mainly `create_pay_request`)
> 
> These two are not strictly needed in this PR tho - but for completeness we should have them (on another PR maybe)

I've made a couple of issues to follow up and complete these issues as part of separate PRs, feel free to add any info in it (on linear btw)

solana/accounting-contract/programs/accounting-contract/src/lib.rs
silva-fj: I see, but wouldn't this storage only keep data of the latest transaction? There are a few ways to audit the transactions in solana, and it's way simpler that Ethereum because you don't have to go block by block. A simple way is to use [getSignaturesForAddress](https://solana.com/docs/rpc/http/getsignaturesforaddress) in combination with [getTransaction](https://solana.com/docs/rpc/http/gettransaction)

felixfaisal: > I see, but wouldn't this storage only keep data of the latest transaction? 

No, it doesn't store only the latest transaction, it's address is derived with the beneficiary and nonce, It's sort of like double mapping keys, which is unique for each payout request. 
https://github.com/litentry/heima/blob/109ae6e206ec654a76d7848a7b38679fb848e3f2/solana/accounting-contract/programs/accounting-contract/src/lib.rs#L149

felixfaisal: > A simple way is to use [getSignaturesForAddress](https://solana.com/docs/rpc/http/getsignaturesforaddress) in combination with [getTransaction](https://solana.com/docs/rpc/http/gettransaction)

This still requires filtering for the particular transaction and fetching the details for it, It still requires an upper and lower bound for the blocks. 

We are basically checking which all contexts this address is a part of and then fetching the details of those transactions, correct, wdyt? 
We can remove it if we feel it's not necessary

silva-fj: > > I see, but wouldn't this storage only keep data of the latest transaction? 
> 
> 
> 
> No, it doesn't store only the latest transaction, it's address is derived with the beneficiary and nonce, It's sort of like double mapping keys, which is unique for each payout request. 
> 
> https://github.com/litentry/heima/blob/109ae6e206ec654a76d7848a7b38679fb848e3f2/solana/accounting-contract/programs/accounting-contract/src/lib.rs#L149

Ah, I missed this.

silva-fj: > > A simple way is to use [getSignaturesForAddress](https://solana.com/docs/rpc/http/getsignaturesforaddress) in combination with [getTransaction](https://solana.com/docs/rpc/http/gettransaction)
> 
> 
> 
> This still requires filtering for the particular transaction and fetching the details for it, It still requires an upper and lower bound for the blocks. 
> 
> 
> 
> We are basically checking which all contexts this address is a part of and then fetching the details of those transactions, correct, wdyt? 
> 
> We can remove it if we feel it's not necessary

I think it's fine to keep it

silva-fj (approved):

PR #3283, opened 2025-03-11
Ethereum delegate call and pay fees

kziemianek:
Adds ability to delegate calls using EIP-7702 features so transaction fees can be paid by omni-executor account and user's omni account doesn't need to hold any `eth`.


tee-worker/omni-executor/intent-executors/ethereum-intent-executor/src/delegate_call.rs
```diff
@@ -0,0 +1,63 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+use alloy::primitives::{Address, Uint};
+use alloy::sol_types::SolValue;
+use alloy::{hex, sol};
+
+sol!(
```
Kailai-Wang, line 21: Is there any macro to reuse sol file and avoid copy-up?

tee-worker/omni-executor/intent-executors/ethereum-intent-executor/src/lib.rs
```diff
@@ -16,77 +16,62 @@
...
 pub struct EthereumIntentExecutor {
 	rpc_url: String,
+	#[allow(dead_code)]
+	delegation_contract_address: Address,
 }
 
 impl EthereumIntentExecutor {
-	pub fn new(rpc_url: &str) -> Result<Self, ()> {
-		Ok(Self { rpc_url: rpc_url.to_string() })
+	pub fn new(rpc_url: &str, delegation_contract_address: &str) -> Result<Self, ()> {
```
Kailai-Wang, line 39: `delegation_contract_address` is bound to the constructor, imagine we want to support more than one evm chain, does it mean we'll have per-chain intent-exectuor instance?

tee-worker/omni-executor/intent-executors/ethereum-intent-executor/src/tx.rs
```diff
@@ -0,0 +1,125 @@
...
+					>,
+				>,
+			>,
+			alloy::providers::fillers::WalletFiller<EthereumWallet>,
+		>,
+		alloy::providers::RootProvider<
+			alloy::transports::http::Http<alloy::transports::http::Client>,
+		>,
+		alloy::transports::http::Http<alloy::transports::http::Client>,
+		alloy::network::Ethereum,
```
Kailai-Wang: Is this type annotation needed? It looks very verbose 😂

Kailai-Wang (approved):
Thank you, it looks in general.

This is prepared for EIP7702 right? So it means we could only test it on sepolia for now

tee-worker/omni-executor/intent-executors/ethereum-intent-executor/src/lib.rs
kziemianek: That's not clear yet - but `EthereumIntentExecutor` is already coupled to specific network (`rpc_url`) so I kept it consistent.

kziemianek:
Can be tested on local `anvil` instance or ethereum test networks (like `sepolia`)

PR #3289, opened 2025-03-14
Change chain-spec id to heima

Kailai-Wang:
### Context

As topic, after this change is applied/released all node operators need to rename the database folder, which they need to do anyway as a result of the previous binary name change.


BillyWooo (approved):

m1iktea:
LGTM, When you deploy, please let me know that I need to modify the original storage directory

PR #3290, opened 2025-03-14
Adding support for binance spot trading api

silva-fj:
As topic, this PR adds the most important API methods to support trading with the spot trading api as well as some small refactoring.

I've tested the spot trading api with the test network: https://testnet.binance.vision/


kziemianek (approved):
It looks fine, hovewer worker will need to manage open orders.
Also order book is dynamic and it may not be a taker always.

silva-fj:
> It looks fine, hovewer worker will need to manage open orders. Also order book is dynamic and it may not be a taker always.

Yes, as far as I understand it depends on how we decide to use it. We can set the order type, if it's of type `LIMIT` for example, we have to set the exact price and we'll need to manage the open orders, but if we use order type `MARKET`, then the trade is executed at the market price so we can just wait for the result.

PR #3291, opened 2025-03-17
feat: P-1339 add requests: addAccount, removeAccounts

wli-pro:
publicizeAccount, setPermissions, add chaindata and use it in enclave, generate docs for existing code, upgrade nrwl to nx

### Labels

### Testing Evidences


tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/test-utils/helpers.ts
```diff
@@ -0,0 +1,32 @@
...
+): Promise<CorePrimitivesOmniAccountMemberAccount[]> {
+  // wait account store created
+  let accountStore: Option<Vec<CorePrimitivesOmniAccountMemberAccount>> | undefined;
+  do {
+    try {
+      accountStore = await api.query.omniAccount.accountStore(account);
+    } catch (_) {
+      // do nothing
+    }
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
silva-fj, line 34: This will wait even when getting the accountStore, how about doing a simple `while` loop and `break` it when getting the value?

wli-pro: added while break now.

silva-fj (approved):

PR #3292, opened 2025-03-18
Prefund ethereum account

kziemianek:
Ethereum tx `submit` function accepts additional `prefund` flag.
If `prefund` is set to `true` and `signer_key` account doesn't have eth to cover tx fees then required amount of eth will be sent from `pay_master` account.


tee-worker/omni-executor/intent-executors/ethereum-intent-executor/src/tx.rs
```diff
@@ -35,19 +41,24 @@ pub async fn submit(
 	value: [u8; 32],
 	call_data: Vec<u8>,
 	signer_key: PrivateKeySigner,
-	delegation_details: Option<DelegationDetails>,
-) -> Result<(), ()> {
-	let (tx_signer, delegation_contract_address) =
-		if let Some(delegation_details) = delegation_details {
-			(delegation_details.pay_master, Some(delegation_details.delegation_contract_address))
-		} else {
+	delegation_or_prefund_details: Option<DelegationDetailsOrPrefund>,
```
silva-fj: How about changing this to an optional enum like: `mode:  Option<SubmissionMode>` (or some other name) ? I think this would make the logic a bit easier to understand and will probably make it simpler as well

kziemianek: Sounds like a good idea. I'll add enum `Paymode` with 3 variants: `Self`, `Delegated`, `Prefunded`

silva-fj (approved):

PR #3296, opened 2025-03-20
Setting fixed auth token (JTW) exp time to 1 week.

silva-fj:
This pull request improves the the auth token security by fixing the expiration time to ~ 1 week.


Kailai-Wang (approved):
Thanks!

tee-worker/client-api/parachain-api/prepare-build/interfaces/omniExecutor/definitions.ts
```diff
@@ -12,7 +12,7 @@ export default {
         MrEnclave: "H256",
         NativeCall: {
             _enum: {
-                request_auth_token: "(LitentryIdentity, AuthOptions)",
```
0xverin: Can we release a new client-sdk in this PR? Then I will update it in the sandbox to adapt (remove the JWT expiration time parameter).

silva-fj: Actually, let me revert the changes in the identity client-sdk, this should only affect the omni-executor. When the omni-executor's client sdk is ready we can adapt the demo app

PR #3297, opened 2025-03-20
Add configurable max connections limit for the omni-executor RPC server

silva-fj:
This allows better control over server resource usage and scaling capabilities.


tee-worker/omni-executor/rpc-server/src/server.rs
```diff
@@ -73,7 +73,12 @@ pub async fn start_server<
 	jwt_secret: String,
 ) -> Result<(), Box<dyn std::error::Error>> {
 	let address = format!("0.0.0.0:{}", port);
-	let server = Server::builder().build(address.parse::<SocketAddr>()?).await?;
+	let max_connections: u32 =
+		env::var("RPC_SERVER_MAX_CONNECTIONS").unwrap_or("100".to_string()).parse()?;
```
kziemianek: Should we prefix all env variables with `OE` or `OMNI_EXECUTOR` ?

kziemianek (approved):

tee-worker/omni-executor/rpc-server/src/server.rs
silva-fj: It sounds good

PR #3303, opened 2025-03-24
PumpX API initial setup in the omni-executor

silva-fj:
This PR introduces the initial setup of the Pumpx API, as well as some refactoring of the JWTs issued for pumpx and the user_connect integration.


Kailai-Wang:
Thanks - would you wait for https://github.com/litentry/heima/pull/3319 first and then adjust based on it? 🙌

silva-fj:
> Thanks - would you wait for #3319 first and then adjust based on it? 🙌

Sure, no problem

tee-worker/omni-executor/executor-core/src/native_operation.rs
Kailai-Wang, line 1: I think this file can be completely removed after we switch to `NativeTask`

tee-worker/omni-executor/executor-storage/src/pumpx_auth_token_id.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+impl PumpxAuthTokenIdStorage {
+	pub fn new(db: Arc<DB>) -> Self {
+		Self { db }
+	}
+}
+
+impl Storage<Hash, String> for PumpxAuthTokenIdStorage {
+	fn get(&self, identity_hash: &Hash) -> Option<String> {
```
Kailai-Wang: A general question to `Storage`

shall we key it by `OmniAccount`? Imagine we have more than one identity later, so the passed-in identity might not be the "main" identity.

Just want to be generic enough and stick to the concept that omniAccount is the unique identifer for the user

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -192,11 +196,28 @@ async fn handle_native_task<
...
+						AUTH_TOKEN_ID_TYPE.to_string(),
+						auth_options,
+					)
+				},
+				_ => AuthTokenClaims::new(
+					sender.hash().to_string(),
+					AUTH_TOKEN_ID_TYPE.to_string(),
+					auth_options,
+				),
+			};
```
Kailai-Wang, line 220: Just to make sure that I understand it correctly, this part (`RequestAuthToken`) is actually not used in the pumpx flow, isn't it?

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -525,33 +567,55 @@ async fn handle_native_task<
...
 					log::error!("Failed to send response");
 				}
 				return;
 			};
 
-			let response =
-				NativeTaskResponse::Ok(NativeTaskOk::PumpxJwt { session_token, trade_token });
+			let storage = PumpxAuthTokenIdStorage::new(ctx.storage_db.clone());
+			if storage.insert(sender.hash(), id_token.clone()).is_err() {
+				log::error!("Failed to insert pumpx_auth_token into storage");
```
Kailai-Wang: Can we be more specific about logging, like "failed to insert **id** token" - same for other places

tee-worker/omni-executor/pumpx/src/pumpx_api/types.rs
Kailai-Wang, line 1: Thanks for adding those APIs : )

tee-worker/omni-executor/pumpx/src/pumpx_api/types.rs
```diff
@@ -0,0 +1,101 @@
...
+	pub google_auth_check: bool,
+}
+
+#[derive(Deserialize, Serialize, Encode, Decode, PartialEq, Eq, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct ApiResponse<T: Codec> {
+	code: u32,
+	message: String,
+	data: T,
+}
```
Kailai-Wang, line 40: I like the abstraction of this 👍 We just need to double-check later if every response follows it

Kailai-Wang (commented):
Looks good! Thanks, I only have some small comments

tee-worker/omni-executor/executor-core/src/native_operation.rs
silva-fj: ah, I was re-introduced when fixing the conflicts. I'll remove it

tee-worker/omni-executor/native-task-handler/src/lib.rs
silva-fj: Correct

tee-worker/omni-executor/pumpx/src/pumpx_api/types.rs
silva-fj: All of the responses in the docs I've seen have this shape, hopefully it's consistent

Kailai-Wang (approved):
Thanks!

Does it make sense to apply OmniAccount as key also for other storages? (Not in this PR)

silva-fj:
> Thanks!
> 
> Does it make sense to apply OmniAccount as key also for other storages? (Not in this PR)

The other storages that use identity as key are for the email verification code and oauth2 code but I think they are identity specific. Wdyt?

PR #3304, opened 2025-03-24
Lock account assets

kziemianek:
Adds logic to track locked assets per account.

Implemented strategies:
* `account-wide` - single `bool` flag indicating whether there is any asset locked per account
* `precise` - tracks locked asset amounts per account

We should rely on `account-wide` strategy untill we implement persistence layer for intents.


tee-worker/omni-executor/intent/asset-lock/src/account_wide.rs
```diff
@@ -0,0 +1,169 @@
...
+			return Err(());
+		}
+		*locked = true;
+		Ok(())
+	}
+
+	fn release(&mut self, _asset_id: crate::AssetId, _amount_to_release: u128) -> Result<(), ()> {
+		let mut locked = self.locked.write().unwrap();
+
+		if !*locked {
```
kziemianek: We could go with less restrictive strategy here and return `Ok` even if not locked.

tee-worker/omni-executor/intent/asset-lock/src/precise.rs
```diff
@@ -0,0 +1,244 @@
...
+			}
+			self.locked_assets.insert(asset_id, amount_to_lock);
+		}
+		Ok(())
+	}
+
+	fn release(&mut self, asset_id: AssetId, amount_to_release: u128) -> Result<(), ()> {
+		if let Some(lock) = self.locked_assets.get_mut(&asset_id) {
+			if *lock < amount_to_release {
+				// this is some kind of inconsistency - we should either set lock to 0 and return Ok or return Err
```
kziemianek, line 63: We could go with less restrictive strategy here and return `Ok` even if not locked.

tee-worker/omni-executor/intent/asset-lock/src/account_wide.rs
```diff
@@ -0,0 +1,169 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::sync::RwLock;
+
+use crate::AssetsLock;
+
+// Manages whether any asset is locked for particular account or not.
+pub struct AccountWideAssetsLock {
+	locked: RwLock<bool>,
```
silva-fj: How about using `std::sync::atomic::AtomicBool` here instead? We can avoid the overhead of acquiring and releasing a lock and the code would get simpler, e.g: 

```rust
self.locked.swap(true, Ordering::SeqCst)
...
self.locked.swap(false, Ordering::SeqCst)
```

kziemianek: I've changed it to plain `bool` .

silva-fj (approved):

PR #3305, opened 2025-03-24
Update Intent enum type for pumpx trade

Kailai-Wang:
### Context

As topic

- moves all omni-related primitives to `common/primitives/core/src/omni` mod
- change `Asset` definition to be more generic
- keep only necessary fields in `SwapOrder` - other configs could belong to some `ExtraArgs` struct or so - the `SwapOrder` only records the raw order
- place `SwapOrder` as index 0 for consistent reference for pumpx colleagues, other intents were for demoApp - they could be refactored

cc @silva-fj 

@kziemianek would this be sufficient?


common/primitives/core/src/omni/intent.rs
```diff
@@ -52,10 +50,14 @@ pub struct TransferSolana {
...
 
-#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq, MaxEncodedLen, TypeInfo)]
+#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq, TypeInfo)]
 pub struct Asset {
-    pub id: AssetId,
-    pub chain_id: ChainId,
+    // `name` is intentionally defined as string-like type to:
+    // - allow easier integration with external service. E.g. APIs of binance/pumpx definitely expect strings
+    // - be future-compatible: meme-coins/alt-coins could have arbitrary names
+    pub name: Vec<u8>,
```
kziemianek: This basically doesn't align well with [current token amount query implementation](https://github.com/litentry/heima/blob/dev/tee-worker/omni-executor/intent/token-query/src/lib.rs#L31-L41).

Should we query pumpx for account balance ?

Kailai-Wang: As per requirements, we would only swap "mainstream coins" (= native currency and stable coins) for any coin, specifically on solana it would be [`SOL` and `USDT`](https://linear.app/litentry/issue/P-1392/double-check-supported-sourcedest-chains-token-types#comment-9513941f). So I assume we only need to query balance of those two tokens.

Shall we have a `to_solana_token()` which maps the `name` on Solana to `SolanaToken` type? We'll need to hardcode the mint address of `usdt` in the mapping (https://solscan.io/token/Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB)

Wdyt?

kziemianek: I think mint token / erc20 address should be used instead of `name` in `intent.rs` - this is what uniiquely identiefies token.
We can implement whitelisting if limitation is needed.

If pumpx api cannot accept address and we have to provide name - we should do the mapping in package responsible for integrating with pumpx.

kziemianek: pumpx api expects CA https://github.com/litentry/heima/pull/3303/files#diff-649b162768ba7355b16233723b2ffa0ab2606dc5a87fe55bee1692992219d29bR66

Kailai-Wang: CA is string tho, shall we also use Vec<u8> for it? Especially considering different chains could have different formats.

And what about binance-api? Does it expect token name or CAs? @silva-fj

silva-fj: The Binance API expects token pair symbols e.g: USDTSOL to trade USDT -> SOL

Kailai-Wang: Hmm then we'll need `name` <> `CA` mapping in the code anyway. In this case I prefer we simply use `name` for better readability, not to mention `CA` has to be of type `Vec<u8>` too even if go for it

Kailai-Wang: Wdyt? @kziemianek @silva-fj

kziemianek: But it can be enum `Asset { Solana(SolanaAsset) , Ethereum(EthereumAsset) }` , see asset enums [here](https://github.com/litentry/heima/blob/dev/tee-worker/omni-executor/intent/token-query/src/lib.rs#L31-L41).

Additionaly we will need tickers for only few assets (USDT,SOL,ETH) while we will need `CA` for all of the rest (practically unbounded). If that's correct I would prefer to map address to ticker.

Kailai-Wang: So an enum whose entries are grouped by chain types, sounds like a good idea! Let me adjust it

Kailai-Wang:
Asking for a re-review @silva-fj @kziemianek 

Now the asset is an enum:
```
pub enum ChainAsset {
    Ethereum(u32, EthereumToken), // with chain id
    Solana(SolanaToken),
}
```

For `Ethereum` I had to include the chain_id again - the same token (name) could have different CAs on different evm chains, their native token is different too

kziemianek (approved):

PR #3307, opened 2025-03-25
feat: add accounting contract for bsc 

felixfaisal:
Adds the bsc-accounting contract under `evm-contracts/bsc-accounting-contract/`  

I've more or less used the same logic from the solana program, to test it out: 
1. `forge build` 
2. `forge test` 
I've tried to cover all possible cases in my unit tests 

I'll do the following in next PRs 
1. ~Set up CI that builds the contract and tests it~ (already done in this PR)
2. ~Integration code for the omni-executor~ (also finished in this PR) 
3. Deployment script and testing on bsc-testnet


evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
```diff
@@ -0,0 +1,103 @@
...
+        require(msg.sender == admin, "Unauthorized: not admin");
+        _;
+    }
+
+    modifier onlyWorker() {
+        require(msg.sender == worker, "Unauthorized: not worker");
+        _;
+    }
+
+    function setAdmin(address newAdmin) external onlyOwner {
```
Kailai-Wang: I kind of feel `Owner` and `Admin` are somewhat overlapped, is it better to keep only one?

evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
```diff
@@ -0,0 +1,103 @@
...
+        worker = initialWorker;
+    }
+
+    address public admin;
+    address public worker;
+    mapping(address => uint256) public nonces;
+
+    struct Payout {
+        uint256 amount;
+        bool paid;
```
Kailai-Wang: Is there any case where `paid` is `false`?

evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
```diff
@@ -0,0 +1,103 @@
...
+        Payout memory request = payouts[beneficiary][nonce];
+        require(request.amount == 0, "Already exists");
+
+        (bool sent, ) = beneficiary.call{value: amount}("");
+        require(sent, "Payout failed");
+
+        payouts[beneficiary][nonce] = Payout({amount: amount, paid: true});
+        nonces[beneficiary] = nonce;
+
+        emit PayoutRequestCreated(beneficiary, amount, nonce);
```
Kailai-Wang: It sounds more like `PaidOut` or `PayoutRequestExecuted` - same problem with fn name

Kailai-Wang (approved):
Looks good in general, some small problems

Asking @kziemianek @silva-fj @wli-pro to review it too

tee-worker/omni-executor/accounting-contract-client/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	pub async fn create_pay_request(
+		&self,
+		beneficiary: Address,
+		nonce: U256,
+		amount: U256,
+	) -> Result<(), ()> {
+		let call =
+			AccountingContractCalls::createPayRequest(AccountingContract::createPayRequestCall {
```
wli-pro: This function seems to handle the creation and submission of pay request transactions. Would it worth renaming it to submitPayRequest?

evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    function withdrawFunds(
+        address payable beneficiary,
+        uint256 amount
+    ) external onlyAdmin {
+        require(address(this).balance >= amount, "OutOfBalance");
+        (bool sent, ) = beneficiary.call{value: amount}("");
+        require(sent, "Withdraw failed");
+        emit FundsWithdrawn(beneficiary, amount);
+    }
```
wli-pro: Is there any reentrancy vulnerability? should we update state before interaction? as below:
```
function withdrawFunds(
    address payable beneficiary,
    uint256 amount
) external onlyAdmin {
    require(address(this).balance >= amount, "OutOfBalance");
    
    emit FundsWithdrawn(beneficiary, amount);
    
    (bool sent, ) = beneficiary.call{value: amount}("");
    require(sent, "Withdraw failed");
}
```
Should we use Openzeppelin ReentrancyGuard contract that can help to avoid reentrancy vulnerability?

felixfaisal: Thanks @wli-pro that's a good catch, I think reentrancy is possible here, I'll include the ReentrancyGuard here

felixfaisal: Added the non reentrant contract implementation

tee-worker/omni-executor/accounting-contract-client/src/lib.rs
felixfaisal: Renamed it to execute_pay_out_request

evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
felixfaisal: Renamed the fn to - `executePayOutRequest` and event to `PayOutRequestExecuted`

evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
felixfaisal: There is no such case, i was thinking something else when writing this struct 😅 , i've removed it.

evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
felixfaisal: Yeah, I felt so too, So i updated the contract to use `AccessControls` instead of `Ownable`, I'm still keeping the storage item for worker public key so that when setting new worker, we can also revoke the role for the current worker role holder, to ensure only one public key will always have the `Worker` role.

Kailai-Wang:
The changes look good, thanks.

Only a small (picky) request: the accounting contract is tightly bound to the executor. Does it make sense to move them to
- `tee-worker/omni-executor/accounting-contract/bsc`
- `tee-worker/omni-executor/accounting-contract/solana`

evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
```diff
@@ -0,0 +1,100 @@
...
+        address indexed beneficiary,
+        uint256 amount,
+        uint256 nonce
+    );
+
+    function setAdmin(address newAdmin) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        require(newAdmin != address(0), "New admin cannot be zero address");
+        emit AdminUpdated(msg.sender, newAdmin);
+        grantRole(DEFAULT_ADMIN_ROLE, newAdmin);
+        renounceRole(DEFAULT_ADMIN_ROLE, msg.sender);
```
silva-fj: I think it would be safer to let the new admin accept the role before the old resigns to it. If the new admin address is incorrect or inaccessible, this could result in permanent loss of admin access

silva-fj (commented):
it looks good to me. I just left a small comment

evm-contracts/bsc-accounting-contract/src/AccountingContract.sol
felixfaisal: I frankly, don't think this a concern as the admin public key will most likely be a multi-sig address similar to that of chainbridge, and the same pattern is used for the omni-bridge. There is due diligence required before setting a new admin. 

> I think it would be safer to let the new admin accept the role before the old resigns to it

How does the new admin accept the role? If it's via tx, then it's sort of a syn/ack model i.e. 
1. `admin` sends a tx to `grantRole` to `new_admin` 
2. `new_admin` sends a tx to `acknowledge` the admin role 
3. `admin` sends a tx to `renounceRole` 

This will require three tx instead of one, and also has to monitor cases where the `new_admin` fails to acknowledge role and actively needs to `revoke` it. 

Let me know if you have something else in mind for the implementation, I don't see the advantage in complicating this.

silva-fj: It would require at least 2 transactions, yes. 

It was just a suggestion, In my experience being extra safe in crypto is always good, specially when managing assets. e.g: A multisig won't prevent a typo

Kailai-Wang: Yea I think 2 step is the standard (openzepplin) way of transferring ownership. It makes sure the new admin can actually accept it (sending tx).

Setting admin in one step is substrate style: like `set_sudo`

In the context of evm, I actually agree using a two-step paradigm is safer. Once it goes wrong, you can't modify the (raw) storage like runtime upgrade in substrate as last resort.

@felixfaisal wdyt?

Changing it isn't too hard - we just need to use `AccessControlDefaultAdminRules` - you could refer to our HEI contract:  https://github.com/litentry/HEI-token/blob/main/contracts/heima/HEI.sol

Kailai-Wang: Even if it's hard to go wrong when calling `set_admin`, it's never bad to ask the new admin to actually test it out (by explicitly accepting it)

felixfaisal: I went through it now, Makes sense and is simple, I was just confused about the implementation details and unfamiliar with this 😅 . I'll implement this in the contract right away.

felixfaisal: Done ✅

PR #3308, opened 2025-03-25
Update JWT token issuance

silva-fj:
This pull request updates JWT handling mechanism, switching from using a shared secret to RSA key pairs for encoding and decoding JWTs. Additionally, a new native call was added (`NativeCall::request_pumpx_jwt`) that issues a session token and a trade token

Note: This native call also relies on the existence of the OmniAccount, a follow up PR will change that


tee-worker/client-api/parachain-api/prepare-build/interfaces/omniExecutor/definitions.ts
```diff
@@ -63,9 +63,13 @@ export default {
         CallResponse: {
             _enum: {
                 ExtrinsicReport: "XtReport",
-                AuthToken: "Text",
+                AuthToken: "AuthToken",
```
silva-fj, line 66: FYI @wli-pro

tee-worker/omni-executor/native-task-handler/src/native_call_handlers.rs
```diff
@@ -386,6 +393,76 @@ pub async fn handle_native_call<
...
+			let Ok(trade_token) = jwt::create(&trade_claims, &ctx.jwt_rsa_private_key) else {
+				let response =
+					NativeOperationResponse::Err(NativeOperationError::AuthTokenCreationFailed);
+				if response_sender.send(response.encode()).is_err() {
+					log::error!("Failed to send response");
+				}
+				return;
+			};
+			// TODO: should we track this?
+			let auth_token_requested_call = parentchain_api_interface::tx()
```
kziemianek: I think this should not have such side effects.

kziemianek (approved):

PR #3311, opened 2025-03-27
More tests onmi sdk

wli-pro:
### Context

1 Added tests as part of CI
2 Some missing changes due to package rename from heima to heima-network 

### Labels

### Testing Evidences


0xverin (approved):
over all looks good.

tee-worker/omni-executor/native-task-handler/src/native_call_handlers.rs
```diff
@@ -91,7 +91,24 @@ pub async fn handle_native_call<
...
+					let response =
+						NativeOperationResponse::Err(NativeOperationError::InternalError);
+					if response_sender.send(response.encode()).is_err() {
+						log::error!("Failed to send response");
+					}
+					return;
+				},
+			};
+			// Increment nonce for the next transaction
+			let tx = ctx.transaction_signer.sign(auth_token_requested_call, Some(nonce + 1)).await;
```
Kailai-Wang, line 111: If I look up `ctx.transaction_signer.sign` in the file, I see different code:
- `ctx.transaction_signer.sign(.., Some(nonce) ...)`
- `ctx.transaction_signer.sign(.., Some(nonce + 1) ...)`
- `ctx.transaction_signer.sign(.., None ...)`

Which one is correct? Can we have consistent code for it?
@wli-pro @silva-fj

silva-fj: We should use `None` when we submit only one extrinsic, the signer will get the current nonce and set it. But in cases where we submit more than once (e.g: `RequestIntent`), we need to increment the nonce manually for the second extrinsic. This is where we use the `Some()` variant. (In the first two examples we are doing the same thing, it's just that in the first one, the nonce is incremented before passing it to the signer and the second one it is incremented while passing it to the signer.

Another option could be to have two methods. `sign(..)` and `sign_with_nonce(.., nonce)`, but the logic should be the same

kziemianek: This is something that will very likely cause troubles when more than one thread is using same signer account.

silva-fj: True, we should probably track the nonce in-memory

silva-fj: I'll create an issue for this

PR #3314, opened 2025-03-27
Intent sanity check

kziemianek:
Adds checks before accepting cross chain trade intent.

1. Account has enough balance on source chain
2. Only one cross chain intent is processed at same time.


silva-fj (approved):

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -15,31 +15,170 @@
...
+				else {
+					log::error!("Failed to create rpc client");
+					return Err(());
+				};
+				let available_amount = match &swap_order.from_asset {
+					ChainAsset::Ethereum(chain_id, token) => {
+						let rpc_url =
+							self.rpc_endpoint_registry.get(&Chain::Ethereum(*chain_id)).ok_or(())?;
+						// query pumpx signer wallet
+						let address = EthereumAddress::default();
```
Kailai-Wang: Here it's actually TODOs, right?
Can we do it now, or in another PR?

Kailai-Wang (approved):

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: Updated

common/primitives/core/src/omni/intent.rs
```diff
@@ -16,15 +18,29 @@ pub enum Intent {
...
     #[codec(index = 4)]
     TransferNative(TransferNative),
     #[codec(index = 5)]
     TransferSolana(TransferSolana),
 }
 
+impl Intent {
+    pub fn intent_id(&self) -> &IntentId {
+        match self {
+            Intent::CrossChainSwap(so) => &so.intent_id,
```
BillyWooo: There are new changes from PR: https://github.com/litentry/heima/pull/3340.

BillyWooo (approved):

BillyWooo:
Let's merge it once CI passes.

tee-worker/omni-executor/rpc-server/src/methods/omni/get_next_intent_id.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::ErrorCode;
+use executor_primitives::AccountId;
+use jsonrpsee::{types::ErrorObject, RpcModule};
+use log::error;
+use parentchain_rpc_client::AccountId32;
+
+pub fn register_get_next_intent_id(module: &mut RpcModule<RpcContext>) {
+	module
+		.register_async_method("omni_getNextIntentId", |params, ctx, _| async move {
+			//ss58 encoded account_id
```
Kailai-Wang, line 29: Someday we'll need to unify the omni-account identiifer format - in some places e.g. `sub` field in jwt we use hexed [u8; 32] as omni-account 

cc @silva-fj

tee-worker/omni-executor/rpc-server/src/methods/pumpx/get_next_intent_id.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::ErrorCode;
+use executor_primitives::AccountId;
+use heima_primitives::{Identity, IdentityString};
+use jsonrpsee::{types::ErrorObject, RpcModule};
+use log::error;
+
+pub fn register_get_next_intent_id(module: &mut RpcModule<RpcContext>) {
+	module
+		.register_async_method("pumpx_getNextIntentId", |params, ctx, _| async move {
+			match params.parse::<String>() {
```
Kailai-Wang: Could you use JSON'ed param here for consistency, even tho there's only one parameter?

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -252,7 +256,36 @@ async fn handle_native_task<
...
+					let response = NativeTaskResponse::Err(NativeTaskError::InternalError);
+					if response_sender.send(response.encode()).is_err() {
+						log::error!("Failed to send response");
+					}
+					return;
+				}
+			} else {
+				let response = NativeTaskResponse::Err(NativeTaskError::IntentNonceMismatch);
+				if response_sender.send(response.encode()).is_err() {
+					log::error!(
```
Kailai-Wang, line 281: I this this error log is in the wrong branch

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -252,7 +256,36 @@ async fn handle_native_task<
...
+			let Ok(stored_intent_id) = ctx.intent_id_store.get(&sender.to_omni_account()).await
+			else {
+				log::error!("Failed to read intent from store");
+				let response = NativeTaskResponse::Err(NativeTaskError::InternalError);
+				if response_sender.send(response.encode()).is_err() {
+					log::error!("Failed to send response");
+				}
+				return;
+			};
+			if intent_id != stored_intent_id + 1 {
```
Kailai-Wang: Should be `intent_id == stored_intent_id + 1` here?

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -15,31 +15,176 @@
...
+						})?;
+						query_solana(rpc_url, &pubkey, token).await.map(|v| AmountType::from(v))?
+					},
+				};
+				self.account_asset_lock.check_and_insert(
+					account_id.clone(),
+					swap_order.from_asset.clone(),
+					AmountType::from(swap_order.from_amount),
+					available_amount,
+				)?;
```
Kailai-Wang, line 162: I think we agreed that we don't have to lock the asset (as we do ordered processing: 1. deduct balance 2. binance convert 3. payout), but just need to put the next cross-chain swap request on-hold until this payout is done.

Or did you plan to do it in another PR?

Kailai-Wang: I'm also thinking of a corner case:

1. user requests to do a cross-chain swap with intent-id 1
2. user requests to do a single-chain swap with intent-id 2
3. user requests to do a single-chains wap with intent-id 3
4. user requests to do a cross-chains wap with intent-id 4

In this case, shall we block 2 and 3 until 1 is finished? Probably not.

But then it means:
- we should only lock the cross-chain trade processing part (so we'll need to distinguish between cross-chain swap and single-chain swap)
- the id continuity check isn't simple +1 check, we need to keep track of last unfinished cross-chain trade id (e.g. for intent 4 we need check if payout of intent 1 has finished)

Is it a bit too complicated? 😂

Kailai-Wang (commented):
Thanks! Looks good in general, I have questions on the blocked processing of intents

tee-worker/omni-executor/rpc-server/src/methods/pumpx/get_next_intent_id.rs
kziemianek: I can wrap it within object.

kziemianek: Done

tee-worker/omni-executor/native-task-handler/src/lib.rs
kziemianek: Done

tee-worker/omni-executor/native-task-handler/src/lib.rs
kziemianek: Done

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: > I think we agreed that we don't have to lock the asset (as we do ordered processing: 1. deduct balance 2. binance convert 3. payout), but just need to put the next cross-chain swap request on-hold until this payout is done.
> 
> Or did you plan to do it in another PR?

Currently we are using `AccountWideAssetsLock` which simply locks whole account for the time of intent processing (only one intent at the time can be processed per account). 

Worker:
1. checks intent id
2. updates intent id
3. checks account lock

Kailai-Wang: I misread it sorry, I see it's an account-wide lock. 

But it also means if the next request comes, it will error out instead of waiting, right? Acutally I'm rethinking if we even need a lock here. Can we simply continue processing, if the balance is insufficient, it will error out anyway when we try to deduct user balance?

kziemianek: We can queue intents -> persist them to disk.
Upon intent processing finish - we can check if there is any pending intent and send if for processing.
Some intents can be queue - some not - depends on type <-but this adds another level of complexity. Once intent type is allowed for parallel processing you need to keep it in mind everytime you make any changes to that type processing logic, otherwise you can bring troubles 😈

PR #3318, opened 2025-03-28
Pumpx signer client

kziemianek:
Adds client to interact with `pumpx-signer` service


silva-fj (approved):

tee-worker/omni-executor/pumpx/Cargo.toml
```diff
@@ -5,6 +5,13 @@ authors = ['Trust Computing GmbH <info@litentry.com>']
 edition.workspace = true
 
 [dependencies]
+hex = { workspace = true }
+jsonrpsee = { workspace = true, features = ["client"] }
+serde = { version = "1.0", features = ["derive"] }
```
Kailai-Wang: Shall we use workspace dep?

tee-worker/omni-executor/pumpx/Cargo.toml
```diff
@@ -5,6 +5,13 @@ authors = ['Trust Computing GmbH <info@litentry.com>']
...
+serde_json = { workspace = true }
+serde_with = { version = "3.11", features = ["hex"] }
```
Kailai-Wang: Same here - maybe define it in workspace? I mean it would make sure we use consistent version in the whole project

Kailai-Wang (approved):

PR #3319, opened 2025-03-30
Simplify nativetask calling interface

Kailai-Wang:
### Context

As topic - hopefully this can be used as base for subsequent RPC method implementation for pumpx.

Major changes:
- consistent interface: `omni_submitNativeTask` is used to submit `NativeTask` to worker, a list of `NativeTask`s is defined in `native_task.rs` (basically what we have so far). A `NativeTask` could require authentication and/or encryption - see the `NativeTaskTrait` definition - if the client doesn't provide this, it will error out.
- `RawTask` is the only payload to construct for `omni_submitNativeTask` as parameter. It's an enum type of a plain task (`NativeTaskWrapper`, which is a wrapper to `NativeTask`) or aes-encrypted task (decrypting the aes payload will get you a `NativeTaskWrapper`).
- `AesRequest` => `AesTask` to differ from what we have in id-worker, otherwise compiler would complain when building parachain-api
- The old `NativeCall` and `NativeQuery` are merged - they seem to be handled the same way. The `get_account_store` query is deleted temporarily
- `CallResponse` and `QueryResponse` are merged into `NativeTaskOk`
- use camel cases for `NativeTask` enum, this is rust standard

`omni-client-sdk-test` is disabled for now - @wli-pro @0xverin could you please adjust sdk (based on the changes in this PR) after this PR gets merged?


common/primitives/core/src/omni/account.rs
```diff
@@ -21,6 +21,7 @@ use sp_runtime::RuntimeDebug;
 
 #[derive(Encode, Decode, TypeInfo, Clone, PartialEq, Eq, RuntimeDebug)]
 pub enum OmniAccountAuthType {
+    NoAuth,
```
kziemianek: Can it be wrapped with `Option` instead of adding `None` like variant ?

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -160,15 +159,428 @@ async fn handle_native_task<
...
+					log::error!("Failed to get account nonce: {:?}", e);
+					let response = NativeTaskResponse::Err(NativeTaskError::InternalError);
+					if response_sender.send(response.encode()).is_err() {
+						log::error!("Failed to send response");
+					}
+					return;
+				},
+			};
+			// Increment nonce for the next transaction
+			let tx = ctx.transaction_signer.sign(auth_token_requested_call, Some(nonce + 1)).await;
```
silva-fj, line 227: I don't think we need to increment the nonce here, it can be just `None`. We increment the nonce in the `RequestIntent` case because we submit two extrinsics to the chain, one to track the requested intent and the other to dispatch it (if it's native call) or track the result (if it was on another chain) and the second would fail with the error in the comment above. 

Here we submit one and return the token in the response

tee-worker/omni-executor/native-task-handler/src/types.rs
```diff
@@ -4,7 +4,7 @@ use parity_scale_codec::{Decode, Encode};
 use std::vec::Vec;
 
 #[derive(Encode, Decode, Debug, PartialEq, Eq)]
-pub enum NativeOperationOk {
+pub enum NativeTaskOk {
 	CallResponse(CallResponse),
 	QueryResponse(QueryResponse),
```
silva-fj: Should we merge these as well? `NativeOperationOk` can use the variants of `CallResponse ` directly`. Wdyt?

kziemianek (approved):

common/primitives/core/src/omni/account.rs
Kailai-Wang: That's feasible too - @silva-fj do you definitely need a type in the parachain callbacks? Can it be `None`?

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: It's from @wli-pro 's change : )

https://github.com/litentry/heima/pull/3311/files#diff-09ff3956fe238deae5e520798cbb1baee343ddddc01e96c436596d0a48c08472

I also left a comment about inconsistency there

common/primitives/core/src/omni/account.rs
silva-fj: If this is something we don't want to track it can be `None` yes

tee-worker/omni-executor/native-task-handler/src/types.rs
Kailai-Wang: Ah yes - let me change that

silva-fj (approved):

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: I'll create an issue to review these

PR #3321, opened 2025-03-30
Fix dev ci failures

0xverin:
The libs in @polkadot/api are no longer exported from parachain-api, we should directly exported from @polkadot/api.


Kailai-Wang:
For omni-sdk test, maybe we should wait until https://github.com/litentry/heima/pull/3319 is merged (or you could do adjustments based on that feature branch)

silva-fj (approved):

0xverin:
omni client-sdk ci failure will fix in this [PR](https://github.com/litentry/heima/pull/3323)

PR #3325, opened 2025-03-31
Keep track of the omni executor tx signer nonce in memory

silva-fj:
This pull request includes some improvements focusing on refactoring the transaction signing logic, primarily updating the nonce handling in transaction signing.


tee-worker/omni-executor/parentchain/signer/src/tx_signer.rs
```diff
@@ -86,28 +70,36 @@ impl<
...
-			.map_err(|e| {
-				error!("Could not unseal key: {:?}", e);
-			})
-			.unwrap();
-		Keypair::from_secret_key(secret_key_bytes)
-			.map_err(|e| {
-				error!("Could not create secret key: {:?}", e);
-			})
-			.unwrap()
+	pub async fn update_nonce(&self) {
```
kziemianek, line 94: If instance of `TxSigner` can be shared between threads and we have both in-memory nonce and method to reset it based on onchain nonce wouldn't it cause troubles when multiple threads shares it?

T1 sign (nonce 1)
T2 update (nonce 0)
T1 sign (nonce 1)

silva-fj: We need a mechanism to get the current on-chain nonce if something wrong happens, otherwise the worker will never be able to send any transactions (if the nonce gets outdated for example). We could improve rpc client to add a retry mechanism and only update the nonce when needed, but this can be done in a separate PR

kziemianek (approved):

PR #3331, opened 2025-04-02
Adding methods for trading to PumpxApi

silva-fj:
This PR introduces two new methods in the `PumpxApi` for trading as well as some type updates and small refactorings


kziemianek (approved):
Looks fine. Why we need language header?

silva-fj:
> Looks fine. Why we need language header?

Some (pumpx) NativeTasks will require to return the response we get from the pumpx api and the FE should be able to set the language. It's a requirement on their side

PR #3333, opened 2025-04-02
Skeleton for cleartext RPC

Kailai-Wang:
### Context

Define a dedicated RPC for each request type (currently `omni_requestEmailVerificationCode`, `pumpx_requestJwt` and `pumpx_exportWallet`) and use clear-text JSON'ed params:
- the `verify_auth` is marked as `async` - I removed the extra thread spawning
- the raw `NativeTask` enum is not touched to reuse code when possible, `omni_submitNativeTask` is kept too (but shouldn't be used by pumpx)
- omni-sdk test is disabled

TODOs after this PR is merged:

**1/** returned result, I think it's better to use JSON'ed result too, like:
```
{
  "jsonrpc": "2.0",
  "result": {
    "custom1": "Hello",
    "custom2": [1, 2, 3],
    "custom3": { "nested": true }
  },
  "id": 1
}
```
current it's an opaque hex string.
@kziemianek @silva-fj could you adjust the result struct in your impl (`pumpx_exportWallet`, `pumpx_requestJwt` respectively) - I didn't do it as I'm not as familiar with the result that should be returned

**2/** update apifox: now it makes more sense to document it in apifox.  @kziemianek @silva-fj Ideally you could do it after updating the result struct, otherwise I can do it too. You could refer to the "_Request email verification code_" page.

**3/** (less urgent): I think we don't necessarily need to employ a local thread-pool, jsonrpsee handles it async already. Unless we have CPU-heavy task => needs some discussion

**4/** @wli-pro @0xverin could you think about how to test these APIs? You don't have to delete the old `NativeTask` construction as it's still kept in rust. For `pumpx_exportWallet` we probably need some example code about how to construct the `key` field - but I assume we have it somwhere in the sdk already


tee-worker/omni-executor/rpc-server/src/methods/pumpx/export_wallet.rs
```diff
@@ -0,0 +1,98 @@
...
+					ErrorObject::owned::<()>(
+						DECRYPT_REQUEST_FAILED_CODE,
+						"Shielded value decryption failed",
+						None,
+					)
+				})?;
+			let aes_key: Aes256Key = aes_key.try_into().map_err(|_| {
+				ErrorObject::owned::<()>(
+					DECRYPT_REQUEST_FAILED_CODE,
+					"Shielded value decryption failed",
```
kziemianek: It's already unshielded - should we change error/reason?

tee-worker/omni-executor/rpc-server/src/methods/omni/request_email_verification_code.rs
```diff
@@ -10,6 +10,11 @@ use jsonrpsee::{
 };
 use parentchain_rpc_client::{SubstrateRpcClient, SubstrateRpcClientFactory};
 
+#[derive(Debug, Serialize, Deserialize)]
```
silva-fj: We only need `Deserialize` in this case

tee-worker/omni-executor/rpc-server/src/methods/pumpx/export_wallet.rs
```diff
@@ -0,0 +1,98 @@
...
+			}
+
+			let (response_sender, response_receiver) = oneshot::channel();
+
+			if ctx.native_task_sender.send((wrapper, response_sender)).await.is_err() {
+				log::error!("Failed to send request to native call executor");
+				return Err(ErrorCode::InternalError.into());
+			}
+			match response_receiver.await {
+				// TODO: we should use json field for better readability, either a defined structure, or serde_json::Value
```
kziemianek, line 81: Imho it's fine to return hex encoded string. What would json object contain ?

silva-fj (approved):

tee-worker/omni-executor/rpc-server/src/methods/pumpx/export_wallet.rs
kziemianek: Only in case when method result needs to be encrypted of course.

Kailai-Wang: Like deserialized JSON'ed `AesOutput`? Something like
```
"result": {
  "key" : {
   "ciphertext": "542b...",
   "aad": "35c3...",
   "nonce": "593d..."
  }
}
```

Otherwise it would be a (single) serialized byte array (not to mention we again use scale encode here : ))

PR #3334, opened 2025-04-03
fix: support aes encrypted response from omni-executor in omni-client…

wli-pro:
…-sdk

### Labels

### Testing Evidences


0xverin (approved):
Looks good. I published a test package in this pr, and tested on F/E,  but it returned an error when I request pumx JWT:
image
This may be an issue related to the PumpX API.

PR #3335, opened 2025-04-03
Refactoring JWT issuance and verification

silva-fj:
The main changes in this PR are related to how we handle JWTs, here are the most important changes:

- Now we will use a timestamp instead of block number for the expiration.
- We'll use the OmniAccount (derived from the sender identity) hex string as the token subject. 
- As we don't need to query the current block number from the chain anymore, the `RpcContext` have been simplified, this has also simplified the rpc-server related code.
- We return a json object for the request jwt response instead of hex encoded string
- A language param has been added to `pumpx_requestJwt` rpc method.


Kailai-Wang (approved):
Looks good!

PR #3336, opened 2025-04-03
Integrate verify google code endpoint the PumpxApi

silva-fj:
This PR adds support for pumpx's verify_google_code endpoint and integrates it into the login flow.

### DIFF
--- a/tee-worker/omni-executor/native-task-handler/src/lib.rs
+++ b/tee-worker/omni-executor/native-task-handler/src/lib.rs
@@ -537,7 +537,7 @@ async fn handle_native_task<
 			let tx = ctx.transaction_signer.sign(dispatch_as_omni_account_call, None).await;
 			(response_sender, tx)
 		},
-		NativeTask::PumpxRequestJwt(sender, invite_code, google_code, language) => {
+		NativeTask::PumpxRequestJwt(sender, invite_code, maybe_google_code, language) => {
 			let email = match sender {
 				Identity::Email(ref identity_string) => {
 					let Ok(email) = std::str::from_utf8(identity_string.inner_ref()) else {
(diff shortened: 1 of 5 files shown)


tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -578,13 +578,43 @@ async fn handle_native_task<
...
+			}
+
 			let Ok(user_connect_response) = ctx
 				.pumpx_api
-				.connect_user(&access_token, email.clone(), invite_code, google_code, language)
+				.connect_user(
+					&access_token,
+					email.clone(),
+					invite_code,
+					maybe_google_code,
```
Kailai-Wang, line 588: Ah I just saw `maybe_google_code` is passed into backend here too - I think it still needs a bit of clarification if our previous local verification is redundant or not 😂

Kailai-Wang: but let's leave as it is right now

tee-worker/omni-executor/pumpx/src/pumpx_api/types.rs
```diff
@@ -143,3 +145,14 @@ pub struct NewLimitOrder {
 pub struct LimitOrder {
 	pub order_id: String,
 }
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
```
Kailai-Wang: Hmm should it be just `google_code`? (snake_case)

tee-worker/omni-executor/pumpx/src/pumpx_api/mod.rs
```diff
@@ -66,6 +66,34 @@ impl PumpxApi {
...
+	pub async fn verify_google_code(
+		&self,
+		access_token: &str,
+		google_code: String,
+	) -> Result<VerifyGoogleCodeResponse, Error> {
+		let endpoint = format!("{}/v3/account/verify_google_code", self.base_url);
+		let response = self
+			.http_client
+			.post(&endpoint)
+			.bearer_auth(access_token)
```
Kailai-Wang, line 80: What about `X-Language`? API doc says it's mandatory field

Kailai-Wang (commented):
Thanks!

tee-worker/omni-executor/pumpx/src/pumpx_api/mod.rs
silva-fj: We have a default header in the http_client config. but let me pass it as an argument as well

tee-worker/omni-executor/pumpx/src/pumpx_api/types.rs
silva-fj: Oh, yes. They use camelCase in the other endpoints. I'll double check all of them

Kailai-Wang: I also noticed that - they seem to mixing up camel and snake cases

Kailai-Wang (approved):

PR #3337, opened 2025-04-04
Write both jwts to storage

Kailai-Wang:
### Context

As topic


tee-worker/omni-executor/executor-storage/src/pumpx_jwt.rs
```diff
@@ -0,0 +1,49 @@
...
+		Self { db }
+	}
+}
+
+impl Storage<PumpxJwtStorageKey, String> for PumpxJwtStorage {
+	fn get(&self, key: &PumpxJwtStorageKey) -> Option<String> {
+		match self.db.get(storage_key(STORAGE_NAME, &key.encode())) {
+			Ok(Some(value)) => String::decode(&mut &value[..]).ok(),
+			_ => {
+				log::error!("Error getting pumpx_auth_token from storage");
```
silva-fj: Should we update the log messages as well?

silva-fj (approved):

PR #3338, opened 2025-04-05
Verify google auth code in ExportWallet

Kailai-Wang:
### Context

As topic

It also tries to adjust the exportWallet rpc response to be more readable.

### DIFF
--- a/tee-worker/omni-executor/executor-crypto/src/aes256.rs
+++ b/tee-worker/omni-executor/executor-crypto/src/aes256.rs
@@ -7,6 +7,8 @@ use ring::{
 	},
 	error::Unspecified,
 };
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
 use std::vec::Vec;
 
 // we use 256-bit AES-GCM as request enc/dec key
@@ -19,10 +21,14 @@ pub type Aes256KeyNonce = [u8; NONCE_LEN];
 // metadata that is required for decryption
 //
 // by default a postfix tag is used => last 16 bytes of ciphertext is MAC tag
-#[derive(Debug, Default, Clone, Eq, PartialEq, Encode, Decode)]
+#[serde_as]
+#[derive(Debug, Default, Clone, Eq, PartialEq, Encode, Decode, Serialize, Deserialize)]
 pub struct AesOutput {
+	#[serde_as(as = "serde_with::hex::Hex")]
 	pub ciphertext: Vec<u8>,
+	#[serde_as(as = "serde_with::hex::Hex")]
 	pub aad: Vec<u8>,
+	#[serde_as(as = "serde_with::hex::Hex")]
 	pub nonce: Aes256KeyNonce, // IV
 }
 
--- a/tee-worker/omni-executor/native-task-handler/src/lib.rs
+++ b/tee-worker/omni-executor/native-task-handler/src/lib.rs
@@ -643,17 +643,53 @@ async fn handle_native_task<
 		},
 		NativeTask::PumpxExportWallet(
 			sender,
-			_maybe_google_code,
+			maybe_google_code,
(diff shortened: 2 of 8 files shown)


BillyWooo (approved):
LGTM. Let's merge it.

tee-worker/omni-executor/executor-crypto/src/aes256.rs
```diff
@@ -19,10 +21,14 @@ pub type Aes256KeyNonce = [u8; NONCE_LEN];
 // metadata that is required for decryption
 //
 // by default a postfix tag is used => last 16 bytes of ciphertext is MAC tag
-#[derive(Debug, Default, Clone, Eq, PartialEq, Encode, Decode)]
+#[serde_as]
```
BillyWooo, line 24: One suggestion: maybe we should apply this "serde_as" to quite some log output, such as pubkeys, etc.

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -643,17 +643,53 @@ async fn handle_native_task<
...
+					let response = NativeTaskResponse::Err(NativeTaskError::InternalError);
+					if response_sender.send(response.encode()).is_err() {
+						log::error!("Failed to send response");
+					}
+					return;
+				};
+
+				let verify_result = ctx
+					.pumpx_api
+					.verify_google_code(&access_token, google_code.to_string(), None)
```
BillyWooo, line 666: Should we use `id_token` here, instead of `access_token` ? Looks like `id_token` has higher priority.

BillyWooo (changes requested):

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: We should use access_token here

tee-worker/omni-executor/executor-crypto/src/aes256.rs
Kailai-Wang: It's for serialisation, not normal output - if we have such requirement we should change it directly in the stdout printing

BillyWooo (approved):

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -643,17 +643,53 @@ async fn handle_native_task<
...
 			pumpx_wallet_index,
 			expected_wallet_address,
 		) => {
+			if let Some(ref google_code) = maybe_google_code {
+				let storage = PumpxJwtStorage::new(ctx.storage_db.clone());
+				let Some(access_token) =
+					storage.get(&(sender.to_omni_account(), AUTH_TOKEN_ACCESS_TYPE))
+				else {
+					let response = NativeTaskResponse::Err(NativeTaskError::InternalError);
+					if response_sender.send(response.encode()).is_err() {
```
kziemianek, line 658: Can we log/identify error reason somehow so later if this happens we know why?

kziemianek (approved):

silva-fj (approved):

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: Do you mean more detailed errors than `InternalError`?

kziemianek: It doesn't need to be propagated to client - it can be logged.
Just in case we experience this we can know the reason why request has failed - just by looking into logs for example.

Kailai-Wang: Sure I'll add them

PR #3339, opened 2025-04-07
feat: avoid error while access process.env.*

wli-pro:
support toggle debug log for Enclave, support using user created Enclave instance, add more docs

### Labels

### Testing Evidences


0xverin (commented):
~~Is this PR addressing this [[issue](https://linear.app/litentry/issue/P-1413/updaterewrite-the-environment-variables-for-the-client-sdk)](https://linear.app/litentry/issue/P-1413/updaterewrite-the-environment-variables-for-the-client-sdk)? Or is it just addressing the current access error??~~

~~The actual purpose of the issue is to instantiate a corresponding `enclave` service through external parameters.~~
~~e.g.:~~

```
import { Enclave } from '@heima-network/client-sdk';

const enclave = new Enclave('http://localhost:2100');

export async function requestEmailVerificationCode(args: { email: string },enclave:Enclave): Promise<void> {
  const { email } = args;
  assert(email.length > 0, 'Email is required');

  // send the request to the Enclave
  const rpcRequest: JsonRpcRequest = {
    jsonrpc: '2.0',
    method: 'omni_requestEmailVerificationCode',
    params: [email],
  };

  await enclave.send(rpcRequest);
}
```

~~However, it seems that the config is still needs to be set up manually, but it has indeed resolved the access env error👍.~~

~~Imagine if the F/E devs needs to temporarily test the `hemai-staging` environment (we haven't released the staging npm SDK), then it requires modifying the configuration once and then releasing it again. If other issues arise, it may also require releasing the SDK for all environments, which still does not make the SDK convenient.~~

0xverin (approved):

PR #3340, opened 2025-04-07
On-chain intent lifecycle and storage skeleton

Kailai-Wang:
### Context

As topic - I think in general it's a bit verbose than thought, but IMO it's fine to start from it and see what we actually need in practice.


parachain/pallets/omni-account/src/lib.rs
```diff
@@ -587,6 +669,20 @@ pub mod pallet {
...
+			who: T::AccountId,
+			intent_id: IntentId,
+			intent: Intent,
+		) -> DispatchResult {
+			ensure!(intent_id > Self::accepted_intent_ids(&who), Error::<T>::IntentIdTooSmall);
+			ensure!(!Intents::<T>::contains_key(&who, intent_id), Error::<T>::IntentAlreadyExists);
+			// the continuity should have been checked in the worker already
+			AcceptedIntentIds::<T>::insert(&who, intent_id);
+			Intents::<T>::insert(&who, intent_id, intent.clone());
+			Self::deposit_event(Event::IntentAccepted { who, intent_id, intent });
```
kziemianek, line 682: Does this mean worker should wait for this call finalization before intent execution ?

common/primitives/core/src/omni/intent.rs
```diff
@@ -4,23 +4,30 @@ use scale_info::TypeInfo;
...
-    #[codec(index = 4)]
     TransferNative(TransferNative),
-    #[codec(index = 5)]
+    #[codec(index = 4)]
     TransferSolana(TransferSolana),
+    #[codec(index = 5)]
+    Swap(
+        SwapOrder,
+        Option<CrossChainSwapProvider>,
+        SingleChainSwapProvider,
```
silva-fj, line 29: Should we merge these two? We could use an enum like:

```rust
enum SwapProvider {
  CrossChain(CrossChainSwapProvider)
  SingleChain(SingleChainSwapProvider) 
}
```
Will the FE (or user) choose the provider?

parachain/pallets/omni-account/src/lib.rs
Kailai-Wang: No - worker has a local copy (storage) of these IDs (which should always be updated first)

common/primitives/core/src/omni/intent.rs
```diff
@@ -4,23 +4,30 @@ use scale_info::TypeInfo;
...
 use sp_runtime::{traits::ConstU32, BoundedVec};
 
-pub const CALL_ETHEREUM_INPUT_LEN: u32 = 10 * 1024;
-pub const MAX_REMARK_LEN: u32 = u32::max_value();
+pub type CallEthereumInputLen = ConstU32<{ 10 * 1024 }>;
+pub type RemarkLen = ConstU32<{ u32::max_value() }>;
+pub type IntentStringLen = ConstU32<128>;
 
-#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq, TypeInfo)]
+pub type IntentId = u32;
```
silva-fj, line 11: Do you think we will have other types of intents that will require `intentId` ? We could use `SwapOrderId` (or something similar) and make it part of the `Swap` intent

common/primitives/core/src/omni/intent.rs
Kailai-Wang: That's my imagination (in the future) - or we have some preset "smart" strategy that does it (which can be done either in F/E or worker)

common/primitives/core/src/omni/intent.rs
Kailai-Wang: We have - theoretically all "intents" should have an `IntentId` 🤔  like ethereum contract call - as long as it needs to be tracked and order needs to be preserved.

silva-fj: Well, in the ERC-7683 specs there is no intent id. But I can see the usefulness, depending on the usage

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -194,14 +215,44 @@ pub mod pallet {
...
+	#[derive(Clone, Debug, PartialEq, Encode, Decode, TypeInfo)]
+	pub enum SwapInProcessDetail {
+		SourceChainBalanceDeducted { asset: ChainAsset, amount: u64 },
+		DestChainBalanceAdded { asset: ChainAsset, amount: u64 },
+		SingleChainSwapSubmitted { tx_hash: Vec<u8> },
+	}
+
+	#[derive(Clone, Debug, PartialEq, Encode, Decode, TypeInfo)]
+	pub enum SwapCompletedDetail {
+		// TODO
```
kziemianek: What else we need here ?

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -387,10 +439,13 @@ pub mod pallet {
...
-		pub fn request_intent(origin: OriginFor<T>, intent: Intent) -> DispatchResult {
+		pub fn request_intent(
+			origin: OriginFor<T>,
+			intent_id: IntentId,
+			intent: Intent,
+		) -> DispatchResult {
 			let who = T::OmniAccountOrigin::ensure_origin(origin)?;
-			Self::deposit_event(Event::IntentRequested { who, intent });
-			Ok(())
+			Self::do_accept_intent(who, intent_id, intent)
```
kziemianek: Can cross chain swap be accepted without balance checks?

silva-fj: We could add back the `IntentRequested` event (or just get rid of this call) and only call `accept_intent` after sanity check

common/primitives/core/src/omni/intent.rs
```diff
@@ -58,10 +63,53 @@ pub enum HeimaMultiAddress {
...
+}
+
+// basically copied from pumpx API
+#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
+pub struct PumpxConfig {
+    pub order_type: PumpxOrderType,
+    pub swap_type: u32, // 1：buy 2：sell
+    pub chain_id: u32,  // to align with pumpx: sol:10000 eth:1 bsc:56 base:8453
+    pub token_ca: BoundedVec<u8, IntentStringLen>,
+    pub doulbe_out: bool,
```
silva-fj:
```suggestion
    pub double_out: bool,
```

parachain/pallets/omni-account/src/lib.rs
Kailai-Wang: Like if there's any returned result if successful (e.g. actually received amount)

common/primitives/core/src/omni/intent.rs
```diff
@@ -58,10 +63,53 @@ pub enum HeimaMultiAddress {
...
+    pub is_one_click: bool,
+    pub is_anti_mev: bool,
+    pub is_auto_slippage: bool,
+    pub gas_type: u32, // 1: slow, 2: medium, 3: fast
+    pub slippage: u32,
```
silva-fj, line 101: `is_anti_mev`, `is_auto_slippage`, `gas_type` and `slippage` come from `get_user_trade_info` (PumpxApi)

silva-fj: In `create_market_order_unsigned_tx` there is also a `requestId`, should this come from the FE as well ?

Kailai-Wang: Yes @silva-fj 

- `requestId` is actually the intent-id
- others are from `get_user_trade_info` call, we need to select the right `gas_type` based on the dest-chain

parachain/pallets/omni-account/src/lib.rs
Kailai-Wang: The initial idea was to make it possible to submit intent directly onto chain (for non-pumpx cases), but it could cause chaos in this case.

I'll add back the IntentRequested event as a placeholder, and make some comments

common/primitives/core/src/omni/intent.rs
silva-fj: If a user wants to make a cross chain swap, do they need to send `SingleChainSwapProvider` as well ?

Kailai-Wang: Yes - it's not an `Option` right now, we always need a `SingleChainSwapProvider` to process swaps on the dest-chain

Shall I rename it to `DestChainSwapProvider`?

parachain/pallets/omni-account/src/lib.rs
Kailai-Wang: For now I'll just use generic `Success/Failure` without extra info

common/primitives/core/src/omni/intent.rs
silva-fj: oh I see, I understand now. It sounds good

silva-fj (approved):

parachain/pallets/omni-account/src/lib.rs
```diff
@@ -194,14 +215,42 @@ pub mod pallet {
...
-		/// Intent is requested
-		IntentRequested { who: T::AccountId, intent: Intent },
-		/// Intent is executed
-		IntentExecuted { who: T::AccountId, intent: Intent, result: IntentExecutionResult },
 		/// Member permission set
 		AccountPermissionsSet { who: T::AccountId, member_account_hash: H256 },
 		/// An auth token is requested
 		AuthTokenRequested { who: T::AccountId, expires_at: i64 },
+		/// Intent is requested by some user
+		IntentRequested { who: T::AccountId, intent: Intent },
```
kziemianek, line 223: We might miss 'IntentId' here for the sake of consistency

kziemianek (approved):

parachain/pallets/omni-account/src/lib.rs
Kailai-Wang: Yes - I have comment in the extrinsic definition saying this is unused for now. It probably needs a redesign/refactoring anyway if we start to use it

PR #3341, opened 2025-04-07
P 1419 implement pumpx_addWallet rpc

BillyWooo:
Add `addWallet` rpc call.


tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
```diff
@@ -0,0 +1,57 @@
...
+
+			let (response_sender, response_receiver) = oneshot::channel();
+
+			if ctx.native_task_sender.send((wrapper, response_sender)).await.is_err() {
+				log::error!("Failed to send request to native call executor");
+				return Err(ErrorCode::InternalError.into());
+			}
+
+			match response_receiver.await {
+				Ok(response) => Ok::<String, ErrorObject>(hex_encode(response.as_slice())),
```
silva-fj: The pumpx rpc methods should return plain text (json)

tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
```diff
@@ -0,0 +1,56 @@
...
+
+			let (response_sender, response_receiver) = oneshot::channel();
+
+			if ctx.native_task_sender.send((wrapper, response_sender)).await.is_err() {
+				log::error!("Failed to send request to native call executor");
+				return Err(ErrorCode::InternalError.into());
+			}
+
+			match response_receiver.await {
+				Ok(_response) => Ok::<(), ErrorObject>(()),
```
Kailai-Wang: Ideally we need to forward the pump-api's response to F/E - you can refer to user_connect as example

Kailai-Wang (commented):
Looks good - just the RPC response needs adjustment

Kailai-Wang:
Btw a small thing: could you rename the PR title to be more readable? : )

tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
BillyWooo: Actually there is 'null' returned from pumpx-api. I have discussed with them. I can update later.

felixfaisal (approved):
LGTM, The implementation looks good, 
However the response is empty, are we looking to provide a 200 result without any data?

Kailai-Wang (approved):

PR #3346, opened 2025-04-08
fix: add scripts for interacting with accounting-contract

felixfaisal:
This adds a foundry script and a wrapper bash script for interacting with the accounting-contract, 
I've tested on testnet looks good 
I've set default values that can be used for anvil for testing locally the entire flow of cross-chain


Kailai-Wang (approved):
Thanks!

PR #3350, opened 2025-04-08
Adding pumpx_submitSwapOrder rpc method

silva-fj:
This is the initial setup for the `pumpx_submitSwapOrder` rpc method, there are a couple of TODOs than can be done in the next iteration, once we start integrating all the peaces for the intent processing

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
+++ b/tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
@@ -1,6 +1,6 @@
 use crate::{
-	error_code::*, oneshot, server::RpcContext, verify_auth::verify_auth, Decode, Deserialize,
-	ErrorCode,
+	error_code::*, oneshot, server::RpcContext, verify_auth::verify_auth_token_authentication,
+	Decode, Deserialize, ErrorCode,
 };
 use executor_core::native_task::*;
 use executor_primitives::{BoundedVec, OmniAuth};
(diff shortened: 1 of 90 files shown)


kziemianek (approved):

tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	fn try_get_chain_asset(chain_id: u32, token_ca: Option<String>) -> Result<ChainAsset, ()> {
+		match chain_id {
+			SOLANA_CHAIN_ID => {
+				let solana_token = match token_ca {
+					None => SolanaToken::Native,
+					Some(ref s) if s.is_empty() => SolanaToken::Native,
+					Some(ref mint_address) => {
+						let mint_address: Address32 = mint_address.as_str().try_into()?;
```
Kailai-Wang, line 64: From their API example, they seem to pass in solana address format (base58), like
```
  "token_ca": "2qEHjDLDLbuBgRYvsxhc5D6uDWAivNFZGan56P1tpump",
```

So we need extra steps to convert it to Address32

Kailai-Wang (commented):
Thank you, I only one comment about token_ca format

tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
```diff
@@ -0,0 +1,214 @@
...
+use pumpx::types::SwapType;
+
+// TODO: move this to a central place
+const SOLANA_CHAIN_ID: u32 = 10000;
+const ETHEREUM_CHAIN_ID: u32 = 1;
+const BSC_CHAIN_ID: u32 = 56;
+const BASE_CHAIN_ID: u32 = 8453;
+
+#[derive(Debug, Deserialize)]
+pub struct SubmitSwapOrderParams {
```
Kailai-Wang, line 26: It doesn't have the `usd_worth` field

PR #3351, opened 2025-04-08
P 1420 implement pumpx_transferWithdraw rpc

BillyWooo:
as title: add rpc call `pumpx_transferWithdraw`
because of the merge history, content of #3341 also be here. So it's better to merge that PR first. But there is no conflict if only merge this one, cancel #3341 .


tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -733,6 +733,284 @@ async fn handle_native_task<
...
+			let Some(chain_type) = ChainType::from_pumpx_chain_id(chain_id) else {
+				log::error!("Failed to map pumpx chain_id {}", chain_id);
+				let response = NativeTaskResponse::Err(NativeTaskError::InternalError);
+				if response_sender.send(response.encode()).is_err() {
+					log::error!("Failed to send response");
+				}
+				return;
+			};
+
+			// 4. Use the pumpx_signer_client to sign each tx_data entry in parallel
```
kziemianek: `pumpx-signer` will support signing multiple transactions within one request - could you adjust this ?

BillyWooo: Yes. Just saw the PR on signer. Will update once new rpc call is ready.

kziemianek: It's here https://github.com/litentry/heima/pull/3353

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -733,6 +733,284 @@ async fn handle_native_task<
...
+					));
+					if response_sender.send(response.encode()).is_err() {
+						log::error!("Failed to send response");
+					}
+					return;
+				}
+			}
+
+			// 2. Verify we have a valid Pumpx "access" token for the user
+			let storage = PumpxJwtStorage::new(ctx.storage_db.clone());
```
silva-fj: This should probably be step 1, so we don't get the access_token from storage twice.

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -733,6 +733,284 @@ async fn handle_native_task<
...
+				let response = NativeTaskResponse::Err(NativeTaskError::InternalError);
+				if response_sender.send(response.encode()).is_err() {
+					log::error!("Failed to send response");
+				}
+				return;
+			};
+
+			// 4. Use the pumpx_signer_client to sign each tx_data entry in parallel
+			let mut sign_futures = Vec::new();
+			for tx in tx_data {
```
silva-fj, line 882: we should probably extract this logic to some service we can reuse as this will be used in multiple places. But it can be done in a separate PR

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -733,6 +733,284 @@ async fn handle_native_task<
...
+				signed_tx_data.push(signed_tx);
+			}
+
+			// 5. Send the signed tx to the Pumpx backend
+			let send_res = match ctx
+				.pumpx_api
+				.send_transfer_tx(&access_token, transfer_id, chain_id, signed_tx_data, language)
+				.await
+			{
+				Ok(res) => res,
```
silva-fj: You can send the response and return from here

silva-fj (commented):
it seems there are some changes from your other PR as well

tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(Debug, Deserialize)]
+pub struct AddWalletParams {
+	pub user_email: String,
+	pub auth_token: String,
+}
+
+impl From<AddWalletParams> for NativeTaskWrapper<NativeTask> {
+	fn from(p: AddWalletParams) -> Self {
+		Self {
+			task: NativeTask::PumpxAddWallet(Identity::from_web2_account(
```
kziemianek, line 19: How do you ensure both `user_email` and `auth_token` relates to same `Identity` ?

If you have a JWT token can I impersonate you by specifying your email here and using my JWT token ?

tee-worker/omni-executor/native-task-handler/src/lib.rs
BillyWooo: refactored to use "request_signatures"

tee-worker/omni-executor/rpc-server/src/error_code.rs
```diff
@@ -15,11 +15,17 @@ const AUTH_TOKEN_CREATION_FAILED_CODE: i32 = -32007;
 const INVALID_MEMBER_IDENTITY_CODE: i32 = -32008;
 const VALIDATION_DATA_VERIFICATION_FAILED_CODE: i32 = -32009;
 const UNSUPPORTED_IDENTITY_TYPE_CODE: i32 = -32010;
-const PUMPX_API_GOOGLE_CODE_VERIFICATION_FAILED_CODE: i32 = -32011;
-const PUMPX_API_USER_CONNECTION_FAILED_CODE: i32 = -32012;
-const PUMPX_API_ERROR_CODE: i32 = -32013;
-const PUMPX_API_ADD_WALLET_FAILED_CODE: i32 = -32014;
-const INTENT_NONCE_MISMATCH_ERROR_CODE: i32 = -32015;
+
+const PUMPX_API_GOOGLE_CODE_VERIFICATION_FAILED_CODE: i32 = -32031;
```
kziemianek, line 19: In my opinion this leaks implementation details -> does it make sense to make api consumer to be aware of this ? Do we expect it will conditionally make any actions based on these error codes ? If not why bother.

Kailai-Wang: We verify the result and error out if it's false - at the same time we report back the error code for it.

IMO F/E needs such information - they probabaly display it to user

tee-worker/omni-executor/rpc-server/src/methods/pumpx/transfer_widthdraw.rs
```diff
@@ -0,0 +1,101 @@
...
+	pub token_ca: String,
+	pub amount: String,
+	pub email_code: String,
+	pub google_code: MaybeGoogleCode,
+	pub lang: Option<String>,
+}
+
+#[derive(Serialize, Clone)]
+pub struct TransferWithdrawResponse {
+	pub tx_hash: Option<Vec<String>>,
```
Kailai-Wang: Are these raw responses from backend?

Shall we:
1. put it under field `backend_respones:`
2. use `ApiResponse<>` for it? Take `UserConnectResponse` and `AddWalletResponse` for example

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -791,26 +775,204 @@ async fn handle_native_task<
...
+					);
+					return;
+				},
+			};
+		},
 	};
-	let report = match rpc_client.submit_and_watch_tx_until(&tx, XtStatus::Finalized).await {
-		Ok(report) => report,
+
+	match rpc_client.submit_and_watch_tx_until(&tx, XtStatus::Finalized).await {
```
Kailai-Wang, line 944: Why do we need this?
We don't have to report anything to heima, and `tx` isn't a substrate tx I assume?

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -791,26 +775,204 @@ async fn handle_native_task<
...
+					send_error(
+						format!("Failed to sign transfer tx: {:?}", e),
+						response_sender,
+						NativeTaskError::PumpxSignerError(PumpxSignerError::RequestSignatureFailed),
+					);
+					return;
+				},
+			};
+
+			let signed_tx_data: Vec<String> = signatures.into_iter().map(hex::encode).collect();
```
Kailai-Wang: Do they expect a `0x`-prefixed string? I assume yes

there's some `to_hex` method

tee-worker/omni-executor/rpc-server/src/error_code.rs
BillyWooo: Thank you. good point. I will take this suggestion and remove it.

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -791,26 +775,204 @@ async fn handle_native_task<
...
+				send_error(
+					"Failed to get access_token within NativeTask::PumpxTransferWidthdraw"
+						.to_string(),
+					response_sender,
+					NativeTaskError::InternalError,
+				);
+				return;
+			};
+
+			// 2. Verify google code (if provided).
```
Kailai-Wang, line 802: I think here might be wrong - (so is exportWallet impl @kziemianek)

I don't know if F/E knows if a user has binded GA or not (or only B/E knows it).

If we put it into `Option<>`, it means F/E has to differentiate:
- `None` value => user doesn't have a GA, in this case we'll skip verification
- Empty value / `Some("")` => user does have a GA, but didn't give any input => we need to call B/E to verify it

I'm inclined to F/E not knowing this info, if that's the case - here should be an `String` instead of an `Option<String>` => and we **always** have to call B/E

We'll need to clarify it with pumpx

tee-worker/omni-executor/native-task-handler/src/lib.rs
BillyWooo: I think it is a substrate tx.
This part is just a refactor.

tee-worker/omni-executor/rpc-server/src/methods/pumpx/transfer_widthdraw.rs
BillyWooo: Yes. It was written before the `backend_response` refactor.

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: Will address it in another PR if needed (still waiting for pumpx's response)

Kailai-Wang (approved):

PR #3352, opened 2025-04-09
disable account locks

kziemianek:


Kailai-Wang (approved):

Kailai-Wang:
Thanks - could you maybe add a comment in `tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs` explaining why `always_unlocked` variant is used? (as we guarantee the dest-chain payout happens after the source chain finalisation)

PR #3354, opened 2025-04-09
Implement swap intent for pumpx

silva-fj:
This PR introduces a draft implementation of the swap intent processing for pumpx so we can start testing.

Important notes: 
- Cross chain swaps are not implemented yet. (It will be implemented in a separate PR)
- The integration has not been tested yet, there may be some things that need to be adjusted (we need to test them).
- The assets locking mechanism in the intent-executor (sanity check) is commented out as it needs adjustments. (We will need to do asset conversion as we receive float numbers from pumpx but we need amounts in the asset smallest unit.


tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -82,17 +102,21 @@ impl<
...
 		transaction_signer: Arc<ParentchainTxSigner>,
-		rpc_endpoint_registry: RpcEndpointRegistry,
+		_rpc_endpoint_registry: RpcEndpointRegistry,
 		pumpx_signer_client: Arc<SignerClient>,
+		pumpx_api: Arc<PumpxApi>,
+		storage_db: Arc<StorageDB>,
 	) -> Result<Self, ()> {
 		// there is no need for account/assets locks if we guarantee the dest-chain payout happens after the source chain finalisation
-		let account_asset_lock = AccountAssetLocks::<AlwaysUnlockedAssetsLock>::empty();
+		// let account_asset_lock = AccountAssetLocks::<AlwaysUnlockedAssetsLock>::empty();
```
Kailai-Wang, line 111: Is this just to make it compile?
From logic `AlwaysUnlockedAssetsLock` does basically nothing which can be treated as placeholder

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -176,17 +208,269 @@ impl<
...
+											"Unsupported swap type: {}",
+											pumpx_config.swap_type
+										);
+										return Err(());
+									},
+								},
+								amount_in: from_amount_string.clone(),
+								double_out: pumpx_config.double_out,
+								is_one_click: pumpx_config.is_one_click,
+								address: "todo: what's this??".to_string(),
```
Kailai-Wang: This is the wallet address of tx sender - in this case the omni-account address on this chain, which can be retrieved from signer by `omni-account` + `chain_id` + `wallet_index`

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -375,21 +368,36 @@ async fn handle_native_task<
...
 					let intent_executed_call =
 						parentchain_api_interface::tx().omni_account().intent_completed(
 							omni_account.to_subxt_type(),
-							0, // TODO
+							intent_id,
 							execution_result,
 						);
-					ctx.transaction_signer.sign(intent_executed_call).await
+					let tx = ctx.transaction_signer.sign(intent_executed_call).await;
+					if rpc_client.submit_tx(&tx).await.is_err() {
```
Kailai-Wang, line 395: It doesn't wait for any finalisation, does it? (We don't want to wait for it)

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -60,16 +77,19 @@ pub type ParentchainTxSigner = TxSigner<
 	SubxtMetadataProvider<CustomConfig>,
 >;
 
+// TODO: should we rename this to something like MultiChainIntentExecutor?
```
Kailai-Wang, line 80: Good idea - but probably in separate PR

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -176,17 +208,269 @@ impl<
...
+										return Err(());
+									},
+								},
+								double_out: pumpx_config.double_out,
+								token_cap,
+								price_usd,
+								trailing_percent: pumpx_config
+									.trailing_percent
+									.map(|v| v.to_string()),
+								address: "todo: what's this??".to_string(),
```
Kailai-Wang: same above

Kailai-Wang (commented):
The logic looks good - I have a few comments.

The `address` is the main missing part

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: This specific part no, I commented out everything related to asset lock to be able to compile as we need to refactor it a bit

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: ah I forgot about this one, I only changed it in the other order type 👍🏼

tee-worker/omni-executor/native-task-handler/src/lib.rs
silva-fj: No it does not wait, when we want to wait we use `submit_and_watch_tx_until`

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: done

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: done

tee-worker/omni-executor/executor-core/src/intent_executor.rs
```diff
@@ -29,7 +29,7 @@ pub trait IntentExecutor: Send {
 		account_id: &AccountId,
 		intent_id: IntentId,
 		intent: Intent,
-	) -> Result<(), ()>;
+	) -> Result<Option<Vec<u8>>, ()>;
```
kziemianek, line 32: Can this be generic ?

kziemianek (approved):

tee-worker/omni-executor/executor-core/src/intent_executor.rs
silva-fj: Do you mean like a trait object?

Kailai-Wang (approved):

PR #3355, opened 2025-04-10
Serialize Vec<u8> in RPC struct fields to 0x-prefixed-hex-string

Kailai-Wang:
### Context

As topic - it uses `ethers::types::Bytes` to do this.

Alternatively we could write a manual serializer/deserializer, but I think that's too verbose.

Extra struct is defined to keep the original struct SCALE-codecable as we still use it internally. E.g. `AesOutput` vs `SerdeAesOutput`

### DIFF
--- a/tee-worker/omni-executor/executor-crypto/src/aes256.rs
+++ b/tee-worker/omni-executor/executor-crypto/src/aes256.rs
@@ -1,3 +1,4 @@
+use ethers::types::Bytes;
 use parity_scale_codec::{Decode, Encode};
 use rand::Rng;
 use ring::{
@@ -8,7 +9,6 @@ use ring::{
 	error::Unspecified,
 };
 use serde::{Deserialize, Serialize};
-use serde_with::serde_as;
 use std::vec::Vec;
 
 // we use 256-bit AES-GCM as request enc/dec key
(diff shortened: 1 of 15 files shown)


tee-worker/omni-executor/executor-crypto/src/lib.rs
```diff
@@ -6,18 +6,27 @@ pub mod traits;
...
 
-	#[serde_as]
-	#[derive(Serialize, Deserialize, Debug)]
+	#[derive(Debug, Serialize, Deserialize)]
 	pub struct Rsa3072PubKey {
-		#[serde_as(as = "serde_with::hex::Hex")]
 		pub n: Vec<u8>,
-		#[serde_as(as = "serde_with::hex::Hex")]
 		pub e: Vec<u8>,
 	}
```
wli-pro, line 17: Shall we make Rsa3072PubKey to use fixed-size arrays to guarantee 3072-bit key length as below?:
```
pub struct Rsa3072PubKey {
    pub n: [u8; 384], 
    pub e: [u8; 3], 
}
```

tee-worker/omni-executor/executor-crypto/src/lib.rs
```diff
@@ -6,18 +6,27 @@ pub mod traits;
...
+
+	#[derive(Serialize, Deserialize)]
+	pub struct SerdeRsa3072PubKey {
```
wli-pro, line 20: Would it be better to rename this to SerdeRsaPubKey because the struct can handle RSA public keys of any length, not just 3072-bit?

Kailai-Wang: I mean this struct itself is a counterpart of `Rsa3072PubKey` - so if we want to rename it, we have to rename both.

But `Rsa3072PubKey` is serialised from public key of `ShieldingKey` - which is a 3072 bit rsa key pair. So actually I think it's fine

tee-worker/omni-executor/executor-crypto/src/lib.rs
```diff
@@ -6,18 +6,27 @@ pub mod traits;
...
+		pub n: Bytes,
+		pub e: Bytes,
+	}
+
+	impl From<Rsa3072PubKey> for SerdeRsa3072PubKey {
+		fn from(k: Rsa3072PubKey) -> Self {
+			Self { n: k.n.into(), e: k.e.into() }
```
felixfaisal, line 27: Why do we need the wrapper, if the `Bytes` support conversion to `Vec`, this might get cumbersome for other types we want to ensure `0x`, if only here then i guess it's fine.

felixfaisal (approved):
`Bytes` seems like a quick solution, I do have one question other than that LGTM

tee-worker/omni-executor/executor-crypto/src/aes256.rs
```diff
@@ -21,17 +21,27 @@ pub type Aes256KeyNonce = [u8; NONCE_LEN];
...
 	pub ciphertext: Vec<u8>,
-	#[serde_as(as = "serde_with::hex::Hex")]
 	pub aad: Vec<u8>,
-	#[serde_as(as = "serde_with::hex::Hex")]
 	pub nonce: Aes256KeyNonce, // IV
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+
+pub struct SerdeAesOutput {
```
kziemianek, line 33: We might end up with many `Serde*` types in the future if we take this approach that's why I prefere `serde_with` way which only tackles this cosmetic issue described in PR.

PR #3356, opened 2025-04-10
renaming param

BillyWooo:
very minor PR.


Kailai-Wang:
As discussed, we should keep as it is - F/E only needs to pass in `id_token` in every case

PR #3358, opened 2025-04-10
Add `pumpx_signLimitOrder`

kziemianek:
Adds a rpc methods for signing arbitrary bytes.

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/methods/pumpx/sign_limit_order.rs
+++ b/tee-worker/omni-executor/rpc-server/src/methods/pumpx/sign_limit_order.rs
@@ -1,0 +1,145 @@
+// Copyright 2020-2024 Trust Computing GmbH.
+// This file is part of Litentry.
+//
+// Litentry is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Litentry is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+use executor_primitives::utils::hex::FromHexPrefixed;
+use heima_primitives::Address32;
+use heima_primitives::IntentId;
+use rsa::pkcs1::DecodeRsaPrivateKey;
+use rsa::pkcs1::EncodeRsaPublicKey;
+
+use crate::error_code::get_native_task_error_code;
+use crate::error_code::AUTH_VERIFICATION_FAILED_CODE;
+use crate::server::RpcContext;
+use crate::ErrorCode;
+use executor_core::native_task::NativeTask;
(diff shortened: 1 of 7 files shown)


tee-worker/omni-executor/rpc-server/src/methods/pumpx/sign_limit_order.rs
```diff
@@ -0,0 +1,147 @@
...
+			// this validates jwt
+
+			let Ok(token) =
+				jwt::decode::<AuthTokenClaims>(&params.auth_token, public_key.as_bytes())
+			else {
+				return Err(ErrorCode::ServerError(AUTH_VERIFICATION_FAILED_CODE).into());
+			};
+			if token.sub != "access" {
+				return Err(ErrorCode::ServerError(AUTH_VERIFICATION_FAILED_CODE).into());
+			}
```
Kailai-Wang, line 83: Thank you

Shall we make a per task fn (like `expected_auth_token_type()` and add it as parameter to `verify_auth`) as it should be verified in most RPC methods?

Kailai-Wang (approved):
Thanks!

Would you update apifox once it's merged?

kziemianek:
> Thanks!
> 
> Would you update apifox once it's merged?

apifox should be already up to date 🙃

tee-worker/omni-executor/rpc-server/src/methods/pumpx/sign_limit_order.rs
BillyWooo: is it possible to reuse and keep alignment with other rpc calls?
Something like:
```
let wrapper: NativeTaskWrapper<NativeTask> = params.into();

if wrapper.task.require_auth() && verify_auth(ctx.clone(), &wrapper).await.is_err() {
return Err(ErrorCode::ServerError(AUTH_VERIFICATION_FAILED_CODE).into());
}
```

kziemianek: > is it possible to reuse and keep alignment with other rpc calls? Something like:
> 
> ```
> let wrapper: NativeTaskWrapper<NativeTask> = params.into();
> 
> if wrapper.task.require_auth() && verify_auth(ctx.clone(), &wrapper).await.is_err() {
> return Err(ErrorCode::ServerError(AUTH_VERIFICATION_FAILED_CODE).into());
> }
> ```

Yes but that would require some time for rework.

BillyWooo: Just copy/paste and change fields? can it help?
Like:
```
impl From<RequestJwtParams> for NativeTaskWrapper<NativeTask> {
	fn from(p: RequestJwtParams) -> Self {
		Self {
			task: NativeTask::PumpxRequestJwt(
				Identity::from_web2_account(p.user_email.as_str(), Web2IdentityType::Email),
				p.invite_code,
				p.google_code,
				p.language,
			),
			nonce: None,
			auth: Some(OmniAuth::Email(p.email_code)),
		}
	}
}
```

kziemianek: Nope, simple copy paste doesn't help in this case 😂

kziemianek: The main problem is that `verify_auth` function derives omni_account from sender `Identity`.
In case of JWT `type: access` `sub` field already contains omni_account.

PR #3359, opened 2025-04-10
Add token type validation to AuthToken authentication

silva-fj:
With this PR we enforce the usage of ID tokens for authentication in the worker.

### DIFF
--- a/tee-worker/omni-executor/heima/authentication/src/auth_token.rs
+++ b/tee-worker/omni-executor/heima/authentication/src/auth_token.rs
@@ -38,16 +38,20 @@ impl AuthTokenClaims {
 
 pub struct Validation {
 	pub sub: String,
+	pub typ: String,
 }
 
 impl Validation {
-	pub fn new(sub: String) -> Self {
-		Self { sub }
+	pub fn new(sub: String, typ: String) -> Self {
+		Self { sub, typ }
 	}
 
 	pub fn validate(&self, claims: &AuthTokenClaims) -> Result<(), Error> {
 		if self.sub != claims.sub {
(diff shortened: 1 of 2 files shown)


Kailai-Wang:
Thanks - the problem is not all RPC methods expect `id` token, in fact, some expect `access` token - see https://github.com/litentry/heima/pull/3358/files#diff-eebbfcea2bd8c3066d73e2081b541cce5e11de949c03ef24b3d2849488d47642

silva-fj:
> Thanks - the problem is not all RPC methods expect `id` token, in fact, some expect `access` token - see https://github.com/litentry/heima/pull/3358/files#diff-eebbfcea2bd8c3066d73e2081b541cce5e11de949c03ef24b3d2849488d47642

@Kailai-Wang That looks wrong, why would it expect access token?

Also, this part does not make sense:
```rust
			if token.sub != "access" {
				return Err(ErrorCode::ServerError(AUTH_VERIFICATION_FAILED_CODE).into());
			}
```
token.sub will never be "access"

--
Updated: So, the BE will use access token to call the worker's API.. 🤔

Kailai-Wang:
> Updated: So, the BE will use access token to call the worker's API

Yes ... that's the only token they know 😂 

```
if token.sub != "access" 
```

This was fixed - it should be `typ`

Kailai-Wang:
So such info (which task expects which type of token) should be available already on `verify_auth`, that's why I suggested an `expected_auth_token_type()` for each `NativeTask`.

The precondition is we could build `NativeTask` out of an RPC..

silva-fj:
I'm closing this PR as the token typ validation needs to be enforced/constructed at the rpc method level

PR #3363, opened 2025-04-10
Spawn task handling task concurrently

Kailai-Wang:
### Context

As topic - it should work as we have a bounded channel (buf = 1024)

### DIFF
--- a/tee-worker/omni-executor/native-task-handler/src/lib.rs
+++ b/tee-worker/omni-executor/native-task-handler/src/lib.rs
@@ -150,7 +150,8 @@ pub async fn run_native_task_handler<
 
 	tokio::spawn(async move {
 		while let Some((wrapper, sender)) = receiver.recv().await {
-			handle_native_task(ctx.clone(), wrapper, sender).await;
+			let ctx_cloned = ctx.clone();
+			tokio::spawn(async move { handle_native_task(ctx_cloned, wrapper, sender).await });
 		}
 	});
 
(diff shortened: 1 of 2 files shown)


tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -150,7 +150,8 @@ pub async fn run_native_task_handler<
 
 	tokio::spawn(async move {
 		while let Some((wrapper, sender)) = receiver.recv().await {
-			handle_native_task(ctx.clone(), wrapper, sender).await;
+			let ctx_cloned = ctx.clone();
+			tokio::spawn(async move { handle_native_task(ctx_cloned, wrapper, sender).await });
```
felixfaisal: We need some kind of backpressure here as tasks could get spawned indefinitely and cause memory overhead 
Even though the receiver channel is bounded, it doesn't enforce real backpressure since messages are consumed quickly and tasks are still spawned indefinitely.

Kailai-Wang: Thank you - that's a good point

I use semaphore to limit it. In the long run I think it makes sense to not send the tasks to native-task-handler at all, as jsonrpsee should be handle async registered rpc concurrently already

felixfaisal: Yeah we should try to move away from the native-task-handler this pattern was used because of our limitations with `sgx` dependencies and to isolate from block productions.

felixfaisal (approved):
LGTM

kziemianek (approved):

PR #3364, opened 2025-04-11
Implement cross chain swaps

silva-fj:
TODO:

- [x] Create map for binance assets.
- [x] Transfer from_asset to binance deposit address (from user's OmniAccount)
- [x] Make the trade using binance spot trading api from_asset => BNB and notify BE in case of errors.
- [ ] Transfer assets back to OmniAccount if binance trade fails.
- [x] Call accounting contract on BSC. (may need some help from @felixfaisal)
- [x] Make native trade with pumpx API (using market order)

Note: the following trades are the only ones supported.

USDT@solana -> BNB  
USDC@solana -> BNB  
SOL@solana -> BNB


tee-worker/omni-executor/executor-worker/src/main.rs
```diff
@@ -139,13 +143,33 @@ async fn main() -> Result<(), ()> {
...
+				binance_api_secret,
+				binance_api_base_url,
+			));
+
+			let solana_client = Arc::new(SolanaClient::new(&args.solana_url));
+
+			let ethereum_rpc_provider = ethereum_rpc::AlloyRpcProvider::new(&args.ethereum_url);
+			let accounting_contract_client = AccountingContractClient::new(
+				ethereum_rpc_provider,
+				"TODO:get contract address".parse().unwrap(),
```
Kailai-Wang: We have the contract address now:
`0xb0830ef478a215ed393c20a0c97aa69869a0beea` on bsc

tee-worker/omni-executor/pumpx/src/signer_client_mocks.rs
Kailai-Wang, line 1: Good one, thanks

tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
```diff
@@ -136,20 +137,43 @@ pub fn register_submit_swap_order(module: &mut RpcModule<RpcContext>) {
...
+						.unwrap();
+					HeimaMultiAddress::Address20(Address20::from(
+						pubkey_to_evm_address_bytes(&wallet).unwrap(),
+					))
+					// get ethereum addresss
+				},
+				ChainAsset::Solana(_) => {
+					unimplemented!("Only Ethereum supported");
+				},
+			};
```
Kailai-Wang: This is however not entirely true

The "swap" can be a single-chain swap, so the `to_address` could be some solana address (imagine it's a SOL -> meme-coin swap on solana).

The initial idea of `to_address` in `SwapOrder` is user could specify recipient address and fall back to his omni account on `to_chain` if not specified.

But pumpx doesn't support customisation of `to_address`, so we don't even include this parameter in RPC.

There is `to_chain_type` and `to_wallet_address` already in intent-executor, we can reuse it.

Keeping `SwapOrder.to_address` as `None` is fine IMO - eventually this order will be recorded on chain and **this is how it should be like if user submits order via pumpx api**

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -76,35 +101,48 @@ pub type ParentchainTxSigner = TxSigner<
 	SubxtMetadataProvider<CustomConfig>,
 >;
 
+// TODO: temporary solution
```
Kailai-Wang, line 102: Just curious: why is it temporary? I thought this was **the** solution 😂

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -434,11 +501,31 @@ impl<
...
 				} else {
+					debug!("from and to chain are different, performing cross chain swap");
+					if !matches!(
+						swap_order.to_asset,
+						ChainAsset::Ethereum(pumpx::constants::BSC_CHAIN_ID, _)
+					) {
+						log::error!("Only BSC payout supported");
+					}
+
+					let payout_address = match swap_order.to_address {
```
Kailai-Wang: As mentioned above, maybe `to_wallet_address` is better?

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -504,27 +579,386 @@ impl<
...
+							)
+							.await
+							.map_err(|_| {
+								log::error!("Failed to transfer SPL");
+							})?;
+					}
+					let (trade_symbol, order_side) = match binance_coin_name.as_str() {
+						"USDC" => ("BNBUSDC".to_string(), BinanceOrderSide::BUY),
+						"USDT" => ("BNBUSDT".to_string(), BinanceOrderSide::BUY),
+						"SOL" => ("SOLBNB".to_string(), BinanceOrderSide::SELL),
```
Kailai-Wang, line 653: This is correct right? (direction)

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -504,27 +579,386 @@ impl<
...
+						})?;
+						// TODO: Figure out how to transfer back the asset to the omni account
+						// check https://developers.binance.com/docs/wallet/capital/withdraw
+
+						return Err(());
+					};
+
+					let mut trade_success = false;
+					let mut to_amount = 0;
+					loop {
```
Kailai-Wang, line 688: Could this be endless loop? 😂

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -504,27 +579,386 @@ impl<
...
+								log::error!("Failed to get binance order");
+							})?;
+
+						match trade_order.status {
+							BinanceOrderStatus::FILLED => {
+								log::info!("Binance order filled");
+								for fill in trade_order.fills.unwrap() {
+									let fill_price: u128 = fill.price.parse().unwrap();
+									let fill_amount: u128 = fill.qty.parse().unwrap();
+									to_amount += fill_price * fill_amount;
```
Kailai-Wang: Just to double-check that it's what we want - `to_amount` is the amount of BNBs to pay out? I just don't know if `fill_price * fill_amount` get it right. Like should it multiple it by `fill_price`?

Did it take decimal into consideration?

So if it's 0.01 BNB, `to_amount` will be `(0.01 * 10^18)`, right?

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -504,27 +579,386 @@ impl<
...
+							payout_address,
+							user_nonce,
+							U256::from_str(&to_amount.to_string()).unwrap(),
+						)
+						.await
+						.map_err(|_| {
+							log::error!("Failed to execute pay out request");
+						})?;
+
+					let Some(chain_type) = ChainType::from_pumpx_chain_id(pumpx_config.to_chain_id)
```
Kailai-Wang: We have `to_chain_type` already

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -504,27 +579,386 @@ impl<
...
+					else {
+						log::error!("Unsupported to_chain_id: {}", pumpx_config.to_chain_id);
+						return Err(());
+					};
+
+					let wallet_address: String = match chain_type {
```
Kailai-Wang: And `to_wallet_address` : )

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -504,27 +579,386 @@ impl<
...
+						swap_type: match pumpx_config.swap_type {
+							1 => SwapType::Buy,
+							2 => SwapType::Sell,
+							_ => {
+								log::error!("Unsupported swap type: {}", pumpx_config.swap_type);
+								return Err(());
+							},
+						},
+						// amount received from binance should be used...
+						amount_in: to_amount.to_string(),
```
Kailai-Wang: Here we need to consider decimal, from previous example, backend expects `0.01` here, not `0.01 * (10^18)`

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -504,27 +579,386 @@ impl<
...
+								log::error!("Failed to get nonce");
+							},
+						)?;
+
+					debug!("Received {:?} nonce", user_nonce);
+
+					debug!("Calling accounting contract payout with address {:?}, nonce {:?} and amount {:?}", payout_address, user_nonce, to_amount);
+					//todo : make sure we payout correct amount
+					self.accounting_contract_client
+						.execute_pay_out_request(
```
Kailai-Wang, line 771: Second, is the tx actually signed? 😂

Kailai-Wang (changes requested):
Thank you - I think the flow is fine - I have a few details to discuss

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -504,27 +579,386 @@ impl<
...
+							.binance_api
+							.spot_trading()
+							.get_order(&trade_symbol, Some(binance_order.order_id), None, None)
+							.await
+							.map_err(|_| {
+								log::error!("Failed to get binance order");
+							})?;
+
+						match trade_order.status {
+							BinanceOrderStatus::FILLED => {
```
Kailai-Wang, line 699: So it means it's fully "filled", right? No partial trade could happen

tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
kziemianek: Ok, i'll keep it as `None` 👍🏼

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: If it's not comming from `SwapOrder` then yes.

tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
kziemianek: Should be done

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: @silva-fj ?

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: removed

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: It will faill, we need to provide wallet here :

```
			let ethereum_rpc_provider = ethereum_rpc::AlloyRpcProvider::new(&args.ethereum_url);
			let accounting_contract_client = AccountingContractClient::new_with_wallet(
				ethereum_rpc_provider,
				"TODO:get contract address".parse().unwrap(),
			);
```

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: Now I assume we will get FILLED status with all fills (100% completion)

```
	PARTIALLY_FILLED,
	FILLED,
```
`PARTIALLY_FILLED` status is ignored and we poll untill it's `FILLED`

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: These values are decimals i'll change it

kziemianek: Binance amount received will be multiplied by 10^8 because native BNB has a smalles unit [Jager](https://coinmarketcap.com/academy/glossary/jager)

kziemianek: That's a good catch - i wasn't sure about it. Thanks.

Kailai-Wang: We actually need decimals in pumpx-api 

It just has to match the binance api (what `fill_price` and `fill_amount` stand for)

Maybe you could write a comment naming an example for it - so that everyone that reads the code knows the _intention_ at least

kziemianek: Done

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
Kailai-Wang: I see - will we reach it eventually? Like can it be an endless loop?

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: Conversion to bnb unit (Jager) is done later. I expect current impl will result in `0.01` not `0.01 * 10^8`

Kailai-Wang (approved):
Let's merge it and I'll raise another PR to address the binance-api problem

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: I was just thinking that on test networks for example the contract/mint addresses could be different so we may want to have a config for this

PR #3365, opened 2025-04-11
Make `google_code` required field

Kailai-Wang:
### Context

As topic - according to pumpx, `verify_google_code` must be called - F/E will pass in string with value in any case.

### DIFF
--- a/tee-worker/omni-executor/executor-core/src/native_task.rs
+++ b/tee-worker/omni-executor/executor-core/src/native_task.rs
@@ -39,9 +39,9 @@ pub enum NativeTask {
 
 	// pumpx specific, starting from index 20
 	#[codec(index = 20)]
-	PumpxRequestJwt(Identity, Option<String>, MaybeGoogleCode, Option<String>),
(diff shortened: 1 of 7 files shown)


tee-worker/omni-executor/executor-core/src/native_task.rs
```diff
@@ -39,23 +39,23 @@ pub enum NativeTask {
...
-		String,          // token_contract_address
-		String,          // amount
-		MaybeGoogleCode, // google_code
-		Option<String>,  // language
+		u32,            // chain_id
+		u32,            // wallet_index
+		String,         // recipient_address
+		String,         // token_contract_address
+		String,         // amount
+		String,         // google_code
```
kziemianek, line 57: With type alias 

`type GoogleCode = String` we got the parameter meaning.

kziemianek (approved):

tee-worker/omni-executor/executor-core/src/native_task.rs
Kailai-Wang: I'll do it in a follow-up PR

PR #3368, opened 2025-04-11
Update solana chain_id

Kailai-Wang:
### Context

10000 => 100000

Also move it to a central place


silva-fj (approved):

Kailai-Wang:
I updated the `PumpxOrderType` to always lowercase to be consistent with other values in RPC (lower cases) @silva-fj

BillyWooo (approved):

PR #3369, opened 2025-04-11
Notify backend only for cross-chain swap

Kailai-Wang:
### Context

As topic


BillyWooo (approved):

silva-fj (commented):
Thanks Kai, this is already done in the draft PR. What is missing is the new params

PR #3371, opened 2025-04-12
feat:Update client-sdk tools.

0xverin:
this PR:

- add aes key decryption method.

- update the encryption structure of the shielding key


Kailai-Wang (approved):

Kailai-Wang:
A dumb question: any way (CI/test) to verify the change actually works?

0xverin:
> A dumb question: any way (CI/test) to verify the change actually works?

I actually verified it on the [front end](https://github.com/0xverin/tets-client-sdk/blob/main/src/app/page.tsx#L56). It can actually be hardcoded for verification, as long as the aeskey matches. However, it is not recommended to open the client-sdk's CI now, as it is still unstable.

image

PR #3377, opened 2025-04-13
Take gas fee into account for cross chain trade

Kailai-Wang:
### Context

As topic - we should call pumpx api with adjusted `amount_in`.


kziemianek (commented):
Is it ok that full amount is paid out by accounting contract ?
Shouldn't it take gas fee into amount?

Kailai-Wang:
> Is it ok that full amount is paid out by accounting contract ?
Shouldn't it take gas fee into amount?

@kziemianek The accounting contract is triggered by worker, so the worker will pay gas.
The gas will be considered when using pumpx : )

The only exception is the swap process errors out after paying out but before calling pumpx. In that case, yes we'll lose the gas fee we paid, but we consider the fee minimal and case is rare, so we live with it in the first version

kziemianek (approved):

PR #3379, opened 2025-04-13
Refactor the parachain notification from worker

Kailai-Wang:
### Context

- only notify intent `accepted` and `completed` event (to not spam it before we know the rate)
- "submit and forget": do not care about submission result


silva-fj (approved):

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -396,22 +381,21 @@ async fn handle_native_task<
...
-					if rpc_client.submit_tx(&tx).await.is_err() {
-						log::error!("Failed to submit RequestIntent tx");
-						ctx.transaction_signer.update_nonce().await;
-					}
-					return;
+					(execution_result, should_notify_parentchain)
 				},
 			};
 
-			(response_sender, tx)
```
silva-fj, line 415: Now in the happy path, the `response_sender` is being dropped without sending any message to the channel so it errors out. This is why the test is failing o CI

Kailai-Wang: Good catch, thanks!

Kailai-Wang: Would you make change directly on this branch? I'm working on another pr

@silva-fj

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -298,51 +298,51 @@ async fn handle_native_task<
...
 						value: transfer.value,
 					};
-					let dispatch_as_omni_account_call =
-						parentchain_api_interface::tx().omni_account().dispatch_as_signed(
-							sender.hash().to_subxt_type(),
-							RuntimeCall::Balances(transfer_call),
-							auth_type.map(|t| t.to_subxt_type()),
-						);
-					ctx.transaction_signer.sign(dispatch_as_omni_account_call).await
+					let _ = dispatch_as_omni_account(
```
silva-fj: This call should use `dispatch_as_signed`.  `SystemCall::remark` also needs to be changed

PR #3380, opened 2025-04-14
accounting contract ecdsa keypair

kziemianek:
Adds key store for storing `ECDSA` key used for signing accounting contract transactions.


felixfaisal (approved):
LGTM

PR #3381, opened 2025-04-14
fix: allow overriding with start_block

felixfaisal:
`start_block` already exists in the CLI, we just needed to adjust the logic to allow overriding using the start_block


Kailai-Wang (approved):

kziemianek (approved):

Kailai-Wang:
Thank you - could try it locally (or even with paseo) if haven't done it?

PR #3383, opened 2025-04-14
Forward error code

wli-pro:
### Labels

### Testing Evidences

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
+++ b/tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
@@ -39,8 +39,8 @@ impl From<AddWalletParams> for NativeTaskWrapper<NativeTask> {
 pub fn register_add_wallet(module: &mut RpcModule<RpcContext>) {
 	module
 		.register_async_method("pumpx_addWallet", |params, ctx, _| async move {
-			let params = params.parse::<AddWalletParams>().map_err(|_| {
-				log::error!("Failed to parse params");
+			let params = params.parse::<AddWalletParams>().map_err(|e| {
+				log::error!("Failed to parse params: {:?}", e);
 				PumpxRpcError::from_error_code(ErrorCode::ParseError)
 			})?;
 
--- a/tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
+++ b/tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
@@ -1,14 +1,16 @@
 use crate::{
-	error_code::*, oneshot, server::RpcContext, verify_auth::verify_auth, Decode, Deserialize,
-	ErrorCode,
+	error_code::*, methods::pumpx::PumpxRpcError, server::RpcContext, verify_auth::verify_auth,
+	Deserialize, ErrorCode,
 };
 use executor_core::native_task::*;
 use executor_primitives::OmniAuth;
 use heima_primitives::{Identity, Web2IdentityType};
-use jsonrpsee::{types::ErrorObject, RpcModule};
-use native_task_handler::{NativeTaskError, NativeTaskOk, NativeTaskResponse};
+use jsonrpsee::RpcModule;
+use native_task_handler::NativeTaskOk;
 use pumpx::types::AddWalletResponse;
 use serde::Serialize;
+
+use super::common::{check_pumpx_api_response, handle_pumpx_native_task};
 
 #[derive(Debug, Deserialize)]
 pub struct AddWalletParams {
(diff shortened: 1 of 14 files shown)


tee-worker/omni-executor/rpc-server/src/methods/pumpx/add_wallet.rs
```diff
@@ -37,51 +39,30 @@ impl From<AddWalletParams> for NativeTaskWrapper<NativeTask> {
 pub fn register_add_wallet(module: &mut RpcModule<RpcContext>) {
 	module
 		.register_async_method("pumpx_addWallet", |params, ctx, _| async move {
-			let internal_error: ErrorObject = ErrorCode::InternalError.into();
-			let params = params.parse::<AddWalletParams>()?;
+			let params = params.parse::<AddWalletParams>().map_err(|_| {
+				log::error!("Failed to parse params");
```
kziemianek: Doesn't make sense to log reason here (`e`)  ?

tee-worker/omni-executor/rpc-server/src/methods/pumpx/common.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+impl PumpxRpcError {
+	pub fn from_code_and_message(code: i32, message: String) -> Self {
+		Self { code, message }
+	}
+
+	pub fn from_error_code(error_code: ErrorCode) -> Self {
+		let (code, message) = match error_code {
+			ErrorCode::ParseError => (-32700, "Parse error"),
```
kziemianek: These messages are all in english while the ones returned from pumpx backend will be i18 , is that correct ?

Kailai-Wang (approved):
I actually like this refactoring 👍 

Would fields of all `..ResponseData` need to be in `Option`? As in error case they would be empty.

BillyWooo:
> I actually like this refactoring 👍
> 
> Would fields of all `..ResponseData` need to be in `Option`? As in error case they would be empty.

Yes. They all should be `Option`. 
@wli-pro can you update all of them? and also resolve the conflicts. Besides, there is one more new RPC `pumpx_getOmniAccount` added.

tee-worker/omni-executor/rpc-server/src/methods/pumpx/common.rs
```diff
@@ -0,0 +1,98 @@
...
+pub fn check_pumpx_api_response<T>(
+	response: ApiResponse<T>,
+	name: String,
+) -> Result<(), PumpxRpcError>
+where
+	T: Codec,
+{
+	if response.code != 10000 {
+		log::error!("{} failed: code={}, message={}", name, response.code, response.message);
+		return Err(PumpxRpcError::from_code_and_message(response.code as i32, response.message));
```
Kailai-Wang: Does it mean if we detect errors from pump-api response, we'll use their code and msg as our response and return it to F/E?

Could that be problematic for F/E to tell if the error actually comes from worker or backend? Especially when the error code happens to coincide.

Would this makes more sense? Like
```
{
  "code": -32055, // just as example
  "message": "Pumpx API error",
  "error": {
    "backend_response": { // the raw backend response
       "code": -12345,
       "message": ...,
       "data": {}
     }
  }
}
```

So here `-32055` means we return because backend returns error - so we propagate it

tee-worker/omni-executor/rpc-server/src/methods/pumpx/common.rs
BillyWooo: Here is jsonrpsee internal error code. It's fine.

tee-worker/omni-executor/rpc-server/src/methods/pumpx/common.rs
```diff
@@ -0,0 +1,98 @@
...
+	pub code: i32,
+	pub message: String,
+}
+
+impl PumpxRpcError {
+	pub fn from_code_and_message(code: i32, message: String) -> Self {
+		Self { code, message }
+	}
+
+	pub fn from_error_code(error_code: ErrorCode) -> Self {
```
BillyWooo, line 35: How about to have:
```rust
pub fn from_error_code(error_code: ErrorCode) -> Self {
Self { code: error_code.code(), message: error_code.message().to_string() }
}
```

wli-pro:
> > I actually like this refactoring 👍
> > Would fields of all `..ResponseData` need to be in `Option`? As in error case they would be empty.
> 
> Yes. They all should be `Option`. @wli-pro can you update all of them? and also resolve the conflicts. Besides, there is one more new RPC `pumpx_getOmniAccount` added.

As all ResponseData should be Option, I think there is no need to make any changes, as the response of jsonrpc already has a different response structure for success and error. 

Success response example like below
```
{
	"jsonrpc": "2.0",
	"id": 1,
    "result": {
       "backend_response": {}
     }
}
```
Error response example like below
```
{
	"jsonrpc": "2.0",
	"id": 1,
	"error": {
		"code": -32700,
		"message": "Parse error"
	}
}
```

PR #3386, opened 2025-04-14
Use `user_uid` over `user_email` ubiquitously

Kailai-Wang:
### Context

As topic - hopefully we get the same mapped omni-account as backend this way so we don't have to move assets.
For some cases where email verification code needs to be validated, a `user_email` is still needed.

In `exportWallet` and `withdraw` it verifies if the user_id and user_email match

It also provides a helper rpc `pumpx_getOmniAccount` for cross-check


tee-worker/omni-executor/rpc-server/src/methods/pumpx/export_wallet.rs
```diff
@@ -14,6 +14,7 @@ use sha2::Sha256;
 
 #[derive(Debug, Deserialize)]
 pub struct ExportWalletParams {
+	pub user_uid: String,
```
kziemianek: How do we make sure both `user_id` and `user_email` belong to same user ?

Kailai-Wang: Good point, we'll need to query the uid for an email from backend and verify it 😂  let me adjust it

silva-fj (commented):
I think we should get the user ID from the BE, for us there is no way to verify it and map it to the user's email. We can keep the email authentication as it is, call the BE to get the user ID for that email and issue the JWTs with the OmniAcccount (sub) derived from that ID

tee-worker/omni-executor/rpc-server/src/methods/pumpx/export_wallet.rs
silva-fj: I read this after my comment 😆

tee-worker/omni-executor/rpc-server/src/methods/pumpx/export_wallet.rs
```diff
@@ -60,6 +60,24 @@ pub fn register_export_wallet(module: &mut RpcModule<RpcContext>) {
...
 
+			// verify user_id and user_email matches
+			let Ok(res) = ctx.pumpx_api.get_account_user_id(params.user_email.clone()).await else {
+				log::error!("Failed to call get_account_user_id");
+				return Err(
+					ErrorCode::ServerError(PUMPX_API_GET_ACCOUNT_USER_ID_FAILED_CODE).into()
+				);
+			};
+
+			if res.user_id != params.user_id {
```
kziemianek, line 71: If we query it from backend anyway can we simply not accept it from frontend ?

kziemianek (approved):

silva-fj (approved):

tee-worker/omni-executor/rpc-server/src/methods/pumpx/export_wallet.rs
Kailai-Wang: I think it gives us extra chance to tell F/E something goes wrong early.
If F/E doesn't pass in `user_email` - it could still give the wrong/arbitrary `user_id`. Now at least they need to pass in matching pairs.

PR #3389, opened 2025-04-14
move accounting ecdsa keypair to trusted keystore

kziemianek:
It should be encrypted as specified in the issue.


Kailai-Wang (approved):
Thank you - does it mean we don't need anything in `sgx.allowed_files` now?

BillyWooo (approved):

PR #3392, opened 2025-04-14
Improve binance spot trading logic

silva-fj:
Improving cross chain swaps by waiting until the deposit has been confirmed before attempt to create a market trade order on Binance. Additionaly, if the trade fails, the assets are transferred back to the OmniAccount


tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -647,6 +647,64 @@ impl<
...
+								Ok(deposit_amount) => deposit_amount.to_u64().unwrap_or(0),
+								Err(_) => {
+									log::error!("Failed to parse deposit amount");
+									continue;
+								},
+							};
+							if deposit.status == 1 && // 1 = success
+							   deposit.coin == binance_coin_name &&
+							   deposit.network == binance_network_info.network &&
+                               deposit_amount == amount_to_transfer
```
Kailai-Wang: Hmm so if the user somehow makes two consectutive deposit with the same param (coin, network, amount) - can it be differentiated?

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -669,6 +727,20 @@ impl<
 						self.binance_api.spot_trading().create_order(binance_order_params).await
 					else {
 						log::error!("Failed to create binance order");
+						self.binance_api
+							.wallet()
+							.withdraw(
+								&binance_coin_name,
+								&from_address,
+								amount_to_transfer_decimal.to_string(),
```
Kailai-Wang: Just to confirm, this is **with** multiplier, right?

"decimal" here just means the `Decimal` format, it doesn't mean it's e.g. `0.03` instead of `3 * (10^16)` (just example)

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -677,8 +749,6 @@ impl<
 						self.pumpx_api.cross_fail(&access_token, body).await.map_err(|_| {
 							log::error!("Failed to notify pumpx-signer");
 						})?;
-						// TODO: Figure out how to transfer back the asset to the omni account
```
Kailai-Wang, line 637: There is another `TODO: Figure out how to transfer back the asset to the omni account` after the trade order, do we need to apply the withdraw too?

Could you please check that?

Kailai-Wang (commented):
Thanks! I left a few comments

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -647,6 +647,64 @@ impl<
...
+									"Deposit confirmed on Binance for {} {}",
+									deposit.amount, binance_coin_name
+								);
+								break;
+							}
+						}
+
+						if !deposit_confirmed {
+							debug!("Deposit not confirmed yet, waiting 5 seconds...");
+							sleep(Duration::from_secs(5)).await;
```
felixfaisal, line 672: Is 5 seconds arbitrary? Finalisation delay could vary

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -647,6 +647,64 @@ impl<
...
+						}
+					}
+
+					if !deposit_confirmed {
+						log::error!("Deposit not confirmed within timeout period");
```
felixfaisal, line 677: This would mean deposit not confirmed within timeout period even though we know deposit has been made on the chain and it's taking time for Binance to confirm because of re-orgs or something, should we have an exponential backoffs instead of arbitary 5s delay? wdyt?

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -669,6 +727,20 @@ impl<
...
+						self.binance_api
+							.wallet()
+							.withdraw(
+								&binance_coin_name,
+								&from_address,
+								amount_to_transfer_decimal.to_string(),
+								Some(&binance_network_info.network),
+							)
+							.await
+							.map_err(|_| {
```
felixfaisal: I think it's important to segregate errors here, Like what are the possible errors? Could some types of errors be mitigated by performing retries (like network delay), or if the balance is too low(balance is locked and requires verification) then we have to readjust the amount, it'd be easier to debug and adjust if we segregate this,

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -647,6 +647,64 @@ impl<
...
+							.wallet()
+							.get_deposit_history(Some(binance_coin_name.clone()))
+							.await
+						else {
+							log::error!("Failed to get deposit history");
+							continue;
+						};
+
+						// Check if there's a recent successful deposit
+						for deposit in deposit_history {
```
felixfaisal, line 630: Does this also detects deposits that require more information from the account, if so can we programmatically fill the details of the deposit information that is required by KYC/AML of binance

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: There is a `txId` optional param, let me refactor the solana_client to return the signatures and pass it to the api. But I think we don't need to differentiate them, if the deposit was made the user cannot withdraw the assets. I've also included a filter for the source address to make sure the deposit is from the same user, I will push it soon

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: Yes, this is with multiplier but looking at the API docs again, I think this should be the amount we get in pumpx_config, as binance expects a decimal number, so `0.03` in this example

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: Yep

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: I was thinking about this, the problem is that according to the API docs there are a couple of cases where the deposit will not succeed, so we cannot just assume that It will be available eventually. In one case the deposit could be `rejected` (In this case I don't know what would happen, if binance just returns the assets to the source address or what), another case is `Wrong Deposit` (I don't know what that even means) and there is another `credited but cannot withdraw` which I don't know what we could do (this cannot be resolved with the API). In my experience deposits from Solana in binance are quite fast, but we'll need to revisit this once we start adding support for other networks.

The exponential backoffs sounds like a good improvement, we could get fancier in the next iterations 👍🏼 

I'm trying find something In the docs that would help to implement a more robust solution for these edge cases

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: I agree, it would be awesome to know what the possible errors are but apart from regular network errors, there is no info in the docs about this

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: Can you elaborate a bit more?

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
Kailai-Wang: Eventually it's a centralised service so yes it could be rejected (same for withdrawal - they could have arbitrary reasons to delay/reject it 😂)

These need time to improve, we don't expect a version that works for all corner cases. I'd say I'm fine if it can handle happy path + major error cases

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
felixfaisal: When we receive a deposit on binance, we have to provide information on the depositor until then the deposit is not credited to your account, so does this detect deposits that are pending because of not providing depositor details. 

I'm not sure if you have tested it on production, if this didn't come up during your testing with prod binance api, then ig it's nothing to worry, but if you haven't tested then it's something to look out for.

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
felixfaisal: Makes sense, but then i was wondering if deposit has been made on chain, and it timed out here because of the different cases you mentioned, then how will we be able to refund the user, because we probably don't have access to it anymore. 

I think we will get a better understanding once we start testing but something to look out for

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: I see, I think from Binance perspective you (as the onwer of the binance account) are the depositor, so our account should have been checked already (?). But good point, we need to check all these scenarios when we start testing 👍🏼

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
silva-fj: done

Kailai-Wang (approved):
@silva-fj I'm happy with it mostly - if there're no major TODOs, I'm inclined to merging it and building a new image cc @BillyWooo

silva-fj:
> @silva-fj I'm happy with it mostly - if there're no major TODOs, I'm inclined to merging it and building a new image cc @BillyWooo

I agree, the sooner we start testing the better

PR #3397, opened 2025-04-15
fix manifest + minor changes

BillyWooo:
- fix manifest
- some minor update


kziemianek (approved):
I assume `sgx.allowed_files` is just a workaround.

Kailai-Wang:
> I assume `sgx.allowed_files` is just a workaround.

Yes we should explicit mount them I guess - for now it's fine

PR #3404, opened 2025-04-16
Remove redundant AccountId::from

Kailai-Wang:
### Context

As topic

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/methods/omni/get_next_intent_id.rs
+++ b/tee-worker/omni-executor/rpc-server/src/methods/omni/get_next_intent_id.rs
@@ -18,7 +18,6 @@ use std::str::FromStr;
 
 use crate::server::RpcContext;
 use crate::ErrorCode;
-use executor_primitives::AccountId;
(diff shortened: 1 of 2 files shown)


tee-worker/omni-executor/rpc-server/src/methods/omni/get_next_intent_id.rs
```diff
@@ -18,7 +18,6 @@ use std::str::FromStr;
 
 use crate::server::RpcContext;
 use crate::ErrorCode;
-use executor_primitives::AccountId;
```
silva-fj, line 21: We should probably keep this one and remove `parentchain_rpc_client::AccountId32` (this is the subxt type)

Kailai-Wang: I'll look for something like `from_ss58_str`

silva-fj: If I remember correctly `AccountId::from_str` would do that internally

BillyWooo (approved):

PR #3408, opened 2025-04-17
fix: test alloy provider transaction

wli-pro:
### Labels

### Testing Evidences


felixfaisal:
I think we don't need this anymore, So I'm closing it, Feel free to open it if you think we should include it.

PR #3409, opened 2025-04-17
Toggle parentchain listener through CLI

kziemianek:
By default it's set to false (if flag is not present). It will not run parentchain listener so if there is any logic based on onchain events it won't be triggered.

### DIFF
--- a/tee-worker/omni-executor/executor-worker/src/cli.rs
+++ b/tee-worker/omni-executor/executor-worker/src/cli.rs
@@ -41,6 +41,8 @@ pub struct RunArgs {
 		value_name = "accounting contract address"
 	)]
 	pub accounting_contract_address: String,
+	#[arg(long, value_name = "should sync with parentchain")]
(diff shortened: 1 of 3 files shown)


tee-worker/omni-executor/executor-worker/src/cli.rs
```diff
@@ -41,6 +41,8 @@ pub struct RunArgs {
 		value_name = "accounting contract address"
 	)]
 	pub accounting_contract_address: String,
+	#[arg(long, value_name = "should sync with parentchain")]
+	pub parentchain_sync: bool,
```
BillyWooo, line 45: should we have a default value?

kziemianek: It's a flag (`--parentchain-sync`) so by default it's `false` (flag is not present).

BillyWooo (approved):

PR #3410, opened 2025-04-17
check google auth returned from user_connect

BillyWooo:
as title, a minor change to check google_auth_check return value.

### DIFF
--- a/tee-worker/omni-executor/native-task-handler/src/lib.rs
+++ b/tee-worker/omni-executor/native-task-handler/src/lib.rs
@@ -626,17 +626,6 @@ async fn handle_native_task<
 				return;
 			};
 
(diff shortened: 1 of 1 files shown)


tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -659,6 +648,25 @@ async fn handle_native_task<
 			};
 			log::debug!("Response pumpx user_connect: {:?}", backend_response);
 
+			// check google auth value
+			if let Some(user_connect_res) = backend_response.clone().data {
```
kziemianek: could be `ref user_connect_res` without `clone()`

kziemianek (approved):
Looks fine

Kailai-Wang (approved):

PR #3413, opened 2025-04-17
Cross-chain swap deposit/payout logic improvement 

silva-fj:
This PR adds support for both confirmed and estimated payout flows by adding an execution strategy to prevent waiting for deposit confirmations on binance

### DIFF
--- a/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
+++ b/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
@@ -500,6 +500,8 @@ impl<Provider: EthereumRpcProvider<Transaction = TransactionRequest> + Send + Sy
 					debug!("Response create_cross_order: {:?}", response);
 
 					// 2. transfer from_asset to binance deposit address
+					let should_wait_for_deposit_confirm = false; // Switch for strategy
+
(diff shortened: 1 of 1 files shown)


tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -642,233 +644,281 @@ impl<Provider: EthereumRpcProvider<Transaction = TransactionRequest> + Send + Sy
...
-					// in BUY orders:
-					// - `executedQty` indicates the amount of the base-asset bought
-					// - `cummulativeQuoteQty`` shows the total amount of the quote-asset spent
-					let mut trade_success = false;
-					let mut bnb_received = "".to_string();
-					loop {
-						let trade_order = self
+						debug!("Total received {} bnb", bnb_received);
+					} else {
+						// Estimate BNB payout (simulate spot trade, apply service fee)
```
Kailai-Wang, line 873: Maybe later we should think about how to improve the readability in this file, maybe split into Fns or files

It's hard to find this `else` after a long `if` arm (which is always false 😂 )

Kailai-Wang (approved):

PR #3414, opened 2025-04-17
Refactoring message to sign for web3 authentication

silva-fj:
As topic, this PR updates and simplifies the web3 message to sign as well as the signature verification. A new RPC method has been added to get the message payload to sign.

The `verify_auth` function no longer requires the `NativeTaskWrapper`.

This is part of the omni-executor's auth rework.

Note: This PR only includes changes for web3 authentication, the `AddAccount` (to AccountStore) native call still uses the old format for the validation data


tee-worker/omni-executor/heima/authentication/src/web3.rs
```diff
@@ -0,0 +1,52 @@
...
+/// the specified period, similar to TOTP (Time-based One-Time Password) algorithm.
+///
+/// # Arguments
+///
+/// * `period` - The time period in seconds for which the code remains valid
+///
+/// # Returns
+///
+/// A string containing a 6-digit message code that changes every `period` seconds
+pub fn generate_message_code(period_in_seconds: u64) -> String {
```
kziemianek: This function could take `now` as argument so there would be no need to wait in tests.

tee-worker/omni-executor/heima/authentication/src/web3.rs
```diff
@@ -0,0 +1,52 @@
...
+	use std::time::Duration;
+
+	#[test]
+	fn test_generate_message_code() {
+		let period = 5;
+		let code = generate_message_code(period);
+		assert_eq!(code.len(), 6);
+
+		// Generate the code within the same period
+		sleep(Duration::new(2, 0));
```
kziemianek: This fails ocassionaly. If you change sleep time to 4 it will fail more often.

kziemianek (changes requested):

tee-worker/omni-executor/heima/authentication/src/web3.rs
silva-fj: I've adjusted the code 👍🏼

tee-worker/omni-executor/heima/authentication/src/web3.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn test_message_code_at() {
+		let period = 5;
+
+		let base_ts = 1_000;
```
kziemianek: This test fail if you use current ts `let base_ts = 1745322204;`

kziemianek (dismissed):

tee-worker/omni-executor/heima/authentication/src/web3.rs
silva-fj: Yeah that's because when dividing by that period it gets at the edge of the next window, so just adding 2 secs will make it jump to the next. The function generates the code for the time window using the period regardless of when the first code was requested (just like google authenticator for example). I guess we could return from the RPC method the code + time left for the next one to simulate authenticator apps. Otherwise well have to persist some sort of records to keep track of each code

silva-fj: I've updated the RPC method to return the time left until the next code

kziemianek: Ok I see so in some cases this method will fail (for example if it takes too long for request to reach to server).

tee-worker/omni-executor/heima/authentication/src/web3.rs
```diff
@@ -0,0 +1,91 @@
+use chrono::Utc;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct HeimaMessagePayload {
+	pub message_code: String,
+}
+
+pub const MESSAGE_CODE_PERIOD: u64 = 60; // 1 minute
```
silva-fj: What would be a sensible value here?

silva-fj: Not relevant anymore

silva-fj:
I'll include these changes in another PR

PR #3416, opened 2025-04-18
fix: check contract balance before deposit

felixfaisal:
It checks if there's enough balance in the contract for a payout, else it exits before making a deposit.

### DIFF
--- a/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
+++ b/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
@@ -1078,7 +1078,7 @@ async fn estimate_bnb_amount(
 	from_amount_decimal: Decimal,
 ) -> Result<String, ()> {
 	let price_str =
-		binance_api.spot_trading().get_symbol_price(&trade_symbol).await.map_err(|_| {
+		binance_api.spot_trading().get_symbol_price(trade_symbol).await.map_err(|_| {
 			log::error!("Failed to get symbol price for {}", trade_symbol);
 		})?;
 
--- a/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
+++ b/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
@@ -613,13 +613,23 @@ impl<Provider: EthereumRpcProvider<Transaction = TransactionRequest> + Send + Sy
 						},
 					};
 
-					let estimated_bnb_receive = estimate_bnb_amount(&self.binance_api, &trade_symbol, &binance_coin_name, from_amount_decimal).await?;
+					let estimated_bnb_receive = estimate_bnb_amount(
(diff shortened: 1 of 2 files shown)


tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -603,6 +603,36 @@ impl<Provider: EthereumRpcProvider<Transaction = TransactionRequest> + Send + Sy
...
+					};
+
+					let estimated_bnb_receive = estimate_bnb_amount(
+						&self.binance_api,
+						&trade_symbol,
+						&binance_coin_name,
+						from_amount_decimal,
+					)
+					.await?;
+					let payout_amount = str_to_u256(&estimated_bnb_receive, 18).ok_or(())?;
```
silva-fj: We can also move the payout_amount conversion we do in the next steps here, so we also notify pumpx BE  and resolve the TODO

This part: 
```rust
let payout_amount = match str_to_u256(&bnb_to_receive, 18) {
	Some(a) => a,
	None => {
		log::error!("Fail to convert bnb amount {} to U256", bnb_to_receive);
		let body = CrossFailBody {
			request_id: intent_id,
			fail_reason:
			"Fail to construct payout request due to U256 conversion error"
				.to_string(),
		};
		self.pumpx_api.cross_fail(&access_token, body).await.map_err(|_| {
			log::error!("Failed to notify pumpx-signer");
		})?;
		// TODO: what to do with user asset?
		return Err(());
	},
};
```

felixfaisal: I've updated the code, to inform `pumpx` BE, but `payout` calculation still need to take place again if we're waiting for the trade to be completed based on trade fill data, unless i'm missing something here. 

I do have one question though, what is the criteria for informing `pumpx` BE, I couldn't find a pattern here, some parts of the code we are simply logging the error and returning `Err()`, and in some we are informing `pumpx` BE.

silva-fj: > I've updated the code, to inform pumpx BE, but payout calculation still need to take place again if we're waiting for the trade to be completed based on trade fill data, unless i'm missing something here.

Yes, you are right. Actually I found a couple of bugs while looking at this. I have opened a small PR to your branch to address them: https://github.com/litentry/heima/pull/3421

> I do have one question though, what is the criteria for informing pumpx BE, I couldn't find a pattern here, some parts of the code we are simply logging the error and returning Err(), and in some we are informing pumpx BE.

Yes we need to iterate on this, we have covered only the happy path

felixfaisal: I've merged it, Let me know if there's something else i need to look at.

silva-fj (approved):

PR #3417, opened 2025-04-18
Enforce token decimal precision in amount_in calculation

silva-fj:
This pull request adds support for decimal precision in `calculate_amount_in`

### DIFF
--- a/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
+++ b/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
@@ -1095,6 +1095,15 @@ mod tests {
 	use super::*;
 
 	#[test]
+	fn test_str_to_u256() {
+		let amount = "0.0050773374896233926847036101";
+		let decimals = 18;
+
+		let result = str_to_u256(amount, decimals);
+		assert_eq!(result, Some(U256::from_str("5077337489623392").unwrap()));
+	}
+
+	#[test]
 	fn test_calculate_amount_in() {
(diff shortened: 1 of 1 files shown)


tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -1079,12 +1079,37 @@ fn str_to_u256(amount: &str, decimals: u32) -> Option<U256> {
...
+		let amount = "0.0050773374896233926847036101";
+		let decimals = 18;
+
+		let result = str_to_u256(amount, decimals);
+		assert_eq!(result, Some(U256::from_str("5077337489623392").unwrap()));
+	}
+
+	#[test]
+	fn test_calculate_amount_in() {
+		let bnb_to_receive = "0.0050773374896233926847036101";
```
kziemianek, line 1117: Do you know from where do we get value with too many decimals ?

From this test case it seems like it's `decimal_bnb_to_receive`

Kailai-Wang: The `amount_in` that we pass to backend => backend complains it doesn't match the token precision (e.g. > 18 decimal precision for `BNB`)

silva-fj: > Do you know from where do we get value with too many decimals ?

It comes when estimating the bnb to receive, we multiply the from_amount with the price we get from binance api

kziemianek (approved):

PR #3419, opened 2025-04-21
Refactor JWT validation

silva-fj:
This pull request refactors the JWT authentication logic and improves the validation process. Now we validate that the tokens have the correct `typ` and remove the need to include the `sender` when verifying the tokens with `verify_auth_token_authentication`.

Note: This PR should be merged after https://github.com/litentry/heima/pull/3414 is merged. 

TODO: Update base branch after https://github.com/litentry/heima/pull/3414 is merged

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/verify_auth.rs
+++ b/tee-worker/omni-executor/rpc-server/src/verify_auth.rs
@@ -1,12 +1,15 @@
 use crate::server::RpcContext;
 use executor_crypto::hashing::blake2_256;
 use executor_primitives::{
-	signature::HeimaMultiSignature, utils::hex::ToHexPrefixed, Identity, OAuth2Data,
-	OAuth2Provider, OmniAuth, VerificationCode, Web2IdentityType,
+	signature::HeimaMultiSignature, Identity, OAuth2Data, OAuth2Provider, OmniAuth,
+	VerificationCode, Web2IdentityType,
 };
 use executor_storage::{OAuth2StateVerifierStorage, Storage, VerificationCodeStorage};
 use heima_authentication::{
-	auth_token::{AuthTokenValidator, Error as AuthTokenError, Validation},
+	auth_token::{
+		AuthTokenClaims, AuthTokenValidator, Error as AuthTokenError, Validation,
+		AUTH_TOKEN_ID_TYPE,
+	},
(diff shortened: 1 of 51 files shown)


kziemianek (commented):
We have many places where we set `NativeTask.sender` based on params provided to rpc method example (`PumpxAddWallet(Identity)` <- it holds only sender):

```
impl From<AddWalletParams> for NativeTaskWrapper<NativeTask> {
	fn from(p: AddWalletParams) -> Self {
		Self {
			task: NativeTask::PumpxAddWallet(Identity::from_web2_account(
				p.user_id.as_str(),
				Web2IdentityType::Pumpx,
			)),
			nonce: None,
			auth: Some(OmniAuth::AuthToken(p.auth_token)),
		}
	}
}
```

How do we make sure that `sub` from `JWT` and `sender` relates to same user ?
I don't know if it's right PR to address this but this one surerly removes check of `JWT` sub field.

silva-fj:
> We have many places where we set NativeTask.sender based on params provided to rpc method example (PumpxAddWallet(Identity) <- it holds only sender)

In this case we can just use the OmniAccount, in the handler we only use the sender to get it (`sender.to_omni_account()`), I'll check all the pumpx native tasks and update them.

> How do we make sure that `sub` from `JWT` and `sender` relates to same user ? I don't know if it's right PR to address this but this one surerly removes check of `JWT` sub field.

In the cases where we need the sender we can check `if claims.sub == sender.to_omni_account().to_hex()`

silva-fj:
Ok this will require more changes, I'll keep the `sub` validation, we can tackle this in another issue

tee-worker/omni-executor/rpc-server/src/verify_auth.rs
```diff
@@ -100,9 +104,11 @@ pub fn verify_auth_token_authentication(
 	ctx: Arc<RpcContext>,
 	sender: &Identity,
```
kziemianek: What do you thing about getting rid of inner `sender.omni_account()` and require `omni_account: String` explicitly ?

silva-fj: I've updated the code 👍🏼

silva-fj:
I will include al thesel changes in another PR

PR #3420, opened 2025-04-21
Use correct OmniAccount in SignLimitOrder's NativeTask's

silva-fj:
This PR fixes an issue with the OmniAccount's extracted from these NativeTasks: `PumpxSignLimitOrder` and `PumpxNotifyLimitOrderResult`

We create the NativeTask using the OmniAccount itself as the sender: 
```rust
let omni_account = token.sub;
let Ok(address) = Address32::from_hex(&omni_account) else {
	error!("Not a valid address");
	return Err(internal_error);
};

let task_wrapper = NativeTaskWrapper {
	task: NativeTask::PumpxSignLimitOrder(
		Identity::Substrate(address),
		params.chain_id,
		params.wallet_index,
		params.unsigned_tx.iter().map(|tx| tx.to_vec()).collect(),
	),
	nonce: None,
	auth: Some(OmniAuth::AuthToken(params.auth_token)),
};
```
Using `to_omni_account` derives a different OmniAccount

### DIFF
--- a/tee-worker/omni-executor/native-task-handler/src/lib.rs
+++ b/tee-worker/omni-executor/native-task-handler/src/lib.rs
@@ -814,7 +814,18 @@ async fn handle_native_task<
 			return;
 		},
 		NativeTask::PumpxSignLimitOrder(sender, chain_id, wallet_index, unsigned_tx) => {
-			let omni_account = sender.to_omni_account();
+			// This is a workaround, in this case the Substrate identity is the OmniAccount address itself
+			let omni_account: AccountId = match sender {
+				Identity::Substrate(ref account) => account.into(),
+				_ => {
+					send_error(
+						"Invalid sender type for PumpxSignLimitOrder".to_string(),
+						response_sender,
+						NativeTaskError::InternalError,
+					);
+					return;
+				},
+			};
 			let Some(chain) = ChainType::from_pumpx_chain_id(chain_id) else {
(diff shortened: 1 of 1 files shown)


Kailai-Wang (commented):
Thank you, I see the problem

Does it make sense to have an `Identity::Omni` type? And we differentiate the case in `to_omni_account` fn

silva-fj:
> Thank you, I see the problem
> 
> Does it make sense to have an `Identity::Omni` type? And we differentiate the case in `to_omni_account` fn

I had a similar idea, but this could be a bit confusing if we start using the AccountStore, we will also have to add rules like "you cannot add Omni identities" for example as it's the OmniAccount itself and maybe some other cases like this. 

Another solution could be to use use the AccountStore local storage or create a new one to map OmniAccount -> Identity when creating the tokens so we don't have to do these workarounds.

We only have this problem for the rpc methods called by the BE, as they don't have the Identity, we can only extract the OmniAccount from the JWT

BillyWooo (approved):

Kailai-Wang:
> We only have this problem for the rpc methods called by the BE, as they don't have the Identity, we can only extract the OmniAccount from the JWT

This makes sense, I'm thinking if we should have omni_account as the unified (and unique) identifier acrossc communication in the whole system

Kailai-Wang (approved):

PR #3425, opened 2025-04-23
duplicate rpc call with new name

BillyWooo:
as title. Mainly for Heima demo usage.


tee-worker/omni-executor/rpc-server/src/methods/omni/common.rs
```diff
@@ -0,0 +1,121 @@
+use jsonrpsee::types::{ErrorCode, ErrorObjectOwned};
+use parity_scale_codec::Codec;
+use pumpx::types::ApiResponse;
```
Kailai-Wang, line 3: Either we put in a level higher, or just `omni::types` - same for `PumpxRpcError` and other structs below

Kailai-Wang (approved):
There're still quite many `pumpx` structs, we'd need to fix those in the follow-up PRs

The goal is to have no `pumpx*` stuff in the `omni_*` methods, except for pumpx NativeTasks (which need a bigger refactoring)

PR #3426, opened 2025-04-23
add omni_getOmniAccount rpc call

BillyWooo:
miss one rpc call

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/methods/omni/get_omni_account.rs
+++ b/tee-worker/omni-executor/rpc-server/src/methods/omni/get_omni_account.rs
@@ -1,0 +1,44 @@
+// Copyright 2020-2024 Trust Computing GmbH.
+// This file is part of Litentry.
+//
+// Litentry is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


tee-worker/omni-executor/rpc-server/src/methods/omni/get_omni_account.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::ErrorCode;
+use crate::{methods::omni::common::PumpxRpcError, server::RpcContext};
+use executor_primitives::{utils::hex::ToHexPrefixed, Web2IdentityType};
+use heima_primitives::Identity;
+use jsonrpsee::{types::ErrorObject, RpcModule};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct GetOmniAccountParams {
+	pub user_email: String,
```
silva-fj, line 26: We should probably check in all methods that the email address is valid (Not in this PR)

silva-fj (approved):

0xverin (approved):

PR #3427, opened 2025-04-24
fix: P-1485 update all fields to optional in ...ResponseData of pumpx…

wli-pro:
… api to avoid decode error

### Labels

### Testing Evidences


BillyWooo:
Please remove the `Option` for the `data` field:
```
pub struct ApiResponse<T: Codec> {
	pub code: u32,
	pub message: String,
	pub data: Option<T>,
}
```

Besides, there is another new set of `omni` rpc calls. Basically they are copied over from `pumpx`. Please double check and also adapt the error code there.

BillyWooo (approved):
thanks.

PR #3429, opened 2025-04-24
Include Sender in OmniAuth 

silva-fj:
This pull request refactors the `OmniAuth` enum and updates its usage across the codebase to improve consistency in authentication handling, and separating Task execution from authentication. It also simplifies the `verify_auth` function by removing the `sender` parameter and adjusts related method calls accordingly.


silva-fj:
I will include al thesel changes in another PR

PR #3430, opened 2025-04-25
Set `Content-Length` explicitly for http request to pumpx API

BillyWooo:
as title. a tiny PR. 
Looks like the database has set up strict rules about it.
Any idea about it? @Kailai-Wang @silva-fj

### DIFF
--- a/tee-worker/omni-executor/pumpx/src/pumpx_api/mod.rs
+++ b/tee-worker/omni-executor/pumpx/src/pumpx_api/mod.rs
@@ -113,6 +113,7 @@ impl PumpxApi {
 			.post(endpoint)
 			.header("X-Language", language.unwrap_or("en".to_string()))
 			.bearer_auth(access_token)
+			.json(&serde_json::json!({}))
 			.send()
 			.await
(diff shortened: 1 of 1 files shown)


silva-fj (commented):
> Looks like the database has set up strict rules about it.

What do you mean? What error do you get?

tee-worker/omni-executor/pumpx/src/pumpx_api/mod.rs
```diff
@@ -113,6 +113,7 @@ impl PumpxApi {
 			.post(endpoint)
 			.header("X-Language", language.unwrap_or("en".to_string()))
 			.bearer_auth(access_token)
+			.json(&serde_json::json!({}))
```
kziemianek: If this needs to be merged please put a comment here so we know that this is something required and should not be removed in the future.

Kailai-Wang (approved):

Kailai-Wang:
Otherwise we get 411 error

kziemianek (approved):

PR #3431, opened 2025-04-25
OmniAccount auth rework

silva-fj:
In this PR we are refactoring the OmniAccount authentication mechanism without changing it's behaviour. The changes consist of simplifying the data structures and separation of concerns (putting together auth related data in `OmniAuth`.

It includes the following sub-tasks:
- [X]  https://linear.app/litentry/issue/P-1453/re-design-and-implement-the-msg-for-web3-authentication
- [X]  https://linear.app/litentry/issue/P-1454/include-all-information-in-omniauth
- [X] https://linear.app/litentry/issue/P-1456/jsoned-auth-struct

**Note:** The new json structs for authentication (including `Identity` and `HeimaMultiSignature`) have been included, but the params of the RPC methods have not been changed as we have some inconsistencies in the way we handle authentication, mostly because of some pumpx specific logic. We should refactor the params once we define the (generic) API we want to expose for other potential clients in a follow up PR.


tee-worker/omni-executor/rpc-server/src/methods/omni/get_message_code.rs
```diff
@@ -0,0 +1,34 @@
...
+			let params = params.parse::<GetMessageCodeParams>()?;
+			let omni_account = AccountId::from_str(&params.omni_account).map_err(|_| {
+				log::error!("Could not parse AccountId: {:?}", params.omni_account);
+				ErrorCode::InvalidParams
+			})?;
+			let verification_code_storage = VerificationCodeStorage::new(ctx.storage_db.clone());
+			let message_code = generate_otp(8);
+
+			verification_code_storage
+				.insert(&omni_account.hash(), message_code.clone())
```
kziemianek: My two concerns here:
* this is an only growing storage (old codes are not pruned - can be only overriden - it's relatively cheap for somebody to flood worker with prepared requests  - I haven't measured the impact)
* people can override code for each other

silva-fj: > this is an only growing storage (old codes are not pruned - can be only overriden - it's relatively cheap for somebody to flood worker with prepared requests - I haven't measured the impact)

This is a good point. I'll remove the entry after the code is verified.

> people can override code for each other

Should we make this idempotent until the code is used? If there is a code in storage we just return it instead of inserting a new one?

kziemianek: > This is a good point. I'll remove the entry after the code is verified.

Still unused codes remain in storage - if that could be an issue maybe we should have a cron job for cleaning them up

> Should we make this idempotent until the code is used? If there is a code in storage we just return it instead of inserting a new one?

perfect 🙃

silva-fj: > Still unused codes remain in storage - if that could be an issue maybe we should have a cron job for cleaning them up

Yeah it would be a good improvement

tee-worker/client-api/src/omni/interfaces/omniExecutor/definitions.ts
```diff
@@ -41,10 +41,10 @@ export default {
     },
     OmniAuth: {
       _enum: {
-        Web3: "(HeimaMultiSignature)",
+        Web3: "(Identity, HeimaMultiSignature)",
         Email: "(Text, Text)",
```
kziemianek, line 46: Can all variants contain `Identity` as a first value ?

tee-worker/omni-executor/executor-primitives/src/auth.rs
```diff
@@ -1,36 +1,217 @@
...
+/// A serializable representation of HeimaMultiSignature for JSON interchange.
+/// ```json
+/// {
+///   "type": "Ed25519",
+///   "data": "base64-encoded-signature"
+/// }
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(tag = "type", content = "data")]
+pub enum HeimaMultiSignatureSerde {
+	Ed25519(String),  // base64-encoded
```
kziemianek: I would suggest keeping hex encoding here.
These are short length byte arrays and I see no benefit from using base64, do you?

tee-worker/omni-executor/executor-primitives/src/auth.rs
```diff
@@ -1,36 +1,217 @@
...
+}
+
+/// A serializable representation of HeimaMultiSignature for JSON interchange.
+/// ```json
+/// {
+///   "type": "Ed25519",
+///   "data": "base64-encoded-signature"
+/// }
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(tag = "type", content = "data")]
```
kziemianek: If you move this serde macro to `HeimaMultiSignature` and enable `sp-core` `serde` feature then you can json-serialize `HeimaMultiSignature` directly to:

```
{"type":"Ed25519","data":"3e1994ba3589f8ae95bbe118ba30186d641b95c44205de8c161088ef9a1685604f02b46a9670740b0623200491f03682cec1c839f170237ae261aee7dd0d6202"}
```

If we can stick to hex encoding maybe `HeimaMultiSignatureSerde` is not needed at all?

kziemianek (approved):
It looks fine. I've just left some cosmetic comments.

tee-worker/omni-executor/executor-primitives/src/auth.rs
silva-fj: Done ✅

tee-worker/client-api/src/omni/interfaces/omniExecutor/definitions.ts
silva-fj: They could, one thing is a bit annoying is that we'll have to add conversions/checks for the identity. But we can revisit this when we actually start using the serde types (so the clients send json objects) instead of these internal types

PR #3432, opened 2025-04-28
pumpx api trait

kziemianek:
Extracts `PumpxApi` trait defining api interface for interacting with pumpx backend. 
This will bring us closer to having `CrossChainIntentExecutor` unit tests.


Kailai-Wang (approved):
Great, thanks. CI test needs to be fixed

PR #3436, opened 2025-04-29
Split pumpx RPC struct definition per file

0xverin:
resolve https://linear.app/litentry/issue/P-1459/split-pumpx-rpc-struct-definition-per-file


tee-worker/omni-executor/pumpx/src/pumpx_api/pumpx_types/common.rs
```diff
@@ -0,0 +1,56 @@
+use parity_scale_codec::{Codec, Decode, Encode};
```
kziemianek: I though we want to make pumpx-api `scale-codec` agnostic. Why we need this ?

BillyWooo (commented):
I think it's fine to skip the xxx_impl layer. e.g. move `add_wallet` directly inside add_wallet.rs. The same apply to other functions.

BillyWooo (approved):

0xverin:
> I think it's fine to skip the xxx_impl layer. e.g. move `add_wallet` directly inside add_wallet.rs. The same apply to other functions.

This is a trait that cannot be skipped.

PR #3438, opened 2025-05-05
Upgrade polkadot-sdk to `stable2412`

BillyWooo:
update polkadot-sdk to `stable2412`. 

Note: 
Polkadot-sdk@stable2412 has introduced fork_aware_txpool. Together with existed single_state_txpool(BasicPool), there is a new wrapper introduced: `TransactionPoolWrapper` to wrap these two types. 
But in our node/src/service.rs, we don't use fatxpool yet. There is a workaround to continue use BasicPool. 

A following issue to tackle fork_aware_txpool: [P-1546: use `fork_aware_txpool`](https://linear.app/litentry/issue/P-1546/use-fork-aware-txpool).


tee-worker/client-api/parachain-api/src/lookup.d.ts
```diff
@@ -0,0 +1,2 @@
+declare type Lookup76 = any;
```
0xverin: For xcmv5 support, polakdot.js-api devs have submitted a [PR](https://github.com/polkadot-js/api/pull/6142).

0xverin: FYI https://github.com/polkadot-js/api/releases/tag/v15.10.2
I think you could update polkadot.js realted deps first. After that, delete the declaration file and regenerate the types to try again.

parachain/Cargo.toml
```diff
@@ -320,3 +318,9 @@ pallet-investing-pool = { path = "pallets/collab-ai/investing-pool", default-fea
 
 [patch.crates-io]
 ring = { git = "https://github.com/betrusted-io/ring-xous", branch = "0.16.20-cleanup" }
+
```
Kailai-Wang, line 321: As we updated the `webpki` - do we still need this `ring` override?

parachain/Cargo.toml
```diff
@@ -320,3 +318,9 @@ pallet-investing-pool = { path = "pallets/collab-ai/investing-pool", default-fea
...
+[patch."https://github.com/rust-ethereum/evm"]
+evm = { git = "https://github.com/litentry/evm", rev = "e81732d6bb47e3d3d68d233e43919c4522598361" }
```
Kailai-Wang, line 323: Can we avoid our fork? Do we have our own change?

parachain/precompiles/omni-bridge/src/lib.rs
```diff
@@ -82,7 +84,7 @@ where
 			};
 
 		let call = pallet_omni_bridge::Call::<Runtime>::pay_in { req: pay_in_request };
-		RuntimeHelper::<Runtime>::try_dispatch(handle, Some(origin).into(), call)?;
+		RuntimeHelper::<Runtime>::try_dispatch(handle, Some(origin).into(), call, 0)?;
```
Kailai-Wang, line 87: Do you know if `storage_growth` needs to be benchmarked, or are we fine with always 0?

parachain/runtime/heima/Cargo.toml
```diff
@@ -5,6 +5,7 @@ name = 'heima-runtime'
 version = '0.1.0'
 
 [dependencies]
+getrandom = { version = "0.2.15", default-features = false, features = ["js"] }
```
Kailai-Wang, line 8: This doesn't seem to be right (although it compiles)

I assume that's because some crate pulled in std deps also in a wasm target - I highly recommend we resolve that instead of adding this crate with `js` feature

parachain/runtime/heima/src/lib.rs
```diff
@@ -227,16 +227,16 @@ impl_opaque_keys! {
...
+	impl_name: alloc::borrow::Cow::Borrowed("heima"),
 	authoring_version: 1,
 	// same versioning-mechanism as polkadot: use last digit for minor updates
 	spec_version: 9242,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 2,
 	// https://hackmd.io/JagpUd8tTjuKf9HQtpvHIQ
-	state_version: 1,
+	system_version: 1,
```
Kailai-Wang, line 239: Does it need any runtime upgrade?

Kailai-Wang (commented):
Looks good in general, thanks!

I mainly have concerns of pulling in `getrandom` crate manually

parachain/Cargo.toml
BillyWooo: We don't have local changes. Mainly it's because we want this fixed commit. Newer version evm has changes, which requires changes in `frontier` too.

parachain/precompiles/omni-bridge/src/lib.rs
BillyWooo: We are fine with `0`. But I think it's worth to do a new benchmark. There is [some strange error](https://github.com/litentry/heima/actions/runs/15262942876/job/42924229289#step:8:1641). Still looking into it.

parachain/runtime/heima/src/lib.rs
BillyWooo: Yes. But I think we need to do upgrade anyway after a new release, right?

Kailai-Wang: Sorry I meant runtime migration

parachain/Cargo.toml
Kailai-Wang: We can set the `rev` on the original git url too - so it seems we don't have to fork it

parachain/Cargo.toml
BillyWooo: We still need it. Otherwise there will be `duplicate lang item` error.

parachain/Cargo.toml
BillyWooo: `patch` doesn't allow it to set to original git url. Because the url is the same.

Kailai-Wang: I'm a bit confused - then why do we need patch? 😂 
Why not use the original `evm` crate? (and following crates too)

Kailai-Wang (approved):
Let's get it merged first

Will address issues with follow-up PRs

PR #3441, opened 2025-05-08
fix: sign the unsigned tx within omni-executor

felixfaisal:
So this should allow us to
1. Get the unsigned tx from Backend 
2. Have the unsigned tx signed from Signer 
3. Construct signed tx with the signature 
4. send the order tx to Backend 

A major assumption, I've made here since i could not test is that, `marshallBinary()` in Golang `RLPEncodes` the transaction, thereby which it should be easy to decode into a `TxLegacy` as `alloy` has implemented `RLPDecodable` and `RLPEncodable`, So if this assumption stands then this code should work seamlessly. 

While this PR gets reviewed, i will perform a miniature test to see what `marshallBinary()` gives us is indeed what we expect by trying to decode it into `TxLegacy`, and vice versa. It'd be great if someone already has some fixtures i can test again. 

This should most likely break the unit tests if it's been merged already, i'll fix it in this or in the follow up PR 😅

### DIFF
--- a/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
+++ b/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
@@ -1041,6 +1041,7 @@ impl<BinanceClient: BinanceApi> IntentExecutor for CrossChainIntentExecutor<Bina
 						},
 						slippage: pumpx_config.slippage,
 						wallet_index: pumpx_config.wallet_index,
+						recipient_address: String::from("")
 					};
 					debug!("Calling pumpx create_market_order_tx, body: {:?}", body);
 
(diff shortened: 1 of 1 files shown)


tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -1074,6 +1075,69 @@ impl<BinanceClient: BinanceApi> IntentExecutor for CrossChainIntentExecutor<Bina
...
+		account_id: &AccountId,
+	) -> Result<SendOrderTxResponse, ()> {
+		let wallet_index = order.wallet_index.clone();
+		let response = self
+			.pumpx_api
+			.create_market_order_unsigned_tx(access_token, order)
+			.await
+			.map_err(|_| log::error!("Failed to get unsigned market order tx"))?;
+
+		let unsigned_tx_string = response.data.tx_data.ok_or_else(|| ())?;
```
kziemianek: Can we have an error logged here ? In other places too.

kziemianek (approved):

Kailai-Wang:
Can we also get the CI passed? Then we're good to test it : )

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
felixfaisal: I've logged errors

PR #3443, opened 2025-05-12
Cross chain swap instant payout

kziemianek:
For small trade amounts trade from accounting contract account to user's account so it's faster because there is no need to trade and withdraw tokens from Binance.


Kailai-Wang:
Would you resolve the conflict? 😂  Then I'll look into it 🙌

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
```diff
@@ -97,20 +119,72 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
...
+			omni_account,
+			swap_order.from_asset.clone(),
+			from_amount_decimal,
+			from_address.clone(),
+			pumpx_config.wallet_index,
+			should_wait_for_deposit_confirm,
+		)
+		.await?;
+
+		if instant {
```
silva-fj: Why do we only lock and release assets (before and after binance deposit) if `instant` ?

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
```diff
@@ -97,20 +119,72 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
...
+				).map_err(|e| 	log::error!("Could not release asset lock for account: {:?}, amount: {:?}, asset: {:?}, reason: {:?}", account_id, swap_order.from_asset, amount_to_release, e))?;
+		}
 
-		self.do_payout(&payout_address, &payout_amount_u256).await?;
+		let mut payout_amount =
+			self.estimate_payout(swap_order.from_asset.clone(), from_amount_decimal).await?;
+		let mut payout_amount_u256 = str_to_u256(&payout_amount, 18)?;
+
+		// trade only if deposit is confirmed
+		if should_wait_for_deposit_confirm {
```
silva-fj: Is this correct? It looks like we'll never do the binance trade

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
```diff
@@ -97,20 +119,72 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
...
+				.request_wallet(to_chain_type, pumpx_config.wallet_index, omni_account)
+				.await
+				.map_err(|e| error!("Could not get to_wallet from pumpx-signer: {:?}", e))?;
+
+			let to_address = pubkey_to_address(to_chain_type, &to_wallet)?;
+
+			let payout_address = Address::from_str(&to_address).map_err(|_| {
+				error!("Failed to parse payout address");
+			})?;
+			self.do_payout(&payout_address, &payout_amount_u256).await?;
```
silva-fj: What does `instant` mean? should we do the payout if `instant` is true?

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
```diff
@@ -361,126 +389,270 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
...
+
+		// init `payout_amount` with estimated-amount-to-receive
+		let payout_amount =
+			estimate_bnb_amount(&self.binance_api, &trade_symbol, &binance_coin_name, from_amount)
+				.await?;
+		Ok(payout_amount)
+	}
+
+	// TODO: move `should_wait_for_deposit_confirm` to provider config
+	async fn do_binance_trade(
```
silva-fj: This comments seems to be outdated. `should_wait_for_deposit_confirm` is not used in this function

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: For instant flow we do not trade assets and withdraw them from binance to user's wallet. 
Once they are deposited they are controlled by binance account owner -> so lock can be released.

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: This is the current flow:
https://github.com/litentry/heima/blob/dev/tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs#L314

In my opinion it makes sense -> you simply have to wait for deposit confirmation before making any trade.

But i'm not sure whether it's correct - i tried to keep it unchanged.

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: There is not payout for instant flow (hovewer PR title and requirements description in linear issue may be confusing).
Accounting contract is used as source account for single chain swap.

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
silva-fj: Yeah I got that. My question is, do we need to lock assets if the flow is not instant?

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: I decided to kept it https://github.com/litentry/heima/blob/c95859a569531175a04d1f18c6d21cc9477ba41a/tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs#L158

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: For non instant flow where assets needs to be deposited -> traded -> withdrawn there is no way to spent more than currently owning.

For instant flow accounting contract is used as a source account for single chain swap so user could submit multiple swap intents and total amount traded could eventually exceed his balance.

kziemianek: Currently we do not need to lock assets for non instant flow.

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
silva-fj: I was referring to how we set the value a few the lines above:
```rust
// TODO: this should be abstracted away
// todo we only transfer to binance without swap

let should_wait_for_deposit_confirm = false;
```
We never change this value so we'll never actually do the binance trade. Is it a placeholder? I don't remember this in the original implementation

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
silva-fj: The TODO is ok in the other functions, what I mean is that `should_wait_for_deposit_confirm` is not used in `do_binance_trade`, so the comment is confusing

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: It's hardcoded to `false` on `dev` branch 
https://github.com/litentry/heima/blob/c95859a569531175a04d1f18c6d21cc9477ba41a/tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs#L107

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: But it's moved in this PR (there is only one place where this todo exists), do you have any idea which place would be better to keep it ?

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
silva-fj: I see, I guess it is some sort of placeholder then

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
silva-fj: I think it would be a good idea to put it where `should_wait_for_deposit_confirm` is defined. What I don't know is when/how this value will change and when are we doing the binance trades

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: I'll keep it as it is.

tee-worker/omni-executor/intent/asset-lock/src/lib.rs
```diff
@@ -45,15 +52,25 @@ impl<AL: AssetsLock> AccountAssetLocks<AL> {
...
 		available_amount: AmountType,
 	) -> Result<(), ()> {
-		let mut account_lock = self.locks.write().unwrap();
-
-		if let Some(account) = account_lock.get_mut(&account_id) {
-			account.lock(asset_id, amount_to_lock, available_amount)?;
+		let key = Key { account_id: account_id.clone(), asset_id: asset_id.clone() };
+		if let Some(account_lock) = self
+			.storage
+			.get(&key)
```
Kailai-Wang, line 59: Not in this PR - in the long term do you prefer using substrate's way of constructing storage keys in general? Like copying pallet's storage - that uses hashed keys under the hood

I think it makes the key-length consistent and thus a better search speed. It probably only makes a difference when we have larger DB tho 

cc @silva-fj too

silva-fj (approved):

tee-worker/omni-executor/executor-storage/src/lib.rs
```diff
@@ -24,6 +24,9 @@ mod pumpx_jwt;
 pub use pumpx_jwt::PumpxJwtStorage;
 mod intent_id;
 pub use intent_id::IntentIdStorage;
+mod asset_lock;
+pub use asset_lock::AssetLockStorage;
+pub use asset_lock::Key;
```
Kailai-Wang: Can we make `Key` not directly visible from the crate root?  It would be a bit too general (while this `Key` is actually specific to AL)

tee-worker/omni-executor/intent/asset-lock/src/precise.rs
```diff
@@ -14,273 +14,255 @@
...
-use std::collections::HashMap;
+use parity_scale_codec::{Decode, Encode};
+
+use crate::{AmountType, AssetsLock};
 
 // Manages asset locks per account. Used for precise tracking of currently locked assets.
+#[derive(Encode, Decode)]
 pub struct PreciseAssetsLock {
-	locked_assets: HashMap<AssetId, AmountType>,
+	amount: AmountType, // locked_assets: HashMap<AssetId, u64>,
```
Kailai-Wang: Is the comment correct? It seems to be a scalar type

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
```diff
@@ -53,36 +63,48 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
...
 			error!("Unsupported to_chain_id: {}", pumpx_config.to_chain_id);
 			return Err(());
 		};
 
-		let from_wallet = self
-			.pumpx_signer_client
-			.request_wallet(from_chain_type, pumpx_config.wallet_index, omni_account)
-			.await
-			.map_err(|e| error!("Could not get from_wallet from pumpx-signer: {:?}", e))?;
+		let from_asset_binance_coin_name = determine_binance_coin_name(&swap_order.from_asset)?;
```
Kailai-Wang: This probably needs to be resolved with Will's PR (depending on which PR is merged first)

I suggested to have binance related traits and implement `ChainAsset` for it

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -268,16 +255,28 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait> IntentExecutor
...
 
-				// self.account_asset_lock.release(
-				// 	account_id.clone(),
-				// 	swap_order.from_asset.clone(),
-				// 	AmountType::from_str_radix(&from_amount_string, 10).map_err(|_| {
-				// 		log::error!("Failed to parse from_amount_string");
-				// 	})?,
-				// )?;
+				// depost here and unlock assets
+				if let Some(details) = instant_flow_details {
```
Kailai-Wang, line 261: Hmm this is still synchronous - can we push the task into a queue which is processed in (another) dedicated thread and return as early as possible?

Does it make sense?
Not necessarily in this PR

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
```diff
@@ -96,21 +118,78 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
...
-		let (payout_amount, payout_amount_u256) = self
-			.do_binance_swap(
+		let mut payout_amount =
+			self.estimate_payout(swap_order.from_asset.clone(), from_amount_decimal).await?;
+		let mut payout_amount_u256 = str_to_u256(&payout_amount, 18)?;
+
+		if instant {
+			Ok((
+				self.apply_gas_fee(&payout_amount, access_token, pumpx_config).await?,
+				self.accounting_contract_client.get_address().await.to_string(),
```
Kailai-Wang: I'm not sure that's correct

This address will be used as `address` field when calling `create_market_order_unsigned_tx` and it has to be an EOA (able to sign), but here it seems the accounting contract address :(

Kailai-Wang (changes requested):
Thanks

I think one thing might be wrong is the contract address should be used in backend API call.. could you double-check it?

Others are merely some suggestions (especially the async deposit)

tee-worker/omni-executor/executor-storage/src/lib.rs
kziemianek: The root cause of problem is that all domain specific storages are put into single crate.

I think asset-lock storage should be located within `intent-asset-lock` lock crate -> so key doesn't leak.

Currently we have following `Storage` trait definition: 

```
pub trait Storage<K: Encode, V: Codec> {
	fn get(&self, key: &K) -> Result<Option<V>, ()> {
```

So key type needs to be known both by this crate and code which is using this interface.

There are three ways to solve this:
1. make key public (as it is in this PR)
2. duplicate key type definition (in client crate - this is the shortest path to achieve what you are asking here)
3. move `AssetLockStorage` to `intent-asset-lock` crate so both logic + storage are located withing one crate responsible for single feature (asset locks) -> but I've done that previously with `IntentIdStorage` and it was moved by somebody else to `executor-storage` storage in another PR -> so I assume we don't want that but I prefer it.

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain.rs
kziemianek: should be resolved

tee-worker/omni-executor/executor-storage/src/lib.rs
silva-fj: I'm ok with both approaches (keep all storages in a single crate or separate them into their specific domains). I think the issue that @Kailai-Wang is referring to is that the name `Key` is too generic, you could export it as `AssetLockStorageKey` so it is specific to `AssetLockStorage`

tee-worker/omni-executor/intent/asset-lock/src/lib.rs
silva-fj: How we construct the keys does not make much of a difference in the way we use the database, we simply use it a as key value storage. Substrate adds a few layers on top, in a pallet storage it matters because they put the the data in a data structure called Patricia Trie to reduce the space used, and having hashed keys (with prefixes) helps to keep the trie balanced.

tee-worker/omni-executor/executor-storage/src/lib.rs
kziemianek: I've changed exported name.

tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
kziemianek: It should make intent processing faster - we can add such improvement in separate PR.

kziemianek: I've managed to do that in this PR - should be resolved now.

tee-worker/omni-executor/executor-storage/src/lib.rs
Kailai-Wang: Yes I merely meant the exported names (could be misused if being referenced from other crates)

Kailai-Wang (approved):
Looking good - thanks!

PR #3444, opened 2025-05-12
P-1499 Separate the types of client api

0xverin:
context:https://litengtechnologies.slack.com/archives/C05328L0GAF/p1743191470803999

This PR:
Builds the packaging of the `identity` `omni` `sidechain-api` within the client-api.

This can solve the problem of type conflicts (now any type name can be named in `omni`, even the same as `identity`), but if there is a type associated with `identity` in `omni`, we still need to define it once in `omni`.

When using three types, distinguish them by the path, for example:
```
import {Web3Network} from “@heima-network/api-argument/identity”
import { MemberAccount } from '@heima-network/api-argument/omni’;
```
[usage](https://github.com/litentry/heima/blob/d7fa3b5d61016250429d52bdfcd2cf9ab07db19f/tee-worker/client-api/README.md)


tee-worker/omni-executor/ts-tests/integration-tests/package.json
```diff
@@ -38,9 +38,9 @@
         "micro-base58": "^0.5.1",
         "mocha": "^10.6.0",
         "mocha-steps": "^1.3.0",
-        "parachain-api": "file:../../../client-api/parachain-api",
+        "parachain-api": "file:../../../client-api/src/identity",
```
0xverin: I will make corrections this path(e.g. `"@heima-network/api-argument/identity":"file:..."`). For now, let's see if CI can pass the test.

tee-worker/client-api/src/identity/interfaces/identity/definitions.ts
```diff
@@ -0,0 +1,154 @@
...
+        Discord: "IdentityString",
+        Github: "IdentityString",
+        Substrate: "Address32",
+        Evm: "Address20",
+        Bitcoin: "Address33",
+        Solana: "Address32",
+        Email: "IdentityString",
+      },
+    },
+    LitentryIdentity: "Identity",
```
silva-fj, line 34: Should we rename this and other types to `Heima*` ?  (Not as part of this PR)

tee-worker/client-api/src/identity/interfaces/trusted_operations/definitions.ts
```diff
@@ -0,0 +1,222 @@
...
+        Finalized: null,
+        Usurped: null,
+        Dropped: null,
+        Invalid: null,
+        TopExecuted: "Bytes",
+      },
+    },
+
+    /**
+     * Native tasks (OmniAccount) results
```
silva-fj, line 151: Native tasks are not really used in the identity worker, we should probably clean them up and remove these types (In a separate issue/PR)

tee-worker/omni-executor/ts-tests/integration-tests/utils/signer.ts
```diff
@@ -4,7 +4,7 @@ import { bufferToU8a, hexToU8a, isString, stringToU8a, u8aToHex } from '@polkado
 import { KeyringPair } from '@polkadot/keyring/types';
 import { blake2AsU8a } from '@polkadot/util-crypto';
 import { ApiPromise } from '@polkadot/api';
-import { Identity } from 'parachain-api';
+import { Identity } from '@heima-network/api-argument/identity';
```
silva-fj: Should this come from `@heima-network/api-argument/omni` ?  or maybe I misunderstood the separation

0xverin: You didn't misunderstand; I made a mistake in my writing. :)

tee-worker/client-api/src/identity/interfaces/trusted_operations/definitions.ts
0xverin: Ok, I also need to confirm this here, and then organize all types.

tee-worker/client-api/src/identity/interfaces/identity/definitions.ts
0xverin: https://linear.app/litentry/issue/P-1514/replace-litentry-with-heima-in-client-api-definitions

tee-worker/client-api/src/identity/interfaces/trusted_operations/definitions.ts
silva-fj: What I meant is that we (partially) implemented native tasks in the identity worker, but nobody is using this as we moved to the omni-executor worker. So, I think we can clean that up eventually.

silva-fj (approved):

Kailai-Wang:
A dumb Q:
can `heima-network/api-argument/omni` types work without any dependency (direct or indirect) to `identity` types?

0xverin:
> A dumb Q: can `heima-network/api-argument/omni` types work without any dependency (direct or indirect) to `identity` types?

No,we can't do this, any non-basic type must have a defined source(omni types depend on identity types), which needs to be established during the type generation process.

PR #3446, opened 2025-05-14
fix: hash the msg before requesting signer

felixfaisal:
So as per the title, 
I think i missed this part that the signer expects hashed messages, this should fix the signer errors we received.

### DIFF
--- a/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
+++ b/tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
@@ -34,6 +34,7 @@ use heima_authentication::auth_token::AUTH_TOKEN_ACCESS_TYPE;
 use rust_decimal::prelude::*;
 use rust_decimal::Decimal;
 use solana::{signer::RemoteSigner, SolanaClient as SolanaClientTrait};
+use sp_core::keccak_256;
 use std::str::FromStr;
 use tokio::{
 	runtime::Handle,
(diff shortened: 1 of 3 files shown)


tee-worker/omni-executor/intent/executors/cross-chain/src/lib.rs
```diff
@@ -1110,7 +1111,8 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
 			.iter()
 			.map(|s| {
 				let hex_str = s.trim_start_matches("0x");
-				hex::decode(hex_str).expect("Invalid hex string")
+				let hex_decoded = hex::decode(hex_str).expect("Invalid hex string");
+				keccak_256(&hex_decoded).to_vec()
```
Kailai-Wang, line 1115: hold on, in the `request_signatures` impl, there's already:

```
		let signature =
			self.request_signer.sign_prehashed(&keccak_256(&serialized_multi_sign_wallet)).0;
```

could you please verify if the hashing is still needed?

felixfaisal: the hashing is being done here only for signing(to get signature), but the payload will still have the messages without hashing being done i.e. it is being passed as it is.

felixfaisal: Code ref 
https://github.com/litentry/heima/blob/4d2b4e415b33158f157178a7545c8fd0f0f8b4d5/tee-worker/omni-executor/pumpx/src/signer_client.rs#L237

Kailai-Wang (approved):
Thanks, let's get it merged and tested

PR #3448, opened 2025-05-14
Fix ts-related high/critical dependabot warning

0xverin:
dependabot warning is mainly due to
1. deps in ts need to be updated
2. different deps in ts use the same peerdeps, but the versions are inconsistent(some are outdated.), requiring the use of overrider to force override

### DIFF
--- a/parachain/scripts/ts-utils/package.json
+++ b/parachain/scripts/ts-utils/package.json
@@ -1,0 +1,26 @@
+{
+    "name": "ts-utils",
+    "version": "1.0.0",
+    "description": "",
+    "main": "index.js",
+    "scripts": {
+        "format": "pnpm exec prettier --write '**.ts'"
+    },
+    "keywords": [],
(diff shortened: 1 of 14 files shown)


parachain/scripts/ts-utils/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "license": "ISC",
+    "dependencies": {
+        "@polkadot/api": "^15.10.2",
+        "@polkadot/util-crypto": "^13.0.2",
+        "colors": "^1.4.0",
+        "ethers": "^6.13.2",
+        "exceljs": "^4.3.0",
+        "prettier": "^2.8.7",
+        "ts-node": "^10.9.1",
+        "typescript": "^5.0.4"
```
silva-fj, line 23: NIT: These last dependencies can be moved to `devDependencies`

silva-fj (approved):

PR #3450, opened 2025-05-14
Refactor swap code logic and fix single-chain market order

Kailai-Wang:
### Context

As topic

The PR tries to split the big `CrossChainIntentExecutor::execute` into smaller logic units/files, so that it's more readable and easy to maintain.

It populates the `recipient_address` in every case too.

There will be a few more readability improvements incoming, but much smaller than this one.

### DIFF
--- a/tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain_swap_tests.rs
+++ b/tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain_swap_tests.rs
@@ -102,7 +102,7 @@ async fn simple_cross_chain_swap() {
 			mockall::predicate::eq(pumpx_wallet_index),
 			mockall::predicate::eq(pumpx_wallet_omni_account),
 		)
-		.times(1)
+		.times(2)
 		.returning(|_, _, _| {
 			Ok([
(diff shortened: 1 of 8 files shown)


tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain_swap_tests.rs
```diff
@@ -102,7 +102,7 @@ async fn simple_cross_chain_swap() {
 			mockall::predicate::eq(pumpx_wallet_index),
 			mockall::predicate::eq(pumpx_wallet_omni_account),
 		)
-		.times(1)
+		.times(2)
```
kziemianek, line 105: This PR introduced one redundant call.

kziemianek (approved):
Let's merge it.

tee-worker/omni-executor/intent/executors/cross-chain/src/cross_chain_swap_tests.rs
Kailai-Wang: True - I'll improve it further in the upcoming PRs where I put the common value parsing ahead

PR #3452, opened 2025-05-15
Fix ts-related high/critical dependabot warning one more

0xverin:
According to the dependabot alerts, override some affectedversion deps.  
e.g.
image

solve:https://github.com/litentry/heima/compare/fix-dependbot-one-more?expand=1#diff-d674fa433b003321c681bec456cdd2259d05883a3cfa06c49021c3d101f10772R15


Kailai-Wang:
It seems to cause some runtime errors

0xverin:
> It seems to cause some runtime errors

Upgrading base-x to version 5.x will cause compatibility issues. Downgrading to version 4.0.1 to fix.

silva-fj (approved):

PR #3453, opened 2025-05-15
Include omni-executor's mrenclave in release notes

silva-fj:
This pull request introduces functionality to extract the `MRENCLAVE` value in the `omni-executor`, along with some code refactoring. The changes primarily focus on enabling the `MRENCLAVE` value to be extracted, passed through workflows, and included in release notes. Additionally, a new CLI command and helper function are added to support this feature.


.github/workflows/create-release-draft.yml
```diff
@@ -366,7 +377,9 @@ jobs:
 
       - name: Generate release notes
         run: |
+          echo "MRENCLAVE: ${{ needs.build-omni-executor.outputs.mrenclave }}" > mrenclave.txt
           ./parachain/scripts/generate-release-notes.sh ${{ github.workspace }}/.github/release_notes.md ${{ needs.set-release-type.outputs.release_type }} ${{ env.DIFF_TAG }}
+          cat mrenclave.txt >> ${{ github.workspace }}/.github/release_notes.md
```
Kailai-Wang: I believe the end of it isn't the right place - you could check the release notes so far, it always has a list of PRs.

You have to really adjust `generate-release-notes.sh` to insert it into the right place

tee-worker/omni-executor/parentchain/attestation/src/lib.rs
```diff
@@ -20,39 +21,50 @@ type ParentchainTxSigner = TxSigner<
...
-	client_factory: Arc<SubxtClientFactory<CustomConfig>>,
-	signer: Keypair,
-	transaction_signer: Arc<ParentchainTxSigner>,
-	worker_url: &str,
-	shielding_pubkey: Vec<u8>,
-) -> Result<MrEnclave, ()> {
-	let mut quote = vec![];
-	let mut attestation_type = AttestationType::Dcap(DcapProvider::Intel);
-	let mut mrenclave = MrEnclave::default();
+pub async fn get_attestation_data(signer: Keypair) -> Result<(Vec<u8>, MrEnclave), ()> {
```
Kailai-Wang: Hmm it's just a getter - why do we have writing operations like

```
		let mut f = File::create("/dev/attestation/user_report_data").unwrap();
		let content = signer.public_key().0;
		f.write_all(&content).unwrap();
```

Is the `signer` actually required?

BillyWooo:
I tried it out. Errors. Send you in DM.

.github/workflows/create-release-draft.yml
silva-fj: I moved the logic to generate-release-notes.sh script

tee-worker/omni-executor/parentchain/attestation/src/lib.rs
silva-fj: This is specific to gramine, you need to open the `user_report_data` and write into it, under the hood, gramine uses the EREPORT hardware instruction to generate an [SGX Report](https://gramine.readthedocs.io/en/stable/sgx-intro.html#term-SGX-Report). Afterwards the quote file can be opened for read. 

This is the report we sent to the parachain so we include the signer

silva-fj:
> I tried it out. Errors. Send you in DM.

I tested the new command with gramine and without gramine. Both work for me

kziemianek:
Have you considered using gramine tools for extracting mrenclave from `*.sig` file ?

```
gramine-sgx-sigstruct-view omni-executor.sig 
Attributes:
    mr_signer: <signer_key>
    mr_enclave: 76a0513d375050995d838d132cd402b8382155ed91fc121504047296e116ec2f
    isv_prod_id: 0
    isv_svn: 0
    debug_enclave: False
```

silva-fj:
> Have you considered using gramine tools for extracting mrenclave from `*.sig` file ?
> 
> ```
> gramine-sgx-sigstruct-view omni-executor.sig 
> Attributes:
>     mr_signer: <signer_key>
>     mr_enclave: 76a0513d375050995d838d132cd402b8382155ed91fc121504047296e116ec2f
>     isv_prod_id: 0
>     isv_svn: 0
>     debug_enclave: False
> ```

I was not aware of this, what's the benefit? how do we generate the `*.sig` file?

kziemianek:
It's created by `gramine-sgx-sign` during build process so:
* you don't need to run binary in order to get mrenclave
* you can use already present tools to extract mrenclave instead of adding logic to binary

Kailai-Wang:
`gramine-sgx-sigstruct-view omni-executor.sig` sounds like a robuster way to me 🙌

parachain/scripts/generate-release-notes.sh
```diff
@@ -209,12 +209,15 @@ if is_omni_executor_release; then
   WORKER_VERSION=$(grep '^version' tee-worker/omni-executor/executor-worker/Cargo.toml | head -n1 | sed -E 's/^version *= *["'\''](.*)["'\'']/\1/')
   WORKER_BIN=$(grep '^name' tee-worker/omni-executor/executor-worker/Cargo.toml | head -n1 | sed -E 's/^name *= *["'\''](.*)["'\'']/\1/')
   WORKER_RUSTC_VERSION=$(cd tee-worker/omni-executor && rustc --version)
+  docker run litentry/omni-executor:$OMNI_EXECUTOR_DOCKER_TAG print-mr-enclave > mrenclave.txt
```
BillyWooo, line 212: this line doesn't output mrenclave value. Please double check it.

silva-fj:
I'll open a new PR for this

kziemianek:
MRENCLAVE is also printed during build 

```
Measurement:
    d62bcc583c066b1f1ac96a848aeaa53d182a050c3aa62b553e475dcf07362775
```

PR #3454, opened 2025-05-16
change accounting contract monitored address

kziemianek:
It should monitor signer address, not contract address.


BillyWooo (approved):
LGTM.

PR #3456, opened 2025-05-17
fix: signing sol tx from within worker

felixfaisal:
This PR as title says it is 
A few things to note: 
1. Solana requires multiple signatures depending on the instructions within the message, but since we pretty much request for the same signature, we're filling the placeholders with the same signature similar to what pumpx is doing here - https://github.com/dexs-k/dexs-backend/blob/9404cdb153b0a412d198874b2df40fd5d2891779/apps/trade/internal/chain/solana/txManager.go#L364
2. We don't have to prehash the message since the `signer` expects a raw message and hashes it itself before signing it 
3. I was considering why not use `solana_sdk` primitives for signing in the `TEE Signer` instead of `sp_core` as it would be more consistent.


tee-worker/omni-executor/intent/executors/cross-chain/src/single_chain.rs
```diff
@@ -42,36 +42,18 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
 
 				// evm market order => worker constructs, signs and sends it
 				// solana market order => pumpx (backend) constructs, signs and sends it => TODO
```
silva-fj: should we update these comments?

tee-worker/omni-executor/intent/executors/cross-chain/src/single_chain.rs
```diff
@@ -272,6 +294,68 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
...
+		omni_account: [u8; 32],
+		order_id: u32,
+		chain_id: u32,
+	) -> Result<Vec<u8>, ()> {
+		// Note: Based on the go code, It is going to be mostly a single Transaction
+		let unsigned_tx: Vec<solana_sdk::transaction::Transaction> = unsigned_tx_string
+			.iter()
+			.map(|tx| {
+				let unsigned_tx_bytes = hex::decode(tx.trim_start_matches("0x")).unwrap();
+				bincode::deserialize(&unsigned_tx_bytes[..]).expect("Failed to decode solana tx")
```
silva-fj: How about if we handle these errors?
```rust
let unsigned_tx: Vec<solana_sdk::transaction::Transaction> = unsigned_tx_string
    .iter()
    .map(|tx| {
	let unsigned_tx_bytes = hex::decode(tx.trim_start_matches("0x"))
	    .map_err(|_| log::error!("Invalid hex string"))?;
	bincode::deserialize(&unsigned_tx_bytes[..]).map_err(|_| {
	    log::error!("Failed to deserialize Solana transaction");
	})
    })
    .collect::<Result<Vec<_>, _>>()?;
```

tee-worker/omni-executor/intent/executors/cross-chain/src/single_chain.rs
```diff
@@ -112,55 +94,55 @@ impl<BinanceClient: BinanceApi, SolanaClient: SolanaClientTrait>
...
-		debug!("Calling pumpx create_market_order_tx, body: {:?}", body);
-		let response =
-			self.pumpx_api.create_market_order_tx(access_token, body).await.map_err(|_| {
-				log::error!("Failed to create market order tx");
-			})?;
-
-		debug!("Response create_market_order_tx: {:?}", response);
-		Ok(response.encode())
-	}
+	// async fn pumpx_do_market_order(
```
silva-fj: Are we planning to use this later? Can this be removed?

Kailai-Wang:
> I was considering why not use solana_sdk primitives for signing in the TEE Signer instead of sp_core as it would be more consistent

Probably because ed25519 is also one of the native substrate signing schemes, so no additional crate was introduced..

tee-worker/omni-executor/intent/executors/cross-chain/src/single_chain.rs
Kailai-Wang: Yes please : )

tee-worker/omni-executor/intent/executors/cross-chain/src/single_chain.rs
Kailai-Wang: I think either we remove it completely - or just keep it uncommented. I think this method could still be used to quickly verify that if it's backend problem or not (e.g. on staging)

Kailai-Wang (commented):
Thank you - could you resolve Frans' comments too?

tee-worker/omni-executor/intent/executors/cross-chain/src/single_chain.rs
felixfaisal: Updated ✅

tee-worker/omni-executor/intent/executors/cross-chain/src/single_chain.rs
felixfaisal: Done ✅

tee-worker/omni-executor/intent/executors/cross-chain/src/single_chain.rs
felixfaisal: I'm keeping it uncommented for now with `clippy(allow(unused))` derive macro to not get complains from `cargo clippy`, I'm keeping it around as i'm not fully sure if we may or may not need it later.

Kailai-Wang (approved):
Let's get it tested

PR #3470, opened 2025-05-22
update enclave RPC endpoint

0xverin:
context:https://litengtechnologies.slack.com/archives/C05328L0GAF/p1747808500820799
packages on npmjs:https://www.npmjs.com/package/@heima-network/chaindata/v/0.0.6

### DIFF
--- a/tee-worker/omni-executor/client-sdk/packages/chaindata/package.json
+++ b/tee-worker/omni-executor/client-sdk/packages/chaindata/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@heima-network/chaindata",
-  "version": "0.0.5",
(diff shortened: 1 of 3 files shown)


BillyWooo (commented):
How about `heimaDev` and `heimaProd`? Do we need them?

0xverin:
> How about `heimaDev` and `heimaProd`? Do we need them?

Confirmed with simple that `dev` and `prod` have not been updated yet.

BillyWooo (approved):
I think there is no `test` env. I would suggest simply remove `heimaDev` and related code. 
Only `staging-dex-worker` and `dex-worker`. That's enough.

PR #3471, opened 2025-05-22
Refactor PumpxConfig and swap providers for simpler string handling

silva-fj:
This PR adds some refactoring to the `PumpxConfig`, `SingleChainSwapProvider`, and related components to improve type safety, simplify string handling, and reduce boilerplate code in cross-chain and single-chain swap functionality. The key changes include replacing `BoundedVec` with `String` for string fields in `PumpxConfig`, introducing conversion traits between `PumpxConfig` and `OnChainPumpxConfig`, and updating the `Intent` enum to use the new `OnChainSingleChainSwapProvider`. Additionally, test cases and dependencies have been updated to reflect these changes.

### DIFF
--- a/common/primitives/core/Cargo.lock
+++ b/common/primitives/core/Cargo.lock
@@ -2581,9 +2581,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.133"
+version = "1.0.138"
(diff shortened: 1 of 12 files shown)


Kailai-Wang (approved):
Looks good, thanks!

We'll need a parachain runtime upgrade for it, right? Could you also bump the runtime version?

PR #3478, opened 2025-05-24
Conditionally do id-worker related CI

Kailai-Wang:
### Context

As topic - it also tries to run everything without container


.github/workflows/ci.yml
```diff
@@ -268,25 +268,47 @@ jobs:
...
 
-  tee-check:
+  identity-worker-check:
     runs-on: ubuntu-22.04
     needs:
       - fmt
       - set-condition
       - sequentialise
-    container: "litentry/litentry-tee-dev:latest"
+    if: needs.set-condition.outputs.rebuild_identity == 'true'
```
BillyWooo, line 277: why don't we use that docker image any more?

.github/workflows/ci.yml
```diff
@@ -268,25 +268,47 @@ jobs:
...
+          wget https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key && cat intel-sgx-deb.key | sudo tee /etc/apt/keyrings/intel-sgx-keyring.asc > /dev/null
+          sudo apt-get update
+          sudo apt-get install -yq \
+            openssl clang libclang-dev cmake protobuf-compiler \
+            libsgx-dcap-quote-verify-dev \
+            libsgx-enclave-common-dev libsgx-dcap-ql-dev libsgx-dcap-default-qpl-dev
+
+      - name: Install intel sgx libs
+        run: |
+          wget https://download.01.org/intel-sgx/sgx-linux/2.18.1/distro/ubuntu22.04-server/sgx_linux_x64_sdk_2.18.101.1.bin
```
BillyWooo: We are using version 2.19.

Kailai-Wang: Thank you, updated

.github/workflows/ci.yml
Kailai-Wang: So that we don't have to maintain extra dependency

BillyWooo (approved):

PR #3484, opened 2025-05-29
Add support for Authorization header to handle JWT authentication

silva-fj:
This pull request introduces some refactoring, middleware integration, authentication improvements, as well as some dependency updates and some clean up.

Relevant changes:
* Introduced `HttpMiddleware` and `RpcMiddleware` for handling HTTP headers and RPC authentication, respectively, with corresponding unit tests to validate functionality.
* Updated `start_server` in `server.rs` to incorporate middleware layers for enhanced request processing. 

To keep the PR as small as possible, the refactoring of the RPC methods to use Authorization header as authentication mechanism will be done in a separate PR(s).


kziemianek (approved):

BillyWooo (approved):

tee-worker/omni-executor/rpc-server/src/methods/mod.rs
```diff
@@ -7,6 +7,9 @@ use pumpx::*;
 mod omni;
 use omni::*;
 
+// TODO: list all protected methods here once the interface has been updated
+pub const PROTECTED_METHODS: [&str; 1] = ["omni_testProtectedMethod"];
```
Kailai-Wang, line 11: It will be updated in the follow-up PR I assume?

tee-worker/omni-executor/rpc-server/src/middlewares/rpc_middleware.rs
```diff
@@ -0,0 +1,93 @@
...
+	S: RpcServiceT<'a> + Send + Sync + Clone,
+{
+	type Future = ResponseFuture<S::Future>;
+
+	fn call(&self, mut req: Request<'a>) -> Self::Future {
+		if PROTECTED_METHODS.contains(&req.method_name()) {
+			if let Some(http_extensions) = req.extensions().get::<HttpExtensions>() {
+				let token =
+					http_extensions.authorization_header.trim_start_matches("Bearer ").trim();
+				match verify_auth_token_authentication(
```
Kailai-Wang, line 61: Hmm I don't understand this one 😂 

As we put jwt at `Authorization` header, we should verify its validity even without decoding the message (to JSON request), shouldn't we? Why do we do it in the rpc middleware? 

It should also save the compute (reject it with 401 Unauthorized for example), no need to attach it into extensions either

silva-fj: The problem is that at the http level (middleware) we don't have access to the rpc method (we do but we have to manually parse the body to extract the method, this also consumes the request so we have to re-instantiate it afterwards to pass it down the stack (to the next middlewares, services or handlers) and in the rpc middleware we don't have access to the headers as the request have been parsed and transformed already. So, in the http middleware we extract the value of the Authorization header if present and put it in the extensions so we can use it in the rpc middleware to verify it if the rpc method is protected (and if the value was set), if the verification fails we return an error, no need to reach the handlers. Then in the handlers, if the sender is present it means the user is authenticated. 

Note that setting the Authorization header is currently optional, but later we can make it mandatory and only support the other types of authentication in `omni_userLogin` the get the token

tee-worker/omni-executor/rpc-server/src/methods/mod.rs
silva-fj: Yes

tee-worker/omni-executor/rpc-server/src/middlewares/rpc_middleware.rs
Kailai-Wang: I see, thanks

Kailai-Wang (approved):

PR #3486, opened 2025-05-30
Adding omni_userLogin rpc method

silva-fj:
This PR introduces a new rpc method to login (get JWTs) to be able to make subsequent requests to the worker, as well as some small refactorings.

There following structs have been introduced: `UserId`, `UserAuth` and `ClientAuth` to handle authentication and simplify the API params.

TODO: 
- [x] Update `heima_post_login`'s path in pumpx api module


tee-worker/omni-executor/executor-primitives/src/auth.rs
```diff
@@ -92,12 +91,45 @@ pub enum OmniAuth {
...
+				error!("Pumpx identity is not supported in UserId conversion");
+				Err(())
+			},
+		}
+	}
+}
+
+#[derive(Deserialize, Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub enum UserAuth {
+	Web3(HeimaMultiSignature),
```
Kailai-Wang: I think it's a good opportunity to rewrite (flattern) it a bit, no `Web3` wrapper, just `Evm`, `Substrate`, `Solana`... variants
- to be consistent with `UserId`
- to avoid confusions like `ecdsa` vs `ethereum`, actually `ecdsa` is in substrate context, ethereum signatures might be wrapped with EIP191

tee-worker/omni-executor/executor-primitives/src/auth.rs
```diff
@@ -1,23 +1,26 @@
...
-///  "type": "Email",
-///  "data": "test@test.com"
+///  "Email": "test@test.com",
 /// }
 /// ```
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
-#[serde(tag = "type", content = "data")]
-pub enum IdentitySerde {
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Encode, Decode)]
+pub enum UserId {
```
Kailai-Wang, line 29: Shall we use snake_case everywhere? (to be consistent with other rpc fields)? The first char will become small letter by default I assume

tee-worker/omni-executor/executor-primitives/src/auth.rs
silva-fj: Yeah it sounds good 👍🏼

Kailai-Wang (approved):

PR #3487, opened 2025-05-30
Remove explicit `getrandom` dep

Kailai-Wang:
### Context

As topic - trying to resolve https://github.com/litentry/heima/pull/3438#discussion_r2109192262

We need to pin `webpki`


BillyWooo (approved):
Thank you.

PR #3488, opened 2025-05-30
Support fork-aware txpool

Kailai-Wang:
### Context

As topic.

To use it we need to add `--pool-type=fork-aware` to parachain node command line.

### DIFF
--- a/parachain/node/src/rpc.rs
+++ b/parachain/node/src/rpc.rs
@@ -37,7 +37,6 @@ use sc_client_api::{
 use sc_network::service::traits::NetworkService;
 use sc_network_sync::SyncingService;
 pub use sc_rpc::SubscriptionTaskExecutor;
-use sc_transaction_pool::{ChainApi, Pool};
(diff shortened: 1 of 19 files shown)


parachain/node/src/rpc.rs
```diff
@@ -96,13 +97,13 @@ where
...
 /// Full client dependencies
-pub struct FullDeps<C, P, A: ChainApi> {
+pub struct FullDeps<C, P> {
 	/// The client instance to use.
 	pub client: Arc<C>,
 	/// Transaction pool instance.
 	pub pool: Arc<P>,
 	/// Graph pool instance.
-	pub graph: Arc<Pool<A>>,
+	pub graph: Arc<P>,
```
BillyWooo, line 106: even checked the Astar upper stream, still don't understand why both `pool` and `graph` exists together. Why one wouldn't be enough?

BillyWooo (approved):
thank you.

PR #3490, opened 2025-05-31
Use unified workspace setting for primitives and parachain

Kailai-Wang:
### Context

As topic - it makes sure the old `common/` folder (now moved into `parachain/`) joins the workspace management so that no separate deps need to be managed.

Not having a repo-wide workspace setting as it would be too large, plus `parachain` and `omn-executor` aren't tightly dependant.

It renames some `litentry-*` to `heima-*` too


BillyWooo (approved):

.github/workflows/ci.yml
BillyWooo, line 1: actually I got a small question, why to remove `--all` parameter for cargo ?

PR #3495, opened 2025-06-03
Include client_id as aud claim in jwt

silva-fj:
This pull request introduces changes to improve JWT handling and refactor constants usage across the codebase. The most significant updates include adding an audience (`aud`) claim to JWTs, centralizing authentication-related constants, and updating references to these constants in multiple modules.

### DIFF
--- a/tee-worker/omni-executor/native-task-handler/src/lib.rs
+++ b/tee-worker/omni-executor/native-task-handler/src/lib.rs
@@ -18,7 +18,13 @@ use executor_storage::{
 	IntentIdStorage, MemberOmniAccountStorage, PumpxJwtStorage, PumpxProfileStorage, Storage,
 	StorageDB,
 };
-use heima_authentication::auth_token::*;
+use heima_authentication::{
+	auth_token::*,
+	constants::{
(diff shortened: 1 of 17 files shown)


tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -618,6 +626,7 @@ async fn handle_native_task<
 			let access_token_claims = AuthTokenClaims::new(
 				omni_account.to_hex(),
 				AUTH_TOKEN_ACCESS_TYPE.to_string(),
+				CLIENT_ID_PUMPX.to_string(),
```
Kailai-Wang, line 629: We'll need to replace it with that from rpc parameters

Kailai-Wang (approved):
Thanks! CI still needs a tweak

tee-worker/omni-executor/native-task-handler/src/lib.rs
silva-fj: Yes, this is just to make CI pass, I'll update this once the new `userLogin` is ready

PR #3497, opened 2025-06-04
Add dependabot setting for ts-test

0xverin:
resolve https://github.com/litentry/heima/pull/3497
 This PR ignores the npm checks under identity-worker

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,12 +31,6 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     labels: ["automated-pr"]
-    ignore:
-      # Ignore all npm dependencies in identity-worker related directories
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -27,13 +27,37 @@ updates:
...
+    groups:
+      typescript-dependencies:
+        patterns:
+          - "*"
+    reviewers:
+      - "litentry/parachain"
+    schedule:
+      interval: "weekly"
+
+  # Add configuration for tee-worker/identity
```
Kailai-Wang: This is not covered in the previous config with directory `/`?

0xverin: Yes, you are right,  wrote 2 checks step to fix.

silva-fj (approved):

PR #3499, opened 2025-06-04
Update web3 sign-in message format

silva-fj:
This PR updates the web3 sign-in message format to include `client_id` and `omni_account`.

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/verify_auth.rs
+++ b/tee-worker/omni-executor/rpc-server/src/verify_auth.rs
@@ -47,8 +47,8 @@ impl Display for AuthenticationError {
 
 pub async fn verify_auth(ctx: Arc<RpcContext>, auth: &OmniAuth) -> Result<(), AuthenticationError> {
 	match auth {
-		OmniAuth::Web3(ref signer, ref signature) => {
-			verify_web3_authentication(ctx.storage_db.clone(), signer, signature)
+		OmniAuth::Web3(ref client_id, ref signer, ref signature) => {
+			verify_web3_authentication(ctx.storage_db.clone(), client_id, signer, signature)
 		},
 		OmniAuth::Email(ref email, ref verification_code) => {
 			verify_email_authentication(ctx, email, verification_code)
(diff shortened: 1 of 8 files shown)


0xverin (commented):
LGTM,Thanks. but there is a small issue. Why does prettier modify some formats, such as double quotes change to single quotes? Don't we have a unified format in this ts-tests?

Kailai-Wang:
> Why does prettier modify some formats, such as double quotes change to single quotes?

Can this be enforced via prettify format?

tee-worker/omni-executor/rpc-server/src/verify_auth.rs
```diff
@@ -79,7 +80,11 @@ pub fn verify_web3_authentication(
 	verification_code_storage
 		.remove(&storage_key)
 		.map_err(|_| AuthenticationError::VerificationCodeNotFound)?;
-	let message = HeimaMessagePayload { message_code };
+	let message = HeimaMessagePayload {
+		client_id: client_id.to_string(),
+		omni_account: signer.to_omni_account().to_string(),
```
Kailai-Wang, line 85: Does `to_string()` have `0x` prefix? Could you check? (We want `0x prefix)

Kailai-Wang (approved):
Thanks - just to confirm: `OmniAuth` is actually an internal struct, isn't it (not exposed as RPC parameters)

silva-fj:
> LGTM,Thanks. but there is a small issue. Why does prettier modify some formats, such as double quotes change to single quotes? Don't we have a unified format in this ts-tests?

Good question, I don't know. The editor picks the rules of the project and format the code based on that

tee-worker/omni-executor/rpc-server/src/verify_auth.rs
silva-fj: I'll get it merged and fix it in the loginUser PR, I need some changes from this there

silva-fj:
> Thanks - just to confirm: `OmniAuth` is actually an internal struct, isn't it (not exposed as RPC parameters)

Correct

PR #3504, opened 2025-06-06
Account abstraction contracts

kziemianek:
Adds Forge project with `ERC-4337` AA contracts

Most of the smart contract code is taken from `eth-infinitism/account-abstraction` repo.
Notable changes:

* `SmartAccount.sol` and `SmartAccountFactory.sol`
* Project config & layout - upstream is based on `Hardhat` but here we rely on `Forge` .Unit tests were rewritten from `ts` to `solidity`
* `PackedUserOperation` has been extended with `sessionAccount`, `sessionExpiration` and `sessionAccountProof`


.github/file-filter.yml
```diff
@@ -48,5 +48,8 @@ omni_executor_test: &omni_executor_test
   - "tee-worker/omni-executor/client-sdk/**"
   - "tee-worker/omni-executor/ts-tests/**"
 
-evm_contracts_src:
+evm_accounting_contracts_src:
   - "tee-worker/omni-executor/accounting-contract/bsc/**"
+
+evm_aa_contracts_src:  
+- "tee-worker/omni-executor/aa-contracts/**"
```
BillyWooo: missing intend?

kziemianek: thank you, resolved :)

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
```diff
@@ -0,0 +1,166 @@
...
+ *  smart wallet account
+ *  has an owner which is represented by sha256(0x01+hexencoded-address)
+ *  has and root address, this account can generate session keys that will pass signature validation
+ *  has list of allowed signers who can sign messages
+ *  has execute, eth handling methods
+ */
+contract SmartAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Initializable {
+    bytes32 public owner;
+    bytes32 public clientId;
+    address public root;
```
kziemianek: one or many ?

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
```diff
@@ -0,0 +1,166 @@
...
+    mapping(address => bool) public allowedSigners;
```
kziemianek: signers like passkey

.gitmodules
```diff
@@ -4,3 +4,12 @@
 [submodule "tee-worker/omni-executor/accounting-contract/bsc/lib/forge-std"]
 	path = tee-worker/omni-executor/accounting-contract/bsc/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
+[submodule "tee-worker/omni-executor/aa-contracts/aa-contracts/lib/forge-std"]
+	path = tee-worker/omni-executor/aa-contracts/aa-contracts/lib/forge-std
```
Kailai-Wang: Eventually we should be able to merge these (to only include `forge-std` submodule once), right?

tee-worker/omni-executor/aa-contracts/.github/workflows/test.yml
Kailai-Wang, line 1: Is it actually run in our CI? Or was it just copied from some template?

tee-worker/omni-executor/aa-contracts/src/Counter.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+contract Counter {
```
silva-fj, line 4: Is this part of the project initialization template? should we remove it?

.gitmodules
silva-fj: Is this path correct? Also, it looks like the submodule is duplicated

tee-worker/omni-executor/aa-contracts/src/Counter.sol
kziemianek: We use this contract in tests.

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
```diff
@@ -0,0 +1,185 @@
...
+                || isRootSigner(msg.sender),
+            "account: not Owner or EntryPoint or allowed signer or root"
+        );
+    }
+
+    /**
+     * convert sender to oa bytes
+     */
+    function _determineOa(address sender) internal view returns (bytes32) {
+        bytes1 oaType = 0x01;
```
Kailai-Wang: Here it should be `b"evm"`

See https://github.com/litentry/heima/blob/e1b710a74202d097e0428504d6f50222936d61a3/parachain/primitives/src/identity.rs#L427

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
```diff
@@ -0,0 +1,185 @@
...
+import "@openzeppelin/contracts/utils/Strings.sol";
+import "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol";
+import "../core/BaseAccount.sol";
+import "../core/Helpers.sol";
+import "./callback/TokenCallbackHandler.sol";
+
+/**
+ *  smart wallet account
+ *  has an owner which is represented by sha256(0x01+hexencoded-address)
```
Kailai-Wang: A small correction about the derivation formula - see below

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
```diff
@@ -0,0 +1,185 @@
...
+ *  has and root address, this account can generate session keys that will pass signature validation
```
Kailai-Wang: Typo? : ) "has and"

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
```diff
@@ -0,0 +1,185 @@
...
+ *  has list of allowed signers who can sign messages
+ *  has execute, eth handling methods
+ */
+contract SmartAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Initializable {
+    bytes32 public owner;
+    bytes32 public clientId;
+    mapping(address => bool) public rootSigners;
+    mapping(address => bool) public allowedSigners;
```
Kailai-Wang: If it's ECDSA signer (of `address` type), should it be merged into `rootSigners`?

Having two storages could be confusing, essentially they are both "authorised keys"

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
```diff
@@ -0,0 +1,185 @@
...
+        require(_determineOa(msg.sender) == owner || msg.sender == address(this), "only owner");
+    }
+
+    /**
+     * @dev The _entryPoint member is immutable, to reduce gas consumption.  To upgrade EntryPoint,
+     * a new implementation of SimpleAccount must be deployed with the new EntryPoint address, then upgrading
+     * the implementation by calling `upgradeTo()`
+     * @param anOwner the owner (signer) of this account
+     */
+    function initialize(bytes32 anOwner, bytes32 aClientId, address aRoot) public virtual initializer {
```
Kailai-Wang, line 66: Is it better to make `ClientId` per entrypoint? If so it should be either read from the entrypoint contract, or at least check against it

Kailai-Wang: Shall we make `aRoot` optional-like? It should have value only for non-evm users.

For evm users it should be empty (or 0) by default as we want users to explicitly authorise them

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccountFactory.sol
```diff
@@ -0,0 +1,63 @@
...
+        senderCreator = _entryPoint.senderCreator();
+    }
+
+    /**
+     * create an account, and return its address.
+     * returns the address even if the account is already deployed.
+     * Note that during UserOperation execution, this method is called only if the account is not deployed.
+     * This method returns an existing account address so that entryPoint.getSenderAddress() would work even after account creation
+     */
+    function createAccount(bytes32 oa, bytes32 clientId, address root) public returns (SmartAccount ret) {
```
Kailai-Wang, line 31: As mentioned, shall we store `clientId` into the entrypoint storage? Maybe even as ctor param

tee-worker/omni-executor/aa-contracts/src/core/EntryPoint.sol
Kailai-Wang, line 1: I haven't reviewed this contract - is it standard template?

Kailai-Wang (commented):
Thanks!
Hopefully I'm not wrong - I assume `SmartAccount` and `SmartAccountFactory` are the main files to review (with their tests)?

CI fails too : (

tee-worker/omni-executor/aa-contracts/.github/workflows/test.yml
kziemianek: removed

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: changed

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: they have same type (`address`) but slightly different meaning.

signer can only sign userOp
root can also generate session proof

should we merge them so for example passkey can also generate sessions?

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: changed

tee-worker/omni-executor/aa-contracts/src/core/EntryPoint.sol
kziemianek: This is standard template

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: but this will leak to worker - worker that submits tx will need to know mapping client_id -> entry_point
also each new client cannot join seamlessly - it will require to:
1. deploy entry point
2. update worker mapping

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: updated, thanks

tee-worker/omni-executor/aa-contracts/src/core/EntryPoint.sol
kziemianek: Hovewer we have some changes compared to 4337 spec so we are not compatible and we cannot reuse existing infrastructure (entrypoints/bundlers)

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
Kailai-Wang: I think it does no harm to (implicitly) allow that

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
Kailai-Wang: > worker that submits tx will need to know mapping client_id -> entry_point

Do you want to use the same entrypoint contract for all clients? If not then worker needs to know the mapping anyway (if the worker is shared among clients)

tee-worker/omni-executor/aa-contracts/src/core/EntryPoint.sol
Kailai-Wang: Can't we keep the bundler as it is? I mean in worker we probably don't send it to an RPC backed by bundler, but just call the entrypoint's handleUserOp directly

kziemianek: Yes, we don't send it to bundler - worker submits tx to rpc node - but clients cannot use infrastructure provided by for example alchemy to submit `UserOp` because it requires different entrypoint instance.

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: Ok, I've removed `allowedSigners`

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: Yes, it's not client specific so I see no point in having instance per client. If we move `clientId` to `EntryPoint` then we need such relationship.

.gitmodules
kziemianek: I've moved all shared modules to `tee-worker/omni-executor/forge-libs`

kziemianek: I've moved all shared modules to `tee-worker/omni-executor/forge-libs`

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
Kailai-Wang: That should work too. If we really want a different logic for a specific client, we can always deploy a new one.

Kailai-Wang (approved):
Thanks!

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: ok, if needed we will adjust later 👍🏼

PR #3508, opened 2025-06-09
Make client_id Dynamic

silva-fj:
This PR implements dynamic `client_id` support across the omni-executor RPC system to enhance authentication context and improve client identification.

Key Changes:

- Added `client_id` field to `NativeTaskWrapper` and authentication middleware (RpcExtensions)
- Authentication Enhancement: Updated `User` struct and `check_auth` function to handle dynamic client identification
- RPC Method Updates: Integrated `client_id` parameter into all omni and pumpx RPC methods
- Updated TypeScript definitions and task creators to support the new `client_id` parameter

TODO: 
- [x] Update rpc methods in Apifox

### DIFF
--- a/tee-worker/omni-executor/executor-core/src/native_task.rs
+++ b/tee-worker/omni-executor/executor-core/src/native_task.rs
@@ -24,12 +24,13 @@ pub struct NativeTaskWrapper<T: NativeTaskTrait> {
 	pub task: T,
 	pub nonce: Option<Nonce>,
 	pub auth: Option<OmniAuth>,
+	pub client_id: String,
 }
 
 impl<T: NativeTaskTrait + Debug> NativeTaskWrapper<T> {
-	pub fn new(task: T, nonce: Option<Nonce>, auth: Option<OmniAuth>) -> Self {
+	pub fn new(task: T, nonce: Option<Nonce>, auth: Option<OmniAuth>, client_id: String) -> Self {
 		let id: String = Uuid::new_v4().into();
-		Self { id, task, nonce, auth }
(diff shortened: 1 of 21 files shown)


tee-worker/omni-executor/executor-core/src/native_task.rs
```diff
@@ -24,12 +24,13 @@ pub struct NativeTaskWrapper<T: NativeTaskTrait> {
...
 	pub auth: Option<OmniAuth>,
+	pub client_id: String,
 }
 
 impl<T: NativeTaskTrait + Debug> NativeTaskWrapper<T> {
-	pub fn new(task: T, nonce: Option<Nonce>, auth: Option<OmniAuth>) -> Self {
+	pub fn new(task: T, nonce: Option<Nonce>, auth: Option<OmniAuth>, client_id: String) -> Self {
 		let id: String = Uuid::new_v4().into();
-		Self { id, task, nonce, auth }
+		Self { id, task, nonce, auth, client_id }
```
Kailai-Wang, line 33: Just to make sure I understand it correctly - now that we verify the auth outside of a `NativeTask`, do we still need the `sender` Identity parameter?

Not saying we need to tackle this in this PR

Kailai-Wang (approved):

tee-worker/omni-executor/executor-core/src/native_task.rs
silva-fj: I think we still need this in some situations, mostly the tasks related to the AccountStore and request intents. So the tasks that send extrinsics to the parachain

PR #3509, opened 2025-06-09
Use OmniAccount hash as storage key for email verification code

silva-fj:
Refactors the email verification code flow to use the omni account hash (derived from email and client_id) as the storage key in VerificationCodeStorage, instead of the email identity hash and updates all relevant RPC methods.

### DIFF
--- a/tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/__test__/account-store.request.test.ts
+++ b/tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/__test__/account-store.request.test.ts
@@ -459,7 +459,7 @@ describe('account-store', () => {
     console.log('Step 1: create account store');
     await (async () => {
       // Request email verification code for next account store creation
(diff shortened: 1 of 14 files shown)


tee-worker/omni-executor/client-sdk/packages/client-sdk/src/lib/requests/__test__/account-store.request.test.ts
```diff
@@ -459,7 +459,7 @@ describe('account-store', () => {
     console.log('Step 1: create account store');
     await (async () => {
       // Request email verification code for next account store creation
-      await requestEmailVerificationCode({ email });
```
0xverin, line 462: In fact, we may not need to update any test cases in the client-sdk, because F/E team are currently only referencing apifox. If we must add, it might be better to add in ts-tests. Just a reminder, the client-sdk should not be maintained further in the future.

Kailai-Wang (approved):
Thank you - could you also update apifox accordingly? (Including the previous changes like `omni_userLogin` etc)

PR #3511, opened 2025-06-10
Make OmniAccount derivation on chain verifiable

silva-fj:
In this PR we update how the OmniAccount is derived on-chain, it's a follow up of the changes already made in the worker. The main change is the inclusion of the `client_id` in the OA derivation.


parachain/pallets/omni-account/runtime-api/src/lib.rs
```diff
@@ -16,6 +16,15 @@
...
 
+#[cfg(not(feature = "std"))]
+extern crate alloc;
+
+#[cfg(not(feature = "std"))]
+use alloc::string::String;
+
+#[cfg(feature = "std")]
+use std::string::String;
+
```
Kailai-Wang, line 27: I think you could simply use `sp_std`

Kailai-Wang (approved):
Thanks!

parachain/pallets/omni-account/runtime-api/src/lib.rs
silva-fj: I thought the same thing too, but there is no String exported from it

PR #3512, opened 2025-06-11
fix: save user access token to storage

wli-pro:
before submitting swap order intent

### Labels

### Testing Evidences

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
+++ b/tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
@@ -6,7 +6,7 @@ use crate::{
 use executor_core::native_task::*;
 use executor_primitives::OmniAuth;
 use executor_storage::{HeimaJwtStorage, Storage};
-use heima_authentication::constants::{AUTH_TOKEN_ACCESS_TYPE, AUTH_TOKEN_ID_TYPE};
+use heima_authentication::constants::{AUTH_TOKEN_ACCESS_TYPE, AUTH_TOKEN_ID_TYPE, CLIENT_ID_HEIMA};
 use heima_primitives::{
 	AccountId, Address20, Address32, BinanceConfig, BoundedVec, ChainAsset, CrossChainSwapProvider,
 	EthereumToken, Identity, Intent, PumpxConfig, PumpxOrderType, SingleChainSwapProvider,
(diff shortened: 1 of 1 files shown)


tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
```diff
@@ -234,6 +234,13 @@ pub fn register_submit_swap_order(module: &mut RpcModule<RpcContext>) {
             })?);
            	let user_identity =
 				Identity::from_web2_account(&params.user_id, Web2IdentityType::Pumpx);
+
+			// Save access token to storage before submitting swap order intent, as it's required by the cross-chain intent executor
+			let user_omni_account = user_identity.to_omni_account_with_client_id(CLIENT_ID_HEIMA);
+			if storage.insert(&(user_omni_account, AUTH_TOKEN_ACCESS_TYPE), access_token.clone()).is_err() {
```
silva-fj, line 240: I don't understand this change, why are you deriving the OmniAccount again and saving the access_token (also again) ? We save the access token in storage when the user requests the Jwts (Also in the new `omni_loginUser` rpc method). Please also keep in mind the changes in this [PR](https://github.com/litentry/heima/pull/3508)

silva-fj (changes requested):
Can you elaborate a bit on why you needed this change?

Kailai-Wang:
Was it due to some bug found in the test?
Could you elaborate the context?

tee-worker/omni-executor/rpc-server/src/methods/pumpx/submit_swap_order.rs
wli-pro: The omni account is derived with CLIENT_ID_HEIMA as below in native task handler of NativeTask::RequestIntent before:
```
let omni_account = sender.to_omni_account_with_client_id(CLIENT_ID_HEIMA);
```
And in submit_swap_order.rs#L161, the access_token is read from storage using the derived omni account above. But because the access_token hasn't been saved to storage for this omni account yet, an error will occur.

```
let storage = HeimaJwtStorage::new(ctx.storage_db.clone());
let Ok(Some(access_token)) =
	storage.get(&(omni_account, AUTH_TOKEN_ACCESS_TYPE))
else {
	error!("Failed to get access token from storage");
	return Err(PumpxRpcError::from_error_code(ErrorCode::InternalError));
};
```
So, I save the access_token for the derived omni account before submitting the intent request in my PR.

After the omni account is not derived with CLIENT_ID_HEIMA in this [PR](https://github.com/litentry/heima/pull/3508), this PR is no longer needed.

wli-pro:
this recent [PR](https://github.com/litentry/heima/pull/3508) implemented that the omni account is not derived from CLIENT_ID_HEIMA, hence current PR is not needed

PR #3517, opened 2025-06-12
Enable omni-account integration tests and fix API package references

silva-fj:
## Summary
- Re-enable omni-executor integration tests in CI workflow by removing the disabled condition
- Fix TypeScript import paths from `@heima-network/api-argument` to `@heima-network/api-augment` across test files
- Update omni account utilities to support `client_id` parameter for enhanced functionality
- Fix package dependencies and type creators for proper authentication handling


Kailai-Wang (approved):
Great, thanks!

PR #3521, opened 2025-06-13
add migration logic for Paseo and Heima

BillyWooo:
migration for Paseo and Heima:
- pallet-identity -> StorageVersion
- pallet-xcm -> V2 to V5
    - because v2 was removed in stable2412, and there is only one entry inside PolkadotXcm.Queries storage. So to keep it simple, just manually remove old value and insert a new value.


parachain/runtime/heima/src/migration/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+		let known_query_id: u64 = 0;
+		let known_new_value: QueryStatus<BlockNumberFor<T>> = QueryStatus::VersionNotifier {
+			origin: VersionedLocation::V5(Location { parents: 0, interior: Junctions::Here }),
+			is_active: true,
+		};
+
+		// Build Blake2_128Concat key
+		let encoded_key = known_query_id.encode();
+		let mut final_key = blake2_128(&encoded_key).to_vec();
+		final_key.extend(encoded_key);
```
Kailai-Wang: Is there a built-in method for these?

Kailai-Wang (approved):
Looks fine - have you tried with try-runtime?

Kailai-Wang:
Could you also fix the PR title?

BillyWooo:
> Looks fine - have you tried with try-runtime?

Yes. [Here](https://github.com/litentry/heima/actions/runs/15664945428).
There is a `try-runtime.yml`. It's not really used any more. How about removing it? @Kailai-Wang

parachain/runtime/heima/src/migration/mod.rs
BillyWooo: Didn't find any built-in. Maybe @silva-fj do you have any idea?

silva-fj: Yes, you can use `Blake2_128Concat::hash`

parachain/runtime/heima/src/migration/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+
+pub type Migrations<Runtime> = (IdentityUpdateStorageVersion<Runtime>, RewriteXcmQueries<Runtime>);
+
+const IDENTITY_LOG_TARGET: &str = "runtime::identity";
+const XCM_LOG_TARGET: &str = "runtime::xcm";
+
+pub struct IdentityUpdateStorageVersion<T>(PhantomData<T>);
+impl<T> OnRuntimeUpgrade for IdentityUpdateStorageVersion<T>
+where
+	T: frame_system::Config + pallet_identity::Config,
```
silva-fj, line 48: I think I mentioned this in our last daily, these are not our pallets, we should not be manually writing storage migrations for them, we should just use the pallet's own migrations. e.g: [identity_pallet](https://paritytech.github.io/polkadot-sdk/master/pallet_identity/migration/index.html)

silva-fj:
I think the xcm pallet's migration should work, they added a [fix](https://github.com/paritytech/polkadot-sdk/pull/6148) for the V2 issue some time ago

parachain/runtime/heima/src/migration/mod.rs
Kailai-Wang: That makes sense

If there's official migration code from upstream, we should use it @BillyWooo

parachain/runtime/heima/src/migration/mod.rs
BillyWooo: updated.

parachain/runtime/heima/src/migration/mod.rs
BillyWooo: @Kailai-Wang @silva-fj , I created a new branch: something-new-try-runtime. But it doesn’t work out directly. I get error like:
```
[2025-06-16T09:26:35Z ERROR runtime::frame-support] ParachainIdentity: On chain storage version StorageVersion(1) doesn't match in-code storage version StorageVersion(2).
[2025-06-16T09:26:35Z INFO  runtime::frame-support] ⚠️ MultiBlockMigrations declares internal migrations (which *might* execute). On-chain `StorageVersion(0)` vs in-code storage version `NoStorageVersionSet`
```

Looks like the pallet_migrations’ logic was not triggered before the pallet_identity checking. Maybe miss some configuration.

I would suggest to merge this PR. It doesn't worth to spend too much time on this topic.

BillyWooo:
> I think the xcm pallet's migration should work, they added a [fix](https://github.com/paritytech/polkadot-sdk/pull/6148) for the V2 issue some time ago

This fix is not released.

PR #3531, opened 2025-06-17
Remove IAS attestation support 

silva-fj:
This PR removes Intel Attestation Service (IAS) support from the codebase and to only support DCAP attestation. The changes include:

  - Removed all IAS-related code from teebag pallet, identity-worker, and SGX verification modules
  - Updated tests to remove IAS verification and fix SGX build mode expectations
  - Replaced deprecated AttestationType::Ias with Ignore in mocks and benchmarks
  - Removed unused sign_type parameters from remote attestation functions
  - Cleaned up unused imports and code formatting


Kailai-Wang (commented):
Thank you for tidying these up - I have two comments:

1. would it need runtime migration as we removed the `Ias` from `AttestationType`? Would it make life easier if we keep it in the enum and make it as deprecated, id-worker would always register with `Dcap` or `Ignore`?

2. could you further check if `ring` and `webpki` crates can be left out? In both parachain and worker directory - it would greatly help with resolve the dependabot and make `cargo update` easier

silva-fj:
> 1. would it need runtime migration as we removed the `Ias` from `AttestationType`? Would it make life easier if we keep it in the enum and make it as deprecated, id-worker would always register with `Dcap` or `Ignore`?

Good point, if we remove it we would need a runtime migration. So we could keep it as deprecated and revert a few changes for now.
The identity worker use `Dcap` if the flag is enable (I think this is the default), otherwise it will use `Ignore`.

> 2. could you further check if `ring` and `webpki` crates can be left out? In both parachain and worker directory - it would greatly help with resolve the dependabot and make `cargo update` easier

I'll take a look at this again, I think it was use by other part as well.

Kailai-Wang:
> I'll take a look at this again, I think it was use by other part as well.

This might need more time - I can take a look too

For now we could just commit the source code change @silva-fj

Kailai-Wang (approved):

PR #3535, opened 2025-06-18
Standardize authentication for omni_transferWithdraw and omni_exportWallet

silva-fj:
## Summary

Authentication Flow Updates
  - export_wallet and transfer_withdraw now use Authorization header authentication
  - All methods extract the OA from the user's JWT
  - Removed redundant user_email parameters

 Refactored protected RPC methods to extract AccountId from authenticated users and remove email parameter:
  - omni_exportWallet 
  - omni_transferWithdraw 
  - omni_addWallet 
  - omni_submitSwapOrder 
  - omni_signLimitOrder 
  - omni_notifyLimitOrderResult 

Refactored related NativeTask's to use OA instead of Identity.

Note: We need to refactor the omni-account pallet to adjust how intent calls are dispatched. We currently send the Identity but we should send the actual OA. Because of this, some code has been commented out (as well as the integration tests)


BillyWooo (approved):

tee-worker/omni-executor/rpc-server/src/methods/omni/transfer_widthdraw.rs
```diff
@@ -16,14 +17,12 @@ use super::common::{check_omni_api_response, handle_omni_native_task};
 #[derive(Debug, Deserialize)]
 pub struct TransferWithdrawParams {
 	pub user_email: String,
-	pub client_id: String,
```
Kailai-Wang, line 19: There should be no `user_email` anymore

tee-worker/omni-executor/rpc-server/src/methods/omni/export_wallet.rs
```diff
@@ -17,18 +18,16 @@ use super::common::handle_omni_native_task;
 
 #[derive(Debug, Deserialize)]
 pub struct ExportWalletParams {
-	pub client_id: String,
```
Kailai-Wang, line 20: There should be no `user_email` anymore

Kailai-Wang: We just extract the OA from JWT if we need it. 
If we need the `user_email` to construct the `NativeTask` - then it's wrong, NativeTask shouldn't need the `Identity` anymore - as we verify the auth before processing `NativeTask`

Kailai-Wang (changes requested):
Thanks - I think it requires a small refactoring of `NativeTask`

I assume checks against other methods would come in a follow-up PR?

tee-worker/omni-executor/rpc-server/src/methods/omni/export_wallet.rs
silva-fj: I thought about that but it probably requires a broader refactoring as all tasks are expected to have the sender's Identity. I'll try to make the changes here independently, then create a follow up PR if needed

silva-fj:
> I assume checks against other methods would come in a follow-up PR?

All other methods have been checked already. Some require no authentication and the others already use the Authorization header

tee-worker/omni-executor/rpc-server/src/methods/omni/export_wallet.rs
silva-fj: it turns out we were not using the sender method from the NativeTask so the change won't affect the other tasks that require the sender's identity

tee-worker/omni-executor/executor-core/src/native_task.rs
```diff
@@ -41,7 +39,7 @@ pub type PumpxChainId = u32;
 #[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
 pub enum NativeTask {
 	RequestAuthToken(Identity),
-	RequestIntent(Identity, IntentId, Intent),
+	RequestIntent(AccountId, IntentId, Intent),
 	CreateAccountStore(Identity),
```
kziemianek, line 43: Is it possible that remaining task types will accept `AccountId` instead of `Identity` ?

Kailai-Wang: @kziemianek do you mean old tasks (non-dex-related)? If so we could do it in another PR - maybe together with AccountStore cleanup 🤔

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -1030,33 +1012,33 @@ fn send_ok(sender: ResponseSender, ok_res: NativeTaskOk) {
...
-	RpcClient: SubstrateRpcClient<Header> + Send + Sync + 'static,
->(
-	client: &mut RpcClient,
-	signer: Arc<ParentchainTxSigner>,
-	sender: Identity,
-	call: RuntimeCall,
-	auth_type: Option<OmniAccountAuthType>,
-) {
-	let call = parentchain_api_interface::tx().omni_account().dispatch_as_signed(
-		sender.hash().to_subxt_type(),
```
Kailai-Wang, line 1044: IIUC the problem here is the `hash()` right? We could later use the hash of the OmniAccount - either blake2_256 or another sha2 if we want it to be evm-friendly.

Can do together with AccountStore cleanup

Kailai-Wang (approved):
Looks good, thanks.

Could you (or ask Verin) to make sure the apifox is up-to-date?

tee-worker/omni-executor/executor-core/src/native_task.rs
kziemianek: Yes, ok

wli-pro (approved):

PR #3538, opened 2025-06-19
Fix typo for env vars

0xverin:
correct `OE_ETHERUM_URL` to `OE_ETHEREUM_URL`


tee-worker/omni-executor/executor-worker/src/env_config.rs
BillyWooo, line 1: This file should not be here.

BillyWooo (approved):

PR #3541, opened 2025-06-23
Try to remove `ring` and `webpki` dep from parachain

Kailai-Wang:
### Context

Finally we may rejoice.

This PR is able to replace the `ring` and `webpki` usage with wasm-friendly crates like `p256`/`ecdsa`. It should resolve most of the dependabot issues and open the door of free `cargo update`.


kziemianek (commented):
Would it be possible to keep fuzz tests?

BillyWooo:
> Would it be possible to keep fuzz tests?

Looks like we never used it. Did we?

BillyWooo (approved):
LGTM

Kailai-Wang:
> Would it be possible to keep fuzz tests?

I could, but it's not run anywhere in CI (not even in upstream) - and we haven't spent time maintaining it.

Do you see the value to maintain it?

kziemianek:
> > Would it be possible to keep fuzz tests?
> 
> I could, but it's not run anywhere in CI (not even in upstream) - and we haven't spent time maintaining it.
> 
> Do you see the value to maintain it?

It's related to sgx attestation so i would say yes.

Kailai-Wang:
Sure let me add it back

BillyWooo:
> > > Would it be possible to keep fuzz tests?
> > 
> > 
> > I could, but it's not run anywhere in CI (not even in upstream) - and we haven't spent time maintaining it.
> > Do you see the value to maintain it?
> 
> It's related to sgx attestation so i would say yes.

Hi @kziemianek , can you show me how it's used? Maybe I missed something implicit. I couldn't find how our code / test code / ci uses this part of code.

Kailai-Wang:
I added back fuzz/ @kziemianek 

Tested by `cargo fuzz run decode_quote` (as one of the target examples)

parachain/pallets/account-fix/src/lib.rs
```diff
@@ -17,6 +17,7 @@
 //! A pallet for temporary fix of onchain accountInfo.
 //! No storage for this pallet and it should be removed right after fixing.
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::useless_conversion)]
```
silva-fj, line 20: Why do we need this now?

silva-fj (approved):

parachain/pallets/account-fix/src/lib.rs
Kailai-Wang: I assume it's a false positive for the newer rustc (as they are indeed different types).

I've double-checked the upstream polkadot-sdk - it has the same warning as it's not our code's problem.

kziemianek (approved):

PR #3542, opened 2025-06-23
Client for interacting with AA contracts

kziemianek:
* Adds `EntryPointClient` for integration with onchain `EntryPoint` instances
* Adds `SmartWalletClient` for integration with onchain `SmartWallet` instances
* Adds `SimplePaymaster` contract code to enable gas sponsorship. Only tx's coming from whitelisted addresses (offchain workers) will be sponsored.
* Changes OA calculation method in contracts to align with Rust code
* Adds `get_code_at` method to ethereum `RpcProvider`
* Adds `local-deploy.sh` for setting up local `anvil` node with AA contracts deployed


tee-worker/omni-executor/aa-contracts/src/core/BasePaymaster.sol
```diff
@@ -0,0 +1,147 @@
...
+        entryPoint.depositTo{value: msg.value}(address(this));
+    }
+
+    /**
+     * Withdraw value from the deposit.
+     * @param withdrawAddress - Target to send to.
+     * @param amount          - Amount to withdraw.
+     */
+    function withdrawTo(address payable withdrawAddress, uint256 amount) public onlyOwner {
+        entryPoint.withdrawTo(withdrawAddress, amount);
```
0xverin, line 108: I think we need add ` require(amount > 0, "Amount must be greater than 0");`

tee-worker/omni-executor/aa-contracts/src/core/BasePaymaster.sol
```diff
@@ -0,0 +1,147 @@
...
+    // IEntryPoint of this paymaster
+    function _validateEntryPointInterface(IEntryPoint _entryPoint) internal virtual {
+        require(
+            IERC165(address(_entryPoint)).supportsInterface(type(IEntryPoint).interfaceId),
+            "IEntryPoint interface mismatch"
+        );
+    }
+
+    /// @inheritdoc IPaymaster
+    function validatePaymasterUserOp(PackedUserOperation calldata userOp, bytes32 userOpHash, uint256 maxCost)
```
0xverin, line 40: Do we need to add a pause function, so that we can pause this part at any time in case there are issues?  
```
  function validatePaymasterUserOp(...)
        external
        override
        whenNotPaused 
        returns (bytes memory context, uint256 validationData)
    {
        // ...
    }
    
    function emergencyPause() external onlyOwner {
        _pause();
    }
    function unpause() external onlyOwner {
        _unpause();
    }

File: tee-worker/omni-executor/aa-contracts/src/core/BasePaymaster.sol (hunk 1 above)
Comment (kziemianek) [reply tee-worker/omni-executor/aa-contracts/src/core/BasePaymaster.sol:108]: added

File: tee-worker/omni-executor/aa-contracts/src/core/BasePaymaster.sol (hunk 2 above)
Comment (kziemianek) [reply tee-worker/omni-executor/aa-contracts/src/core/BasePaymaster.sol:40]: added. 

It could be paused by removing all `authorizedBundlers` from `SimplePaymaster` but I've added pausable feature in `BasePaymaster`

File: tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs [hunk 3]
```diff
@@ -0,0 +1,547 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::types::{
+	createAccountCall, depositToCall, getSenderAddressCall, getUserOpHashCall, handleOpsCall,
+	SenderAddressResult,
+};
+use crate::utils::{
+	build_call_transaction, build_payable_transaction, calculate_smart_account_address,
```
Kailai-Wang: `calculate_smart_account_address` => `calculate_smart_wallet_address`, as the other client is called `smart_wallet_client` - we try to keep it consistent

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
```diff
@@ -0,0 +1,547 @@
...
+			sessionExpiration: U256::from(0),
+			sessionAccountProof: Bytes::new(),
+			signature: Bytes::new(),
+		})
+	}
+
+	/// Create PackedUserOperation using local CREATE2 address calculation
+	/// This is more efficient as it avoids the EntryPoint.getSenderAddress RPC call
+	#[allow(clippy::too_many_arguments)]
+	pub async fn create_packed_user_operation_with_local_address(
```
Kailai-Wang, line 143: Later we should be able to extract common code of the two fn

tee-worker/omni-executor/aa-contracts/src/core/BasePaymaster.sol
Kailai-Wang, line 1: Is this optional or mandatory in reality?

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
```diff
@@ -86,8 +86,8 @@ contract SmartAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Ini
      * convert sender to oa bytes
      */
     function _determineOa(address sender) internal view returns (bytes32) {
-        bytes memory oaType = bytes("evm");
-        return sha256(abi.encodePacked(oaType, clientId, sender));
+        bytes3 oaType = 0x65766d;
```
Kailai-Wang, line 90: would it make difference to `bytes("evm")`? If no then no magic number is prferred, alternatively you could leave a comment too

Kailai-Wang (approved):
Looks good, well done, thanks!

tee-worker/omni-executor/aa-contracts/src/accounts/SmartAccount.sol
kziemianek: Proposed version should result in slightly less gas usage but we can restore previous one or add comment.

kziemianek: I've added comment

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
kziemianek: renamed

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
kziemianek: extracted

tee-worker/omni-executor/aa-contracts/src/core/BasePaymaster.sol
kziemianek: optional

kziemianek: comes from https://github.com/eth-infinitism/account-abstraction/tree/develop/contracts/core as few other files. It's mentioned in `README.md`

PR #3543, opened 2025-06-23
Bump runtime and remove migrations

Kailai-Wang:
### Context

As topic


BillyWooo:
runtime/heima/src/migration/mod.rs also need to be removed (clean the file content, similar like paseo)

BillyWooo (approved):

PR #3556, opened 2025-06-24
stable2412 should use version number 0.9.25

BillyWooo:
as title


parachain/runtime/heima/src/lib.rs
```diff
@@ -245,7 +245,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: alloc::borrow::Cow::Borrowed("heima"),
 	authoring_version: 1,
 	// same versioning-mechanism as polkadot: use last digit for minor updates
-	spec_version: 9244,
+	spec_version: 9252,
```
Kailai-Wang: I think we can start from 9250 as this is the first stable2412 runtime - it's not related to tag version

Kailai-Wang (approved):

PR #3560, opened 2025-06-25
extend one omni-executor to have multiple SendGrid key/email/sendname

BillyWooo:
extend one omni-executor to have multiple SendGrid key/email/sendname, so that multiple frontend/backend can connect to one worker. 

`client_id` is used as the key to differentiate.


kziemianek:
These changes are coming from outside (clients) so ideally:
* we move it onchain so they can manage it by themselves
* we make it restart -free

Have you considered that?

BillyWooo:
> These changes are coming from outside (clients) so ideally:
> 
> * we move it onchain so they can manage it by themselves
> * we make it restart -free
> 
> Have you considered that?

Didn't consider that. But, yes that's a better working mechanism. Also need more effort to design. 
@Kailai-Wang what about your opinion? Do you think it's worth of it? 
@silva-fj Wdyt?

silva-fj:
> > These changes are coming from outside (clients) so ideally:
> > 
> > * we move it onchain so they can manage it by themselves
> > * we make it restart -free
> > 
> > Have you considered that?
> 
> Didn't consider that. But, yes that's a better working mechanism. Also need more effort to design. @Kailai-Wang what about your opinion? Do you think it's worth of it? @silva-fj Wdyt?

It depends, by "themselves" do you mean that they will provide the API KEY, sender email, etc ? if so, then it would be nice to have an interface for them to register, probably in the worker as the API KEY should not be exposed, so clients and configure their own mailer. (Later we could even support multiple mailer services so clients can decide what to use and not lock in with Sendgrid).

If not, then it won't help much as we need to manually configure the sender email, etc in the mailer service.

Also, please read sendgrid's documentation, I think we have use one account (One API KEY) and configure different email senders for the different clients (also different email templates)

BillyWooo:
> > > These changes are coming from outside (clients) so ideally:
> > > 
> > > * we move it onchain so they can manage it by themselves
> > > * we make it restart -free
> > > 
> > > Have you considered that?
> > 
> > 
> > Didn't consider that. But, yes that's a better working mechanism. Also need more effort to design. @Kailai-Wang what about your opinion? Do you think it's worth of it? @silva-fj Wdyt?
> 
> It depends, by "themselves" do you mean that they will provide the API KEY, sender email, etc ? if so, then it would be nice to have an interface for them to register, probably in the worker as the API KEY should not be exposed, so clients and configure their own mailer. (Later we could even support multiple mailer services so clients can decide what to use and not lock in with Sendgrid).
> 
> If not, then it won't help much as we need to manually configure the sender email, etc in the mailer service.
> 
> Also, please read sendgrid's documentation, I think we have use one account (One API KEY) and configure different email senders for the different clients (also different email templates)

Regarding sendgrid's doc, yes, we can and we are using in such a way. There is only one SendGrid account, where we have multiple api keys. Each key can bind together with different sender-email and sender-name.

The current solution, I just think it's sufficient. In case in the future we need to support client-control mailer, we can improve it. That's my opinion.

tee-worker/omni-executor/config-loader/src/config.rs
```diff
@@ -75,47 +80,103 @@ fn get_env_value(var: &EnvVar) -> Option<String> {
...
+	// Find all mailer client IDs by looking for CLIENT_ID env vars
+	let mut client_ids = std::collections::HashSet::new();
+	for (key, _) in &env_vars {
+		if key.starts_with("OE_SENDGRID_") && key.ends_with("_CLIENT_ID") {
+			// Extract the mailer identifier (e.g., "1" from "OE_SENDGRID_1_CLIENT_ID")
+			if let Some(mailer_id) =
+				key.strip_prefix("OE_SENDGRID_").and_then(|s| s.strip_suffix("_CLIENT_ID"))
+			{
+				client_ids.insert(mailer_id.to_string());
+			}
```
Kailai-Wang: I think having the digit as infix is a bit awkward - can we simply append the client_id to the end of the env variable? Like  `OE_SENDGRID_API_HOST_WILDMETA`

Kailai-Wang: Also I don't think we need `OE_SENDGRID_0_CLIENT_ID`

Kailai-Wang (commented):
I'm generally fine with server-centric control just for now

tee-worker/omni-executor/config-loader/src/config.rs
BillyWooo: `OE_SENDGRID_0_CLIENT_ID` we still need something as flag to differentiate.

Kailai-Wang: Hmm... why?

Doesn't `OE_SENDGRID_API_HOST_WILDMETA` indicate "wildmeta" already? Same for other envs - can e.g. client-id "heima" use the env `OE_SENDGRID_API_HOST_WILDMETA`? If not, then the env name itself is sufficient as indicator: we always expect `{var_name}_{client_id}` as the full env name

silva-fj:
> Regarding sendgrid's doc, yes, we can and we are using in such a way. There is only one SendGrid account, where we have multiple api keys. Each key can bind together with different sender-email and sender-name.
> 

This is what I was referring to, I think you don't need multiple api keys if you are using the same account you can just use one, and configure the different allowed sender's. (I think I've done this in the past but I don't remember where, that's why I suggested to check the docs)

BillyWooo:
> > Regarding sendgrid's doc, yes, we can and we are using in such a way. There is only one SendGrid account, where we have multiple api keys. Each key can bind together with different sender-email and sender-name.
> 
> This is what I was referring to, I think you don't need multiple api keys if you are using the same account you can just use one, and configure the different allowed sender's. (I think I've done this in the past but I don't remember where, that's why I suggested to check the docs)

Oh, I didn't try that. 
Adding another KEY is just as two clicks simply. And it's even better for reviewing the statistics in the future. So I do it in this way. But anyway, this is not an issue.

tee-worker/omni-executor/config-loader/src/config.rs
```diff
@@ -75,47 +80,107 @@ fn get_env_value(var: &EnvVar) -> Option<String> {
...
+		warn!("No mailer configurations found. Looking for legacy single mailer config...");
+		client_ids.insert("default".to_string());
+	}
+
+	for mailer_id in client_ids {
+		// Use mailer_id as lowercase client_id
+		let client_id = mailer_id.to_lowercase();
+
+		let api_host_key = if mailer_id == "default" {
+			"OE_SENDGRID_API_HOST".to_string()
```
Kailai-Wang: We insert `default` to `client_id`, but the naming of these envs are inconsistent - I would expect something like `OE_SENDGRID_DEFAULT_API_HOST`

This might simplify code too

tee-worker/omni-executor/config-loader/src/config.rs
```diff
@@ -75,47 +80,107 @@ fn get_env_value(var: &EnvVar) -> Option<String> {
...
+			// Extract the mailer identifier from patterns like OE_SENDGRID_HEIMA_API_HOST, OE_SENDGRID_WILDMETA_API_KEY, etc.
+			// Split by underscores and take the part after OE_SENDGRID_ and before the last part
+			let parts: Vec<&str> = key.split('_').collect();
+			if parts.len() >= 4 && parts[0] == "OE" && parts[1] == "SENDGRID" {
+				// Extract client ID (e.g., "HEIMA" from "OE_SENDGRID_HEIMA_API_HOST")
+				let client_id = parts[2];
+				client_ids.insert(client_id.to_string());
+			}
+		}
+	}
```
Kailai-Wang: I'm not sure why we need an infix here (instead of e.g. suffix) - it makes parsing unnecessarily complicated..

tee-worker/omni-executor/config-loader/src/config.rs
```diff
@@ -267,4 +329,8 @@ impl ConfigLoader {
 			binance_api_base_url: get("binance_api_base_url"),
 		}
 	}
+
+	pub fn get_mailer_config(&self, client_id: &str) -> Option<&MailerConfig> {
```
Kailai-Wang: We seem to have two getters? 
`load_mailer_configs` vs `get_mailer_config`? When to use which?

BillyWooo: yes, can be removed. Not used right now.

tee-worker/omni-executor/config-loader/src/config.rs
BillyWooo: It's only for backward compatibility. I think I can remove it.

tee-worker/omni-executor/config-loader/src/config.rs
BillyWooo: I don't do it specifically. I just think this can work well. So add it.

tee-worker/omni-executor/config-loader/src/config.rs
```diff
@@ -75,47 +80,76 @@ fn get_env_value(var: &EnvVar) -> Option<String> {
...
+	let env_vars: Vec<_> = std::env::vars().collect();
+
+	// Find all mailer client IDs by looking for SENDGRID env vars
+	let mut client_ids = std::collections::HashSet::new();
+	for (key, _) in &env_vars {
+		if key.starts_with("OE_SENDGRID_") {
+			// Extract the mailer identifier from patterns like OE_SENDGRID_HEIMA_API_HOST, OE_SENDGRID_WILDMETA_API_KEY, etc.
+			// Split by underscores and take the part after OE_SENDGRID_ and before the last part
+			let parts: Vec<&str> = key.split('_').collect();
+			if parts.len() >= 4 && parts[0] == "OE" && parts[1] == "SENDGRID" {
```
Kailai-Wang: I assume the check of `parts[0]` and `parts[1]` are redundant as we have `if key.starts_with("OE_SENDGRID_")` already

tee-worker/omni-executor/config-loader/src/config.rs
```diff
@@ -75,47 +80,76 @@ fn get_env_value(var: &EnvVar) -> Option<String> {
...
+		let api_host_key = format!("OE_SENDGRID_{}_API_HOST", mailer_id);
+		let api_key_key = format!("OE_SENDGRID_{}_API_KEY", mailer_id);
+		let from_email_key = format!("OE_SENDGRID_{}_FROM_EMAIL", mailer_id);
+		let from_name_key = format!("OE_SENDGRID_{}_FROM_NAME", mailer_id);
+
+		// Check if at least one configuration variable exists for this mailer
+		let has_config = std::env::var(&api_key_key).is_ok()
+			|| std::env::var(&api_host_key).is_ok()
+			|| std::env::var(&from_email_key).is_ok()
+			|| std::env::var(&from_name_key).is_ok();
```
Kailai-Wang: Do they all need to exist in order to work?
Is this check covered by `if key.starts_with("OE_SENDGRID_")` already? (So one of these must exist)

tee-worker/omni-executor/config-loader/src/config.rs
```diff
@@ -75,47 +80,76 @@ fn get_env_value(var: &EnvVar) -> Option<String> {
...
+
+	// Find all mailer client IDs by looking for SENDGRID env vars
+	let mut client_ids = std::collections::HashSet::new();
+	for (key, _) in &env_vars {
+		if key.starts_with("OE_SENDGRID_") {
+			// Extract the mailer identifier from patterns like OE_SENDGRID_HEIMA_API_HOST, OE_SENDGRID_WILDMETA_API_KEY, etc.
+			// Split by underscores and take the part after OE_SENDGRID_ and before the last part
+			let parts: Vec<&str> = key.split('_').collect();
+			if parts.len() >= 4 && parts[0] == "OE" && parts[1] == "SENDGRID" {
+				// Extract client ID (e.g., "HEIMA" from "OE_SENDGRID_HEIMA_API_HOST")
```
Kailai-Wang: We extract it from `OE_SENDGRID_HEIMA_API_HOST`, but `api_host` itself can be optional - does it mean we could miss some clients?

tee-worker/omni-executor/config-loader/src/config.rs
Kailai-Wang: I mean `OE_SENDGRID_API_HOST_HEIMA` is much easier to process than`OE_SENDGRID_HEIMA_API_HOST`

Kailai-Wang (commented):
I think I have some problems regarding the current parsing logic, also which fields are mandatory.

I suggest we make it simple - to mark some fields as mandatory, if such field doesn't exist, then no matching `client_id` will be parsed and stored.

IMO the `default` variant should be used in `MailerConfig` getter when no matching `client_id` is found when we try to send out a mail, so something like here:

`let mailer = match ctx.get_mailer(&params.client_id) {`

Only if we allow fallback ofc.

Maybe @silva-fj @wli-pro could take a look too

tee-worker/omni-executor/rpc-server/src/mailer_factory.rs
```diff
@@ -0,0 +1,120 @@
...
+	/// Get or create a mailer for the specified client
+	pub fn get_mailer_for_client(
+		&self,
+		client_id: &str,
+	) -> Result<Arc<dyn MailerTrait + Send + Sync>, Box<dyn std::error::Error>> {
+		let client_key = client_id.to_lowercase();
+
+		// Try to get from cache first
+		{
+			let cache =
```
silva-fj: why crating a new scope for this?

tee-worker/omni-executor/rpc-server/src/mailer_factory.rs
```diff
@@ -0,0 +1,120 @@
...
+		let config = self
+			.config_loader
+			.get_mailer_config(client_id)
+			.ok_or_else(|| format!("No mailer configuration found for client '{}'", client_id))?;
+
+		let mailer = Self::create_mailer(config)?;
+
+		// Cache the mailer
+		{
+			let mut cache =
```
silva-fj: same here

tee-worker/omni-executor/.env.example
```diff
@@ -1,7 +1,22 @@
-# Sendgrid
-OE_SENDGRID_API_KEY=
-OE_SENDGRID_FROM_EMAIL=
-OE_SENDGRID_FROM_NAME=
+# Multi-Client Mailer Configuration
+# Console mailer for testing/development
+OE_MAILER_TYPE_CONSOLE=console
+OE_MAILER_FROM_EMAIL_CONSOLE=test@example.com
+OE_MAILER_FROM_NAME_CONSOLE='Console Mailer'
```
silva-fj, line 5: These `_CONSOLE` variables seem unnecessary

silva-fj (approved):

tee-worker/omni-executor/.env.example
BillyWooo: I would say it's better to have it. Keep the consistency and easier for parser to handle. No harm any way.

PR #3567, opened 2025-06-26
Fix ccsp

wli-pro:
### Labels

### Testing Evidences


tee-worker/omni-executor/accounting-contract-client/src/solana.rs
```diff
@@ -150,25 +150,18 @@ impl AccountingContractApi for AccountingContractClient {
...
-		let (account_nonce, _bump) =
-			Pubkey::find_program_address(&[user.to_bytes().as_ref(), b"nonce"], &self.program_id);
-
 		let program = client.program(self.program_id).map_err(|e| {
 			error!("Failed to create program client: {:?}", e);
 		})?;
 
 		let nonce_account: NonceAccount =
-			tokio::task::spawn_blocking(move || program.account(account_nonce))
+			tokio::task::spawn_blocking(move || program.account(user))
```
felixfaisal: Hey @wli-pro we're looking for the nonce of the account within the smart account, not the overall nonce of the account itself. 
When you first time call this instruction it will error because account has not been created, so if err, it needs to default to `0` 

I've done something similar here - https://github.com/litentry/heima/blob/787aac097769bbdf89ed5550b540d195d7f50900/tee-worker/omni-executor/accounting-contract/solana/tests/src/main.rs#L226 

You can refine it for better error handling accordingly as this was just CLI i didn't want to spend more time on that.

felixfaisal: Also why are we use `spawn_blocking` here?

wli-pro: > Also why are we use `spawn_blocking` here?

spawn_blocking is used to convert blocking operations into asynchronous tasks

wli-pro: > Hey @wli-pro we're looking for the nonce of the account within the smart account, not the overall nonce of the account itself. When you first time call this instruction it will error because account has not been created, so if err, it needs to default to `0`
> 
> I've done something similar here -
> 
> https://github.com/litentry/heima/blob/787aac097769bbdf89ed5550b540d195d7f50900/tee-worker/omni-executor/accounting-contract/solana/tests/src/main.rs#L226
> 
> You can refine it for better error handling accordingly as this was just CLI i didn't want to spend more time on that.

Thanks, I fixed this logic in the new commit here: https://github.com/litentry/heima/pull/3567/commits/555cc2bbd178a00dd6f6e4025496c1c44da416a7

felixfaisal: ~Looks good, But i do have one last question, have you tried using the `anchor-client` with the `async` feature enabled? I think this way you won't have to do `spawn_blocking` and can just use `await` on it.~

I saw your comment on another PR, i think you're already aware.

felixfaisal (approved):
LGTM

PR #3571, opened 2025-06-27
Add AA contract deployment scripts

kziemianek:
This PR adds a comprehensive deployment infrastructure for Account Abstraction contracts, supporting deployment to any EVM-compatible network with configurable options and proper deployment management.


tee-worker/omni-executor/aa-contracts/DEPLOYMENT.md
```diff
@@ -0,0 +1,203 @@
...
+
+Create a `.env` file in the project root:
+
+```bash
+# Required
+PRIVATE_KEY=0x1234567890abcdef...  # Your deployer private key
+RPC_URL=https://your-rpc-endpoint   # Network RPC URL
+
+# Optional
+PAYMASTER_INITIAL_DEPOSIT=1000000000000000000  # 1 ETH in wei (default: 1 ETH)
```
Kailai-Wang, line 53: If I understand the logic correctly - only half of the deposit goes into entrypoint `desposit`, the rest goes to `stake`

tee-worker/omni-executor/aa-contracts/DEPLOYMENT.md
```diff
@@ -0,0 +1,203 @@
...
+INITIALIZE_PAYMASTER=true                      # Whether to initialize paymaster (default: true)
```
Kailai-Wang, line 54: IIURC, our paymaster should always be used together with bundler? (I see authorised bundler list in the contract) - given the networks we want to support/deploy: bsc/arbi/hyperEVM/..., I don't think a public bundler can handle them all.

I think we basically have two options:
1. we just don't use paymaster/bundler - a TEE controlled account send handleUserOps to entrypoint contract directly - without considering preflight or gas consumption
2. we write a mini-embedded-bundler in worker, it doesn't need to have full-fledged features (e.g. no `eth_sendUserOperation` rpc impl, no UserOp mempool or gas estimation), but need to at least handle the preflight/simulation, so that we can return early if user's Op can't be dry-run

Kailai-Wang (commented):
I think paymaster/bundler problem is the most important thing to discuss, additionally I met some problem during deploying on sepolia, check DM (or maybe you could try to deploy it on sepolia first)

Kailai-Wang:
The script itself looks good now : )

tee-worker/omni-executor/aa-contracts/DEPLOYMENT.md
kziemianek: yes - we can adjust if needed

tee-worker/omni-executor/aa-contracts/DEPLOYMENT.md
kziemianek: Our AA contracts are not eip-4337 compilant as UserOperation contains additional fields related to sessions so we cannot use public bundlers. Also we need our own isntance of `EntryPoint`.

Yes - we should provide relayer/bundler instance.

tee-worker/omni-executor/aa-contracts/DEPLOYMENT.md
kziemianek: We can change deposit to 100% and stake to 0%, wdyt ?

Kailai-Wang (approved):

tee-worker/omni-executor/aa-contracts/DEPLOYMENT.md
Kailai-Wang: Yes please : )

PR #3572, opened 2025-06-30
Add support for WILDMETA in token validation

silva-fj:
Extend the allowed client IDs for JWT validation to include CLIENT_ID_WILDMETA. This enables authentication for tokens issued to the Wildmeta client, alongside existing support for Heima and PumpX.


0xverin (approved):
Thanks!

PR #3573, opened 2025-06-30
tidy up tee-worker/common

BillyWooo:
As `common` folder is only used by `identity` worker, it makes sense to merge it back inside `identity` folder.


Kailai-Wang (approved):
Looks good, thanks!

silva-fj (approved):
👍🏼

PR #3576, opened 2025-07-01
split pumpx and omni swap

kziemianek:
- Restructured cross-chain executor with modular `pumpx/` and `omni/` directories
  - Moved existing cross-chain logic into `pumpx/` module
  - Addes `omni/` module for AA based swap logic


silva-fj (approved):

Kailai-Wang:
Would you resolve the conflict and CI?

PR #3540, opened 2025-06-23
Construct EVM transactions inside Executor

felixfaisal:
So this PR as title attemps to construct transactions inside the executor itself using aggregators such as okx, 1inch and kyberswap. 

Majority of the code has been referenced from the pumpx backend, I've not integrated this with the current cross-chain execution and requesting signer, because i haven't been able to do a test with the aggregators, and I've created an issue for it P-1590, I've created integrations tests under examples/ to test them. 

This PR also does not cover four meme aggregator, it'll be covered under a separate issue P-1591

PR #3581, opened 2025-07-02
fix: using heima router contract

felixfaisal:
As per title, will have this ready for review once #3540 gets merged


felixfaisal:
I think i will merge this into a single PR

PR #3584, opened 2025-07-07
 feat: P-1587 add some mocks for evm api and unit test

wli-pro:
### Labels

### Testing Evidences

### DIFF
--- a/tee-worker/omni-executor/ethereum-rpc/src/lib.rs
+++ b/tee-worker/omni-executor/ethereum-rpc/src/lib.rs
@@ -138,10 +138,8 @@ impl RpcProvider for AlloyRpcProvider {
 		let mut tx = raw_tx.from(signer_address);
 		// Bump the gas by 10%
 		// TODO: Set gas fees via CLI
-		let gas_estimate = self.estimate_gas(tx.clone()).await?;
-		tx.gas = Some(gas_estimate * 110 / 100);
-		let gas_price = self.get_gas_price().await?;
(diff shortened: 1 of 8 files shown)


0xverin (approved):
LGTM, thanks! But I have a dumb question: When I call `omni_submitSwapOrder`, will I be using this mock server?

tee-worker/omni-executor/ethereum-rpc/src/lib.rs
```diff
@@ -138,8 +138,10 @@ impl RpcProvider for AlloyRpcProvider {
 		let mut tx = raw_tx.from(signer_address);
 		// Bump the gas by 10%
 		// TODO: Set gas fees via CLI
-		tx.gas = Some(self.estimate_gas(tx.clone()).await? * 110 / 100);
-		tx.gas_price = Some(self.get_gas_price().await?);
+		let gas_estimate = self.estimate_gas(tx.clone()).await?;
```
BillyWooo: maybe a dumb question: is there a special reason to break this line into two lines?

wli-pro: well spot it's no need, I have reverted

wli-pro:
> LGTM, thanks! But I have a dumb question: When I call `omni_submitSwapOrder`, will I be using this mock server?

You need start local omni-worker with mock feature via below command, then you can use the mock server
make build-docker-test
make start-test

PR #3589, opened 2025-07-08
Pack customised fields of `PackedUserOperation` into signature

Kailai-Wang:
### Context

As topic - to not break the compatibility/standard of PackedUserOp.

cc @silva-fj - it might affect demo-app code too - please adjust accordingly


tee-worker/omni-executor/aa-contracts/src/core/Primitives.sol
Kailai-Wang, line 1: I'm not sure creating a new `Primitives.sol` is the best paradigm for such primitive data struct definitions - or shall we put it in `interfaces/...`, or some existing file?

Kailai-Wang: cc @kziemianek @silva-fj

kziemianek: I would suggest to move it to `interfaces/UserOpSigType.sol`

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
```diff
@@ -102,23 +103,64 @@ contract OmniAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Init
...
-        address signer = ECDSA.recover(userOpHash, userOp.signature);
-        if (owner == _determineOa(signer) || isRootSigner(signer)) {
-            return SIG_VALIDATION_SUCCESS;
+        require(userOp.signature.length >= 1, "signature too short");
+
+        UserOpSigType sigType = UserOpSigType(uint8(userOp.signature[0]));
+        bytes calldata sig = userOp.signature[1:];
+
+        if (sigType == UserOpSigType.EOA) {
+            return _validateEOA(userOpHash, sig);
```
kziemianek: eoa | root | passkey - they are all same signature types (only ecdsa signature) but checked against different address (owner, one of root keys, one of passkey keys).

Maybe we should have only two signature types ?

Kailai-Wang: This is more explicit tho - so it bascially ensures the caller knowing what he is doing

tee-worker/omni-executor/aa-contracts/src/core/Primitives.sol
Kailai-Wang: Sure I'll do it

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
kziemianek: kk, just a small detail - all 3 (eoa, root, passkey) are **eoa** imho 
so i would rename eoa to owner in order to keep it consistent

kziemianek (approved):

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
Kailai-Wang: That's a good point, I'll rename it.

Also does it make more sense to call it `UserOpSigner` enum? @kziemianek 
As you mentioned it's same signature type (secp256k1), just different signer types/roles

PR #3590, opened 2025-07-08
`omni_submitUserOp` RPC method

kziemianek:
Contains several changes required for UserOp validation/simulation/signing and submission, mainly:

* New RPC endpoint for submitting ERC-4337 user operations
* User operation ownership validation against on-chain OmniWallet contracts
* New simulate_validation method using state override to deploy EntryPointSimulations contract
* New Solidity contracts: EntryPointSimulations.sol and IEntryPointSimulations.sol
* `omni-cli` binary for interacting with `omni_*` rpc methods


tee-worker/omni-executor/executor-core/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+	pub nonce: [u8; 32],                // U256 as byte array
+	pub init_code: Vec<u8>,             // Bytes as Vec<u8>
+	pub call_data: Vec<u8>,             // Bytes as Vec<u8>
+	pub account_gas_limits: [u8; 32],   // FixedBytes<32> as byte array
+	pub pre_verification_gas: [u8; 32], // U256 as byte array
+	pub gas_fees: [u8; 32],             // FixedBytes<32> as byte array
+	pub paymaster_and_data: Vec<u8>,    // Bytes as Vec<u8>
+	pub session_account: [u8; 20],      // Address as byte array
+	pub session_expiration: [u8; 32],   // U256 as byte array
+	pub session_account_proof: Vec<u8>, // Bytes as Vec<u8>
```
Kailai-Wang: We don't need these anymore

tee-worker/omni-executor/executor-core/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+	pub signature: Vec<u8>,             // Bytes as Vec<u8>
```
Kailai-Wang: Should the RPC UserOp be unsigned? (= without this field)

tee-worker/omni-executor/executor-core/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+use parity_scale_codec::{Decode, Encode};
+use serde::{Deserialize, Serialize};
+
+/// Serializable version of PackedUserOperation for use in NativeTask
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, Serialize, Deserialize)]
+pub struct SerializablePackedUserOperation {
+	pub sender: [u8; 20],               // Address as byte array
+	pub nonce: [u8; 32],                // U256 as byte array
```
Kailai-Wang: Hmm why isn't it integer?

tee-worker/omni-executor/executor-core/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Litentry.  If not, see <https://www.gnu.org/licenses/>.
+
+use parity_scale_codec::{Decode, Encode};
+use serde::{Deserialize, Serialize};
+
+/// Serializable version of PackedUserOperation for use in NativeTask
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, Serialize, Deserialize)]
+pub struct SerializablePackedUserOperation {
+	pub sender: [u8; 20],               // Address as byte array
```
Kailai-Wang: Is it possible to deserialise all fields with 0x prefix? We expect them coming in with 0x (we said that should be standard for hex-string)

tee-worker/omni-executor/executor-core/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+	pub nonce: [u8; 32],                // U256 as byte array
+	pub init_code: Vec<u8>,             // Bytes as Vec<u8>
+	pub call_data: Vec<u8>,             // Bytes as Vec<u8>
+	pub account_gas_limits: [u8; 32],   // FixedBytes<32> as byte array
+	pub pre_verification_gas: [u8; 32], // U256 as byte array
+	pub gas_fees: [u8; 32],             // FixedBytes<32> as byte array
```
Kailai-Wang: Same - should they be integers?

tee-worker/omni-executor/executor-primitives/src/lib.rs
```diff
@@ -78,3 +78,75 @@ impl GetEventId<EventId> for BlockEvent {
...
 }
+
+/// Represents supported blockchain networks for omni operations
+#[derive(
+	Clone, Debug, Hash, Encode, Decode, PartialEq, Eq, serde::Serialize, serde::Deserialize,
+)]
+pub enum Chain {
+	Evm(u64),
+	Solana,
+}
```
Kailai-Wang: Does it make sense to reuse `heima_primitives::omni::ChainType`?

Or maybe just `evm_chain_id` or even the old `pumpx_chain_id` if we want a single integer definition?

Kailai-Wang (commented):
I'd like to get the struct aligned first, when possible

tee-worker/omni-executor/executor-core/src/types.rs
kziemianek: It can be both signed / unsigned.

tee-worker/omni-executor/executor-core/src/types.rs
kziemianek: resolved

tee-worker/omni-executor/executor-core/src/types.rs
Kailai-Wang: Then should this field be optional?

tee-worker/omni-executor/executor-core/src/types.rs
Kailai-Wang: I think we have such example already in e.g. `SubmitSwapOrderParams`, where we use `decode_hex` to do the der

tee-worker/omni-executor/executor-core/src/types.rs
kziemianek: resolved

tee-worker/omni-executor/executor-core/src/types.rs
kziemianek: resolved

tee-worker/omni-executor/executor-core/src/types.rs
kziemianek: resolved

tee-worker/omni-executor/executor-core/src/types.rs
kziemianek: changed to optional 👍🏼

tee-worker/omni-executor/executor-primitives/src/lib.rs
kziemianek: Yes, I've replaced it with u64

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
```diff
@@ -0,0 +1,192 @@
...
+						error!("Invalid sender address '{}': {}", op.sender, e);
+						PumpxRpcError::from_error_code(ErrorCode::ParseError)
+					})
+				})
+				.collect::<Result<HashSet<_>, _>>()?;
+
+			// Validate each unique address once
+			for sender_address in unique_addresses {
+				if let Err(e) = validate_user_operation_ownership(
+					&user,
```
silva-fj: How about if we pass `address_bytes` here so we don't have to extracted again from `user`?

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
```diff
@@ -0,0 +1,192 @@
...
+					address_bytes.len()
+				);
+				return Err(PumpxRpcError::from_error_code(ErrorCode::InternalError));
+			}
+
+			let mut address = [0u8; 32];
+			address.copy_from_slice(&address_bytes);
+
+			// Collect unique sender addresses to avoid redundant validation calls
+			let unique_addresses: HashSet<Address> = params
```
silva-fj: Is it valid if the `user` send multiple user_operations with different sender addresses (`op.sender`)?

silva-fj: Can `op.sender` be different than `user.omni_account` ?

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
```diff
@@ -0,0 +1,192 @@
...
+		error!("No RPC client found for chain ID: {}", chain_id);
+		return Err(PumpxRpcError::from_error_code(ErrorCode::InternalError));
+	};
+
+	// Create a new client with the specific wallet address
+	let omni_client =
+		aa_contracts_client::OmniAccountClient::new(*sender_address, rpc_client.clone());
+
+	// Query the OmniWallet contract directly to get its stored OA
+	let stored_oa = match omni_client.get_owner().await {
```
silva-fj: Why do we need to query the contract? can we just compare `user.omni_account` with `op.sender` ?

silva-fj (commented):
I looks good to me in general. I just have a couple of questions. (I just want to understand better how everything should work)

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
kziemianek: `sender` is an OmniAccount instance address .

We need to query `owner` from the contract so we can make sure the OmniWallet belongs to user from request's context. 

We could remove this check and sign user ops blindly - then they should fail during OmniWallet validation phase but I still think it's worth to keep such authorization gate, so worker submit only UserOps that account legitimately own.

silva-fj: I think we should do the check, my question is if we can compare both here instead of making the rpc call? or `sender` is not the same as `owner` ?  

Also, should we check that the user is root signer instead of owner?  (Sorry, I'm just a bit confused about how everything should work)

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
```diff
@@ -40,19 +47,106 @@ impl<P: RpcProvider<Transaction = TransactionRequest, Addr = Address>> EntryPoin
...
+	/// Get EntryPointSimulations deployed bytecode (cached on first access)
+	///
+	/// This function loads the bytecode from the filesystem on first call and caches it
+	/// in memory for subsequent calls, improving performance.
+	fn get_simulation_bytecode() -> Result<String, ()> {
+		SIMULATION_BYTECODE_ONCE.call_once(|| {
+			// Use deployed bytecode, not creation bytecode
+			let bytecode_path = concat!(
+				env!("CARGO_MANIFEST_DIR"),
+				"/src/bytecode/EntryPointSimulations_deployed_bytecode.hex"
```
Kailai-Wang: Could you write a brief comment how this is generated? Either here or in readme

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
```diff
@@ -0,0 +1,192 @@
...
+			if address_bytes.len() != 32 {
+				error!(
+					"Invalid omni account length: expected 32 bytes, got {}",
+					address_bytes.len()
+				);
+				return Err(PumpxRpcError::from_error_code(ErrorCode::InternalError));
+			}
+
+			let mut address = [0u8; 32];
+			address.copy_from_slice(&address_bytes);
```
Kailai-Wang: Seems we can save this `address` variable and resue `address_bytes` directly?

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
Kailai-Wang: > Can op.sender be different than user.omni_account ?

`op.sender` should be the smart wallet **contract** address, which should be related to oa, but not directly comparable

> Is it valid if the user send multiple user_operations with different sender addresses (op.sender)?
This is theoretically possible, but we don't think we need to bother that here? Right now I don't think we'd have multiple smart wallets for the same user on a single chain - especially for UserOps submitted via this RPC.

We _could_ check:
- all UserOps have the same sender
- it's the smart wallet address (as per locally calculated)

but since it's not strictly needed (the simulation would fail if sent to the wrong address) and extra checking means more processing time, we could trust backend and save this.

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
Kailai-Wang: As mentioned before IMO we could skip the check (as we preflight the simulation anyway) - sure, it helps abort early if this check fails, but we are not like a public bundler that deals with all wild requests (in fact, it's quite possible that we keep the communication of backend <> worker internal - it's still in discussion), so we could assume these fields should be correctly filled in most cases

Kailai-Wang: And yes - if we want to check it - it's preferred to calculate locally instead of querying the contract to avoid network communication

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -921,3 +1096,120 @@ async fn verify_google_code(
...
+
+	Ok(aa_contracts_client::PackedUserOperation {
+		sender: Address::from_str(&user_op.sender)
+			.map_err(|e| format!("Invalid sender address '{}': {}", user_op.sender, e))?,
+		nonce: U256::from(user_op.nonce),
+		initCode: Bytes::from(
+			decode_hex(&user_op.init_code).map_err(|e| format!("Invalid init_code hex: {}", e))?,
+		),
+		callData: Bytes::from(
+			decode_hex(&user_op.call_data).map_err(|e| format!("Invalid call_data hex: {}", e))?,
```
Kailai-Wang, line 1150: Just want to make sure that if `initCode` would be correctly parsed (not errored error) if:
- left empty
- just with value `0x`

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -788,7 +808,162 @@ async fn handle_native_task<
...
+					send_error(
+						"Failed to get wallet address from EntryPoint client".to_string(),
+						response_sender,
+						NativeTaskError::InternalError,
+					);
+					return;
+				},
+			};
+
+			// Run simulation for each UserOperation before submission
```
Kailai-Wang, line 942: Hmm shall we actaully simulate the whole `handleOps()`?

I mean each Op could be pre-checked individually (for obvious errors), but we probably want to simulate the whole operation - it's possible that it's ok when checked separately, but problematic as a whole (e.g. exceeding whole gas limit)

Kailai-Wang (commented):
Thank you - I left some comments

Additional Q:
- shall we check the `preVerificationGas` field? Or at least have a hardcoded minimum
- shall we also pre-check the user/paymaster balance (via `entrypoint.balanceOf`) is >= `getRequiredPrefund + some-buffer-constant`?  The simulation would detect it too, but pre-check should be cheap and easy.

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -788,7 +808,162 @@ async fn handle_native_task<
...
+						.request_signature(
+							ChainType::Evm,
+							wallet_index,
+							omni_account.clone().into(),
+							message_to_sign,
+						)
+						.await;
+
+					let signature = match signature_result {
+						Ok(sig) => sig,
```
Kailai-Wang, line 894: I think we should prepend it with the `SignerType`:
https://github.com/litentry/heima/blob/81f3554c2583b3a792ed87448927241bcbe331e3/tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol#L114-L127

In this case `UserOpSigner.RootKey`

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
kziemianek: resolved

tee-worker/omni-executor/native-task-handler/src/lib.rs
kziemianek: Should be fixed

tee-worker/omni-executor/native-task-handler/src/lib.rs
kziemianek: added unit tests

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
kziemianek: removed

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_user_op.rs
kziemianek: resolved

Kailai-Wang (approved):
Thanks - will address the unresolved convos in next PRs

PR #3592, opened 2025-07-10
Migrate AA demo app from npm to pnpm

silva-fj:
This PR migrates the AA demo app from npm to pnpm for consistency with the rest of the monorepo.

Changes:
- Replace npm lockfile with pnpm lockfile
- Update README instructions to use pnpm commands
- Update demo setup script to use pnpm
- Remove aa-demo-app from dependabot npm configuration


silva-fj:
@Kailai-Wang This PR reverts the changes to dependabot, now it won't be needed as we'll use the same package manager as the rest on the monorepo

0xverin (approved):
Thanks!

PR #3596, opened 2025-07-10
Define primitives of passkey support

Kailai-Wang:
### Context

As topic - this PR introduces some primitive definition for passkey:
- login handle: should be `user_id` - we expect unique `user_id` across the system, he/she could have multiple credential_ids and keypairs tho
- `PasskeyData` for auth verification
- Solidity struct definitions and verification impl

Some details and implementations are yet to be finalised (in Issue 1613), e.g. what type the authenticator sends in `PasskeyData`, how often/when should user is prompted to sign using passkey


tee-worker/omni-executor/aa-contracts/src/interfaces/Passkey.sol
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        // 18. If user verification was determined to be required, verify that
+        // the UV bit of the flags in authData is set. Otherwise, ignore the
+        // value of the UV flag.
+        if (requireUserVerification && (flags & AUTH_DATA_FLAGS_UV) != AUTH_DATA_FLAGS_UV) {
+            return false;
+        }
+
+        // 19. If the BE bit of the flags in authData is not set, verify that
```
silva-fj, line 83: What does these numbers mean? (`17.`, `18.`, `19.`)

silva-fj (approved):
LGTM

kziemianek (approved):

tee-worker/omni-executor/aa-contracts/src/interfaces/Passkey.sol
Kailai-Wang: Ah it was copied from the referenced contract (link in the code comment: `Adapted from ...`) - I just kept them as it is

Kailai-Wang:
```
    /// Verifies the authFlags in authData. Numbers in inline comment
    /// correspond to the same numbered bullets in
    /// https://www.w3.org/TR/webauthn-2/#sctn-verifying-assertion.
```

PR #3604, opened 2025-07-15
Add trading test case examples

0xverin:
as topic

For the tests related to `submitSwapOrder`, we have skipped them directly in the ts test code. The reason is that the authentication token verification requires additional logic in the middleware.

From the front-end perspective, it is sufficient to understand the parameter structure. Therefore, we will maintain the current approach for now and optimize it later as needed for more detailed testing.


silva-fj (approved):
Should we create an issue for the skipped tests? So we can track this

PR #3607, opened 2025-07-18
add a new OmniAccountFactoryV2 to allow upgrade oa conract without ad…

wli-pro:
…dress change

### Labels

### Testing Evidences


Kailai-Wang:
Thank you for proposing this @wli-pro 

After careful consideration we probably have to ditch the "upgradable" factory, as:
- it complicates things a lot
- it breaks the security and auditability => "upgradable" means a centric entity (EOA / multisig / gov) would be able to control and push the upgrade (and thus control the wallet logic indirectly). It would be highly risky if the admin is malicious and/or compromised. It's very hard to make it future-proof: auditting could only check current impl, but not future visions.

Based on that I suggest we close this PR

PR #3617, opened 2025-07-22
Dynamic max gas price estimation

kziemianek:
Implements adaptive gas pricing for ERC-4337 EntryPoint transactions with network-specific configurations. The system now dynamically calculates gas fees based on current network conditions instead of using hardcoded values.


Kailai-Wang (approved):
Thank you!

But the gasConfig isn't applied to worker yet, is it? I see `new_with_config` is only called in tests

silva-fj (approved):

PR #3624, opened 2025-07-25
add Passkey validation for aa contract

BillyWooo:
only add the logic for aa contract and aa contract client. 
worker integration will come later.


silva-fj (commented):
It looks good to me in general. It would probably be a good idea to add some tests for this

tee-worker/omni-executor/aa-contracts-client/src/omni_account_client.rs
```diff
@@ -75,6 +78,74 @@ impl<P: RpcProvider<Transaction = TransactionRequest>> OmniAccountClient<P> {
...
+
+	pub async fn remove_passkey_signer_from_string(&self, pubkey_str: &str) -> Result<(), String> {
+		let pk = parse_passkey_public_key(pubkey_str)?;
+		self.remove_passkey_signer(pk)
+			.await
+			.map_err(|_| "Failed to remove passkey signer".to_string())
+	}
+}
+
+fn parse_passkey_public_key(pubkey_str: &str) -> Result<PasskeyPublicKey, String> {
```
Kailai-Wang: It depends on what format/encoding this `pubkey_str` has - I believe normally WebAuthn has a COSE encoded one, so probably we still need to change this and/or add some preprocessing.

Kailai-Wang (commented):
Yes, we still need tests (at least forge test) for the contract and then we are good to go

tee-worker/omni-executor/aa-contracts-client/src/omni_account_client.rs
```diff
@@ -75,6 +78,74 @@ impl<P: RpcProvider<Transaction = TransactionRequest>> OmniAccountClient<P> {
 			FixedBytes::<32>::abi_decode(&result).map_err(|_| error!("Could not decode owner"))?;
 		Ok(owner)
 	}
+
+	pub async fn add_passkey_signer(&self, pk: PasskeyPublicKey) -> Result<(), ()> {
+		let call_data = addPasskeySignerCall { pk }.abi_encode();
+		let tx = build_call_transaction(self.address, call_data);
+		self.rpc_client.send_transaction(tx).await.map_err(|_| ())?;
```
Kailai-Wang, line 85: These `self.rpc_client.send_transaction` calls probably won't work - as mentioned it won't pass the `onlyOwner()` check

tee-worker/omni-executor/aa-contracts/test/ValidationTests.t.sol
```diff
@@ -0,0 +1,697 @@
...
+import {UserOpSigner} from "../src/interfaces/UserOpSigner.sol";
+import {Passkey} from "../src/interfaces/Passkey.sol";
+import {Counter} from "../src/Counter.sol";
+import {OmniAccountTestUtils} from "./OmniAccountTestUtils.sol";
+import {TestUtils} from "./TestUtils.sol";
+import {PackedUserOperation} from "../src/interfaces/PackedUserOperation.sol";
+import {SIG_VALIDATION_FAILED, SIG_VALIDATION_SUCCESS} from "../src/core/Helpers.sol";
+import "@openzeppelin/contracts/utils/Base64.sol";
+
+contract ValidationTests is Test {
```
kziemianek: Can we have `OmniAccountAsPassKey.t.sol`  ? We've got similiar for `Owner` and `Root`

tee-worker/omni-executor/aa-contracts/test/ValidationTests.t.sol
Kailai-Wang, line 1: It seems to be OmniAccount contract specific - I suggest we rename it to `OmniAccountValidation.t.sol` - `t.sol` stands for test already- we don't have have add `*Tests` as suffix : )

tee-worker/omni-executor/aa-contracts/test/ValidationTests.t.sol
```diff
@@ -0,0 +1,697 @@
...
+        bytes memory emptySig = "";
+
+        vm.expectRevert("Owner signature length invalid");
+        testableAccount.validateOwnerPublic(userOpHash, emptySig);
+    }
+
+    // ============================================================
+    // ROOT KEY VALIDATION TESTS
+    // ============================================================
+
```
Kailai-Wang: I assume this and following secion was already covered in `OmniAccountAsRoot.t.sol`? Could you double-check it? 

Shall we merge it into that file? If so, maybe we should move the code snippet above to a separate file like `OmniAccountAsPasskey.t.sol`

tee-worker/omni-executor/aa-contracts/test/ValidationTests.t.sol
```diff
@@ -0,0 +1,697 @@
...
+        testMetadata = Passkey.Metadata({
+            authData: authData,
+            clientDataJSON: '{"type":"webauthn.get","challenge":"dGVzdA","origin":"https://example.com"}',
+            challengeIndex: 36,
+            typeIndex: 8,
+            userVerificationRequired: false
+        });
+    }
+
+    function test_ValidatePasskey_Direct_UnauthorizedPasskey() public view {
```
Kailai-Wang: A small hiccup: 

the name style of these functions aren't consistent with what we have in other test scripts

tee-worker/omni-executor/aa-contracts/test/ValidationTests.t.sol
```diff
@@ -0,0 +1,697 @@
...
+
+        bytes32 userOpHash = keccak256("test");
+        bytes memory sig = abi.encode(testPublicKey, testSignature, testMetadata);
+
+        // The test will fail at P256.verify since we don't have valid signature data
+        // But it tests that the authorization check passes
+        uint256 validationData = testableAccount.validatePasskeyPublic(userOpHash, sig);
+
+        // This will be SIG_VALIDATION_FAILED due to invalid P256 signature,
+        // but at least we verified authorization logic works
```
Kailai-Wang: How do we know that? I assume we can't tell 😂

tee-worker/omni-executor/aa-contracts/test/ValidationTests.t.sol
```diff
@@ -0,0 +1,697 @@
...
+            address(0), // P256 precompile address would be different
+            abi.encodeWithSignature("verify(bytes32,bytes32,bytes32,bytes32,bytes32)"),
+            abi.encode(true)
+        );
+
+        vm.prank(address(entryPoint));
+        uint256 validationData = account.validateUserOp(packedOp, userOpHash, 0);
+
+        // This will still fail due to P256 verification, but tests our authorization logic
+        assertEq(validationData, SIG_VALIDATION_FAILED, "Would pass with valid P256 signature");
```
Kailai-Wang: This is the problem: we want to test the happy path **with real data**.

It's stated as `Would pass with valid P256 signature`, but I bet not:
- the signature doesn't seem to be correct at first glance
- we are not using p256 precompile, so the `vm.mockCall` is supposed to have no effect

Kailai-Wang (commented):
Thank you, I left some comments, main problems:
- the validation tests are distributed in both separate files and this single test file, we should stick to one way of positioning
- we lack the happy path (verification ok) test with real data (signature/metadata)

tee-worker/omni-executor/aa-contracts/test/ValidationTests.t.sol
Kailai-Wang: I see - we'd need a working example in any case

tee-worker/omni-executor/aa-contracts/test/OmniAccountAsSessionKey.t.sol
Kailai-Wang, line 1: I think most tests are already covered in `OmniAccountAsRoot.t.sol`?

tee-worker/omni-executor/aa-contracts/test/OmniAccountAsPasskey.t.sol
```diff
@@ -0,0 +1,428 @@
...
+        packedOp.signature = fullSig;
+
+        // Mock the P256.verify call to return true for this test
+        // Note: In a real test, you would need actual valid P256 signature data
+        // For now, this tests the authorization check logic
+        vm.mockCall(
+            address(0), // P256 precompile address would be different
+            abi.encodeWithSignature("verify(bytes32,bytes32,bytes32,bytes32,bytes32)"),
+            abi.encode(true)
+        );
```
Kailai-Wang: I don't think we need this

tee-worker/omni-executor/aa-contracts/test/OmniAccountAsPasskey.t.sol
```diff
@@ -0,0 +1,428 @@
...
+        });
+
+        bytes memory sig = abi.encode(validPublicKey, validSignature, metadata);
+
+        // Mock the actual staticcall made by OpenZeppelin's P256 library
+        // When P256.verify is called, it makes a staticcall to the precompile
+        vm.mockCall(
+            address(0x100),
+            abi.encodeWithSignature(""),
+            abi.encode(uint256(1))
```
Kailai-Wang: Is it possible to test the happy path without any mock? This is what we want to test..

We need an ok case with `account.validateUserOp()`

tee-worker/omni-executor/aa-contracts/test/OmniAccountAsPasskey.t.sol
```diff
@@ -0,0 +1,428 @@
...
+        assertTrue(Passkey.checkAuthFlags(0x05, true), "Should pass with UP and UV flags when UV required");
+        assertFalse(Passkey.checkAuthFlags(0x01, true), "Should fail with only UP flag when UV required");
+
+        // Test BE/BS flags
+        assertTrue(Passkey.checkAuthFlags(0x09, false), "Should pass with UP and BE flags");
+        assertFalse(Passkey.checkAuthFlags(0x11, false), "Should fail with UP and BS but no BE");
+    }
+
+    function test_ValidatePasskeyDirectSucceed() public {
+        // Use the exact public key from JavaScript generation
```
Kailai-Wang: I think we can remove these `Use the exact .. from JavaScript generation` - as we can't see the js (or you attach it)

Kailai-Wang (commented):
Thanks! IMO we still miss the success case with `account.validateUserOp` 🤔

tee-worker/omni-executor/aa-contracts/test/OmniAccountAsPasskey.t.sol
BillyWooo: > We need an ok case with account.validateUserOp()

OK.

tee-worker/omni-executor/aa-contracts/test/OmniAccountAsRoot.t.sol
```diff
@@ -147,6 +149,143 @@ contract OmniAccountAsRoot is Test {
         account.removeRootSigner(0x0000000000000000000000000000000000000000);
     }
 
+    function test_ValidateOpPasskey() public {
```
Kailai-Wang: Would you move these passkey related tests to `OmniAccountAsPasskey.t.sol`?

tee-worker/omni-executor/aa-contracts-client/src/omni_account_client.rs
```diff
@@ -75,10 +78,238 @@ impl<P: RpcProvider<Transaction = TransactionRequest>> OmniAccountClient<P> {
...
+/// 1. Attempts COSE parsing first (CBOR-encoded keys with proper validation)
+/// 2. Falls back to raw EC formats for backward compatibility
+/// 3. Supports both hex-prefixed and non-prefixed formats
+///
+/// # COSE support:
+/// - ✅ Full CBOR parsing with `ciborium` dependency
+/// - ✅ P-256 COSE key validation (kty=2, alg=-7, crv=1)
+/// - ✅ Proper extraction of x and y coordinates
+/// - ✅ Comprehensive error handling for invalid COSE keys
+pub fn parse_passkey_public_key(pubkey_str: &str) -> Result<PasskeyPublicKey, String> {
```
Kailai-Wang, line 126: This should not be probably in some utils or helper file, but not in client - it's not used anyway

Kailai-Wang (approved):
Thanks!

@silva-fj @kziemianek could take a look too

tee-worker/omni-executor/aa-contracts/test/OmniAccountAsRoot.t.sol
BillyWooo: There is no such file anymore. 
So `Passkey` or `SessionKey` are all tested within the theme of `OmniAccountAsRoot`. Is it OK?

tee-worker/omni-executor/aa-contracts-client/src/omni_account_client.rs
BillyWooo: It's used just above.

tee-worker/omni-executor/aa-contracts/test/OmniAccountAsRoot.t.sol
Kailai-Wang: Better to create such a file - they represent different access control

SessionKey was embeded in rootkey as it needs siganture from rootkey and is always used together with rootkey

tee-worker/omni-executor/aa-contracts-client/src/omni_account_client.rs
Kailai-Wang: I see - thanks!
I still believe it's not the best place for it 😂 (even `add_passkey_signer_from_string` itself) - as it's meant to be aa wallet client only - these long helpers would worsen the readability a lot

Anyways it's not that critical in this PR

PR #3628, opened 2025-07-29
Introduce `OwnerType` into `OmniAccount.initialize`

Kailai-Wang:
### Context

To record the owner type

@silva-fj could you help check if there're more places in ts are affected?


silva-fj:
Yeah there are a few things we need to update in the demo app. I'll make the changes and push them here

tee-worker/omni-executor/aa-contracts/src/interfaces/OwnerType.sol
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * The owner type
+ * Required as a single byte in the `OmniAccount.initialize` to allow for more fine-grained access control
+ *
+ * See P-1644 for more context
+ *
+ * The fields of this enum are 1 to 1 mapped to the rust `UserId` type in tee-worker/omni-executor/executor-primitives/src/auth.rs
+ */
+enum OwnerType {
```
kziemianek, line 14: I just wanted to confirm that we keep prefix [here](https://github.com/litentry/heima/blob/dev/tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol#L97) unchanged, right ?

Kailai-Wang: Yes

kziemianek (approved):

PR #3629, opened 2025-07-30
Versioned AA contracts

kziemianek:
Prepare initial release of Account Abstraction contracts by adding V1 suffix and proper attribution.


tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccountFactory.sol
```diff
@@ -69,4 +69,8 @@ contract OmniAccountFactory {
             )
         );
     }
+
+    function version() public pure returns (string memory) {
+        return "1.0.0";
```
Kailai-Wang, line 75: Hmm it has the same version as `OmniAccountFactoryV1` - shall we simply delete the old version and just keep the `V1` variants? (same for other contracts)

Kailai-Wang (approved):
Thank you - I'm wondering if we should simply delete the old versions (only keeping `V1` variants)

What about paymaster? We'll probably have an ERC20 Paymaster later - should it be made as in the same series with `SimplePaymaster` (so also following `PaymasterV*` style)?

kziemianek:
> Thank you - I'm wondering if we should simply delete the old versions (only keeping `V1` variants)
> 
> What about paymaster? We'll probably have an ERC20 Paymaster later - should it be made as in the same series with `SimplePaymaster` (so also following `PaymasterV*` style)?

If deployed then yes.

PR #3630, opened 2025-07-31
Enhance OmniAccount access control with signer hierarchy

kziemianek:
- Non-EVM Owner Support: Root signers can call onlyOwner functions for non-EVM owner types (Email, Twitter, etc.)
- Passkey Priority: When passkey signers exist, they take precedence over root signers for onlyOwner access
- Enhanced Validation: Added function selector validation to prevent indirect calls to restricted functions
- EntryPoint Integration: Allow direct onlyOwner calls from EntryPoint with validation safety


tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
```diff
@@ -63,8 +73,16 @@ contract OmniAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Init
     }
 
     function _onlyOwner() internal view {
-        // Directly from EOA owner, or through the account itself (which gets redirected through execute())
-        require(_determineOa(msg.sender) == owner || msg.sender == address(this), "only owner");
+        // Directly from EOA owner, through the account itself (which gets redirected through execute()),
+        // or from EntryPoint (safe because _validateSignature ensures only owner-signed UserOps can call restricted functions)
```
Kailai-Wang: I think it complicates stuff a bit, if we can do restricted call directly in callData (i.e. via entrypoint), is there any case we'd like to wrap it with `execute()`?

Kailai-Wang: Hold on, I realise a potential risk:

`msg.sender == address(this)` is actually problematic in 7702 context, isn't it? A random EOA (non-owner) could construct 7702 tx which would bypass this check and make restricted call easily.

Shall we remove `msg.sender == address(this)` check, as well as `_isExecuteWithRestrictedCall`? So basically we allow restricted call from:
- entrypoint, which enforces validation check
- evm-owner or rootSigner for non-OCVI user (for convenience)

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
```diff
@@ -128,10 +145,32 @@ contract OmniAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Init
...
         } else if (signer == UserOpSigner.Passkey) {
+            // For non-EVM owner types with passkey signers, allow passkeys to call restricted functions
+            if (ownerType != OwnerType.Evm && passkeySignerCount > 0) {
+                // Passkey signers can act as owners for non-EVM accounts
+                return _validatePasskey(userOpHash, sig);
+            }
+            // Otherwise, check if the UserOp is trying to call a restricted function
+            if (_isRestrictedCall(userOp.callData)) {
+                return SIG_VALIDATION_FAILED;
+            }
```
Kailai-Wang: There's a edge case that can be a design debate: an EVM-owner could also bind a passkey: confirming with FaceID can be much more convenient than wallet signing.

If we treat passkey as privileged as owner - here we should simplify the logic to just `return _validatePasskey(userOpHash, sig);`

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
```diff
@@ -32,9 +32,19 @@ contract OmniAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Init
...
     mapping(bytes32 => bool) public passkeySigners;
+    uint256 public passkeySignerCount;
 
     IEntryPoint private immutable _entryPoint;
 
+    // Selectors for restricted functions that only owner should call
+    bytes4 private constant ADD_ROOT_SIGNER_SELECTOR = bytes4(keccak256("addRootSigner(address)"));
+    bytes4 private constant REMOVE_ROOT_SIGNER_SELECTOR = bytes4(keccak256("removeRootSigner(address)"));
+    bytes4 private constant ADD_PASSKEY_SIGNER_SELECTOR = bytes4(keccak256("addPasskeySigner((uint256,uint256))"));
+    bytes4 private constant REMOVE_PASSKEY_SIGNER_SELECTOR = bytes4(keccak256("removePasskeySigner((uint256,uint256))"));
```
Kailai-Wang: Should it be flattened fn sig like:
- `addPasskeySigner(uint256,uint256)`
- `removePasskeySigner(uint256,uint256)`
? 

We'll need tests for it

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
```diff
@@ -32,9 +32,19 @@ contract OmniAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Init
...
+    bytes4 private constant WITHDRAW_DEPOSIT_SELECTOR = bytes4(keccak256("withdrawDepositTo(address,uint256)"));
+    bytes4 private constant UPGRADE_TO_SELECTOR = bytes4(keccak256("upgradeTo(address)"));
```
Kailai-Wang: It seems no `upgradeTo()` interface is present but w/e

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
```diff
@@ -32,9 +32,19 @@ contract OmniAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, Init
...
+    bytes4 private constant UPGRADE_TO_AND_CALL_SELECTOR = bytes4(keccak256("upgradeToAndCall(address,bytes)"));
```
Kailai-Wang: I'm not sure if it's easy to add test case for it too (maybe in another PR)

Kailai-Wang (changes requested):
Thanks! I'm mainly concerned in risks exposed

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
kziemianek: The real function signature is `addPasskeySigner(Passkey.PublicKey)` but raw function selectors are not compiled so they cannot reference struct -> that's why it needs to be turned into tuple.

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
kziemianek: That's the case only for only owner function calls which don't happen frequently.

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
Kailai-Wang: The above statement is incorrect.

7702 can't bypass it - it will make `msg.sender == address(this)` hold, but only on the attacker's own code, not victims.

So there's no vulnerability - but I'd still suggest we ditch `execute()` case for simplicity

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
Kailai-Wang: True - just for consistency

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccount.sol
Kailai-Wang: I see - thanks for the explanation

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccountV1.sol
```diff
@@ -233,20 +271,65 @@ contract OmniAccountV1 is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, In
...
+    function _isRestrictedCall(bytes calldata callData) internal pure returns (bool) {
+        if (callData.length < 4) return false;
+
+        bytes4 selector = bytes4(callData[0:4]);
+
+        // Check against all restricted selectors
+        return selector == ADD_ROOT_SIGNER_SELECTOR || selector == REMOVE_ROOT_SIGNER_SELECTOR
+            || selector == ADD_PASSKEY_SIGNER_SELECTOR || selector == REMOVE_PASSKEY_SIGNER_SELECTOR
+            || selector == WITHDRAW_DEPOSIT_SELECTOR || selector == UPGRADE_TO_AND_CALL_SELECTOR
+            || _isExecuteWithRestrictedCall(selector, callData);
```
Kailai-Wang: Is `_isExecuteWithRestrictedCall` still needed at all? Can we ditch it too?

tee-worker/omni-executor/aa-contracts/src/accounts/OmniAccountV1.sol
```diff
@@ -92,11 +108,11 @@ contract OmniAccountV1 is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, In
         emit AccountInitialized(_entryPoint, owner, ownerType, clientId, aRoot);
     }
 
-    // Require the function call went through EntryPoint or be signed by [owner|root]
+    // Require the function call went through EntryPoint or be signed by owner
     function _requireForExecute() internal view virtual override {
         require(
-            msg.sender == address(entryPoint()) || _determineOa(msg.sender) == owner || isRootSigner(msg.sender),
-            "account: not Owner or EntryPoint or root"
+            msg.sender == address(entryPoint()) || msg.sender == address(this) || _determineOa(msg.sender) == owner,
```
Kailai-Wang: When would `msg.sender == address(this)` be checked?

Shall we remove it for simplicity?

Kailai-Wang (commented):
Thanks!

Only two small things about `execute()`:

since we have the control logic in the validation phase, can we just make it straightforward in `execute()` (like only owner or entrypoint, or even just use `BaseAccount._requireForExecute()` which allows entrypoint only)? It helps with reduce the complexity so no strange compound permission problem might occur - although I can't think of one now

Kailai-Wang (approved):
Thanks!

PR #3631, opened 2025-07-31
feat: add robust transaction retry mechanism for handleOps submission

silva-fj:
## Summary
- Implement automatic retry with gas escalation for failed `handleOps` transactions
- Add chain-specific retry configurations (mainnet, L2s, BSC, HyperEVM)
- Handle transient gas price issues with exponential backoff and 5-20% gas increases per retry

## Solution
Based on research following some examples of bundlers and DEX aggregators:
- **Retry with gas escalation**: 10-20% gas increase per attempt.
- **Exponential backoff**: Prevents network spam while allowing quick retries (2s → 4s → 8s)
- **Chain-specific configs**: Optimized for each network's characteristics (e.g., HyperEVM fast finality needs fewer retries)

**TODO**: Smart error classification: Only retry gas-related failures, avoiding unnecessary attempts on permanent errors

### DIFF
--- a/tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
+++ b/tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
@@ -214,27 +214,49 @@ impl<P: RpcProvider<Transaction = TransactionRequest, Addr = Address>> EntryPoin
 
 	/// Calculate dynamic gas fees based on current network conditions
 	async fn calculate_gas_fees(&self) -> Result<(U256, U256), ()> {
-		// Get current gas price from network
-		let current_gas_price = self
-			.rpc_client
-			.get_gas_price()
-			.await
-			.map_err(|_| error!("Failed to fetch gas price"))?;
+		// Try EIP-1559 estimation first
+		match self.rpc_client.estimate_eip1559_fees().await {
+			Ok(eip1559_estimate) => {
+				// Use EIP-1559 fees with buffer
+				let buffer_multiplier = 100 + self.gas_config.gas_price_buffer_percent;
 
-		// Apply buffer to current gas price for max_fee_per_gas
-		let buffer_multiplier = 100 + self.gas_config.gas_price_buffer_percent;
-		let max_fee_per_gas = U256::from(current_gas_price)
-			.saturating_mul(U256::from(buffer_multiplier))
-			.checked_div(U256::from(100))
-			.unwrap_or(U256::from(current_gas_price));
+				let max_fee_per_gas = U256::from(eip1559_estimate.max_fee_per_gas)
+					.saturating_mul(U256::from(buffer_multiplier))
+					.checked_div(U256::from(100))
+					.unwrap_or(U256::from(eip1559_estimate.max_fee_per_gas));
 
-		// Calculate priority fee (tip) with bounds
-		// Use 10% of current gas price as priority fee, bounded by min/max
-		let priority_fee = U256::from(current_gas_price / 10)
-			.max(U256::from(self.gas_config.min_priority_fee))
-			.min(U256::from(self.gas_config.max_priority_fee));
+				// Use the EIP-1559 priority fee with bounds
+				let priority_fee = U256::from(eip1559_estimate.max_priority_fee_per_gas)
+					.max(U256::from(self.gas_config.min_priority_fee))
+					.min(U256::from(self.gas_config.max_priority_fee));
 
-		Ok((max_fee_per_gas, priority_fee))
+				Ok((max_fee_per_gas, priority_fee))
+			},
+			Err(_) => {
+				// Fallback to legacy gas price calculation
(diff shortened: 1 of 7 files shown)


tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
```diff
@@ -96,24 +96,112 @@ impl GasPriceConfig {
...
+		match chain_id {
+			1 => Self::mainnet(),    // Ethereum mainnet
+			137 => Self::l2(),       // Polygon
+			42161 => Self::l2(),     // Arbitrum
+			10 => Self::l2(),        // Optimism
+			8453 => Self::l2(),      // Base
+			56 => Self::bsc(),       // BSC
+			999 => Self::hyperevm(), // HyperEVM
+			998 => Self::hyperevm(), // HyperEVM Testnet
+			_ => Self::default(),
```
Kailai-Wang, line 175: So all testnets except for hyperEVM uses default setting - is it intended? (either way is ok - just want to confirm)

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
```diff
@@ -247,6 +335,112 @@ impl<P: RpcProvider<Transaction = TransactionRequest, Addr = Address>> EntryPoin
...
+		let total_buffer = self.gas_config.gas_price_buffer_percent + additional_buffer_percent;
+		let buffer_multiplier = 100 + total_buffer;
+
+		let max_fee_per_gas = U256::from(current_gas_price)
+			.saturating_mul(U256::from(buffer_multiplier))
+			.checked_div(U256::from(100))
+			.unwrap_or(U256::from(current_gas_price));
+
+		// Calculate priority fee (tip) with bounds
+		let priority_fee = U256::from(current_gas_price / 10)
```
Kailai-Wang: Just wondering if that's the common way of estimation?

Kailai-Wang (approved):
Looks good from my side!

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
```diff
@@ -247,6 +335,112 @@ impl<P: RpcProvider<Transaction = TransactionRequest, Addr = Address>> EntryPoin
...
+						info!(
+							"Transaction succeeded after {} retries with gas buffer {}%",
+							attempt, gas_buffer_adjustment
+						);
+					}
+					return Ok(tx_hash);
+				},
+				Err(_) => {
+					// Since RPC provider returns unit error, we treat all errors as potentially gas-related
+					// TODO: improve error handling, classify retryable vs non-retryable errors
```
kziemianek, line 396: When do we plan to resolve this todo ?

kziemianek (approved):

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
silva-fj: Good question, this is the logic we currently have so I didn’t research it. (I just added the `additional_buffer_percent` to the formula). To me it looks fine but I don't know if it is a common way of estimation.

I found that alloy provides this method: [estimate_eip1559_fees](https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.estimate_eip1559_fees), how about if we use it instead? c.c @kziemianek

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
silva-fj: I think that since hyperEVM testnet/mainnet both have the same characteristics, we could use the same config

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
Kailai-Wang: Could you have an issue for it?

silva-fj: Yes, I'll create an issue for this

tee-worker/omni-executor/aa-contracts-client/src/entry_point_client.rs
Kailai-Wang: Good catch - I read through the doc, it seems `estimate_eip1559_fees` should be used for chains that support EIP1559 (most chains nowadays), `get_gas_price` is meant for legacy tx only. They use different `eth_*` calls under the hood: `eth_feeHistory` + `eth_maxPriorityFeePerGas` vs `eth_gasPrice`.

My suggestion:

- if chain supports EIP1159 => use `estimate_eip1559_fees`
- if not => use `get_gas_price`

silva-fj: Done

PR #3635, opened 2025-08-04
Improve Ethereum RPC client error handling

silva-fj:
## Summary

- Replace unit type errors with typed error enums throughout ethereum-rpc and aa-contracts-client
- Add error classification with retryability checks for better retry logic
- Improve error propagation and handling of contract reverts
- Update all dependent modules to handle new error types


Kailai-Wang (approved):
It looks good!

Won't be able to verify every case, but it seems to be carefully implemented from the logic

PR #3642, opened 2025-08-06
P-1527 Do not send nativeTask over the channel

felixfaisal:
As per title, We are avoiding sending the tasks through the channels and handling it in another thread as it is redundant to do it since we can handle the async tasks directly with jsonrpsee which uses tokio already. 

So for this PR, I've moved all parameters of `TaskContext` to `RpcContext` so that we can directly use `handle_native_task` function, this can be further improved by moving the `handle_native_task` to the `rpc-server` crate and removing the `TaskContext` completely, for now I'm not doing that. 

There are a couple of unit tests i've commented out, I will fix them as this PR gets reviewed. There is also tracing span that i've removed which i will add back after testing it.


Kailai-Wang:
Thanks @felixfaisal  - could you also resolve conflict and refactor the newest rpc method `omni_getHyperliquidSignatureData` too?

Kailai-Wang:
@gemini-cli /review

tee-worker/omni-executor/rpc-server/src/methods/omni/get_approve_agent_wallet_data.rs
BillyWooo, line 1: should remove this empty file?

BillyWooo (approved):
LGTM. One empty file needs to be removed.
Looks like Claude doesn't recognise this empty file.

tee-worker/omni-executor/rpc-server/src/methods/omni/get_approve_agent_wallet_data.rs
felixfaisal: Looks like it creeped in while resolving conflicts, crazy how the model didn't find this xD

felixfaisal: removed

kziemianek (approved):

PR #3654, opened 2025-08-07
Implement omni_estimateUserOpGas RPC method

silva-fj:
Implements the `omni_estimateUserOpGas` RPC method for ERC-4337 UserOperation gas estimation.

- Added EstimateUserOpGas native task types and handlers
- Implemented gas estimation logic following ERC-4337 specifications
- Added helper functions for packing and unpacking gas limits
- Comprehensive test coverage for gas estimation and helper functions


Kailai-Wang (approved):
I'm good with the logic - just a bit concerned about speed of binary search in estimating call gas limit, as well as the multiple `eth_call` calls that would drain the RPC API usage fast (probably we'll need a paid plan for oe)

silva-fj:
> I'm good with the logic - just a bit concerned about speed of binary search in estimating call gas limit, as well as the multiple `eth_call` calls that would drain the RPC API usage fast (probably we'll need a paid plan for oe)

I was thinking about this, another way is to have multiple free api keys and rotate them on rate limit errors. Alloy provides a built-in retry layer (at the transport level) that we could use to do this: https://github.com/alloy-rs/examples/blob/main/examples/layers/examples/retry_layer.rs

Kailai-Wang:
> > I'm good with the logic - just a bit concerned about speed of binary search in estimating call gas limit, as well as the multiple `eth_call` calls that would drain the RPC API usage fast (probably we'll need a paid plan for oe)
> 
> I was thinking about this, another way is to have multiple free api keys and rotate them on rate limit errors. Alloy provides a built-in retry layer (at the transport level) that we could use to do this: https://github.com/alloy-rs/examples/blob/main/examples/layers/examples/retry_layer.rs

Yes - this retry layer might be sufficient for low frequency and few key rotations. If we really have many users, we'll need to do it in a compound way: paid plan + key rotation 😂 

Btw I believe the callData is quite predictable, currently it's either token transfer, or some bridge contract call - maybe backend should cache the gas estimation by grouping the calldata type and only update it when it's not fit (e.g. fails)

PR #3672, opened 2025-08-14
Wildmeta backend auth

kziemianek:
This PR adds support for a new WildmetaBackend authentication method to the omni-executor RPC server.

 Key Features:
  - New ClientAuth variant: Added WildmetaBackend { signature } to support backend signature verification
  - Enhanced RPC validation: Added wallet_index and client_id validation for WildmetaBackend auth
  - Signature verification: Implemented backend signature verification against user operation hashes using ECDSA recovery


tee-worker/omni-executor/.env.example
```diff
@@ -60,6 +60,9 @@ OE_ENTRY_POINT_ADDRESS=
 # Wildmeta API
 OE_WILDMETA_API_URL=
 
+# Heima Backend ECDSA Public Key (for WildmetaBackend auth signature verification)
+OE_HEIMA_BACKEND_ECDSA_PUBKEY=0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```
BillyWooo: do you think it's better to name it `WILDMETA_BACKEND`,  instead of `HEIMA_BACKEND` ?
And all the following related places.

tee-worker/omni-executor/executor-crypto/src/secp256k1.rs
```diff
@@ -75,3 +75,27 @@ pub fn secp256k1_ecdsa_recover_compressed(
...
 	Ok(pubkey.serialize())
 }
+
+/// Sign a message with a SECP256k1 private key.
+///
+/// - `private_key` is a 32-byte private key
+/// - `msg` is the 32-byte hash to sign
+///
+/// Returns the 65-byte signature in RSV format.
+pub fn secp256k1_ecdsa_sign(
```
BillyWooo, line 85: This function is only used in test code?

BillyWooo (commented):
In general, it looks good to me.
Left two comments.

tee-worker/omni-executor/rpc-server/src/auth_utils.rs
```diff
@@ -43,6 +49,69 @@ pub fn verify_wildmeta_signature(
...
+		error!("Invalid signature length: expected 65 bytes, got {}", signature_bytes.len());
+		return Err(ErrorObject::from(ErrorCode::ParseError));
+	}
+
+	let signature_array: [u8; 65] = signature_bytes
+		.try_into()
+		.map_err(|_| ErrorObject::from(ErrorCode::ParseError))?;
+
+	// Convert first user operation to PackedUserOperation for hashing
+	let first_user_op = convert_to_packed_user_op(user_operations[0].clone()).map_err(|e| {
```
BillyWooo: here only verify the first op may be not enough. After a short discussion with @Kailai-Wang , how about we verify all ops? Then we change `signature` -> `vector of signatures` ? one by one with exact match.

Kailai-Wang: I'm concerned with replay attack risk here - it would be very dangerous if the signature is leaked, as attacker could attach any number of malicious userOps and the auth would be successful 🤔 

It doesn't matter if the first UserOp validation would eventually fail (because of e.g. duplicate nonce), the UserOps are independent with each other.

Maybe we should simply turn the signature into an array too (1:1 to UserOp array element)

tee-worker/omni-executor/executor-crypto/src/secp256k1.rs
kziemianek: yes

tee-worker/omni-executor/rpc-server/src/auth_utils.rs
```diff
@@ -43,6 +49,78 @@ pub fn verify_wildmeta_signature(
...
+		})?;
+
+		let user_op_hash =
+			calculate_user_operation_hash(&packed_user_op, entry_point_address, chain_id);
+		combined_hash_data.extend_from_slice(&user_op_hash.0);
+	}
+
+	// Hash the combined data using keccak256 to create a single 32-byte hash
+	use alloy::primitives::keccak256;
+	let user_op_hash = keccak256(&combined_hash_data);
```
Kailai-Wang, line 96: Compound hash works too - we just have to make sure backend is aware of that even for 1-element array

tee-worker/omni-executor/rpc-server/src/auth_utils.rs
```diff
@@ -43,6 +49,78 @@ pub fn verify_wildmeta_signature(
...
+
+	// Convert user op hash to 32-byte array
+	let user_op_hash_array: [u8; 32] = user_op_hash.0;
+
+	// Recover public key from signature
```
Kailai-Wang: I think just verifying would be sufficient (and slightly faster : )), see e.g. https://github.com/litentry/dex-signer/pull/13

tee-worker/omni-executor/rpc-server/src/auth_utils.rs
kziemianek: fixed

kziemianek: fixed

tee-worker/omni-executor/.env.example
kziemianek: changed

BillyWooo (approved):

PR #3679, opened 2025-08-20
config OE_WILDMETA_BACKEND_ECDSA_PUBKEY through github secret

BillyWooo:
as title, introduce new PUBKEY as github secret and load it during release process


kziemianek:
It's hardcoded by design so it cannot be changed freely on host after deployment. Would you like to explain why in this case we loosen it?

kziemianek (approved):
Looks fine.
Alternative way of setting this which I recommend is to set it in job definition - public key is not necessary meant to be secret.

Kailai-Wang (approved):

PR #3687, opened 2025-08-26
Initial impl of an ERC20 Paymaster

Kailai-Wang:
### Context

As topic.

An initial implementation of an ERC20 Paymaster:
- it only checks ERC20 allowance and payment, not native token (ETH) => there's no allowance/spender for native token, paymaster can't "pull" ETH from user wallet unless user explicitly deposits/funds it
- if the userOp itself is an `approve` userOp, it:
   - checks if user has enough balance to cover gas in **validation** phase
   - charges user for actual cost in **postOp** phase (it could fail - but there's no good way around it)
- if the userOp is a non-approve userOp, it:
   - checks if user has enough balance to cover gas in **validation** phase
   - checks if user has granted paymaster sufficient allowance in **validation** phase
   - charges user for `maxCost` in **validation** phase
   - refunds user if actualCost is less than prefund in **postOp** phase
 - it checks the postOpGasLimit against a minimum `MIN_POST_OP_GAS_LIMIT` (currently set to 50_000)

Reference:
- https://github.com/pimlicolabs/singleton-paymaster/blob/master/src/SingletonPaymasterV7.sol
- https://github.com/eth-infinitism/account-abstraction-samples/blob/master/contracts/TokenPaymaster.sol

See [tee-worker/omni-executor/aa-contracts/ERC20PaymasterV1-README.md](./tee-worker/omni-executor/aa-contracts/ERC20PaymasterV1-README.md) for more


tee-worker/omni-executor/aa-contracts/src/core/ERC20PaymasterV1.sol
```diff
@@ -0,0 +1,494 @@
...
+            revert InsufficientDeposit();
+        }
+
+        // Decode paymaster data first to validate structure
+        PaymasterData memory data = _decodePaymasterData(userOp.paymasterAndData);
+
+        // Extract and validate postOp gas limit after data validation
+        uint256 postOpGasLimit = userOp.unpackPostOpGasLimit();
+        if (postOpGasLimit < MIN_POST_OP_GAS_LIMIT) {
+            revert InsufficientDeposit(); // Reuse existing error for simplicity
```
kziemianek: I agree it's not necessary related to deposit and may be misleading.

tee-worker/omni-executor/aa-contracts/src/core/ERC20PaymasterV1.sol
```diff
@@ -0,0 +1,494 @@
...
+    using UserOperationLib for PackedUserOperation;
+    using Math for uint256;
+
+    // Minimum required postOp gas limit to prevent paymaster losses
+    uint256 private constant MIN_POST_OP_GAS_LIMIT = 50000;
+
+    // Struct to decode paymaster data
+    struct PaymasterData {
+        address token; // ERC20 token address (must not be address(0))
+        uint256 exchangeRate; // Exchange rate: how many token units per 1 wei of ETH
```
kziemianek, line 30: Who sets this value?

tee-worker/omni-executor/aa-contracts/src/core/ERC20PaymasterV1.sol
Kailai-Wang: I see, let me give it another name

tee-worker/omni-executor/aa-contracts/src/core/ERC20PaymasterV1.sol
Kailai-Wang: I talked with backend.

Currently (in reality) all `UserOp`s are submitted from backend, regardless of clientAuth type (`WildmetaHl` or `WildmetaBackend`). But it's technically possible for FE/user to construct such a request using agentWallet signature (so `WildmetaHl`), the worker would accept it too.

So the safest solution right now is that worker fills the `exchangeRate`.

tee-worker/omni-executor/aa-contracts/src/core/ERC20PaymasterV1.sol
Kailai-Wang: Fixed and added a test case

tee-worker/omni-executor/aa-contracts/src/core/ERC20PaymasterV1.sol
Kailai-Wang: It will be implemented in worker code, the contract itself is not affected. So I'll resolve this convo now

kziemianek (approved):

silva-fj (approved):

PR #3691, opened 2025-09-02
Fill the `exchange_rate` within worker for ERC20Paymaster

Kailai-Wang:
### Context

As topic.

The worker detects if it's an ERC20 paymaster data based on the length of `userOp.paymasterAndData`. Then it will fetch the token price, calculates the `exchangeRate` and populates it.

### DIFF
--- a/tee-worker/omni-executor/Cargo.toml
+++ b/tee-worker/omni-executor/Cargo.toml
@@ -55,7 +55,6 @@ chrono = "0.4.40"
 ciborium = "0.2"
 clap = "4.5.17"
 email_address = "0.2"
-env_logger = "0.11.5"
 ethers = "2.0"
 frame-metadata = { version = "23.0.0", features = ["decode"], default-features = false }
 frame-support = { version = "39.0.0", features = ["std"] }
--- a/tee-worker/omni-executor/Cargo.toml
+++ b/tee-worker/omni-executor/Cargo.toml
@@ -54,8 +54,8 @@ bs58 = "0.5"
 chrono = "0.4.40"
 ciborium = "0.2"
 clap = "4.5.17"
-env_logger = "0.11.5"
 email_address = "0.2"
+env_logger = "0.11.5"
 ethers = "2.0"
 frame-metadata = { version = "23.0.0", features = ["decode"], default-features = false }
 frame-support = { version = "39.0.0", features = ["std"] }
--- a/tee-worker/omni-executor/binance-api/src/lib.rs
+++ b/tee-worker/omni-executor/binance-api/src/lib.rs
@@ -41,6 +41,14 @@ pub trait BinanceApi: Send + Sync {
 		T: 'static + serde::de::DeserializeOwned;
 }
 
+// Dyn-compatible trait for ERC20 paymaster operations
+#[async_trait]
+pub trait BinancePaymasterApi: Send + Sync {
+	async fn get_symbol_price(&self, symbol: &str) -> Result<String, Error>;
+	async fn get_exchange_info_for_symbol(&self, symbol: &str) -> Result<Option<u8>, Error>;
+	async fn get_all_trading_symbols(&self) -> Result<Vec<String>, Error>;
+}
+
 #[derive(Debug, Clone)]
 pub struct BinanceApiClient {
 	client: Client,
@@ -194,12 +202,55 @@ impl BinanceApi for BinanceApiClient {
 	}
 }
 
+#[async_trait]
+impl BinancePaymasterApi for BinanceApiClient {
+	async fn get_symbol_price(&self, symbol: &str) -> Result<String, Error> {
+		use crate::spot_trading_api::types::SymbolPrice;
+
+		let endpoint = "/api/v3/ticker/price";
+		let mut params = HashMap::new();
+		params.insert("symbol".to_string(), symbol.to_string());
+
+		let price_info: SymbolPrice = self.make_public_get_request(endpoint, Some(params)).await?;
+		Ok(price_info.price)
+	}
+
+	async fn get_exchange_info_for_symbol(&self, symbol: &str) -> Result<Option<u8>, Error> {
+		use crate::spot_trading_api::types::ExchangeInfo;
+
+		let endpoint = "/api/v3/exchangeInfo";
+		let mut params = HashMap::new();
+		params.insert("symbol".to_string(), symbol.to_string());
+
+		let exchange_info: ExchangeInfo =
+			self.make_public_get_request(endpoint, Some(params)).await?;
+
+		// Find the trading pair and return quote asset precision
+		if let Some(symbol_info) = exchange_info.symbols.first() {
+			Ok(Some(symbol_info.quote_asset_precision))
+		} else {
+			Ok(None)
+		}
+	}
+
+	async fn get_all_trading_symbols(&self) -> Result<Vec<String>, Error> {
+		use crate::spot_trading_api::types::ExchangeInfo;
+
+		let endpoint = "/api/v3/exchangeInfo";
+
+		let exchange_info: ExchangeInfo = self.make_public_get_request(endpoint, None).await?;
+
+		// Return all trading symbols
+		Ok(exchange_info.symbols.iter().map(|symbol| symbol.symbol.clone()).collect())
+	}
+}
+
 #[cfg(feature = "mocks")]
 pub mod mocks {
-	use crate::BinanceApi;
 	use crate::Error;
 	use crate::HashMap;
 	use crate::Method;
+	use crate::{BinanceApi, BinancePaymasterApi};
 	use async_trait::async_trait;
 	use mockall::mock;
 
@@ -228,6 +279,14 @@ pub mod mocks {
 			) -> Result<T, Error>
 			where
 				T: 'static + serde::de::DeserializeOwned;
+
+		}
+
+		#[async_trait]
+		impl BinancePaymasterApi for BinanceApiClient {
+			async fn get_symbol_price(&self, symbol: &str) -> Result<String, Error>;
+			async fn get_exchange_info_for_symbol(&self, symbol: &str) -> Result<Option<u8>, Error>;
+			async fn get_all_trading_symbols(&self) -> Result<Vec<String>, Error>;
 		}
 	}
 }
--- a/tee-worker/omni-executor/binance-api/src/lib.rs
+++ b/tee-worker/omni-executor/binance-api/src/lib.rs
@@ -205,14 +205,8 @@ impl BinanceApi for BinanceApiClient {
 #[async_trait]
 impl BinancePaymasterApi for BinanceApiClient {
 	async fn get_symbol_price(&self, symbol: &str) -> Result<String, Error> {
-		use crate::spot_trading_api::types::SymbolPrice;
-
-		let endpoint = "/api/v3/ticker/price";
-		let mut params = HashMap::new();
-		params.insert("symbol".to_string(), symbol.to_string());
-
-		let price_info: SymbolPrice = self.make_public_get_request(endpoint, Some(params)).await?;
-		Ok(price_info.price)
+		use crate::spot_trading_api::SpotTradingApi;
+		SpotTradingApi::new(self).get_symbol_price(symbol).await
 	}
 
 	async fn get_symbol_precision(&self, symbol: &str) -> Result<u8, Error> {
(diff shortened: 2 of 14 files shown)


tee-worker/omni-executor/Cargo.toml
```diff
@@ -54,8 +54,8 @@ bs58 = "0.5"
 chrono = "0.4.40"
 ciborium = "0.2"
 clap = "4.5.17"
-env_logger = "0.11.5"
 email_address = "0.2"
+env_logger = "0.11.5"
```
kziemianek: i think this can be safely removed

tee-worker/omni-executor/binance-api/src/lib.rs
```diff
@@ -194,12 +202,55 @@ impl BinanceApi for BinanceApiClient {
 	}
 }
 
+#[async_trait]
+impl BinancePaymasterApi for BinanceApiClient {
+	async fn get_symbol_price(&self, symbol: &str) -> Result<String, Error> {
```
kziemianek, line 207: I don't see this being used - am I missing anything?

tee-worker/omni-executor/Cargo.toml
Kailai-Wang: True - I'll remove it

> i think this can be safely removed

tee-worker/omni-executor/binance-api/src/lib.rs
```diff
@@ -194,12 +202,55 @@ impl BinanceApi for BinanceApiClient {
...
+
+		let endpoint = "/api/v3/ticker/price";
+		let mut params = HashMap::new();
+		params.insert("symbol".to_string(), symbol.to_string());
+
+		let price_info: SymbolPrice = self.make_public_get_request(endpoint, Some(params)).await?;
+		Ok(price_info.price)
+	}
+
+	async fn get_exchange_info_for_symbol(&self, symbol: &str) -> Result<Option<u8>, Error> {
```
silva-fj: Should we rename this method to something like `get_asset_precision` ? Otherwise it should probably return `ExchangeInfo`.

tee-worker/omni-executor/binance-api/src/lib.rs
```diff
@@ -194,12 +202,55 @@ impl BinanceApi for BinanceApiClient {
...
+		let endpoint = "/api/v3/exchangeInfo";
+		let mut params = HashMap::new();
+		params.insert("symbol".to_string(), symbol.to_string());
+
+		let exchange_info: ExchangeInfo =
+			self.make_public_get_request(endpoint, Some(params)).await?;
+
+		// Find the trading pair and return quote asset precision
+		if let Some(symbol_info) = exchange_info.symbols.first() {
+			Ok(Some(symbol_info.quote_asset_precision))
```
silva-fj: How about if we just return `Result`  so it's easier to handle?`-> Result<u8, Error>`

Kailai-Wang: Sounds good - let me update this

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -817,6 +1138,36 @@ pub async fn handle_native_task<
...
+				if !packed_user_op.paymasterAndData.is_empty() {
+					match process_erc20_paymaster_data(
+						ctx.binance_api_client.as_ref() as &dyn BinancePaymasterApi,
+						&packed_user_op.paymasterAndData,
+						chain_id,
+					)
+					.await
+					{
+						Ok(Some(updated_paymaster_data)) => {
+							packed_user_op.paymasterAndData = updated_paymaster_data;
```
kziemianek: I think we should not change user op once it's signed.

tee-worker/omni-executor/binance-api/src/lib.rs
Kailai-Wang: I renamed it to `get_symbol_precision`

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: Good catch, I've updated the logic (also it should happen before signature request)

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -759,13 +1078,47 @@ pub async fn handle_native_task<
 						},
 					};
 
-				// Check if UserOperation is signed
+				// Check if UserOperation is signed, if not:
```
kziemianek: If I construct and sign UserOp by my own that means i can specify `exchange_rate` - but this PR should ensure it's filled by worker. Does it mean we've got a gap to fill?

Kailai-Wang: That's a good point.
It can imply a broader problem tho => does it mean that user could use our bundler freely by signing the userOp? (as bundler would just verify the auth and accept it)

Shall we disallow paymaster usage if userOp.signature is non-empty?

kziemianek: If we want to control `paymasterData` for `Erc20` paymaster we should read it's address from `UserOp` and if matches then:
1. reject UserOp if signed
2. fill exchange rate if not signed

Kailai-Wang: But I'm not talking about ERC20 paymaster only - but SimplePaymaster that we currently have too.

The user (FE) could construct a userOp with signature and rpc-auth that passes the bundler's check and bundler will submit it to entrypoint (so paymaster would sponsor the gas), right?

kziemianek: That's true - this point was raised before and we decided to keep it as it is. 
If we want to change that also then that's the way to go (but we only check for our paymasters instances usage).
But this might be still not enough - we should check what are we going to sponsor (payload) because client can still construct unsigned tx with any call and we will sponsor it anyway.

So in my opinion there are two things:
* who specifies exchange rate (we should ensure it's correct - out of user's control)
* what calls are sponsored (we may eventually want to ensure we sponsor only selected functions)

Kailai-Wang: After discussion, we'll manage it in a simple way:  **use of our paymaster requires an unsiged userOp**.

It means if the user submits a **signed** userOp, we expect that the target paymaster isn't ours (read from config)

Kailai-Wang: I updated the logic, see the comment @kziemianek :

```
// Whitelisted paymaster addresses that we support (deployed by heima), the logic is:
// - Unsigned userOp: If paymaster specified, **must** be whitelisted
// - Signed userOp: Only allowed if paymasterAndData is empty (technically we could still relay it, but we are a private bundler and don't want to relay arbitrary userOps)
//
// These addresses are assumed to be the same across all EVM chains
//
// Note: the SimplePaymaster should be gradually deprecated in favor of the ERC20PaymasterV1.
//       Theoretically user could construct an unsiged userOp to use SimplePaymaster "freely"
```

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -41,12 +42,369 @@ use pumpx::{
...
+	chain_id: u64,
+	binance_api_client: Option<&dyn BinancePaymasterApi>,
+) -> Option<(String, u8)> {
+	// First try the existing hardcoded lookup
+	if let Some(result) = get_token_info_from_mapping(token_address, chain_id).await {
+		return Some(result);
+	}
+
+	// If not found in hardcoded list and we have a Binance API client, try dynamic discovery
+	if let Some(binance_client) = binance_api_client {
```
kziemianek, line 243: Not sure if this is needed 🤔 will it work ? It will probably always return same symbol

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -41,12 +42,369 @@ use pumpx::{
...
+// ============================================================================
+// ERC20 Paymaster Exchange Rate Processing
+// ============================================================================
+
+// Constants for ERC20 paymaster processing
+// paymasterAndData format: paymaster_address (20) + validation_gas_limit (16) + postop_gas_limit (16) + paymaster_data
+// paymaster_data: token(20) + exchangeRate(32) + validUntil(32) + validAfter(32)
+const MIN_ERC20_PAYMASTER_DATA_LENGTH: usize = 52 + 20 + 32 + 32 + 32; // 168 bytes minimum
+const PAYMASTER_DATA_OFFSET: usize = 52; // paymaster address (20) + validation_gas_limit (16) + postop_gas_limit (16)
+const EXCHANGE_RATE_FEE_PERCENT: f64 = 0.0; // 0% fee for now
```
kziemianek, line 61: This is something we might want to adjust in the future  - what do you think about providing it from env ?

We can also keep 0  hardcoded for now and implement this once really needed.

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -41,12 +42,369 @@ use pumpx::{
...
+//
+// These addresses are assumed to be the same across all EVM chains
+//
+// Note: the SimplePaymaster should be gradually deprecated in favor of the ERC20PaymasterV1.
+//       Theoretically user could construct an unsiged userOp to use SimplePaymaster "freely"
+//
+// TODO: add ERC20 paymaster addresses
+const WHITELISTED_PAYMASTER_ADDRESSES: &[&str] = &[
+	// SimplePaymaster
+	"0x6255B9F4A4E80BC20eE389fD35DE9d2c029D5912", // staging-v1
```
kziemianek, line 75: Could be provided by env too.

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -41,12 +42,369 @@ use pumpx::{
...
+// Whitelisted paymaster addresses that we support (deployed by heima), the logic is:
+// - Unsigned userOp: If paymaster specified, **must** be whitelisted
+// - Signed userOp: Only allowed if paymasterAndData is empty (technically we could still relay it, but we are a private bundler and don't want to relay arbitrary userOps)
+//
+// These addresses are assumed to be the same across all EVM chains
+//
+// Note: the SimplePaymaster should be gradually deprecated in favor of the ERC20PaymasterV1.
+//       Theoretically user could construct an unsiged userOp to use SimplePaymaster "freely"
+//
+// TODO: add ERC20 paymaster addresses
```
kziemianek, line 72: Should we remove this TODO ?

kziemianek (approved):

tee-worker/omni-executor/binance-api/src/lib.rs
```diff
@@ -194,12 +202,54 @@ impl BinanceApi for BinanceApiClient {
 	}
 }
 
+#[async_trait]
+impl BinancePaymasterApi for BinanceApiClient {
+	async fn get_symbol_price(&self, symbol: &str) -> Result<String, Error> {
+		use crate::spot_trading_api::types::SymbolPrice;
+
+		let endpoint = "/api/v3/ticker/price";
```
silva-fj: I've just realized we already have methods for these endpoints in the `SpotTradingApi`

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: Yea I had the same thought but also the concern it could be modified locally on the host

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: You mean the build or runtime env? Same concerns if using runtime env

If build env, then it's the same as hardcoding it IMO 😂 - we'll need to rebuild it in case of changes

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: Good point - by a second look the `token_address` is not used

I looked through binance-api, it seems that "retrieving token symbol from token address" is not supported by binance-api :(

Then we either need to rely on a third-party service (CMC?) or remove the misleading dynamic discovery and expand the hardcoded mapping. I prefer the latter @kziemianek @silva-fj

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: Not yet - only after the ERC20 paymaster is deployed and address is known

tee-worker/omni-executor/binance-api/src/lib.rs
Kailai-Wang: Good catch, I removed the duplicate

tee-worker/omni-executor/native-task-handler/src/lib.rs
Kailai-Wang: Let me merge it and raise a separate PR to update that

PR #3693, opened 2025-09-03
feat: disable omni_submitUserOpTest in production worker

silva-fj:
## Summary

Added `test-endpoints` feature flag to control test-only RPC methods in omni-executor. The `omni_submitUserOpTest` endpoint is now excluded from production builds by default and only included when explicitly enabled with the feature flag.

### DIFF
--- a/tee-worker/omni-executor/Makefile
+++ b/tee-worker/omni-executor/Makefile
@@ -3,6 +3,7 @@
 
 ARCH_LIBDIR ?= /lib/$(shell $(CC) -dumpmachine)
 SGX ?= 1
+TEST_BUILD ?= 0
 
 BIN_NAME = omni-executor
 COMPILATION_FEATURES =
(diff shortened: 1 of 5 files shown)


tee-worker/omni-executor/Makefile
```diff
@@ -22,6 +22,11 @@ $(BIN_NAME):
 	cargo build --release --locked --features=$(COMPILATION_FEATURES)
 	cp target/release/executor-worker $(BIN_NAME)
 
+.PHONY: $(BIN_NAME)-test
+$(BIN_NAME)-test:
+	cargo build --release --locked --features=$(COMPILATION_FEATURES),test-endpoints
+	cp target/release/executor-worker $(BIN_NAME)
```
Kailai-Wang: May I know how should I make build the staging worker now?

So far it's been `make SGX=1 SGX_DEBUG=0`

silva-fj: We can do: `make omni-executor-test SGX=1 SGX_DEBUG=0`

Kailai-Wang: > We can do: `make omni-executor-test SGX=1 SGX_DEBUG=0`

But it won't build sgx metadata files like `omni-executor.manifest.sgx` / `omni-executor.sig` etc, right?

silva-fj: Right. I've change it so we can do: `make SGX=1 SGX_DEBUG=0 TEST_BUILD=1` now

kziemianek: This means we no logner can deploy releases to staging :(

Kailai-Wang: On staging we deploy it manually anyway 😂 

I'm fine with `make SGX=1 SGX_DEBUG=0 TEST_BUILD=1`

Kailai-Wang (approved):

PR #3694, opened 2025-09-08
Add Wildmeta email verification template

silva-fj:
## Summary
- Added custom email verification template for Wildmeta client with branded styling
- Created dedicated function to send Wildmeta verification emails
- Updated RPC handler to conditionally use Wildmeta template based on client ID


silva-fj:
It needs to be tested on staging, my sendgrid api key is not valid anymore

Kailai-Wang (approved):
Thanks - will deploy on staging to see the effect

PR #3705, opened 2025-09-10
feat: tracing metrics

felixfaisal:
This PR allows us to collect tracing metrics via a Jaeger UI so that we have better observability of our service. With this being hosted we will not have to ssh into the production server to view logs and also by separating the spans, it gets easier to search for a specific trace event rather than having to grep. 

For now the tracing is present only in the block syncer and rpc methods, we can further instrument other lower level function to get a better view of the system. I've included a jaegar docker-compose file to test it out. 

To test it out, first start the jaegar UI using `docker-compose` using the below commands: 
```
docker compose -f docker/jaegar.yml up 
```
and then run the following command to start the omni-executor with jaeger enabled
```
RUST_LOG=info cargo run --release --features mock-server --bin executor-worker -- run --metrics-port 9091 --parentchain-sync --enable-jaeger
```

and then you can view the jaegar UI in http://localhost:16686/

Here i'm attaching a few example pictures of tracing via the Jaeger UI
Screenshot 2025-09-10 at 12 50 48 PM

Screenshot 2025-09-10 at 11 31 26 AM


tee-worker/omni-executor/executor-worker/src/tracing_config.rs
```diff
@@ -0,0 +1,21 @@
+use opentelemetry::trace::TracerProvider;
+use opentelemetry_otlp::{SpanExporter, WithExportConfig};
+use opentelemetry_sdk::trace::{SdkTracer, SdkTracerProvider};
+use opentelemetry_sdk::Resource;
+
+pub fn init_trace(
+	port: u16,
+) -> Result<SdkTracer, Box<dyn std::error::Error + Send + Sync + 'static>> {
+	let endpoint = format!("http://localhost:{}", port);
```
kziemianek: is localhost hardcoded here ok?

tee-worker/omni-executor/rpc-server/src/methods/omni/get_hyperliquid_signature_data.rs
```diff
@@ -211,221 +219,244 @@ pub fn register_get_hyperliquid_signature_data<
...
-			// Unified authentication logic
-			let main_address = if let Some(user_auth) = &params.user_auth {
-				// User authentication provided
-				let auth =
-					to_omni_auth(user_auth, &params.user_id, &params.client_id).map_err(|e| {
-						error!("Failed to convert to OmniAuth: {:?}", e);
-						DetailedError::new(PARSE_ERROR_CODE, "Failed to convert authentication data")
-							.with_field("user_auth")
-							.with_reason(format!("OmniAuth conversion error: {:?}", e))
+		verify_auth(ctx.clone(), &auth).await.map_err(|e| {
```
kziemianek, line 247: this instrumented method verifies auth while `handle_export_wallet_request` not - why?

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_swap_order.rs
```diff
@@ -106,6 +106,250 @@ struct BackendResponse {
 	pub market_order_response: Option<SendOrderTxResponse>,
 }
 
+#[tracing::instrument(skip(params, user, ctx), fields(
```
kziemianek, line 109: As you can see this requires a lot of refactor - have you considered using simply 
```
    let span = span!(Level::INFO, "my_function", arg);
    let _enter = span.enter();
``` 
?

tee-worker/omni-executor/executor-worker/src/tracing_config.rs
felixfaisal: I've added another cli arg to avoid hardcoding it, There is a possibility that we might not have jaeger hosted in the same server as the one worker is deployed, So best to have this as an argument too.

tee-worker/omni-executor/rpc-server/src/methods/omni/get_hyperliquid_signature_data.rs
felixfaisal: I don't think the original implementation uses `verify_auth()` in it's implementation - https://github.com/litentry/heima/blob/3209f738f70454dce866fda64bb2d1d3033c8b72/tee-worker/omni-executor/rpc-server/src/methods/omni/export_wallet.rs#L73 

It's using `check_auth()` , if you notice, i've kept all `check_auth()` outside the `handlers` as it simply extracts credentials from the authenticated requests.

While `verify_auth()` seems to validates cryptographic signatures or emails, so i felt this makes more sense to be inside the `handlers`. 

Let me know what you think.

tee-worker/omni-executor/rpc-server/src/methods/omni/submit_swap_order.rs
felixfaisal: I'm already using this for block syncing inside the `listener.rs` - https://github.com/litentry/heima/blob/e40d5eabf9322d52bef989b80030e7591b1a78f9/tee-worker/omni-executor/executor-core/src/listener.rs#L108

However I decided to instrument functions instead of using this pattern as i wanted to decouple `register_async_methods` from the business logic, as I noticed couple of unit tests having to start an `rpc-server` to test the business logic. 

Here - https://github.com/litentry/heima/blob/3209f738f70454dce866fda64bb2d1d3033c8b72/tee-worker/omni-executor/rpc-server/src/methods/omni/get_shielding_key.rs#L111
https://github.com/litentry/heima/blob/3209f738f70454dce866fda64bb2d1d3033c8b72/tee-worker/omni-executor/rpc-server/src/methods/omni/get_shielding_key.rs#L65
https://github.com/litentry/heima/blob/3209f738f70454dce866fda64bb2d1d3033c8b72/tee-worker/omni-executor/rpc-server/src/methods/omni/get_health.rs#L72

This felt bloated to me, as we're starting a different/new RPC server for each unit test and requesting it within tests, separating the `rpc-server` from the business logic will allow us to unit test individual functions now and in the future without having to start an `rpc-server` for it. 

I've already done the refactor this way, Let me know what you think.

felixfaisal:
I don't think the team requires it as of now, So for now i'll close this PR, Feel free to open it if you think it adds value.

PR #3721, opened 2025-09-17
feat: add ERC20 paymaster token cost estimation

silva-fj:
## Summary
- Implement token cost estimation for ERC20 paymaster operations
- Fetch real-time token prices from Binance API
- Calculate and return estimated token costs in gas estimation responses


tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -1525,16 +1640,42 @@ async fn estimate_user_op_gas(
...
+		call_gas_limit,
+		verification_gas_limit,
+		pre_verification_gas,
+		paymaster_verification_gas_limit,
+		paymaster_post_op_gas_limit,
+		estimated_token_cost: None, // Will be updated if ERC20 paymaster is detected
+	};
+
+	// Step 6: Calculate token cost if ERC20 paymaster is present
+	let estimated_token_cost = if !user_op.paymasterAndData.is_empty() {
```
kziemianek, line 1678: Does it always mean erc20 paymaster is being used ?

silva-fj: No, we check if it's present in `calculate_erc20_token_cost`. Let me move the comment inside the if block to make it clear

BillyWooo (approved):

Kailai-Wang:
The logic looks fine - I'm just wondering if this is actually required/used by backend (or if it's too verbose to calculate and return such information), I'm feeling the logic is heavier and heavier

silva-fj:
> The logic looks fine - I'm just wondering if this is actually required/used by backend (or if it's too verbose to calculate and return such information), I'm feeling the logic is heavier and heavier

This is something the backend could calculate on their own, they would just need to use the estimate gas rpc method then call binance api or any other api to get the exchange rate and estimate the cost in erc20 token.

Kailai-Wang (approved):

PR #3723, opened 2025-09-19
increase preVerificationGas to cover bundler overhead

kziemianek:
User ops together should cover all `handleOps` execution fees.

### DIFF
--- a/tee-worker/omni-executor/native-task-handler/src/lib.rs
+++ b/tee-worker/omni-executor/native-task-handler/src/lib.rs
@@ -1069,7 +1069,8 @@ pub async fn handle_native_task<
 			// Calculate bundler overhead compensation accounting for existing preVerificationGas
 			const BASE_BUNDLER_OVERHEAD_GAS: u128 = 180_000;
 			const OVERHEAD_SAFETY_MULTIPLIER: f64 = 1.15; // 15% buffer for failed UserOps
-			let total_overhead_with_buffer = (BASE_BUNDLER_OVERHEAD_GAS as f64 * OVERHEAD_SAFETY_MULTIPLIER) as u128;
+			let total_overhead_with_buffer =
+				(BASE_BUNDLER_OVERHEAD_GAS as f64 * OVERHEAD_SAFETY_MULTIPLIER) as u128;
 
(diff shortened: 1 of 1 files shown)


BillyWooo (approved):

tee-worker/omni-executor/native-task-handler/src/lib.rs
```diff
@@ -1061,19 +1061,83 @@ pub async fn handle_native_task<
...
+			// Count unsigned and signed operations
+			let unsigned_ops_count =
+				serializable_user_ops.iter().filter(|op| op.signature.is_none()).count();
+			let signed_ops_count = serializable_user_ops.len() - unsigned_ops_count;
+
+			// Calculate bundler overhead compensation accounting for existing preVerificationGas
+			const BASE_BUNDLER_OVERHEAD_GAS: u128 = 180_000;
+			const OVERHEAD_SAFETY_MULTIPLIER: f64 = 1.15; // 15% buffer for failed UserOps
+			let total_overhead_with_buffer =
+				(BASE_BUNDLER_OVERHEAD_GAS as f64 * OVERHEAD_SAFETY_MULTIPLIER) as u128;
```
Kailai-Wang, line 1073: As per my understanding, we should not check the sum of the preVerificationGas against some threshold, but each one individually. Yes they are eventually batched into a single `handleOps` call, but each userOp must provide enough `preVerificationGas` to cover its share of calldata + overhead, and the real transaction’s overhead should be roughly equal to the sum of it.

Kailai-Wang: My assumption is naive and simple: just check `preVerificationGas` in each UserOp against some threshold (can be static or dynamic - e.g. against the calldata byte-size), and reject it if the check fails.

I'd avoid modifying the userOp when possible - meaning the backend/client needs to fill in the correct gas estimation. This is what `omni_estimateUserOpGas` for (it's possible that we need to improve the logic for `preVerificationGas` estimation from our side, cc @silva-fj )

Kailai-Wang: Wdyt?

PR #3726, opened 2025-09-22
fix anvil chain id

kziemianek:
It uses `31337`

```
Chain ID
==================

31337
```


BillyWooo (approved):

BillyWooo (commented):
Please update the test code.

PR #3733, opened 2025-09-23
Clean up on-chain storage for old bridge

Kailai-Wang:
### Context

We've switched to omni-bridge completely, so the old bridge storage is of no use (if we need historical state, we could query the archive rpc node).

Before removing the old-bridge pallets from runtime, this PR removes the on-chain storage first.

Tested via `try-runtime` CLI:
```
try-runtime --runtime ./target/debug/wbuild/heima-runtime/heima_runtime.wasm on-runtime-upgrade --blocktime 12000   live --uri wss://rpc.heima-parachain.heima.network:443
```

### DIFF
--- a/parachain/runtime/heima/src/migration/migration.md
+++ b/parachain/runtime/heima/src/migration/migration.md
@@ -1,5 +1,5 @@
 # Migrate to remove old bridge pallets storage
-p9151.rs
+p9251.rs
 This migration safely removes all on-chain storage for the legacy bridge implementation (ChainBridge, BridgeTransfer, AssetsHandler) after migration to omni-bridge.
 The pallets are completely independent with no shared storage.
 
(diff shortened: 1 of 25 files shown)


parachain/runtime/heima/src/migration/migration.md
```diff
@@ -1,3 +1,8 @@
+# Migrate to remove old bridge pallets storage
+p9151.rs
```
BillyWooo: one typo: p9251.rs

Kailai-Wang: Thanks!

parachain/runtime/paseo/src/migration/migration.md
```diff
@@ -1,3 +1,8 @@
+# Migrate to remove old bridge pallets storage
+p9151.rs
```
BillyWooo: same typo here: p9251.rs

BillyWooo (approved):
LGTM, except two small typos.

silva-fj (approved):

PR #3738, opened 2025-09-25
Add more logs and webhook listener for sendgrid

Kailai-Wang:
### Context

As topic

The py script will be running on worker server and eventually forwards the sendgrid webhook responses to slack channel.

Our staging and prod worker share the same sendgrid, so I only run it on prod-worker to avoid duplicate messages


tee-worker/omni-executor/sendgrid_webhook_listener.py
```diff
@@ -0,0 +1,230 @@
...
+logging.basicConfig(
+    level=logging.INFO,
+    format='%(asctime)s %(levelname)s %(message)s',
+    datefmt='%Y-%m-%d %H:%M:%S'
+)
+
+logger = logging.getLogger(__name__)
+
+# Slack webhook URL - REPLACE THIS WITH YOUR ACTUAL SLACK WEBHOOK URL
+SLACK_WEBHOOK_URL = "https://hooks.slack.com/services/YOUR/SLACK/WEBHOOK"
```
silva-fj, line 36: should this come from an env variable?

Kailai-Wang: It depends on how the script is deployed/started => I run the script locally so probably it doesn't make a difference 😂

silva-fj (approved):

PR #3741, opened 2025-09-29
Prefund local ecdsa signer

kziemianek:
Adds local ecdsa key and changes prefund scripts so bundler will be prefunded automatically after local env run with docker compose is set up.


tee-worker/omni-executor/docker/docker-compose.yml
```diff
@@ -77,7 +77,7 @@ services:
     image: ghcr.io/foundry-rs/foundry
     container_name: fund-eth-account
     command:
-      - "cast send --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 0x370Cd7A68F5762bdB4c922930445a58F39c46C3b --value 10ether -r http://ethereum-node:8545"
```
Kailai-Wang, line 80: What's the old `0x370Cd7A68F5762bdB4c922930445a58F39c46C3b`? Just want to make sure it's unused

Kailai-Wang (approved):

tee-worker/omni-executor/docker/docker-compose.yml
kziemianek: I think that's the old locally generated ecdsa address.

PR #3743, opened 2025-09-29
Use hyperliquid-rust-sdk actions to replace native defined

silva-fj:
## Summary
- Add hyperliquid-rust-sdk dependency to omni-executor 
- Replace custom EIP712 implementation with SDK-provided types (ApproveAgent, Withdraw3, ApproveBuilderFee)
- Remove redundant HyperliquidEip712Signature trait implementation


tee-worker/omni-executor/Cargo.toml
```diff
@@ -59,6 +59,7 @@ ethers = "2.0"
 frame-metadata = { version = "23.0.0", features = ["decode"], default-features = false }
 frame-support = { version = "39.0.0", features = ["std"] }
 futures = "0.3.31"
+hyperliquid_rust_sdk = { git = "https://github.com/silva-fj/hyperliquid-rust-sdk", branch = "eip712-public-trait" }
```
silva-fj, line 62: I have opened a PR in the official SDK for this, but I have the impression that the official sdk is not well maintained, the last release was like 5 months ago

Kailai-Wang: Hmm but your PR was merged into `master`, couldn't we simply use the master branch?

tee-worker/omni-executor/rpc-server/src/methods/omni/get_hyperliquid_signature_data.rs
```diff
@@ -58,42 +54,9 @@ pub struct HyperliquidSignatureData {
 #[derive(Serialize, Clone)]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum HyperliquidAction {
```
Kailai-Wang, line 56: Will it be a breaking change for the client, or is the data struct (layering, names, data types etc) compatible?

Kailai-Wang (commented):
Thanks - this definitely makes it more solid

tee-worker/omni-executor/rpc-server/src/methods/omni/get_hyperliquid_signature_data.rs
BillyWooo: I don't think so. Actually official sdk and we are doing the same thing. They got updated at the same time as we implemented this feature before. So just to use official rust sdk makes our life easier.

tee-worker/omni-executor/Cargo.toml
silva-fj: This is another PR, the `Eip712` trait is currently not public

tee-worker/omni-executor/rpc-server/src/methods/omni/get_hyperliquid_signature_data.rs
silva-fj: Yep, it's the same structure we had

Kailai-Wang (approved):

PR #3746, opened 2025-09-30
feat: add omni_exportBundlerPrivateKey RPC method

silva-fj:
Adds a new RPC method `omni_exportBundlerPrivateKey` to export the bundler private key (accounting ECDSA signer key) for internal use.

## Implementation
- Protected by ECDSA signature verification with timestamp-based replay attack prevention
- Uses `OE_BUNDLER_KEY_EXPORT_AUTHORIZED_PUBKEY` env var to configure authorized public key
- Timestamp must be within 5-minute window and cannot be reused
- Client provides RSA-encrypted AES key in request parameters
- Server decrypts AES key using shielding key's RSA private key
- Bundler private key is encrypted with AES before returning as `SerdeAesOutput`

## Security
- Two-factor security: ECDSA signature authentication + RSA-encrypted AES transport
- Bundler private key never transmitted in plain text
- Signature verification ensures only authorized key holder can call method
- Timestamp validation prevents replay attacks via persistent storage
- Future timestamp check prevents clock manipulation


BillyWooo (commented):
Hi @silva-fj , thanks. It looks quite good for me. 
Besides, can you also check and update the ci yml files, dockerfiles, and gramine manifest file?  Just search around `WILDMETA_BACKEND_ECDSA_PUBKEY` for reference.

BillyWooo (commented):
Hi @silva-fj , thank you. The implementation is really good.

Besides, can you also add one example / or extend current subcommand to generate out the rpc call command? Similar like we did in Signer/Revvy:
1. user call omni_getShieldingKey
2. use the ShieldingKey to create the signature, create new RSA-public key, and  construct the full curl command and print it out
3. user can copy and execute the full curl command directly to get encrypted bundler private key
4. use RSA-private key to decrypt out bundler private key.

silva-fj:
> Hi @silva-fj , thank you. The implementation is really good.
> 
> Besides, can you also add one example / or extend current subcommand to generate out the rpc call command? Similar like we did in Signer/Revvy:
> 
> 1. user call omni_getShieldingKey
> 2. use the ShieldingKey to create the signature, create new RSA-public key, and  construct the full curl command and print it out
> 3. user can copy and execute the full curl command directly to get encrypted bundler private key
> 4. use RSA-private key to decrypt out bundler private key.

Yes, we can extend the cli to do that. How about if we have a separate issue for this?

BillyWooo (approved):

Kailai-Wang (commented):
Thanks!

Yes - we would need a CLI to generate the export payload, given e.g. path to the private key file (similar to builder key import)

Additionally, I don't see `OE_BUNDLER_KEY_EXPORT_AUTHORIZED_PUBKEY` being written when building the release image - could you check it?

PR #3749, opened 2025-10-06
Remove Pumpx related functionality

silva-fj:
This PR removes all PumpX related functionality from the omni-executor, including:

- Removed PumpX API client crate (pumpx/)
- Removed PumpX RPC methods and module
- Removed cross-chain intent executor used by PumpX (intent/executors/cross-chain/)
- Removed swap-related RPC methods (add_wallet, export_wallet, submit_swap_order, etc.)
- Removed CrossChainIntentExecutor from RPC server and native task handler
- Removed swap dependencies from executor-worker
- Updated workspace configuration

## New Environment Variables

During the cleanup, pumpx-specific URLs were replaced with generic configuration:

- **`OE_SIGNER_URL`** (replaces `OE_PUMPX_SIGNER_URL`)
  - Default: `http://localhost:3456/jsonrpc`
  - Purpose: URL for the signer service

- **`OE_WORKER_URL`** (replaces `OE_PUMPX_WORKER_URL`)
  - Default: `ws://localhost:2100`
  - Purpose: Worker RPC server URL

These generic variables maintain the same functionality but with clearer, non-pumpx-specific naming.


silva-fj:
There will be merging conflicts after https://github.com/litentry/heima/pull/3747 is merged.

Kailai-Wang (commented):
A quick skim-through looks good to me - for the RPC methods we'll need to double-check with backend to make sure none of the removed RPCs is still used

@claude takes a second look too - and double-check if the issues found in the first review have been addressed

silva-fj:
Apparently all the `omni_*` rpc methods removed are still used by wildmeta, so I'll open another PR to only remove the `pumpx_*` methods with keep the shared infrastructure. It would be simpler than restoring everything on this PR after the refactoring made

PR #3753, opened 2025-10-09
Remove pumpx RPC methods

silva-fj:
This PR removes all pumpx-specific RPC endpoints from the rpc-server while preserving the pumpx infrastructure (API client, types, storage) that is still used by omni RPC methods.

**Changes:**
- Removed all pumpx RPC methods from `rpc-server/src/methods/pumpx/`
- Removed pumpx client SDK code, tests, and documentation
- Removed unused error codes (`USER_EMAIL_ID_MISMATCH_CODE`, `PUMPX_API_GET_USER_TRADE_INFO_FAILED_CODE`)

**Preserved:**
- Pumpx API client and infrastructure (used by `omni_signLimitOrder`, `omni_submitSwapOrder`, `omni_exportWallet`)


BillyWooo (approved):

Kailai-Wang (approved):
Thanks!

So we can't remove the deps or business logic code as the `omni_*` variants still use them

PR #3757, opened 2025-10-15
Add Apple OAuth2 authentication support

silva-fj:
## Summary

Add Apple OAuth2 authentication support and refactor OAuth2 flow.

### Changes

- Add Apple as OAuth2 provider (alongside existing Google support)
- Add Apple identity type to primitives (Identity, UserId, Web2IdentityType)

### OAuth2 Flow Refactoring (Implicit Flow)

**Security enhancements:**
- Add nonce generation and verification (prevents replay attacks)
- Verify id_token from client before code exchange (aud + nonce checks)
- Store both state and nonce in `OAuth2VerificationData`

**API changes:**
- Rename: `omni_getOAuth2AuthorizationUrl` → `omni_getOAuth2AuthorizationData`
- Return structured data object (url, state, nonce, scope, etc.) instead of just URL
- Add `id_token` parameter to `OAuth2Data`
- Remove Identity/sender from `OmniAuth::OAuth2` variant

### Environment Variables
```bash
OE_APPLE_CLIENT_ID_{CLIENT}=<client_id>
OE_APPLE_CLIENT_SECRET_{CLIENT}=<client_secret>
```


tee-worker/omni-executor/heima/identity-verification/src/web2/apple/mod.rs
BillyWooo, line 1: `apple/mod.rs` and `google/mod.rs` are almost the same (Only the different `IdToken` definition, right?). Can we merge them to one trait implementation?

tee-worker/omni-executor/rpc-server/src/methods/omni/login_with_oauth2.rs
BillyWooo, line 1: the `jwt` token created here is of type `AUTH_TOKEN_ACCESS_TYPE`. But in `verify_auth()`, it's verifying against `AUTH_TOKEN_ID_TYPE`.  Because later the jwt will be used to do `verify_auth()` when attaching passkey. Any solution?

silva-fj: Ah yeah, I forgot we issue two types of token. It's fixed now 👍🏼

BillyWooo (approved):
@Kailai-Wang  @kziemianek Please also take a look. Thank you.

kziemianek (approved):

PR #3759, opened 2025-10-20
Corewriter submit watch for requestLoan

Kailai-Wang:
### Context

As topic - it implements an initial version of "happy path" without complicated dynamic handling:
- it validates all preconditions before taking actions: price, size, minNtl ...
- it then executes 3 actions sequentially, only moving to the next action when the previous one finishes:
  - spot sell
  - usd transfer (into perps)
  - open position
- simply error out when any of the action fails
- tested ok on testnet (until order is opened)


tee-worker/omni-executor/hyperliquid/src/hypercore_api.rs
```diff
@@ -0,0 +1,660 @@
...
+	pub name: String,
+	pub index: u32,
+	#[serde(rename = "isCanonical")]
+	pub is_canonical: bool,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct SpotToken {
+	pub name: String,
+	#[serde(rename = "szDecimals")]
```
kziemianek: `#[serde(rename_all = "camelCase")]` ?

tee-worker/omni-executor/rpc-server/src/methods/omni/request_loan.rs
```diff
@@ -0,0 +1,166 @@
...
+
+			debug!("Received omni_requestLoan, params: {:?}", params);
+
+			validate_chain_id(params.chain_id as u32, Some("evm")).map_err(PumpxRpcError::from)?;
+
+			validate_wallet_index(params.wallet_index).map_err(PumpxRpcError::from)?;
+
+			// Validate smart wallet address
+			let smart_wallet_address_bytes =
+				validate_omni_account_hex(&params.smart_wallet_address, "smart_wallet_address")
```
kziemianek: Should this be `validate_ethereum_address` instead ? 
Then 20 len check is redundant

tee-worker/omni-executor/rpc-server/src/methods/omni/request_loan.rs
```diff
@@ -0,0 +1,166 @@
...
+						.with_suggestion(
+							"Provide a valid collateral size in human-readable format",
+						),
+				));
+			}
+
+			// Validate lending_ratio is between 0 and 100 (0-100%)
+			if params.lending_ratio > 100 {
+				return Err(PumpxRpcError::from(
+					DetailedError::new(PARSE_ERROR_CODE, "Lending ratio must be between 0 and 100")
```
kziemianek: Does it make sense to use `0` as lending ratio?

tee-worker/omni-executor/rpc-server/src/methods/omni/request_loan.rs
Kailai-Wang: I'm sorry - I should have not uploaded this file - please check `request_loan_test` only
I'll remove this file

kziemianek (approved):

PR #3777, opened 2025-10-21
Fix: Remove OAuth2 verification data after use

silva-fj:
## Summary
- Removes OAuth2 verification data from storage immediately after retrieval


Kailai-Wang (approved):

BillyWooo (changes requested):
Let's hold it there. 
Frontend team would prefer this `nonce` to exist for some time (like more than 5 minutes).

silva-fj:
> Let's hold it there. Frontend team would prefer this `nonce` to exist for some time (like more than 5 minutes).

The nonce is included in the id token. The authorization code can be used only once, so if they try again the verification will fail anyways

BillyWooo (approved):

PR #3780, opened 2025-10-22
Refactor to remove native task wrapper

Kailai-Wang:
### Context

Fixes P-1717

This PR tries to remove teh NativeTaskWrapper (and thus `native-task-handler`) completely by inlining the handling directly in rpc method handler.

With help-check of claude It tries to keep the error handling backwards compatible/same so we don't return e.g. different error code, but much of the code is not used so I'm not 100% sure if it covers all cases strictly, as I indeed find some unused error codes in `error_code.rs`

Some more clean-ups:
- the `client_id` in middleware auth is actually not used, so I unwrap the `struct User` to use the underlying `omni_account` directly
- `omni_submitNativeTask` is removed, together with task primitives definitions
- `ethereum-intent-executor` and `solana-intent-executor` are removed from `start_server` and `RpcContext`


silva-fj (approved):
LGTM

PR #3781, opened 2025-10-22
extend HyperliquidAction with UserDexAbstraction variant

BillyWooo:
extend HyperliquidAction with UserDexAbstraction variant

Add support for the UserDexAbstraction action type in the Hyperliquid signature data generation.

Changes:
- Import UserDexAbstraction from hyperliquid-rust-sdk
- Add UserDexAbstraction variant to HyperliquidActionType enum with user address and enabled flag
- Add UserDexAbstraction variant to HyperliquidAction enum
- Implement handling logic with address validation and EIP-712 signature generation
- Add comprehensive unit tests for signature generation and JSON deserialization

This extends the existing Hyperliquid action support to enable DEX abstraction functionality, following the same pattern as ApproveAgent, Withdraw3, ApproveBuilderFee, and SendAsset.

Also a new [PR](https://github.com/hyperliquid-dex/hyperliquid-rust-sdk/pull/159) for upstream.


tee-worker/omni-executor/Cargo.toml
```diff
@@ -58,7 +58,7 @@ hex = "0.4"
 hex-literal = "0.4"
 hmac = "0.12.1"
 http = "1.3.1"
-hyperliquid_rust_sdk = { git = "https://github.com/BillyWooo/hyperliquid-rust-sdk", branch = "usdClassTransfer_sendAsset" }
+hyperliquid_rust_sdk = { git = "https://github.com/BillyWooo/hyperliquid-rust-sdk", branch = "fork_new_feature" }
```
Kailai-Wang, line 60: Shall have a consistent branch on which we accumulatively apply our changes?

Kailai-Wang (approved):

tee-worker/omni-executor/Cargo.toml
BillyWooo: Yes, this is the consistent one. 
The previous ones only have specific names. But it looks like the upstream is always late. So `fork_new_feature` will be the consistent one.

PR #3784, opened 2025-10-23
Add loan record storage

Kailai-Wang:
### Context

As topic - currently it's written to storage after spot sell is filled


kziemianek (approved):

kziemianek:
I agree we might need pagination for accounts with many loan records.

PR #3789, opened 2025-10-27
use `sub` the unique identifier to replace email

BillyWooo:
minor PR, as title. 
`sub` is the unique identifier.


silva-fj (approved):
You need to let the frontend (or backend) know that this identifier is what will be used to derive the omni account, just in case they use the google account for example to do so in the app. (they are not the same)

PR #3793, opened 2025-10-27
Initial implementation of payback loan

Kailai-Wang:
### Context

As topic - it follows the action of:
- cancelling the order or closing the position
- transfer to spot
- spot by the collateral

It fixes the price calculation too (using meta-and-asset-ctxs instead of mid price)

I detected some small issues in tests - still investigating them, but want to make changes after this PR is reviewed/merged to make it more clear

Edit:
I had a few more commits trying to fix the bugs that I found:
- I add the `position_size` into the loanRecord, which will be used when closing the position - we can't fully rely on the `position_size` from the clearingHouseState as it could be the sum of multiple positions (of the same ticker)
- it seems that the `markPx` in the metadata could denote either the highest buy order or lowest sell order, to figure that out, we'll need to compare it with `midPx`
- the price should be re-fetched when doing action 3
- recalculate the spot-buy-size: how much can we buy with available USDC?

**Note:**
occasionally, I still see position opening order gets silently ignored (I get `unknownOid` if I query it) - the encoding of price and size looked ok. I don't know the reason yet..


Kailai-Wang:
As agreed with @kziemianek - we'll merge this PR first and address problems in separate PRs.

PR #3796, opened 2025-10-29
Verify provider token matches client token - OAuth2

silva-fj:
Verify that the OAuth2 provider's token matches the client-provided token by comparing their sub claims.

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/verify_auth.rs
+++ b/tee-worker/omni-executor/rpc-server/src/verify_auth.rs
@@ -24,6 +24,7 @@ pub enum AuthenticationError {
 	VerificationCodeNotFound,
 	InvalidVerificationCode,
 	OAuth2Error(String),
+	OAuth2SubClaimMismatch,
 	AuthTokenError(AuthTokenError),
 }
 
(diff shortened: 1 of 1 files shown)


tee-worker/omni-executor/rpc-server/src/verify_auth.rs
```diff
@@ -230,16 +234,49 @@ async fn verify_oauth2_provider<
...
+						"Could not decode Apple id token from provider".to_string(),
+					)
+				})?;
+
+			id_token.sub
+		},
+	};
+
+	if client_sub != provider_sub {
+		tracing::warn!(
```
BillyWooo, line 266: it's better to be "error", instead of "warn".

BillyWooo (approved):

tee-worker/omni-executor/rpc-server/src/verify_auth.rs
silva-fj: Why do you think we should log it as error? The worker is handling this as expected. It would be the client in this case that is sending incorrect data

BillyWooo: you confused me. This is the error case. The follow line even returns "Err". Why not log the message as error type?

silva-fj: Yes it returns an error but it is expected to return an error to the client if these conditions are met. It's just validation, the worker is working as expected. You should log errors to signal issues that need to be addressed, for example reading the storage failed because the db file was not found or some external service like the signer is down

PR #3797, opened 2025-10-30
Add more loan record status

Kailai-Wang:
### Context

This is a big refactoring to the current impl:
- it introduces various `LoanState` which will be updated after each action
- it tries to use a waterfall pattern to be able to resume from any state where it was left off (like state machine).
- it improves the readability by separating the logic into smaller functions and outlining the flow in the main handler
- it adds two test rpcs for easier debugging of position opening and closing:
  - `openPositionTest`
  - `closePositionTest`

Please beware of the rpc parameter change and struct change, examples:
- no `client_id` in any of the RPCs
- an optional `loan_nonce` in `requestLoanTest` - if we wish to resume from some intermediate state of a specific loan, pass in that loan nonce, for fresh new loans you either leave it out or use the same nonce as in the userop
- `to_perp_move_tx_hash` is added into requestLoan response
- the `LoanRecord` struct change

### DIFF
--- a/tee-worker/omni-executor/rpc-server/src/methods/omni/close_position_test.rs
+++ b/tee-worker/omni-executor/rpc-server/src/methods/omni/close_position_test.rs
@@ -19,7 +19,6 @@ pub struct ClosePositionTestParams {
 	pub chain_id: ChainId,
 	pub wallet_index: u32,
 	pub omni_account: String,
-	pub client_id: String,
 	pub ticker: Option<String>, // If None, close all positions
 }
 
@@ -198,7 +197,6 @@ pub fn register_close_position_test<
 					params.chain_id,
 					params.wallet_index,
 					close_calldata,
-					&params.client_id,
 				)
 				.await?;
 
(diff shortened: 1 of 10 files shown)


tee-worker/omni-executor/rpc-server/src/methods/omni/close_position_test.rs
```diff
@@ -0,0 +1,266 @@
...
+
+			let omni_account = to_omni_account(&params.omni_account).map_err(|_| {
+				error!("Failed to parse omni account");
+				PumpxRpcError::from(DetailedError::new(
+					PARSE_ERROR_CODE,
+					"Failed to parse omni account",
+				))
+			})?;
+
+			let smart_wallet = &params.sender;
```
silva-fj, line 63: What's the difference between `smart_wallet` and the `omni_account` above?

Kailai-Wang: omni_account is bound to your login identity
smart wallet is your smart wallet address (op.sender)

silva-fj (approved):

PR #3798, opened 2025-10-30
Implement CREATE2 factory for deterministic contract deployments

silva-fj:
## Summary

Implements a CREATE2 factory system for deterministic contract deployments across multiple EVM chains.

### Changes

- Configure Foundry for deterministic bytecode (`bytecode_hash = none`, `cbor_metadata = false`)
- Add `Create2Factory` contract with deployment tracking and redeployment prevention
- Add comprehensive test suite
- Add deployment scripts for factory and AA contracts via CREATE2
- Add factory addresses registry for tracking deployments across networks
- Document CREATE2 deployment strategy in DEPLOYMENT.md


tee-worker/omni-executor/contracts/aa/foundry.toml
```diff
@@ -3,9 +3,13 @@ src = "src"
 out = "out"
 libs = ["lib", "../forge-libs"]
 solc_version = "0.8.28"
+evm_version = "cancun"
 optimizer = true
 optimizer_runs = 1000000
 via_ir = true
+# Deterministic bytecode settings for CREATE2 deployments
+bytecode_hash = "none"
+cbor_metadata = false
```
Kailai-Wang, line 12: Have you tried to deploy contracts with create2 factory? I'm not sure if would cause any verification problem (or you'll need manually verify that) as etherscan probably expects some metadatas to verify the contract

tee-worker/omni-executor/contracts/aa/script/DeployWithCreate2.s.sol
Kailai-Wang, line 1: IIUC this is to deploy the contracts in series - can you have some script (either bash or forge script) to deploy an individual contract?

Imagine I have a simple wildlend contract and I want it to be deployed via c2f (create2factory)

tee-worker/omni-executor/aa-contracts/src/core/Create2Factory.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.28;
+
+/**
+ * @title Create2Factory
+ * @notice Factory contract for deterministic contract deployments using CREATE2
+ * @dev Provides deterministic address generation across multiple EVM chains
+ *      This contract should be deployed using a fresh EOA on each network
+ */
+contract Create2Factory {
```
Kailai-Wang: Should it be versioned? Like `Create2FactoryV1`

tee-worker/omni-executor/aa-contracts/src/core/Create2Factory.sol
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * @notice Generates a sender-protected salt for deployment
+     * @param contractName The name/identifier of the contract
+     * @param version The version string
+     * @param sender The address of the deployer (typically msg.sender)
+     * @return salt The generated salt
+     * @dev Sender-protected salts prevent frontrunning by including the deployer address
+     *      This ensures only the specified sender can deploy to the computed address
+     */
+    function generateSalt(string memory contractName, string memory version, address sender)
```
Kailai-Wang: Which "contract version" should I use here?

I see in the script that the salt is calculated as `bytes32 salt = factory.generateSalt("SimplePaymaster", contractVersion, deployer);` for example, so it seems `contractVersion` is used for all salt computation.

Shall we actually dismiss it?

Kailai-Wang: Why I don't think involving `sender` is a good idea - it would then require always the same deployer across networks, also the address is not off-chain predictable.

So I'd also suggest we remove it, so that anyone can call the c2f

Kailai-Wang: Maybe we could check some other battle-tested c2f (e.g. openzeppelin, safe's singletonFactory, uniswap's UniversalDeployer)

IMO a simple utf8 string (`contractName`) is enough for the salt

Kailai-Wang (commented):
Thank you, I left some comments - mainly questions around the salt generation

tee-worker/omni-executor/contracts/aa/foundry.toml
silva-fj: I tried deploying a contract locally on Anvil. I got these values from the docs [here](https://getfoundry.sh/guides/deterministic-deployments-using-create2#configure-your-foundrytoml)

tee-worker/omni-executor/contracts/aa/src/core/Create2FactoryV1.sol
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * @notice Deploys a contract using CREATE2
+     * @param salt The salt for deterministic address generation
+     * @param bytecode The creation bytecode of the contract to deploy
+     * @return deployed The address of the deployed contract
+     */
+    function deploy(bytes32 salt, bytes memory bytecode) external payable returns (address deployed) {
+        // Predict the deployment address
+        deployed = computeAddress(salt, bytecode);
```
kziemianek, line 34: Just to be 100% sure we should add `owner` and a check here if we want to prevent adress sniping.

PR #3817, opened 2025-11-13
add multi-platform origin validation for WebAuthn (Web/iOS/Android)

BillyWooo:
minor PR, as title.


tee-worker/omni-executor/rpc-server/src/methods/omni/mod.rs
```diff
@@ -191,9 +191,26 @@ pub fn get_rp_id_for_client(client_id: &str) -> &str {
...
+/// 4. Format as: android:apk-key-hash:<encoded-hash>
+pub fn get_allowed_origins_for_client(client_id: &str) -> Vec<&str> {
 	match client_id {
-		"wildmeta" => "https://app.wildmeta.ai",
-		_ => "http://localhost:3000", // Development/testing
+		"wildmeta" => vec![
+			// Web and iOS origin (both use HTTPS format)
+			"https://app.wildmeta.ai",
+			// Android origin (apk-key-hash format)
+			"android:apk-key-hash:JFBzaEkXsWLOfuRK5TApIS-cJ_Vf2KYs9YBBd9-VmCw",
```
Kailai-Wang, line 212: Just to double-check it's ok to publicize this

Kailai-Wang (approved):

tee-worker/omni-executor/rpc-server/src/methods/omni/mod.rs
BillyWooo: Yes. It makes sense. No harm yet. But It's better not to publicize it. 
Let me do some refactor.

PR #3835, opened 2025-11-24
Back to `wasm-override` for runtime upgrade CI

Kailai-Wang:
### Context

I have spent much time on it but couldn't find a perfect solution chopsticks, it either has the problem of extracting the runtime version, or injecting the scheduler directly.

See `chopsticks-runtime-upgrade-issue.md` for more information.

I also filed a report in chopsticks repo: https://github.com/AcalaNetwork/chopsticks/issues/979


BillyWooo (approved):
do we still need `paseo-relaychain.yml` ?

PR #4019, opened 2026-05-31
feat(omni-executor): privacy pool / confidential invoice (privacy-invoice-demo, aligned with dev)

Kailai-Wang:
> **Draft** — opened for review/CI of the `privacy-invoice-demo` work, now realigned on top of the latest `dev`.

## What this is

Brings the `privacy-invoice-demo` branch onto current `dev` and opens it for review. It is a **large feature branch**, not a small demo: it adds a ZK privacy-pool / confidential-invoice flow to omni-executor.

- **62 commits ahead** of `dev`, now **0 behind** (latest `origin/dev` merged in).
- After the submodule cleanup below, the diff is **~181 files** (was ~1360 — the rest was vendored third-party code).

### Core (non-demo) code touched

- `crypto/`: new `privacy_pool.rs`, `confidential.rs` (Groth16 / BN254 ZK helpers).
- `rpc-server/methods/omni/`: new `confidential_invoice`, `pay_invoice`, `withdraw_pool`, `export_bundler_private_key`, `get_shielding_key`, plus changes to `settle_user_op`, `verify_user_op`, `server.rs`.
- `storage/`: `confidential_invoice.rs`.
- `vendor/ark-circom/`: vendored dependency for in-enclave proof generation.
- `contracts/privacy-pool/`: Solidity pool + circom `withdraw` circuit + compiled artifacts (`.zkey`/`.wasm`/`verification_key.json`).

## Alignment with `dev`

Merged `origin/dev` in. The only conflict was `omni-executor/Cargo.lock` (`rand` version — `dev` carried security bumps). Resolved by taking `dev`'s lockfile and regenerating with `cargo update --workspace`; `cargo metadata --locked` passes, so the lockfile matches `Cargo.toml`.

**Build check:** `SKIP_WASM_BUILD=1 cargo check --workspace --locked` passes (exit 0) after the alignment. Only a pre-existing `unused variable: join` warning in `bin/src/main.rs:256` (not introduced here). Full CI on this PR remains the real gate.

## privacy-pool/lib → git submodules

The privacy-pool Foundry libs were checked in as plain directories (~1184 files). Converted to **git submodules**, matching how omni-executor's other contracts reference forge-libs in `.gitmodules`. Pinned refs:

| lib | ref |
|---|---|
| circomlib | `v2.0.5` |
| forge-std | `v1.15.0` |
| openzeppelin-contracts | `v5.6.1` |
| poseidon-solidity | content-matching commit (no upstream tag) |

privacy-pool only imports OZ `ERC20`/`IERC20`/`SafeERC20` and poseidon `PoseidonT3` — all present and unchanged at the pinned refs, and the foundry remappings resolve correctly, so **contract compilation is unaffected**. (The prior OZ snapshot carried a few post-5.6.1 experimental crosschain contracts that privacy-pool never imported; pinning to the clean `v5.6.1` release drops those.)

## EIP-4337 / omni-account contracts — untouched

No `contracts/aa/*.sol` is modified — the deployed **EntryPoint / OmniAccountFactory / Paymaster / OmniAccount** contracts are unchanged. The privacy-pool `Deploy.s.sol` does not deploy or touch any of them; the privacy pool is a separate, standalone set of contracts. The old `contracts/aa/aa-demo-app/` (a Next.js demo front-end) is **removed** in this branch — it is not referenced by the live worker, and the standalone `heima-sandbox` project does not depend on it.

## Sensitive-info review (public repo)

Scanned the full diff (excluding third-party `lib/`/`vendor/`/lockfiles):
- **No real private keys, mnemonics, or API keys committed.**
- `.env` / `cache/` / `out/` / `broadcast/` are git-ignored; only `.env.example` templates (empty values) are tracked.
- The only key-looking strings are benign: the BN254 field prime in `privacy_pool.rs`, an RSA **public** shielding-key example in a doc comment, and `0xdeadbeef`/`0x1234…` test placeholders.

## Still needs

- Human review of the confidential / privacy-pool logic (ZK proof handling, key export endpoint) before un-drafting.
- Green CI.


Kailai-Wang:
## CI triage

Two checks failed on the previous run; addressed / explained:

**1. `omni-executor-check` (Cargo clippy) — fixed ✅**
CI runs `cargo clippy --workspace -- -D warnings`, which gated this branch's code for the first time. Fixed all lints with mechanical, semantics-preserving changes (needless borrows, redundant closures, `get(0)`→`first()`, `len()<1`→`is_empty()`, `&[x.clone()]`→`slice::from_ref(&x)`, range-loop→`enumerate()`, unused `join`→`_join`). `cargo clippy --workspace -- -D warnings` now passes locally (exit 0). Pushed in the latest commit.

**2. `contract-check` (Check solana accounting contracts) — pre-existing, unrelated to this PR**
The solana contracts' forge tests pass (13 passed, 0 failed). The failure is in the `anchor b` step: `anchor-cli` (installed via `cargo install --git .../anchor`, currently 1.0.2) pulls `anchor-syn 0.32.1`, which fails to compile against the current Rust toolchain (`no method named local_file found for proc_macro::Span`). This is a toolchain/version-drift issue affecting any PR that touches contracts — on `dev` this job is path-skipped, so it never surfaced. It is **not** caused by the privacy-pool changes and shouldn't be fixed inside this feature PR; pinning the anchor-cli version in CI is a separate maintenance fix.

Kailai-Wang:
## CI fix: cargo test link failure (__rust_probestack)

The `omni-executor-check` job's `cargo test` step failed at link time:

```
rust-lld: error: undefined symbol: __rust_probestack
  >>> wasmer_vm_probestack in libwasmer_vm-*.rlib
```

**Root cause:** this branch adds `ark-circom`, whose singlepass backend pulls in `wasmer-vm`. On x86_64 Linux `wasmer_vm_probestack` falls back to referencing `__rust_probestack`, a symbol the Rust compiler no longer emits by default since 1.90. `dev` doesn't carry this dependency, so the failure is specific to this feature.

**Fix:** provide the canonical stack-probe implementation (verbatim from rustc's `compiler-builtins`) in `oe-crypto`, as a **weak** symbol scoped to `x86_64-linux`. Weak linkage prevents a duplicate-symbol clash with any dependency that also defines it (the omni-executor bin test binary hit exactly that, fixed by `.weak`).

Verified locally:
- `cargo test --workspace --no-run` links cleanly (all test binaries build).
- `oe-crypto` tests: 16 passed, 1 ignored.
- `cargo clippy --workspace -- -D warnings` still passes.

Other targets (including SGX builds) use wasmer-vm's own platform branch and are unaffected.

Kailai-Wang:
## CI fix: confidential_invoice tests (ParseError)

After the link issue was resolved, the `cargo test` step surfaced 2 failing tests:
- `confidential_invoice::tests::test_create_and_get_invoice`
- `confidential_invoice::tests::test_mark_invoice_withdrawn_updates_status_and_summary`

Both panicked on `.unwrap()` with `ParseError: "Invalid JSON format or missing required fields"`.

**Root cause (pre-existing in this branch, not from the dev alignment):** the confidential_invoice RPC methods parse params via `parse_rpc_params::<Struct>()`, which expects a JSON **object** (named fields). The tests called them with **positional arrays** via `rpc_params!`, so deserialization into the param struct failed.

**Fix:** switched all RPC calls in both tests to jsonrpsee `ObjectParams` with the field names each method expects (`created_by`, `buyer_identifier`, `token_address`, `chain_id`, `description`, `recipients` / `invoice_id`, `tx_hash`, `seller_address` / `created_by`, `viewer_address`). Removed the now-unused `rpc_params` import.

Verified locally: both tests pass (2 passed, 0 failed), `cargo clippy -p rpc-server -- -D warnings` clean.

Rebased onto the latest branch head (now includes #4020's anchor-cli pin merged via dev).

Kailai-Wang:
## CI fix: contract-check (anchor build)

Supersedes the earlier v0.32.1 anchor pin (wrong direction).

**Real root cause — a three-way version knot:**
- `anchor b --ignore-keys` (used by #3863 to sync program-id and skip the keypair check) needs `--ignore-keys`, **added in anchor 1.0**.
- anchor 1.0+'s `anchor-syn` only compiles when **proc-macro2 < 1.0.95** — 1.0.95 changed the unstable `Span` API anchor-syn relies on (anchor issue #3661).
- The solana contract `Cargo.lock` pinned **proc-macro2 1.0.96**, so the build failed with `no method named local_file for proc_macro::Span`.
- Pinning anchor to 0.32.1 dodged the compile error but dropped `--ignore-keys` → `unexpected argument`.

**Fix:**
- anchor-cli pinned to **v1.0.2** (restores `--ignore-keys`).
- proc-macro2 pinned to **1.0.94** in the solana contract `Cargo.lock` so anchor-syn builds.

Verified `cargo check` on the solana contract passes with proc-macro2 1.0.94; CI runs the full `anchor b --ignore-keys` end-to-end.

Pre-existing CI/toolchain issue (contract-check is path-skipped on `dev`), unrelated to the privacy-pool feature.

